PR #3, opened 2017-11-03
[RFC] New script generated SMD two terminal packages

poeschlr:
I wrote a script with the goal to generate all possible two terminal chip packages. (Resistor, Capacitor, ...)
https://github.com/poeschlr/kicad-footprint-generator/tree/smd_chip_package_rlc/scripts/SMD_chip_package_rlc-etc

Here the first patch of footprints for review.
I used IPC 7351 to arrive at the pad sizes. (The old pads where far too small.)
For the hand solder versions i elongated the pads in x direction by 0.3mm (This resulted in a pad outside the package similar to my personal footprints that i use for handsoldering. I can easily increase this if someone thinks this is not enough.)

Footprints smaller 0603 do not have a silk screen any more. (The old silk screen lines where quite far away. Outside the courtyard area to be precise. This does not make sense to me. But i can add silk lines if it is desired.)

The screenshot shows both the old and new footprints in comparison. (new once always to the right of the old one.)
screenshot from 2017-11-03 20-26-32

Because the screenshot is quite bad at showing the footprints here a pcb file that has all the new and old resistor footprints in it. (remove the txt ending after downloading. github does not accept attaching kicad_pcb files.)
[smd_resistor_fp_test.kicad_pcb.txt](https://github.com/KiCad/kicad-footprints/files/1442298/smd_resistor_fp_test.kicad_pcb.txt)

I still need to include a special case for 0201 footprints that would solve issue https://github.com/KiCad/Resistors_SMD.pretty/issues/22. (I might just do it by hand instead of making the script even more complex)

Edit:
I allow the script to scale the fab reference down to 0.5mm. (Such that it can fit inside small components.)
As soon as it does not fit any more (I want at least 3 chars to fit inside) I move the fab reference to the top of the component. But i still scale it down to the minimum size i configure.

Should we use singular as the lib name or do we keep plural?


poeschlr:
As soon as we are happy with these new footprints i would suggest we declare resistor, capacitor, ... libs as legacy and start moving their content here. (If v5 is really release before FOSSDEM then we have only 2 or 3 month left to finish the transition.)

poeschlr:
My script now rotates the fab references if it would mean the font size can increase this way.
New pcb file with the updated footprints:
[Uploading new_resistor_fp_test.kicad_pcb.txt…]()

SchrodingersGat:
Yes x 1000 :) This is really great. I don't have time right now to review but I will certainly give it some attention soon.

I agree we need to start transitioning the libraries now. I can see that my initial plan of holding off until the last minute will be too much work given that it is essentially just the two of us.

We can start looking at the "oldest" libs (those that have not been touched in a while) and copy those across and reject any new contributions.

poeschlr:
That's basically why i started with the resistor stuff. There is no pull request waiting and the outstanding issues can not really be resolved within the legacy lib. (ok the one about the 0201 resistor could be resolved but i can copy the new 0201 resistor back to the old repo if you think this would be worth it.)

SchrodingersGat:
What I would also like to see is that each directory contains a README.md which briefly describes the contribution efforts. Much like we do in the packages3d-source repo.

SchrodingersGat:
I wouldn't necessarily bother copying the 0201 back. We only have so much energy to expend and I'd rather see it focused on v5 libs

SchrodingersGat:
Rene great work. Let me know if you have any further changes to make otherwise I'll merge.

I'd like to see this script extended to:

* Chip capacitors
* Chip inductors
* Chip diodes (with polarisation marker)
* Chip LEDs (with polarisation marker)

poeschlr:
Diodes might get interesting.
It seems diodes might have not only an EIAJ naming (the size codes used for resistors, ...) but also JEDEC (SMA, ...)
See for example: http://datasheets.avx.com/schottky.pdf

603 = SOD-523
0805 = SOD-323
1206 = SOD-123
2010 = SMA (D0-214AC)
2114 = SMB (D0-214AA)
3220 = SMC (D0-214AB)

poeschlr:
Added capacitor footprints.
[capacitors.kicad_pcb.txt](https://github.com/KiCad/kicad-footprints/files/1443389/capacitors.kicad_pcb.txt)

poeschlr:
The script can now create polarized packages. Here an example for diodes:
For < 0603 i got inspired by [this blog post](https://blogs.mentor.com/tom-hausherr/blog/2010/09/22/pcb-design-perfection-starts-in-the-cad-library-part-2/) 

screenshot from 2017-11-04 23-45-16

And as always a pcb file that holds all of the diodes footprints:
[diodes.kicad_pcb.txt](https://github.com/KiCad/kicad-footprints/files/1443639/diodes.kicad_pcb.txt)

The castellated version is the holds the footprints for diodes with pads like this: (Yes i noticed that there is a typo. I already fixed it in the script and in the footprints i push.) http://datasheets.avx.com/schottky.pdf

poeschlr:
@jkriege, @evanshultz could you guys also take a look at this?

SchrodingersGat:
Nice, I like the idea of the dot for small diode footprints

poeschlr:
Now this includes the requested tantal and led footprints as well.

For tantal again a pcb file for checking if you need it.
(Old stuff on the back side.)

Tantal had footprints for reflow, wafe and hand without giving details about where any of the pad measurements came from. I now only generate reflow (no suffix) and handsolder (pad elongated by 0.3mm)
The handsolder version (as everywhere else) had very long pads. nearly 1mm longer than required for reflow. If one needs pads that long to handsolder, one should buy a better soldering iron.

SchrodingersGat:
We can ignore wave, I think.

poeschlr:
Now all footprints of the effected libs should be contained here. (I did not change all of them.)

Only thing that might be controversial is that i split up the LEDs repo into LED_SMD and LED_THT

poeschlr:
@SchrodingersGat i only need to solve the problem with the paste layer for small components, then this should be merge able.
The old issue about it: https://github.com/KiCad/Resistors_SMD.pretty/issues/22

My main problem is that i am not yet sure how the correct paste layer should look like. (All papers i could find do not give a clear recommendation. They only list measurements. Understandable but still..)

poeschlr:
Can somebody enlighten me with the paste reduction (%) stuff?
When i set it to -10% it results in a 0.8x0.8 paste area for a 1x1mm pad.
Setting it to -50% results in no paste left.

If none of you can answer this i will ask over at the mailing list.

poeschlr:
@SchrodingersGat now this should be ready to merge.

SchrodingersGat:
It looks like you have copied the other models across too? Are there any models in the R / L / C existing in the current (v4 legacy) libs that are *not* in this PR?

poeschlr:
Everything from the following libs should be included here:
- Capacitors_SMD
- Capacitors_Tantalum_SMD
- Diodes_SMD
- Inductors_SMD
- LEDs
- Resistors_SMD

SchrodingersGat:
Great, that's a lot of work :) I'm happy to merge this. Can you please mark all those libs as "no new PRs"? :)

poeschlr:
I already did mark them (In the readme. I can not edit the repo descriptions them selves.)

SchrodingersGat:
Great, nice work :) Hopefully we can get them all across without too much effort.

poeschlr:
I think most of them we will just transfer as they are. (Most of my time for these scripts was spent in learning how the ipc rules work. And finding the correct datasheets that hold the information i needed. Writing the script was surprisingly little work.)

PR #4, opened 2017-11-05
Pin headers

SchrodingersGat:
This PR adds the pin headers from the PinHeaders footprint repository

* Footprints are split based on pitch
* Renamed to match new KLC convention

Once this is merged I shall mark the PinHeaders repository as not accepting any more PRs


poeschlr:
Seems to be ok. Thanks.

Shackmeister:
was these simply renamed or did you change the script accordingly?

Shackmeister:
also should we start renaming/remaking the models in packages3D for this repo?

SchrodingersGat:
@Shackmeister these were simply renamed. We should also rename (and reorganize) the models in kicad-packages3d.

Would this be handled most easily by regenerating the models?

Shackmeister:
arent the 3D model references fine for the Pinheaders? I believe you talk about the sockets.
Ill take a look at the 3D model as you guys moves further with this repository :)

for the 3D models I think this would be the easiest way, also I really hate when the scripts gets out of sync.

SchrodingersGat:
The 3D models will need to be renamed for *both* PinSocket and PinHeader footprints - as the naming conventions have changed for each...

Also note that you should *not* change the models that exist in the kicad-library repo, only the kicad-packages3d repo. kicad-library must remain in sync with the various .pretty repos

Shackmeister:
ah my bad, for some reason I thought of the 3D offsets in the socket scripts

makes sense, thank you :)

PR #5, opened 2017-11-05
Pin socket

SchrodingersGat:
This PR moves the Socket Strips across from the Socket_Strips repository.

* Separated into folders based on pitch
* Renamed to match new convention

When this is merge I shall mark the Socket_Strips repo as not accepting any new PRs


poeschlr:
1mm pitch version is missing.

SchrodingersGat:
Yep, these footprints [have not been created](https://github.com/kicad/socket_strips.pretty)

poeschlr:
So we have them for pin headers but not for socket strips? That is kind of strange. But ok we can only transfer what is there.
@Shackmeister you created these. Care to add the 1mm version as well?

Shackmeister:
@poeschlr I no its kinda strange, I was working on the pinheader scripts (as you probably remember) and suddently needed the 1mm headers at work. I wanted to overhaul the socket script before making those.
@terjeio has been making new socket models, and might be working on the footprints?
in case not Ill be happy to take a look at the socket footprints and also add the 1mm strips

terjeio:
@Shackmeister The footprints should be updated, they are script generated and I have been looking for the script that has been used to make them in order to remove offset and rotation.

No luck yet, see [Issue #9 for the socket strip models](https://github.com/KiCad/Socket_Strips.pretty/issues/9#issuecomment-341794778)

Shackmeister:
Hi @terjeio I added a comment in the link you provided, that should fix it.
should I fix the offset etc or would you like to do it?

terjeio:
@Shackmeister I am not sure it is so simple, see [this comment I made earlier](https://github.com/easyw/kicad-3d-models-in-freecad/issues/112#issuecomment-340390630)

The best option would be to track down the script used to generate the footprints because there has been other changes made as well - maybe by @jkriege2 knows  where it can be found

Shackmeister:
this part was simply getting the script running.

These are the scripts which has been used for the footprint.
The problem is a lot of changes have been done using other python scripts and text editors in order to fix KLC errors. (even the pinheaders have been renamed now which was done with a seperate script)
Getting this script up and running with up to date standards is gonna take a while.
I spend 1 or 2 weeks on the pinheader script, which is why I have postponed this for some time :)
if you are up for the task Ill be happy to ask you, otherwise I can make it sometime hopefully not too distant in the future

terjeio:
@Shackmeister maybe we should wait until @jkriege2 responds? It was he that commited the last PR for the footprints so hopefully he knows the whereabouts of the script used.

Script generated footprints and models should contain a link back to the script used, at least in the PR comments, in order to avoid issues like this.

antoniovazquezblanco:
I have never used the generation scripts and I don't know if it possible to insert comments in the model files themselves but maybe having the file with the name of the generator inside is a good option.

SchrodingersGat:
It should be sufficient to add an entry into the `README.md` file. For large batches of scripted items, you can have:

* Author
* Link to script
* Date generated
* Any special notes

jkriege2:
Hi!

have a look at the original PR: https://github.com/KiCad/Socket_Strips.pretty/pull/6 My scripts have been linked there. The PR for the scripts is: https://github.com/pointhi/kicad-footprint-generator/pull/18 ... i.e. they are part of @pointhi 's kicad-footprint-generator repo ... but I have no clue what was changed after that without adapting these scripts!

The only way to link the scripts would be by adding the link to the description field, which ist usually not that easy to do, as the script-PR is typically merged only AFTER the footpritns have been merged ... so only the PR could be linked!

JAN

terjeio:
Ok - so how to fix this?
The script in question does not run so it cannot be the one that has been used to create the footprints. 

Should it be fixed so it generates exactly the same footprints as the KLC compliant version in the repository and the modified to add the 1mm sockets and match my script generated models?

Shackmeister:
@terjeio I just had a look at my commit back when I rewrote the pinheader part. Apparently I was the one who made the script unrunable, sorry, my bad.

I dont think the script needs to make footprints exactly like the ones in the repo, but they should comply with KLC. I also think the parameters could use an update

poeschlr:
I think adding a readme at the library level should be sufficient. In my new footprints i added the phrase: `generated with kicad-footprint-generator` to the end of the description field. (This should make it really clear to everyone that these are in fact script generated footprints.)

terjeio:
@Shackmeister I can have a go at it but I do not know how long it will take. The code for drawing the lines isn't exactly easy to follow with all the math involved but the framework looks good to me.

Which parameters did you think of? For 3D rendering?

As a side note: it would be nice to be able to drag&drop a footprint file into the footprint editor or an easy way to reload...

@antoniovazquezblanco Adding a link to the script in model files is possible, I have done so by including it in the license text in the header

terjeio:
Why didn't I check the pin headers? This since the core code is shared with the socket strips. I believe @DASFrank has the updated script...

Shackmeister:
DASFrank is my work account, pretty sure he doesn't have any updated script. But I can probably make it

terjeio:
@Shackmeister Oh well, complete confusion for me then. The core script has been [recently updated](https://github.com/KiCad/Pin_Headers.pretty/pull/27)
for the pin headers - you did that? And by modifying footprint_scripts_pin_headers.py?

Shackmeister:
yes I did that, I found some errors in the way some of the silkscreen was handled when making the 1mm header.. therefore the changes in kicad-footprint-generator/scripts/tools/footprint_scripts_pin_headers.py

I also made some changes in kicad-footprint-generator/scripts/pin-headers_socket-strips/make_pin_headers.py
which was simply to add the new distance between the pads

terjeio:
@Shackmeister ok, so you have the updated scripts after all. The scripts int pointhi's repositiory was last changed in july and does not contain code for the 1mm headers. Or am I completely lost now?

Shackmeister:
@terjeio I can see why its pretty confusing.
I have made a PR for Pointhi's repo a little over a week ago
https://github.com/pointhi/kicad-footprint-generator/pull/72

terjeio:
@Shackmeister your changes has now been merged and I am digging into the latest version.
So far I have found a nasty "bug" that took a while to pin down.

from the scripts - this code gives incorrect results:
``` python
pad=[2.75, 1]
slk_offset_pad = pad[1]/2+slk_offset+min_pad_distance
```
this is ok:
``` python
pad=[2.75, 1.0]
slk_offset_pad = pad[1]/2+slk_offset+min_pad_distance
```
Python is more type aware that I thought:

``` python
>>> print 1/2.0
0.5
>>> print 1/2
0
>>> print 1.0/2
0.5
```
There are many *potential* integer by integer divisions like that in footprint_script_pin_headers.py and I do not want to touch that code too much as it is shared with the pin headers, it could be that this behaviour is deliberately relied upon (or coded around?). This could be dangerous though since it is dependent on input data type, a better practice is either force a float division by using a float divisor or, if the intention is to do an integer division, ensure the dividend is an integer. 

I will "fix" this by ensuring input data that can be affected is of type float.

poeschlr:
use python3 or `from __future__ import division` to get the same behavior in python 2.x
http://www.informit.com/articles/article.aspx?p=1439189

PR #6, opened 2017-11-05
Added DSub connectors

SchrodingersGat:
Same again. Moved DSub connectors across.


poeschlr:
Seems to be ok. Thanks

PR #7, opened 2017-11-05
Terminal blocks

SchrodingersGat:
This PR adds the terminal blocks from:

* 4UCON
* Philmore
* Phoenix
* WAGO

The naming convention has been fixed for each group.

The following terminal blocks will be added later, there are issues with the naming convention:

* MetzConnect
* RND


poeschlr:
should we wait for this to be fixed:
https://github.com/KiCad/TerminalBlock_WAGO.pretty/issues/2

poeschlr:
Also the PTSM footprints that currently are located in the connectors_phoenix repo should be moved into the terminal block phoenix lib.

poeschlr:
You use `Straight` as orientation. In the KLC we defined that Horizontal or Vertical should be used instead of Angled or Straight.
(For terminal blocks the direction the cable enters the connector determines what should be used.)

SchrodingersGat:
Good catch, I should not have used `Straight`.

Are there any terminal blocks where the wire enters vertically? In context, perhaps the orientation is not required? Especially given the series defines the orientation.

poeschlr:
Yes there are terminal headers where it enters vertically:
[Phoenix 1868128](https://www.phoenixcontact.com/online/portal/gb/pxc/product_detail_page/!ut/p/z1/3VdLb-IwEP4r7IFbjR3ngbvSHiBow0NtWQEFcolsx0Fe8qAhQJdfv04CKtuqoBXi0EQjxR7PfPPKRGPowhl0Y7qVC5rJJKah2s9dyyN4YHcdQ3t46v8yUevZ6U400h32BxZ8hi50M8rgXMtXoWQpTf_A-YKJOGespK-OiEU0TPJ9EgxtewDnQ6BpACkyCwCZhUIx7XYtE2kklekaCxO-rIHaw6Az-jmtaXdmHdUw0O9MxTwiTs_5p-sWdM-7X-qjT54WuqQ_V_rNTwVGBhwVSfCDkMaLDV2oIAtGspZ5hstdKriQW5Eezjhfvs_Px7Smib_hOcRYvGal4lsVlL0o8Y-23lXgoNoRGZXhupeJqHcicCEgq2oBmVULqFmxgHDVeghXrYdw1XoIV62HelXroV7VPrle1X4KanSqWED4ywWkBDaphPPVKwcJB1Id-4Xkd4Xz4yjVh65kUWPHowZqKNHGAdQrZb1VHqm6B8jfLy9uC7o8ibPcz1nufx3tBKujiEq1PCiu6yiU66xQVGs11HvHod4rhvq195aROrLufQ1z6gPmMwGMwGSAMY0BbDVZE-sGEff5nO6eG8WLSb8QODPKj0TsOW1vOLOd9oViO0043Uqxg5M4SSN1DypLz_P0M8JMw6QEBAa2gGHoPqCUYaCSSZuYEJ9iCrvoggViXGnhArx1U_in28Ibt02OYd4UfnCt9_2zt06kX9ONK84-dOPqf7txFUVEzwlMZvuxYPsTIkg_HJbkPT6a21P6Agr7IJqCZdBpPwLHZmQ3DqJ_X9PWt7_KttXa/dz/d5/L2dBISEvZ0FBIS9nQSEh/#Z7_82KCHG41MOJQ50AVGHU18HPJA2)

SchrodingersGat:
Ok, great :) I will fix these if Jan doesn't get there first :)

antoniovazquezblanco:
Does Travis check the names of the components? Do you think that naming check could be partially automated with regex or similar?

poeschlr:
The naming convention is quite complex and not yet fully spelled out. I don't think full automation can happen before we have a specific rule for every type of component. (If it can happen at all) Then the question arises how to determine what component is in the pull request.

But stuff like warning about the use of Straight/Angled could already be implemented. (Do you have time to do this?)
Maybe some similar checks for other parts of the rule set could be devised.

jkriege2:
So we should just decide who does it ... my scripts are not yet merged into @pointhi 's repo ... I would propose to work on all these issues ... maybe best would be to leave the old repos alone and merge the renamed stuff into the new v5-repo only? 

If that's fine with everybody, I will start with the RND-connectors to get naming and everything straight and then start doing the rest based on that. OK? ... but give it 1-2 weeks until everything is completed ;-)

JAN

poeschlr:
I would agree with declaring the old repos as legacy and only updating the kicad-footprint repo.
This workflow does not break the installations of users that use the github plugin.
(And it does reduce our workload.)
Simply mark of the current libs as legacy once you start with your work. (so that there is no interference with your work by somebody else.)

antoniovazquezblanco:
@poeschlr I am willing to contribute to this but I am struggling to find the time for it. For the time being I will open an issue in the utils repo and will notify if start working on it.

jkriege2:
I added PRs for RND and Metz as a test baloon:
- https://github.com/KiCad/kicad-footprints/pull/10
- https://github.com/KiCad/kicad-footprints/pull/11

Please review, so I can start redoing the others (but might take some time)

SchrodingersGat:
Closing this PR as Jan is fixing the scripts :)

PR #9, opened 2017-11-07
Adding script generated phoenix contact connector footprints.

poeschlr:
This adds the script generated phoenix contact MSTB and MC series connector footprints.
I split the lib into two. (one per series) We can also think about including the 1,5 and 2,5 part of the series name. (would mean MC_0,5 would get it's own lib.)

I included the full human readable part names. (Example MSTB_2,5_2-G for the MSTB_2,5/2-G)
I did not include the order number in the name because some of these footprints are valid for multiple order numbers. (High current version)

script pull request: https://github.com/pointhi/kicad-footprint-generator/pull/78

The remaining footprints in the connectors_phoenix repo are terminal blocks.


SchrodingersGat:
If we are grouping TerminalBlock footprints separately to "Connectors" should we not prefix these with `TerminalBlock` instead of `Conn`?

poeschlr:
This pull request does not include any terminal block footprints.
There are however some terminal block footprints in the current Connectors_Phoenix lib.

SchrodingersGat:
Ah I see, I guess it's a valid distinction that these are "pluggable" connectors.

Are these footprints completely new, or some copied from the old repo? Or are the scripts simply updated?

poeschlr:
I updated the scripts. (All of these connectors have been in the old repo. But with slightly different names. And there was only one large lib instead of two libs.)

First i transferred some parts to a yaml script and tested if the new version does produce equivalent footprints. Then i changed the format strings that determine the footprint and library names.

Why is ',' illegal in footprint names. Is there any operating system that has problems with this char in a file name? (I can replace it easily with any other char. I already replaced spaces and '/')

SchrodingersGat:
We should update the KLC to allow `,` and `+` characters. Surprised we haven't run into this before.

SchrodingersGat:
Nice work, thanks for doing this.

PR #10, opened 2017-11-07
added regenerated terminal blocks from RND

jkriege2:
Please check whether naming is OK ... these are the same footprints that are already in the legacy lib, but with naming conforming to KLC 3 (I hope)


SchrodingersGat:
Naming looks good, thanks Jan :)

PR #15, opened 2017-11-08
Housings_xxx -> Package_xxx

SchrodingersGat:


SchrodingersGat:
KLC issues have not been addressed. Naming conventions have.

jkriege2:
somehow naming for BGAs is a bit inconsistent with the rest: everywhere `_2x2mm` is the size of the package, but for BGAs we use `_B2x2mm`... possibly to distinguish from `_10x10`, which is the number of balls ...

SchrodingersGat:
Hmm this convention is one of the few ones that I have been on the fence about. I would like the body size to *always* be the parameter that takes preference. Instead we have it *sometimes* with the `B` prefix.

Perhaps we could have

`BGA-12_4x3mm_Layout6x2_P0.5mm` 

for example...

Still not sure about this one.

Thoughts? @jkriege2 @poeschlr

jkriege2:
sounds like a good solution!

JAN

poeschlr:
Good for me as well.

evanshultz:
While I can't find the old IR-DirectFET repo, I believe it was a conversation there where we discussed using `Package` instead of `Housing`. I provided a number of links which showed "package" is the most common term used in the industry.

And on that subject, I believe the `IR` should be removed from the repo name too. International Rectifier was purchased by Infineon, so that company name doesn't exist anymore, and DirectFET was never a proprietary package. I say just call it `Package_DirectFET`.

IPC-735x is pretty clear about BGA footprint names. Are we not going to go with that? Is there more IPC material?

SchrodingersGat:
@evanshultz we wish to retain clarity more than anything else. I feel that the leads we have taken from the Library Expert naming convention fall down in this case. In every other case we only have body parameter without a prefix, I think it is important to be consistent.

evanshultz:
You're the boss, boss.

There are a number of remaining KLC errors which seem like low-hanging fruit:
- Missing 3D models
- No F.Fab ref des
- Wrong 3D model directory (still `Housings` or some other error)
- Courtyard off grid
- Missing `${KISYS3DMOD}/` to start model path
- No keywords
- Some footprints have `Housing` in the footprint name
- Footprint off center
- `SOT89` vs `SOT-89` (missing hyphen)
- Allegro_CB* footprint should be removed (they live in a new folder)
- Both SMD and THT placements selected

To me, these types of errors are egregious enough (high severity) or easy enough (check script can fix it) to correct that they should be fixed now. @SchrodingersGat Do you want to do any or all of these now?

SchrodingersGat:
I don't think that the errors are critical, anyone happy to merge?

jkriege2:
I think we should fix as many errors as possible, while moving footprints here, so we get really good quality in the v5 repos! ... could you at least use a --fix-run of our utils to fill in missing 3D models and off-grid courtyards?

poeschlr:
Fixing too much might overstrech us. We need to manage our time. Otherwise we will not be finished in time for the v5 release. Just remember we have also decided to reorganize the symbol libs.

The most important part for now is to rename all footprints according to the new scheme. We can then fix them later without users ever noticing anything. (As long as we do not rename them.)

jkriege2:
I looked at all hole sizes of the RNDs again and set them by the size of the pin, as the "recommended" PCB hole size vary strongly and do not always seem to be connected to the actual measures (typos???), also sometime they recommend a different hole-size than shown in the footprint-sketch in the same DS ... so rationale now is:
- 1mm pin  --> 1.3mm hole
- 0.7x0.5mm pin --> 1.1mm hole

other fixes in RNDs:
- fixed height of RND205-00287... from 10.16mm to 10.6mm
- fixed pin-size of RND_205-00045...
- added 45°-series RND_205-00056...
- added vertical series RND_205-00276
- all other changes are due to changes in keepout-areas around pads (see pad sizes above + algorithm-modifications)
- removed two that do not exist and overlap with other numbers (205-00241/242 with 5.08mm pitch does not exist)
- changed screw size of RND_205-00287...

jkriege2:
yes, but a few of the things @evanshultz mentioned should be changed (and should be doable with our scripts):
- ${KISYS3DMOD}/
- missing 3D models (I think --fixmore does that, just take care it does not "overfix" _HandSoldering-variants and stuff!)
- SOT89 would be great and removing the Housing in names (that's renaming inthe end) ...
- I think Courtyard off gridis fixed by scripts too
- No F.Fab ref des is also done automatically by script

SchrodingersGat:
@jkriege2 some of those are attainable without too much effort - I'll take a look tomorrow.

jkriege2:
That would be great!

evanshultz:
Thanks all. This is what I was saying: let's do some easy fixes now leveraging the check scripts.

I'll submit a PR for https://github.com/KiCad/Housings_DFN_QFN.pretty/pull/74 once this is merged.

poeschlr:
Just a note such that we do not forget it:
It seems none of these repos have been declared legacy. They still have open pull requests. As soon as we merge this we should close the open pull requests and ask the contributors to re submit their contributions here.
And we should add a a note to the readme.

SchrodingersGat:
Lots of KLC fixes have been made. I have also submitted some PRs to the library-utils scripts to improve the fixing process.

evanshultz:
Much better! Thank you!

Here are what I still see that I still believe should be addressed now:
- Missing 3D model
- Footprint not centered
- No keywords
- Still some "SOT89" instead of "SOT-89"
- Allegro_CB* footprints must be removed since they're now in Sensor_Current
- Remove `//home/travis/build/KiCad/kicad-footprints/Package_SOIC.pretty/library-check.sh`

I thought the check scripts could fix some of this, at least the centering, but others (SOT89 and Allegro_CB) are hand-changes that should be made.

@SchrodingersGat Can you address any more of these items before merging?

Also, the RoHM part has ref des too small on F.Fab. This should be a KLC change, I think, since if the layer is printed in a vector format there shouldn't be a limit. It is currently an error but should be a warning. It would be a valid error in silk, but not on the fab drawing.

SchrodingersGat:
More fixes now.

The "footprint not centered" errors are OK for the remaining footprints, I believe. The KLC checks aren't 100% valid in all cases, especially with asymmetric pads.

There are also a couple of keywords errors which I think are false - I checked and there are actually keywords present.

SchrodingersGat:
I'd like to mark all the Housings_xxx libs as deprecated to move all the active PRs over to here. Any further serious objections here? The quality of these libs is now markedly improved

evanshultz:
@SchrodingersGat Thanks again! I'll merge now.

A few notes:
1. Some of these have `P??mm` and some have `Pitch??mm`. Since this is a widespread issue due to new KLC changes I don't think it needs to be dealt with before merging.
1. Perhaps we should have a standardized list of known-good/approved footprint suffix options. For example, what does '_SmallFlag` mean? Or `_BigPads`? KLC could include pictures as examples for later referencing.
1. We could make the check scripts ignore footprint suffixes, so that `_ThermalVias` or `_NoMask` don't cause a warning. Or perhaps only allow some suffixes that we know are OK. Or... since a warning isn't an error, maybe it's OK just like is it?

Please also note the ref des F.Fab text size item noted above.

Lastly, I'm also good with deprecating all the old libs.

SchrodingersGat:
@evanshultz yes based on your notes above we do need to make some KLC adjustments :)

PR #16, opened 2017-11-08
Buttons / Switches

SchrodingersGat:
Simple copy of the buttons and switches libraries. No KLC checking


evanshultz:
Nothing too serious, and easy fixes we can catch later. Most footprints are pretty good so I'll merge.

PR #19, opened 2017-11-08
RF

SchrodingersGat:
Copy:

* RF_Module(s)
* RF_Annena(s)


evanshultz:
Some of these are also pretty bad, and could be fixed relatively easily. Please confirm if you want to merge as-is.

SchrodingersGat:
I have made some KLC fixes but the rest are spurious errors, based on the calculation of center point. I think they are fine.

evanshultz:
Got it. The few remaining errors seem minimal or bogus. Thanks!

PR #21, opened 2017-11-08
Added Buzzer_Beeper

SchrodingersGat:
Direct library copy.

Edit: Now with KLC fixes


evanshultz:
Assuming this is good now with your fixes above and Travis not finding any errors.

PR #22, opened 2017-11-08
Added the Bourns SRN6028 Series

johnfwhitmore:
Previously added to read only library.

https://www.bourns.com/data/global/pdfs/SRN6028.pdf

screenshot from 2017-11-07 15-59-55


SchrodingersGat:
Nice work, thanks :)

PR #23, opened 2017-11-08
Transformer

SchrodingersGat:
Copied transfomers


evanshultz:
Looks like the last commit fixed them all up, even though Travis is still complaining about 3D model path errors.

PR #26, opened 2017-11-09
Adding JST PH series connectors

poeschlr:
As with GH connectors, adapted the generator scripts for the new KLC. (The script accepts the same config file as the GH connectors.)
No connector dimensions have been changed. (Graphical style is also not changed.)


SchrodingersGat:
Great!

PR #30, opened 2017-11-09
Add new Allegro current sensor footprints

evanshultz:
These are the footprints needed for the library update at https://github.com/KiCad/kicad-library/pull/1817.

Datasheet links:
- PSOF: http://www.allegromicro.com/~/media/Files/Datasheets/ACS780-Datasheet.ashx
- QFN: http://www.allegromicro.com/~/media/Files/Datasheets/ACS711-Datasheet.ashx
- QSOP: http://www.allegromicro.com/~/media/Files/Datasheets/ACS726-Datasheet.ashx
- SIP-4: http://www.allegromicro.com/~/media/Files/Datasheets/A1363-Datasheet.ashx
- SIP-3: http://www.allegromicro.com/~/media/Files/Datasheets/A1369-Datasheet.ashx

Image:
image


evanshultz:
A few fixes:
- Make pad 1 of the SIP footprints square
- Fixed line endings
- The QSOP footprints is using the generic 3D model in Package_SSOP

Image:
image

SchrodingersGat:
Only one KLC error to fix then good to merge :)

evanshultz:
@SchrodingersGat 
Thanks, but I don't see this as an error. I'm using the standard QSOP-24 3D model since the footprint differs but not the physical device package. That should be OK. See above; maybe I wasn't clear?

SchrodingersGat:
Sorry I missed your note above. Good job then, thanks :)

PR #31, opened 2017-11-10
Add IDC headers

fauxpark:
This PR adds the IDC headers from [Connectors_IDC.pretty](https://github.com/KiCad/Connectors_IDC.pretty).

I took the opportunity to tweak the descriptions and tags, borrowing from the pin headers again, and added the word "box" as they are also commonly known as box headers.


SchrodingersGat:
Great work @fauxpark  :)

Just one note: We are moving to use `P2.54mm` instead of `Pitch2.54mm` in the new convention - can you change the filenames to match this?

jkriege2:
Also it should be Horizontal instead of ANgled and Vertical instead of Straight

Best,
JAN

SchrodingersGat:
Good catch @jkriege2 I'm still not used to that one :)

fauxpark:
Done and done. I haven't had the chance to investigate the new KLC yet :P

SchrodingersGat:
Great @fauxpark thanks for the contribution :)

PR #35, opened 2017-11-11
Add reworked screwterminals

jkriege2:
OK, I reworked all screw terminals (and added a few derivatives of the existing families, with other pin pitches)

scripts available here: https://github.com/pointhi/kicad-footprint-generator/pull/68

2017-11-11 18_23_20-pcbnew d__kicad_kicad_review-project2_kicad_review-project2 kicad_pcb


poeschlr:
### 4ucon horizontal terminal block
Sides should be 2.1mm from the outermost pins. (at least according to the datasheet http://www.4uconnector.com/online/object/4udrawing/10691.pdf)
screenshot from 2017-11-11 22-22-38

### 4ucon vertical terminal block
Sides should be 2.25mm from the outermost pins. (at least according to the datasheet http://www.4uconnector.com/online/object/4udrawing/10697.pdf)

screenshot from 2017-11-11 22-32-02

poeschlr:
# General questions:
Should we indicate the cable entry direction for horizontal types?

# MetzConnect
### Type011

Silkscreen looks a bit strange:
screenshot from 2017-11-11 22-38-24

### Type086
Distance between side and outer pins should be 1.85 (http://www.metz-connect.com/de/system/files/productfiles/Datenblatt_310861_RT034xxHBLC_OFF-026114K.pdf)
screenshot from 2017-11-11 23-46-45


### Type094
Body width should be 10.5mm (http://www.metz-connect.com/de/system/files/productfiles/Datenblatt_310111_RT055xxHBLC_OFF-022717S.pdf)
screenshot from 2017-11-11 23-46-45

### Type171
Drill diameter should be 1.3mm (http://www.metz-connect.com/de/system/files/productfiles/Datenblatt_311711_RT137xxHBWC_OFF-022811Q.pdf)

## Measurements checked for types:
- [x] Type011
- [x] Type055
- [x] Type059
- [x] Type067
- [ ] Type073 
   - Can't import in FreeCad (Stepup throws an exception.)
- [ ] Type086 
   - Wrong body length. See above
- [ ] Type094
   - Wrong body width. See above
- [x] Type101
- [ ] Type171
    - drill diameter too large
- [x] Type175
- [x] Type701
- [x] Type703

jkriege2:
I'll fix these ... about the strange silkscreen: That's caused by my method of keeping the pads clear: The pads are approximated by a series of rectangle and each line is intersected with such a rectangle ... when it's inside, the line is not drawn ... I'll see whether I can tweak the parameters here a bit (possibly simply making the screw a bit larger)!
2017-11-12 09_51_07-footprint editor no active library

Best,
JAN

jkriege2:
About the others:
1. Meth Type 011: screw is a bit larger now, so no strange effects on screw circle
2. Metz Type 086 is actually correct. The package is 7.3mm and so is the F.Fab-drawing.
2017-11-12 10_03_58-datenblatt_310861_rt034xxhblc_off-026114k pdf geschutzt - adobe acrobat reader
3. Metz 094: I linked the wrong datasheet (correct one is http://www.metz-connect.com/ru/system/files/productfiles/Data_sheet_310941_RT035xxHBLU_OFF-022742T.pdf .... as the footprint name also indicates). The footprint has a eight of 8.3mm (F.Fab) and so does the package
4. Metz 171: fixed

Thanks for spotting all these,
JAN

jkriege2:
added new metz blocks with 45Degree entry:
2017-11-12 10_29_35-footprint editor no active library

poeschlr:
# Round two (It was too late yesterday to finish looking at everything)
It might be a good idea to add a readme to the libs and point to the script there instead of including the complete link in the documentation field. (Long documentation field are hard to read in kicad.)
As an inspiration have a look at my recently merged molex connectors.

## Philmore
everything ok

## Phoenix Contact
Naming convention of the part name: In my Phoenix Connectors i replaced both the space and / with underline instead of minus. (Phoenix does use the minus to distinguish additional naming fields. Like in the 5.08 pitch version MKDS 1,5/xx-5,08)
It is also allowed to include the comma for 1,5. (Phoenix does use the comma internationally here.)

I would suggest that either you change your namings or i should change mine such that we use the same.

I would at least include the name without replacements in the description instead of repeating the information already contained in the footprint name.


## PT series:
Screw graphics look a bit too large:
screenshot from 2017-11-12 10-28-31


### Measurements checked for:
 - [x] MKDS 1,5/xx
 - [x] MKDS 3/xx
    - The footprint is for the 5.08 version. Correct name: MKDS 3/xx-5,08 (`MKDS_3_xx-5,08`)
 - [ ] MPT 0,5/xx
    - Again this is the 2,54 version. MPT 0,5/xx-2,54 (It seems a 2.5mm version does not exist for the mpt series.)
    - Side distance to outer pads should be 1.5mm (currently is 1.27mm). [See dimensional drawing from phoenix webside](https://www.phoenixcontact.com/assets/images_pr/product_drawings/large/22069_1000_int_04.jpg)
 - [x] PT 1,5/xx 
     - Correct naming: PT 1,5/xx-3,5-H

jkriege2:
About naming: I would suggest to use a minus `-` instead of '/' and stuff in the names, as then it is different from our usual name-part-separator underscore '_' and the MPN is perceived as one entity. I will use , instead of . though!

jkriege2:
reworked RND a bit and added a 45degree and a vertical variant
2017-11-12 10_48_12-pcbnew d__kicad_kicad_review-project2_kicad_review-project2 kicad_pcb

jkriege2:
reworked Phoenix footprints ... naming OK like this? ... please recheck (Thanks!)

2017-11-12 11_24_19-pcbnew d__kicad_kicad_review-project2_kicad_review-project2 kicad_pcb

jkriege2:
PS: READMEs are all there (and have been)

poeschlr:
## RND
(The part numbers by RND make it really hard to check this.)
Please re check the drill diameter for all of these.

TerminalBlock_RND_205-00242_1x03_P10.16mm_Horizontal links a datasheet for 00023 to 00033

Checked measurements:
- [ ] 00001 to 00011 (P=5mm)
   - according to [linked datasheet](http://cdn-reichelt.de/documents/datenblatt/C151/RND_205-00001_DB_EN.pdf), width should be 9mm not 9.8mm
- [ ] 00012 to 00022 (P=5mm)
  - same lead diameter as 01 to 011 but smaller drill?
- [x] 00023 to 00023 (P=10mm)
- [x] 00024 to 00033 (P=10)
- [ ] 00045 to 00055 (P=5)
    - drill diameter should be between 1 an 1.1mm (see datasheet)
- [x] 00067 to 00077 (P=7.5)
- [x] 00078 to 00088 (P=10)
- [x] 00232 to 00242 (P=5.08)
    - drill should be 1 o 1.1
- [ ] 00287 to 00297
   - width should be 10.6 (is 10.16)
- [x] 00298 to 00308

I noticed you changed something in RND. Please give a summary about it.

SchrodingersGat:
> I'll fix these ... about the strange silkscreen: That's caused by my method of keeping the pads clear: The pads are approximated by a series of rectangle and each line is intersected with such a rectangle ... when it's inside, the line is not drawn ... I'll see whether I can tweak the parameters here a bit (possibly simply making the screw a bit larger)!

Why can't you just have a simple circle? These multiple lines are just going to increase file size. What are you trying to solve by "keeping the pads clear"?

jkriege2:
@SchrodingersGat The idea is to have a system that allows you to simply draw circles, but have an automatism in the back that transfroms it into a series of arcs that leave out areas near to the pads, so silkscreen does not overlay the pads (+clearance) ... the algorithm goes a bit wild, when the circles are close to a circular pad (which is approximated by a series of small rectangles), but this is solved not. Typically a circle around a square pad is broken down into 4-5 arcs only ... if there are no obstacles, a simple circle is drawn 
2017-11-12 12_34_26-footprint editor no active library

jkriege2:
there are KLC-violations now, which are due to using a comma in the filename .... but as discussed above that should be fine.

I will now recheck the RNDs

poeschlr:
Strange i thought this merged pull request should have fixed that: https://github.com/KiCad/kicad-library-utils/pull/167

jkriege2:
@poeschlr Thanks for looking through them ... and yes, naming is a mess ;-)

## about comments above:
1. 00001 to 00011: height IS 9mm ... where did you measure 9.8mm????
2. 00012 to 00022: see below
3. 00045 to 00055: see below
4. 00287 to 00297: height fixed (typo in script)

## hole sizes
I looked at all hole sizes of the RNDs again and set them by the size of the pin, as the "recommended" PCB hole size vary strongly and do not always seem to be connected to the actual measures (typos???), also sometime they recommend a different hole-size than shown in the footprint-sketch in the same DS ... so rationale now is:
- 1mm pin  --> 1.3mm hole
- 0.7x0.5mm pin --> 1.1mm hole

## other fixes in RNDs:
- fixed height of RND205-00287... from 10.16mm to 10.6mm
- fixed pin-size of RND_205-00045...
- added 45°-series RND_205-00056...
- added vertical series RND_205-00276
- all other changes are due to changes in keepout-areas around pads (see pad sizes above + algorithm-modifications)
- removed two that do not exist and overlap with other numbers (205-00241/242 with 5.08mm pitch does not exist)
- changed screw size of RND_205-00287...

poeschlr:
You are right about 01...10. I might have loaded the wrong footprint into freecad

poeschlr:
Could you also take a look at the PTSM footprints that currently are located in the connectors_phoenix repo? These should be included in the TerminalBlock_Phoenix lib instead of the Conn_Phoenix lib. (This is why i did not include them in my pull requests that transferred all other footprints in this old lib.)

jkriege2:
can do ...

jkriege2:
... about the PTSM: will do, but in a separate PR ... as it will take me some time!

jkriege2:
@poeschlr can this PR be merged?

jkriege2:
Hi @poeschlr I also took a look at the PTSMs ... and the moved/renamed/fixed the old ones + added a few of the THT-variants that were so far missing, see https://github.com/KiCad/kicad-footprints/pull/37

poeschlr:
About Metz connect type 86: i added the wrong screenshot. (The text above the screenshot says that the lenght is wrong. The width is correct as you remarked in your comment.)
screenshot from 2017-11-13 10-50-55

Body lenght Phoenix MPT is still wrong:
screenshot from 2017-11-13 10-55-33

evanshultz:
Regarding the silkscreen conversation at https://github.com/KiCad/kicad-footprints/pull/35#issuecomment-343730729, why not look at pad shape? If the pad is a circle, just draw a circle. If the pad is square, draw the arcs (at least on silk, I suppose fab could still be a circle?). That way the footprint file size doesn't grow so much with all the tiny arcs and it allows targeted silkscreen where needed. Each pad shape would need to be supported separately, but that doesn't seem too much of a burden to me when there is a generic solution (as you've shown above) available.

jkriege2:
@evanshultz That's more or less what's happening. 
1. All this only applies to F.SilkS
2. In the screenshot above, the circular pad is surrounded by a circle, whereas the circle around the square pad is drawn in sections. The nice things about my scripts is that I don't have to code that explicitly. I simply have a list keepouts, which contains all areas where therer should not be silkscreen (e.g. pad-rectangle + a bit of free space). Then special draw commands check against this list: If a line/arc/circle... is outside it is simply draw, if it intersects, it is split into parts. In the screenshot above, e.g. the circle is broken into 5 visible arcs + 4 arcs that would intersect with the keepout-areas, thus they are not drawn. The problem is that I implemented this for rectangles only (didn't want to do all the math for circles and ellipses ... so what I do is approximate the circle as a set of rectangle, see my screenshot with the rectangle all at the top, this costs a bit more time and is not as accurate, as the mathematically exact solution (and leads to artifacts as shown, when sircles are very close to the pad), but works reasonably well without too much effort. Bottomline: NOte that the rectangles I'm talking about only exist in the software ... they never end up in the output (unless for debugging!). Check the filesizes ... they are reasonably small now!

The nice thing about this approach: I only have to declare what I want (a circle/line/rectangle/screw..., with keepouts cut out) and I get it ... withut any additional fuss.

jkriege2:
@poeschlr I will have a look at the measures later today or tomorrow morning!

Best,
JAN

jkriege2:
Hi @poeschlr !

I fixed the Type086 (thanks for spotting), but the MPTs are correct, see datasheet:

2017-11-13 21_04_48-itemdetail_1725656-920552 pdf - adobe acrobat reader dc

Best,
JAN

poeschlr:
Thanks for this large contribution.

poeschlr:
@jkriege2 is there a reason why not all philmore footprints that where in the old lib have been included here? (You only included the 2 and 3 pin version, the old lib had more. Noticed this when i wanted to update the ToDo file.)

jkriege2:
yes, I could only find datasheets (or catalog pages) for the 2- and 3-pin version. Therefore I skipped the rest.

Best,
JAN

poeschlr:
Then all terminal blocks have been transferred. Excellent i will update the Todo

PR #38, opened 2017-11-13
Add Matias/ALPS keyswitches

fauxpark:
Done in the style of the Cherry MX footprints. ALPS has stopped production on these switches, but Matias makes a clone that is well known in the mechanical keyboard community.

Some notes:
- There seems to be no proper datasheet or model number for these switches, but the "blueprints" can be found in the doc link provided
- These switches are fully plate mount, there are no mounting holes for the switch or the stabilisers
- The stem of Cherry switches is a + shape, the stem of Matias/ALPS is a - shape, so it matters which way the switch is oriented, especially for tall keys such as the numpad + and enter, and the ISO enter key

fp
3d


SchrodingersGat:
Nice work, thanks for the contribution :)

PR #41, opened 2017-11-15
Updated and new pin sockets

terjeio:
Added single row SMD for  2.00 and 2.54mm. Removed 3D model offsets and
rotation so many legacy models can no longer be used.

SMD dimensions are from legacy footprints, there are many variants
available and some does not match the legacy land pattern. Perhaps at
least two different land patterns should be made for each?

Many overlapping line segments has been removed and small line segments has been joined when possible, resulting in more compact files.

Examples:

examples

PR for script here [https://github.com/pointhi/kicad-footprint-generator/pull/82](https://github.com/pointhi/kicad-footprint-generator/pull/82)

More variants can be added later such as SMD with guide pins or bottom entry and THT single row with two pins per socket. More work needs to be done collecting and analyzing dimensions for those beforehand.

Scripted 3D models for all variants has been made and a PR for these will be made when this PR is eventually accepted.


Shackmeister:
@terjeio this is mighty fine work! Since Im on my phone I cant so a proper check but I should hopefully have some time tomorrow.
Only thing I noticed (when looking at the screenshot on my phone):
Silkscreen looks like it's on top of fab layer on quite s few

terjeio:
@Shackmeister Wrong impression is due to the low resolution of the screen dump. Here is the project used to create it:
[examples.zip](https://github.com/KiCad/kicad-footprints/files/1477538/examples.zip)

Shackmeister:
@terjeio got some time to look at the footprints
SMD footprints seems to be spot on :) nice!

with the THT footprints there is a general problem of the silkscreen crossing the 0.2mm soldermask clearance. Should I elaborate or add some photos?

PinSocket_1x02_Px.yymm_Vertical_SMD_Pin1Left
PinSocket_1x02_Px.yymm_Vertical_SMD_Pin1Right
is missing for all of the conns

also can you please add these 1.27mm 1xyy connectors? :)
https://gct.co/connector/bd075

perhaps also the 1.0mm sockets if possible :)
https://gct.co/board-to-board-connector/list?pitch=1.27mm%2C1.00mm

please let me know If I can be helpful somehow

terjeio:
@Shackmeister I think I've got it - it was surely not the easiest footprints to start with due to the large range and, for me, largely unreadable code as a starting point. I see now that I should have checked your updated pin headers in more detail too, and not relied too much on the original pin sockets for reference.

As for the new footprints you refer to, should I just add them "as is"? That would be easy to do.
However, what I have found is that there are quite a bit of different land patterns for same pitch sockets.
Eg. I have a 2.54mm socket on my desk that is close to 10mm across (pin tip to pin tip) and thus does not fit the one in the library.

I was thinking of doing a bit more research later to try to come up with land patterns that possibly could match a range of variants, this since I've seen some parts that has the overall width specified as minimum. Also, body width vary a bit too, not to mention height (for the 3D models). Maybe at some point KiCad has to support alternative 3D models for generic footprints like this?

Shackmeister:
I was thinking of doing a bit more research later to try to come up with land patterns that possibly could match a range of variants, this since I've seen some parts that has the overall width specified as minimum. Also, body width vary a bit too, not to mention height (for the 3D models). Maybe at some point KiCad has to support alternative 3D models for generic footprints like this?
@terjeio I totally agree. when I started with the pinheaders I was planning on the exact same thing as you. fixing the 3D scaling and offset, fixing a missing REF** etc. turned out to be pandoras box to make a bulletproof(ish) script which could ensure the silkscreen wasnt overlapping. This is also the reason why the script became so unreasonably big. Perhaps it can be cut down (and be more readable) with some more objectoriented, but I couldnt find a smooth way (but I am a relatively lousy programmer, so somebody probably can)

for the parameters of the footprints I went with some specific footprints which seemed reasonably generic, used the same ones for 3D models eventhough the height is gonna be different all the time.

I think even mentioning different heights for the pinheader 3D models is disallowed untill the size handling has been adressed properly ;)

terjeio:
@Shackmeister 
I have encapsulated some the core framework code in a few classes already, and the script I made is based on that.
I want to add keepouts in a similar vein too, see [this comment](https://github.com/pointhi/kicad-footprint-generator/pull/82#issuecomment-345462900) .

You did not answer my question about adding the new footprints "as is", or did you?

Shackmeister:
did you use the parameters from the "old" footprints? in that case I would suggest to check them against some common connectors. regarding the ones I linked to then yes you can add them as they are

terjeio:
yes and yes I did. However a more systematic check should be done as I suspect it is not possible to make "one size fits all" footprints. It could easily add up to a few thousand variants if coverage is to be complete? Guide pegs, bottom entry and what not - I do not know what you want supported.

Since my last comment I have been tinkering with a wrapper class for keepout code, so far I have managed to add keepout zones for all the pads with a single line of code in the generator script :-) I have to add code in my core wrapper class for the drawing primitives but that should be fairly easy. IMO the framework provided by pointhi is brilliant and easy to extend.

terjeio:
@Shackmeister Sorry to bother you again.

I have made good progress in adding transparent keepout processing and I am currently reimplementing the solution in drawing_tools.py as I found it lacking i how it handled detail. It is based on rectangles to specify the keepout, resulting approximate results for circles and for rectangular features near the corners. I have changed that to a single rectangle (bounding box) for initial hit/miss processing, if a hit then furter processing is done to resolve the exact point to use based on a rounded rectangle for rectangular features and a circle for circular.

This example image shows the rectangular  bounding box and the keepout area inside and how lines get chopped:

bilde

Oval pads should be easy to add support for, maybe triangular too.

A couple of questions then:

Is there any convention on how short silk lines can be before they should be discarded? I cannot find any reference in the KLC for that.

I ran the KLC validation scripts on the footprints before I made the PR and no errors related to the silkscreen was given. I assume this means the keepout rule is not checked for. I have not looked into those scripts yet but I believe the keepout class I am working on could be utilized to automate checks for at least some (most?) cases related to pads. The tricky ones could be rotated (other than right angles) pads but I assume those are not often found. Is this something that should be looked into?

Terje

evanshultz:
@terjeio 
Looks great! Answers:
1. KLC does not specify silk line length. Width is 0.12mm, which is not a problem to manufacture these days. Line length would then also have to be 0.12mm, resulting in a circle. So try it and let's see how it looks, though my gut says about 0.3mm will probably be the shortest reasonable length that results in an obvious line on an actual PCB. It's a good point to sort this out and add it to KLC!
1. Silk-to-pad keepouts are checked, but in a very basic way and not in all cases (pad shapes, angles, etc.). This would be a great place to bolt in a more robust and comprehensive solution!

If you do submit a PR to replace that chunk of keepout detection code, please provide good before-and-after samples in the PR body for clarity. Thanks!

terjeio:
@evanshultz Thanks for your reply.

A little nitpicking: a line length of 0 results in a circle (have not checked though, it might be skipped), line length of length > 0 results in an oval shape. So maybe you really mean 0.18mm when you write 0.3mm, this since the overall length of a 0.18mm line of width 0.12mm is 0.3mm. Or prehaps 0.18 if the fabricators render the lines with butt ends instead of rounded?

I am not intending to update the original keepout code as my version is wrapped in a class. It can be used standalone but when used in combination with the Layer class I am working on it will perform the best: instantiate the class, bind it to the layer, define keepout areas and draw using the normal drawing methods provided by the Layer class (as wrapper methods for the base classes in the framework).

My plan is to add the new classes in a separate file (canvas.py) with the footprint generator code I submit, until it is of such quality that it can be added to the tools or framework proper.

I will look into the KLC validation code later, the submitted footprints has lots of keepout violations (more or less the same as the originals) that did not get reported so surely there is room for improvement.

Shackmeister:
@terjeio no such thing as bothering! :)

In my script I just used pythagoras to calculate where the silkscreen line should end to tangent the solder mask clearance.

regarding the minimum I just used 2xsilkscreenwidth, but I dont really care.

here is the discussion regarding the pinheaders in case there is anything useful for you
https://github.com/KiCad/Pin_Headers.pretty/pull/24

terjeio:
@Shackmeister 
Ok, I see that you have had a long conversation too... And I can see from your code that you have had quite a hard task before yourself.

What I have done is to move all the complicated math into a set of new classes so that next time I do not need to bother about all that math. Adding keepout handling is just a few (2) extra lines of code:

``` 
    silk    = Layer(kicad_modg, 'F.SilkS')
    pads    = PadLayer(kicad_modg, pad, Pad.TYPE_THT, Pad.SHAPE_OVAL, shape_first=Pad.SHAPE_RECT, drill=ddrill)
    keepout = Keepout(silk)
....
    # create pads
    y = 0.0
    for r in range(1, rows + 1):
        x = 0.0
        for c in range(1, cols + 1):
            pads.add(x, y)
            x -= coldist
        y += rm

    # add pads to silk keepout
    keepout.addPads()
...
``` 
and then draw the silk layer and let the code behind take care of splitting the lines for me. 

I am nearly ready to submit a new PR, maybe only one detail left to discuss, 1.00mm THT single row strips has the pins offset from center:

bilde

Is this ok or should:

1. the pin1 marker be extended right and down?

2. a silk line be added on the right hand side?

both respecting the keepout of course - it is the keepout code that shorten them now.

Shackmeister:
looks good to me, isnt it possible to fit some silkscreen on the right side?
and rememer the footprint should be centered on the pin 1

terjeio:
PR updated, KLC issues adressed and added some 1.27 and 1.00mm sockets

Travis is not happy about the new 1.00mm sockets, complaining about missing entry in fp_lib.

bilde

bilde

bilde

bilde

Shackmeister:
Looks very nice! Is it in Purpose the "filled" pin 1 indicator on the angled connectors have a slight gap in the top?
Currently im at my Mother in laws place but ill be back in a few hours and will check it then :)

Shackmeister:
@SchrodingersGat should an entry simply be made in the fp_lib by the contributor or is that something you or other librarians should do?

terjeio:
@Shackmeister  The gap is "copied" over from the originals, can be removed. I did not remove it myself as I did not know why it was drawn like that.

Shackmeister:
@terjeio ok I would suggest removing it unless somebody objects :)

I had a look at the new changes and found a few small "errors"
please see images below:




some thoughts about some slight changes, comments?
suggestion

silkscreen

Shackmeister:
is it on purpose you didnt add the
all the sockets for 1.00 and 1.27? (primarily horizontal)

checked the silkscreen on the solder mask clearance on the rest and its spot on :D

terjeio:
@Shackmeister 
I left out the horizontal SMD sockets for now as a new function has to be written and I need to take a break from this to attend some other business. My plan is to get the current parts done, including the 3D models, and then do some more research into the rest when I have more time (2-3 weeks from now).
IIRC there are horizontal SMD variants for 2.00 and 2.54mm sockets that should be added too, they should not be left out in the cold... ;-)

I have corrected the errors, my keepout code was not correct for rectangular pads and did not handle rounded pads correctly when bounding boxes overlapped. I have changed the code for that but I suspect there are other edge cases that should be investigated if and when I try to add it to the KLC scripts.

I am a but unsure about what to do about the pin1 marker, it seems to me that you suggest that the horizontal line should stop at the bevel corner and the vertical line start at pad center. Is that so?
A rule will be better than tweaking the generator code for specific footprints.

I do not see the  pad mask expansion in the footprint editor, do you know if it is possible to enable it there or should I check footprints in pcbnew? Also, I wonder if it is possible to add a reload button in the footprint editor (or pcbnew) via scripting, anybody know?

---

Short line segments will now be kept even if the keepout bounding boxes overlap:

bilde

Shackmeister:
> I left out the horizontal SMD sockets
I was talking about horizontal THT sockets :)

> I am a but unsure about what to do about the pin1 marker, it seems to me that you suggest that the horizontal line should stop at the bevel corner and the vertical line start at pad center. Is that so?
not sure I get what you mean, I didnt change/suggest anything regarding the pin 1 marker

> I do not see the pad mask expansion in the footprint editor, do you know if it is possible to enable it there or should I check footprints in pcbnew? Also, I wonder if it is possible to add a reload button in the footprint editor (or pcbnew) via scripting, anybody know?

inside FP editor go to: Footprint Properties -> Solder mask clearance and set it to 0.2mm 
remember to reset it to 0mm when you are done checking.

> Short line segments will now be kept even if the keepout bounding boxes overlap
in my script I only added the line segments if they were bigger than 2*silkscreen_width (0.12mm)

terjeio:
At the GCT website you linked to there are no THT horizontal 1.00mm sockets that I can find, only one single row SMD variant - BC075, I will add THT BD091 to the 1.27mm family - I overlooked that.
My brain soon needs some rest from all the new information it has to digest...

Inconsistent wording is used for the clearance?
KLC F5.1.2: _"pad mask expansion"_
Footprint editor: _"Solder mask clearance"_
Footprint files: _"solder_mask_margin"_
Generator framework: _"min_pad_distance"_
I assume these refer to the same property except perhaps the last one?

Pin1 marker is ok, I will use the rule as I stated in my last comment.

The short lines will be removed if I specify min line length > 2 * silkscreen width. They are 0.108mm long when the silkscreen is placed on the fab line with the default offset. For now I set the minimum to 0 in the keepout processor.

bilde

terjeio:
This update concludes my work on these for now, it will be at least another two weeks before I can commit more time if this should be required. 3D model script is ready, I will rerun that and post a PR when updated license text is made available - I am unsure how that should be referred to in the model files.

terjeio:
Example 3D-models, there are a few SMD oddballs - their dimensions are inherited from legacy footprints.
Total size is nearly 2GB!

examples

Shackmeister:
@terjeio truly great work! thank you!
can you please add the 1mm header library to the Fp-table file?
and can you also make sure there is a readme in all the libraries you are comitting to?

Shackmeister:
Great! Thank you for the contribution

SchrodingersGat:
Awesome work guys :)

evanshultz:
Is anyone planning to make footprints (and 3D models?) for headers with locking tabs? Like http://www.oupiin.com.tw/upload/gData/J3-01.jpg. There are vertical and horizontal versions in various pitches. I believe those are sorely missing from the lib.

terjeio:
@evanshultz 
I can make them. However, it will be at least two weeks before I can commit time.

evanshultz:
@terjeio 
Were you able to find time to make the headers with locking tabs? Is this on your to-do list?

evanshultz:
@terjeio 
This had slipped my mind until now. Did you invest any time in making these footprints? Are you still interested in doing so?

I don't find many datasheets showing the width of the latches when they're open, but https://multimedia.3m.com/mws/media/22268O/3mtm-100-in-4-wll-hdr-100x-100l-e-strt-cmpl-pin-ts0478.pdf shows 9.5mm and 11.7mm lengths for short and long latches. I measure even smaller latch lengths from a couple samples on my desk but I think this is a reasonable place to start.

---

I started to work on this myself but had a little trouble. Since the other guys are using python 3 for our scripts I went that route (it appears you used 2.7). I only saw a couple changes needed:
1. Change `utf8` to `utf-8` for file encoding.
2. Use parentheses around `print` statements.

Then I saw that the path to KiCadModTree was wrong for the location of the script in the directory hierarchy. I needed to add `sys.path.append(os.path.join(sys.path[0], "..", ".."))` to the top of `socket_strips.py`.

And then I ran into an issue with the line `from parameters import params` in `main_generator.py`. I see some methods used that are in `cq_base_parameters.py` (`getSampleModels`, `getAllModels`) but I don't see how this access them. Probably I'm just daft, but I don't see how the script could run as is. Would you be able to give me a bit of help one way or another?

matthijskooijman:
I also ran into the same issue trying to run the Conn_PinSocket script. @terjeio, it *looks* like you forgot to commit a `parameters.py` file that loads the YAML file. Could that be the case? If so, could you make it available somewhere?

matthijskooijman:
Ah, found the missing file here: https://github.com/easyw/kicad-3d-models-in-freecad/blob/master/cadquery/FCAD_script_generator/Conn_PinSocket/parameters.py

It needs just a small change to also load the `datasheet` parameter, which is apparently not used in that version of the file. With that, and fixed import path in `socket_strips.py`, the `Conn_PinSocket` script seems to work for me. I'l see if I can prepare a PR for this later.

matthijskooijman:
I fixed the Conn_PinSocket script in https://github.com/pointhi/kicad-footprint-generator/pull/222. Let's stop polluting this old PR now :-)

PR #43, opened 2017-11-16
Fix PLCC THT packages

nils-eilers:
Fix double defined and/or misplaced pads.

Decrease pad size to enable the placement of two signal lines
between them, a width of 8 mil or 0.2032 mm is assumed.
To to so, the drill size is reduced from 0.95 mm to 0.8 mm and
the pad diameter is reduced from 1.75 mm to 1.4224 mm.

plcc


SchrodingersGat:
Thanks for the contribution @nils-eilers

PR #45, opened 2017-11-16
For BGA case with 0.5mm balls ~0.4mm pads should be used instead

krtkr:
Hi all!

Push https://github.com/KiCad/Housings_BGA.pretty/pull/25 into new repo.

I saw you did some renaming to packages according new KLC. Do we want to use name discussed before: 'BGA-256_17.0x17.0mm_Layout16x16_P1.0mm_Ball0.5mm_Pad0.4mm_NSMD'?


evanshultz:
@krtkr 
Yes, see the details at http://kicad-pcb.org/libraries/klc/#_footprint_guidelines. Thanks for sticking with this repo change and re-submitting!

PR #47, opened 2017-11-17
Add newly created JST THT connector footprints.

poeschlr:
This PR includes all the remaining script generated jst connector footprints. (THT versions)
It recreates the earlier committed PH connector footprints. (Text field position has changed because of the new centralized script parts.)

script pull request: https://github.com/pointhi/kicad-footprint-generator/pull/81

The following footprints are newly added by this pull request:
- EH
- J2100
- JWPF
- NV
- PUD
- VH
- XH
- ZE (THT only, SMD has already been included earlier)

I noticed that all footprints that include a boss have contradictions within their datasheets. I trusted the original footprints for the boss positions.

Notable changes to old footprints: 
 - Most footprints had a wrong drill sizes. (JST does give the tolerances a bit strange.)
 - Some footprints had the wrong layers for pads. (Especially for NPTH pads)
 - Pad sizes are now unified as follows: (In case of contradiction: prioritization from top to bottom)
    - standardized minimum annular ring: 0.15mm (as per KLC)
    - standardized minimum pad to pad clearance: 0.8mm
    - standardized maximum "solder space": 0.5mm (elongation of pads where possible.)
- Some scripts calculated the courtyard offset relative to the silk outline instead of relative to the body outline / fab outline

I will add screenshots tomorrow. It is too late today to do this now. (While creating the screenshots i will do an in depth check of them in freecad and kicad. Until now i only made some cursory checks.)


poeschlr:
# JST EH
http://www.jst-mfg.com/product/pdf/eng/eEH.pdf

differences compared to footprints in old repo: 
- more detailed fab layer, 
- larger pad to pad clearance (but increased pad size in y direction to allow easier soldering), 
- corrected drill size (forgotten to fix in pull requests for https://github.com/KiCad/Connectors_JST.pretty/issues/40)

screenshot from 2017-11-17 11-36-21
screenshot from 2017-11-17 11-43-41

poeschlr:
# JST J2100
http://www.jst-mfg.com/product/pdf/eng/eJFA-J2000.pdf

main differences:
- pin names changed to a1.an, b1...bn
- **pin "1" for straight connector might have been wrong** (row marked a is farthest away from mount holes.) @SchrodingersGat you created these originally. Is there a reason you choose that particular pin numbering scheme? (I don't really trust the datasheets that much. Some of them have contradictions within them)
- Drill size slightly increased (put it in the middle of the tolerance range)
- Unified pad sizes as always.

screenshot from 2017-11-17 12-09-32
screenshot from 2017-11-17 12-19-43

SchrodingersGat:
Looks like you're right, I may have originally misread the datasheet. Good spot!

poeschlr:
The JST datasheets are quite confusing. (They don't give all the information needed, sometimes they even contradict themselfes.) Sadly i only have PH, EH and XH in my possession. So for all others i need to trust the datasheet/ The person who originally made the footprints.

poeschlr:
# JST JWPF

http://www.jst-mfg.com/product/pdf/eng/eJWPF1.pdf

main differences:
- connector width has been wrong in the old footprints
- boss drill was wrong (the tolerances again)
- maximum possible pad to pad clearance is 0.7mm (minimum annular ring requirement)

Silk outline is on top of the boss drill hole. Sadly the generator is not written in such a way as to easily allow a change here.

screenshot from 2017-11-17 13-08-22
screenshot from 2017-11-17 13-08-03

poeschlr:
# JST NV

http://www.jst-mfg.com/product/pdf/eng/eNV.pdf

Some measurements missing in datsheet, needed to trust previous contributor.

differences:
- drill size (tolerances again)
- pad size (0.5 mm annular ring should be enough, no need to elongate them in y direction. I think)

screenshot from 2017-11-17 13-50-32
screenshot from 2017-11-17 13-53-47

poeschlr:
# JST PH

http://www.jst-mfg.com/product/pdf/eng/ePH.pdf

no real changes to the functional parts of the footprint. (Not even when compared to the original footprints in the old repo)
Only change was to increase the silk to pad offset. (forgot to add half the silk line width)

screenshot from 2017-11-17 14-15-39
screenshot from 2017-11-17 14-15-41

poeschlr:
# JST PUD

http://www.jst-mfg.com/product/pdf/eng/ePUD.pdf

vertical version with boss not included (was missing in old repo as well.)

changes:
- smaller drill size (was too large previously)
- pin 1 marker on fab moved towards the inside
- increased pad to pad clearance

screenshot from 2017-11-17 14-41-25
screenshot from 2017-11-17 14-41-19

poeschlr:
# JST VH top entry

http://www.jst-mfg.com/product/pdf/eng/eVH.pdf

- pad size increased a bit (We have space)
- drill was too small (tolerances again)

screenshot from 2017-11-17 15-43-49
screenshot from 2017-11-17 15-43-25

pointhi:
please note at least EH and PH missing pin 1 marker on F.Fab

poeschlr:
Are you sure? Both have pin 1 markers. See screenshots above (the line drawing is the fab layer, in the other screenshot it is hidden beneath the silk pin 1 marker for the vertical versions.)

poeschlr:
# JST VH side entry

http://www.jst-mfg.com/product/pdf/eng/eVH.pdf

as with other vh connectors:
- pad size increased (we still have 1.26mm clearance)
- drill size corrected
- some minor fixes to body size

screenshot from 2017-11-17 16-06-38
screenshot from 2017-11-17 16-06-21

pointhi:
ok, I see it now on the picture. I though you use arrows for alle pin1 marker on F.Fab ^

poeschlr:
Pin 1 markers are not really unified. I only changed them where the pin 1 marker was a triangle outside the part. (I like them better inside the component outline.)

I am not really sure about the pin 1 markers for the VH side entry types. (If i am motivated i might look into this after i finished checking the footprints.)

Also remember these footprints are created by separate generator scripts. (Originally made by different persons with different programming styles and under different versions of your generator library.)

It would be way too much work to make one script that can generate all of them. 
I learned that when i made my powerful generator for the smd connector footprints. (And even that is not able to generate everything. Molex duraclick will need a specialized generator because it has a very complex body outline. Dual row connectors also need other scripts because my script can only handle single row connectors.)

evanshultz:
I created the VH and NV scripts. I do have these on hand if you want measurements, and I used the actual parts to get dimensions missing from the datasheets. Hopefully that means all was well.

I would prefer larger annular rings on these parts. You can put quite a bit of force on the THTs when inserting and removing the mating connectors. While round pads may be perfectly fine, I do not believe we want to use the minimum annular ring. It should be at least as big as IPC Level A, and I'd suggest even bigger than that.

VH side entry:
1. Could the F.Fab text always be located in a place that avoids other F.Fab lines?
1. The BxPS series has the pins drawn on F.Fab, while the SxP has the plastic pieces between the pins drawn but not the pins themselves. That was in my original script but I didn't notice it then. Perhaps it would be best to also draw the pins on SxP footprints?
1. Yeah, the pin 1 mark on F.Fab is a bit weird. It's not pointing towards the pin, but I also preferred it inside the body outline. If you do the item above, maybe this arrow could fit inside the lines for the pin?

poeschlr:
Reference mark for VH side entry fixed:
screenshot from 2017-11-17 17-37-47

poeschlr:
# JST XH

http://www.jst-mfg.com/product/pdf/eng/eXH.pdf

Added two more variants of these connectors.
- side entry compact version
- top entry with boss

What else has changed?
- increased pad to pad clearance
- fab outline for side entry now follows the part outline
- drill size now at center of tolerance range

screenshot from 2017-11-17 17-43-39
screenshot from 2017-11-17 17-53-00
screenshot from 2017-11-17 17-58-35

poeschlr:
@evanshultz 
About annular ring:
only the JWPF series is at the minimum as specified in KLC (0.15).
I would not go smaller for JWPF because otherwise it will get quite hard to get a single trace between the pads. (even with the minimum here we only have 0.7mm clearance between them.)

The next smaller annular ring is 0.225mm (PUD and PH). I don't think this would be too little.

poeschlr:
I ran out of time for today.
I will look into the ZE pad sizes tomorrow. (Something is off here.)

poeschlr:
# JST ZE

http://www.jst-mfg.com/product/pdf/eng/eZE.pdf

Main differences to original footprints:
- larger pads but also larger pad to pad clearances (>0.8mm for a 45 degree track between the pads of the top entry type, much larger for side entry)
- fixed drill sizes (middle of tolerance range)
- fixed body size
- removed silk where it interferes with the boss drill

Details about pad size. (Note that using a simple rectangular pad for pad 1 would reduce the clearance down to ~0.5mm. For this reason i made a special pad that is oval towards the other pads but rectangular towards the outside)
screenshot from 2017-11-18 15-18-29

screenshot from 2017-11-18 15-48-21
screenshot from 2017-11-18 15-48-48
screenshot from 2017-11-18 15-47-32

poeschlr:
I checked everything now. So this should be ready to merge.

The travis errors stem from my special pin 1 pads. (Compination of tht and smd pads but without paste -> travis thinks all smd pads should have paste)

SchrodingersGat:
Great work rene, these all look very good. Thanks for all the work you have done on this.

PR #50, opened 2017-11-17
Added symbols

SchrodingersGat:
Copied from Symbols.pretty -> Symbol.pretty

- Fixed some outstanding issues (renamed `REF***` to `REF** e.g.)
- A lot of KLC errors but these are special cases and I do not think any of these errors apply
- Should be OK for direct merge :)


evanshultz:
Then I'll merge directly!

PR #55, opened 2017-11-18
Heatsink

SchrodingersGat:
Copied heatsink footprints

Fixed some KLC issues with script --fix but some remain.


SchrodingersGat:
A lot of improvements here, about as far as I want to go :)

jkriege2:
Hi!

I had a look ... yes that lib is a real headache ... Maybe if I find time, I will do a major workover at some point ... until then: 
1. Is there a reason, why you added courtyards that only surround the pads?
2017-11-18 17_22_32-footprint editor no active library
2. Could you add the default 3D model links? I think `--fixmore` of our scripts does that fo you ... That should also be able to draw a youryard around everything if required ...

Thanks,
JAN

SchrodingersGat:
Uh the courtyards were automatically generated using `--fix`. I didn't see that it was only around the pads.

jkriege2:
Possibly because the auto-courtyard only takes into account pads and Fab ... An older version took into account the silkscreen, when nothing on *.Fab is available ... but could be that has been removed from the code at some point?

poeschlr:
I fear the script might be broken in some other way. The footprint in your screenshot has stuff on the fab layer but it still placed the new courtyard only around the pads.

jkriege2:
Hmmm ... I just deleted the courtyard and added it with the current scripts ... and it works for some:
2017-11-21 22_08_13-footprint editor no active library

The problem with footprints without F.Fab-drawing was that the script did not use the correct layer-name when searching for a drawing. This is fixed in https://github.com/KiCad/kicad-library-utils/pull/179. With that version everything should work:
2017-11-21 22_26_10-footprint editor no active library

poeschlr:
@SchrodingersGat do you have time to fix this or should i do it in a new pull request?

SchrodingersGat:
I don't have time to fix this currentlly, no.

poeschlr:
Replacement: https://github.com/KiCad/kicad-footprints/pull/140

PR #56, opened 2017-11-18
Added crystals

SchrodingersGat:
- Copied from Crystals.pretty


jkriege2:
looks good to me ...

evanshultz:
As mentioned in the "Who's doing what?" thread, I volunteered to import the crystal footprints and fix https://github.com/KiCad/Crystals.pretty/issues/23. Is that issue unimportant? (I realize it's not important in the grand scheme of things when trying to get the libs ready for v5, but I'm wondering about your thoughts on it ever needing addressed.)

BTW, I'm actively fixing the pot footprints and the footprint names will change (angled/straight>horizontal/vertical among other changes), so please nobody import those old footprints here.

SchrodingersGat:
Evan, that's my bad I forgot that you were working on the crystals. Please address the mentioned issue when you have time.

PR #58, opened 2017-11-18
Added Texas_PWSON-6 footprint

DerekGn:
PWSON-6 footprint

image

Updated footprint


SchrodingersGat:
1. Name should be `Texas_S-PWSON-N6`
1. Add silkscreen
1. Add bevel to F.Fab outline to indication location of pin-1
1. Fab outline does not match dimensions of chip (3x3mm)
1. Pad spacing does not match datasheet (pads should be 0.95mm spacing)

Also, I have found two different TI documents which reference `S-PWSON-N6` but with wildly different dimensions:

* http://www.ti.com/lit/ml/mpds176e/mpds176e.pdf
* http://www.ti.com/lit/ds/symlink/tps714.pdf

SchrodingersGat:
Thanks Derek. Manufacturers don't always make it easy.

SchrodingersGat:
Great! Only thing remaining is that the RefDes on the F.Fab layer is too small. A good size would by 0.5x0.5mm with text thickness of 0.1mm

Fix this and we are good to merge :)

DerekGn:
I have increased the size of the refdes.

PR #59, opened 2017-11-18
Added Texas_R-PDSO-N5 footprint

DerekGn:
Texas_R-PDSO-N5 footprint

image

Added to PackageSSOP as it was the closest match for Plastic Small Outline. Fee free to move if needed.
Updated


SchrodingersGat:
@DerekGn a few changes required:

1. Rotate such that pin-1 is in top-left corner - http://kicad-pcb.org/libraries/klc/F4.2/
1. Increase size of smaller RefDes text to 0.3mm
1. Add a bevel to F.Fab outline to indicate pin-1
1. Add silkscreen 

selection_020

SchrodingersGat:
Thanks :)

PR #61, opened 2017-11-19
added TO-92-variants + a special japan molded package

jkriege2:
datasheets:
- http://www.produktinfo.conrad.com/datenblaetter/175000-199999/181293-da-01-de-TO92_Temperatursensor_PT1000_32209225.pdf
- http://rtellason.com/transdata/2sb734.pdf (this paclage does not seem to have a commonly accepted name ... so I went for giving the sizes)
- https://www.diodes.com/assets/Package-Files/TO92L.pdf
- https://media.digikey.com/pdf/Data%20Sheets/Infineon%20PDFs/KT,KTY.pdf
- https://www.diodes.com/assets/Package-Files/TO92S%20(Type%20B).pdf

2017-11-22 20_43_44-3d viewer
2017-11-22 20_43_10-pcbnew d__kicad_test_addlin_test_addlin kicad_pcb


aimylios:
I'm not sure if this footprint is really manufacturer specific, but shouldn't it be named "Heraeus" instead of "Hereaus"?

evanshultz:
@jkriege2 These mostly look great! A few comments:
- The `TO-92L` looks like a normal TO-92 with lead prepping. Maybe `TO-92-3_LeadPrepWide` or something is better? For example, https://www.onsemi.com/pub/Collateral/2N3903-D.PDF has roughly the same package but On Semi just refers to the lead prep to differentiate from a standard TO-92.
- Using the term `Moulded` is a good description, but I wonder if there's a more intuitive name for users?
- Should it be `Mini` instead of `MINI`?
- Could you please check all the 3D model params? Some of them aren't default.

FYI: I was confused at first because I didn't see F.Fab ref des text except on the NEC package in the image, but it's there in the files.

jkriege2:
Hi!

thanks for the review ... 
- I will fix 3&4 ...
- About the TO-92L (seems to be also known as TO-226) ... the footprint is more or less the same as for a TO-92 ... the package is just longer/higher ... there are even all the variants we have for TO-92 (see http://www.ti.com/lit/an/snoa059/snoa059.pdf). So maybe we even don't need this ... I just added it for sake of completeness and so we can distinguish different 3D models ... should we keep it? Have TO-92L and TO-226 in parallel, or mention TO-226 (which seems less common to me) in the description only? Should we have all major variants?
- Molded seems fine to me ... TI even uses that word in the document above ...

Best,
JAN

evanshultz:
@jkriege2 About TO-92L: I agree it should be a discrete footprint because the 3D model is unique. However, straight leads, staggered leads, prepped-outward leads (like above) are all possible so I was thinking that lead prep should be mentioned. Using `TO-92L` is great to me because that captures body size, but what about capturing lead prep?

jkriege2:
I would say: For now I will just duplicate the naming of the normal TO-92 ... we can always rename later (as that also requires a lot of renaming in the symbol lib). OK with you?

JAN

evanshultz:
@jkriege2 Sure! Let me know when you're ready for merge!

jkriege2:
OK, I renamed the TO-92-variants with a naming scheme as for the existing TO-92, but dropping the `_Molded`,, which does not make sense for me ... `_Inline`is pins configured on a line, no modifier is pins on a circle. If that's OK, I can also rename the exitsing TO-92 packages accordingly.

In addition I fixed the 3D model parameters and added some more missing variants. TO-226 is mentioned in the descriptions/tags now.

Best,
JAN

evanshultz:
Thanks Jan!

1. The TO-92L parts have a typo of "also" in the description.
1. Also, Travis couldn't run and it looks like the indenting of the files. Can you take a look?

jkriege2:
Just for Doku: Current status:
2017-11-22 20_43_44-3d viewer
2017-11-22 20_43_10-pcbnew d__kicad_test_addlin_test_addlin kicad_pcb


@evanshultz Can this be merged?

evanshultz:
@jkriege2 See my comment above. There is a typo, and I'm a bit concerned that Travis didn't run.

jkriege2:
Hmmm ... it's strange the script doesn't run, because I fixed that bug yesterday morning () ... locally all my symbols check out:
```
Checking footprint 'TO-92L_Inline_Narrow' - No errors
Checking footprint 'TO-92L_Inline_Wide' - No errors
Checking footprint 'TO-92L_Narrow' - No errors
Checking footprint 'TO-92L_Wide' - No errors
Checking footprint 'TO-92S-2' - No errors
Checking footprint 'TO-92S' - No errors
Checking footprint 'TO-92S_Wide' - No errors
Checking footprint 'Heraeus_TO-92-2' - No errors
Checking footprint 'NEC_Molded_7x4x9mm' - No errors
```

... and I fixed the speling error!

@SchrodingersGat Can you have a look why Travis seems to be using an old version of the scripts? ... Thanks!

jkriege2:
OK, now the scripts work again ... don't know what happened yesterday evening ;-)

The only remaining KLC-issue is the annular ring of TO-92Mini ... but the pins are so close that I didn't see any other option to keep at least a bit of space between the copper pads ... any ideas how to fix this in a better way? If not, can it be merged?

SchrodingersGat:
I will merge this, thanks for your efforts. If we work out a way to improve the annular rings you should revisit this :)

PR #63, opened 2017-11-21
Test probe

SchrodingersGat:
This PR adds a library for **Test Probes**

e.g.

* https://www.harwin.com/products/P13-2123/
* https://www.harwin.com/products/S25-022/

These are separate from Test Points. A few simple ones to get the ball rolling.


poeschlr:
Have these been in one of our old libs or are these new footprints?

SchrodingersGat:
These are all new to the KiCad library.

poeschlr:
How did you come up with the pad dimensions?

poeschlr:
@SchrodingersGat could you clarify how you derived the drill size and pad size for these footprints?

SchrodingersGat:
@poeschlr I'm going to close this for now so we can focus on transferring rather than adding new parts that require attention :)

PR #65, opened 2017-11-21
Transferred USB connectors

SchrodingersGat:
Transferred USB connectors across.


poeschlr:
Travis errors are false alarms. (Datasheet trumps KLC)
Thanks for your continued efforts.

PR #66, opened 2017-11-21
THT Diodes

SchrodingersGat:
Moved THT diodes across, fixing 3D model links


poeschlr:
Will merge it for now and create yet another fix later issue about polarity markers and second reference positions.

PR #67, opened 2017-11-21
THT inductors

SchrodingersGat:
Moved THT inductors to new repo.

Most KLC issues have been fixed, except for a couple of silkscreen issues. @jkriege2 perhaps some scripting readjustment at some stage.

However this is not serious, please merge as is


poeschlr:
Some of these don't have the 3d model stuff filled out. Is it possible to fix this with the check scripts?
For the other reported errors i will again add a new issue.

SchrodingersGat:
Neither the --fix or --fixmore options add a default 3D model.. Some work needed on the scripts!
If you create a new issue, this can be mentioned there too. We should update the scripts to provide an automatic fix for this.

poeschlr:
Will do. Thanks.

Shackmeister:
Perhaps the script could also reset the 3D parameters if they are different than 1 scaling and 0 offset and 0 rotation

evanshultz:
Since making changes would break the 3D viewer, perhaps add a warning if 3D params aren't default?

poeschlr:
We will not change the v4 repos with this option. For the v5 repo only correctly scaled 3d models are accepted anyways.

evanshultz:
Right. So it should be a warning, if not an error, if the 3D params are wrong. And automatically changing the 3D params still seems too dangerous to me.

Shackmeister:
In my opinion it should be an error

poeschlr:
@evanshultz 
We are talking about changes that are made when running the script locally with the --fixmore option. (Travis can/should not automatically fix contributions)

@Shackmeister 
I would agree it should throw an error for all new contributions.

evanshultz:
@poeschlr So was I when I advocated for a warning if not an error. And I also think automatically fixing this when running locally is dangerous because any changes likely break the 3D view. The user should be flagged but not automatically fixed.

PR #69, opened 2017-11-21
THT resistors

SchrodingersGat:
Added THT resistors

Fixed KLC issues


poeschlr:
For some of these the second reference should either be scaled down or it should be placed outside. Do you want to look into this now or should we add this to the fix later list?

SchrodingersGat:
Add to the fix list :)

poeschlr:
Done and thanks

PR #71, opened 2017-11-21
THT capacitors

SchrodingersGat:
Copied from old repo.

A few file fixes were performed, no KLC issues present.


poeschlr:
None of the polarized caps have an external orientation marker. Do you want to fix it now or do we add them as they are but create an issue to remind us about the problem?

SchrodingersGat:
I feel that we are getting bogged down trying to fix everything. Please merge as is. We should track down who scripted these and update the scripts. Thanks

Shackmeister:
Made by J2kriege since modified by me, is the problem lack og polarisation indicator after component placement?

PR #83, opened 2017-11-22
Bosch BME280_680 footprints

sandhan26:


sandhan26:
Can i submit a PR to change the names of all Housings to Packages in kicad-library repo? Or any other way if i can help?

SchrodingersGat:
@sandhan26 we are not yet ready to make that change, but we have a script to do the work for us :) Stay tuned!

SchrodingersGat:
File name does not match footprint name inside file - please fix this :)

PR #84, opened 2017-11-22
changing 3D model reference from .step to .wrl

Shackmeister:


poeschlr:
Damn i did miss that. Thanks for noticing. (That's the problem when we get a lot of false positives in the travis report. Important stuff is missed.)

PR #85, opened 2017-11-22
Added Coilcraft MA5532 RFID transponder coil

openelectron:
I added this to RF Antenna rather than inductors as it is specifically used as an antenna although it could be described as an inductor also.


evanshultz:
Hi Rob,

For every PR include the datasheet and an image of the symbol or footprint. Please do that here and for the other PRs you raised recently. Thank you!

SchrodingersGat:
@openelectron next time a screenshot makes it easier :) Thanks for the contribution

PR #89, opened 2017-11-23
Harwin connectors

SchrodingersGat:
Added existing Harwin connectors
* Rebuilt from scripts
* No KLC


dragonlost:
Good morning all. Why this pull request was not accepted. For now the numbering for harwin connectors is wrong in kicad which is a shame.

chschlue:
Do you mean #323?
Apparently because its author never finished it.

dragonlost:
ha darn

chschlue:
@dragonlost feel free to re-push and fix it.

PR #92, opened 2017-11-23
added ABM10 crystal

Shackmeister:
abm10


evanshultz:
For easy reference, the DS is at http://www.abracon.com/Resonators/ABM10.pdf.

@SchrodingersGat @poeschlr 
Can you take a look at the this footprint name? KLC v3 doesn't cover crystals (that I see). Do we need the `Crystal_SMD_` part? Or just start with `Abracon_ABM10`? Or is this manufacturer specific at all? While in a quick look I didn't turn up any other parts with the same footprint, I found another SMD 2x2.5mm crystal footprint (if not package) that is shared by several different vendors.

jkriege2:
The `Crystal_`-prefix is there to allow for filtering, as most crystal footprints are meant to be used with the generic symbols in device.lib ... I'm not sure whether this is still needed, as I think KiCAD-FPFilter can also filter by .pretty-name ... @SchrodingersGat I think you added that function?

On the other hand I like the prefixes, as they allow to immediately see what a footprint is for (we also use them for R,C,L,D,... basically everything, which is meant for symbols in device.lib)

SchrodingersGat:
The `Crystal_` prefix should be provided for generic devices, but `Manu_MPN` is sufficient. As Jan says the footprint filters can now be set to e.g. `Crystal*:0603` to include the footprint library name

Shackmeister:
@SchrodingersGat so the naming is fine as is?

SchrodingersGat:
Yep, thanks for the contribution :)

PR #99, opened 2017-11-24
Added Footprints for C&K Digitast switch 

Robert46:
digitast
Im not sure if it is wise to place the LED pins. So I'd have to create a new switch-symbol with two LED's wich would be named SW1a or so instead of D1 and D2.
 The alternative would be just to place two crosses on silk screen.
 Suggestions?
 When I save the model via Footprint Editor it always has CR/LF line endings is this normal behaviour?


poeschlr:
If the LEDs are included in the connector then they should be included in the footprint.

For symbol name either use an atomic approach (SW_[manufacturer-name]_[mpn]) and set the footprint field,
Or a generic approach with SW_[switch descriptor (SPST, DPST, SPDT, ..)]_[number LEDs]xLED and set the footprint filter such that it finds all switches (`SW_*`)

Footprint naming:
please use SW_[manufacturer name]_[manufacturer series name]_[mpn]_[additional options]

maybe we could also include the "switch descriptor" (SPST, DPST, SPDT, ..)
would then be:
SW_[descriptor]_[manufacturer name]_[manufacturer series name]_[mpn]_[additional options]

So the switch that is currently called SW_Digitast would be named:
`SW_CK-Switches_Digitast_[SER|SET|SERU]{-L|2L}`


@SchrodingersGat how should we denote C&K? I have (wrongly) used CuK in the footprints i committed (C und K). The webside is ckswitches.com that's why i used that in my naming above.

Robert46:
What does SER SET or SERU mean?
I'm currently at SW_CK-Switches_Digitast_LED_wide, would that be ok?

poeschlr:
Did you read the datasheet? These are the series names listed by C&K

Robert46:
Not since I introduced the part in 1990 :-)
Just committed the new files. Anything else I must do? (this is my first contact with Github)

poeschlr:
Had a quick look at your footprints:

- rotate them by 180 degree such that pin 1 results in the top left corner
- Change the pin 1 shape to rectangle (square)
- Add some form of pin 1 marker to the fab and silk layers. (See image below)
- move the outermost silkscreen lines outwards such that the complete outline is visible after assembly (This assumes the current artwork on the Fab layer represents the exact body outline)
- move the courtyard lines outward. (courtyard clearance to body should be 0.25mm)
- drill size for the led pads should be smaller. (The leads have a diameter of 0.4mm -> drill should be 0.6mm) I am not sure about this one. They also state 1mm drill size in one drawing.
- drill size for the location pins should be a bit larger (the peg has 1mm diameter -> use at least 1.1mm. This is also the drill size given in the sketch found in the datasheet.)
- 3d settings: set scale to 1, offset and rotation to 0 (The 3d models need to be designed such that this is possible.)
- outside dimensions of SER type is wrong. Fab outline should reflect the exact body outline -> should be 12.3x17.5, Is 12.7x17.78
- Fab lines should have a thickness of 0.1mm
- Silk line should have a thickness of 0.12mm

I played with one of your files. Here is how it should look like. (I did not fix the line width, this shoud still be done.)
[SW_CK-Switches_Digitast_SER-2L.kicad_mod.txt](https://github.com/KiCad/kicad-footprints/files/1501827/SW_CK-Switches_Digitast_SER-2L.kicad_mod.txt)

screenshot from 2017-11-24 14-54-26

SchrodingersGat:
**CK** is my vote for manufacturer designator.

Robert46:
Rene, I just uploaded my personal final version of the switch, Maybe we can come to a compromise since I do not have the time (and the will) to make all the changes you request. Some of them even go completely against the grain for me and are just not intuitive.
First point is that grid origin must be at pin 1 which the checker (that runs when I create a pull request) enforces. In a part like this it is natural feeling that the origin is at the geometrical center. There is no meaningful "pin 1" in a switch and it does not make sense to emphasize it in any way, as grid origin, on any layer or with a special pad shape. (I would not even number the pins but call them CO, NC and NO so my perfect switch would not have a "pin 1" at all but the circuit symbol was already there...) comming from a different layout software (BAE) I find it odd that e.g. diode pins are numbered instead of calling them A and C. 1 or 2, what's the cathode? What would you do with parts that do not have a pin 1?
Pin 1 must be at the left topmost position. This enforces that the part cannot be drawn as you normally use it and you would always have to rotate it 180 degs. This is simply not '.pretty'. I would not sacrifice handiness and usability for strict design rules in my library. A switch has to be drawn upright and a display as it is readable. Don't matter where pin 1 ends. Btw. why top left? I never minded this but in most datasheets (esp. DIP-ICs) pin 1 is top left in the functional drawing (so you can read the pin functions more easily) but in package drawings pin1 is usually bottom left. Rotating it 270° is not only unintuitive but ends up in an inverse cinemascome image, not reflecting the screen properties on wich you have to view or edit it.
Pin numbering for me is a relict of _very_ early CAD systems where pins only had numbers. No reason today why a part must have a "pin 1" at all.
Courtyard line width: the checker enforces 0.05mm, that's why I set it. Wonder why KiCad does not enforce this. You can draw anything here but you cannot post it later. Maybe a checkbox "enforce KiCad design rules" somewhere in the preferences?
I believe that the coutyard dimensions should be left as they are now. They are 0.2mm in X and 0.14mm in Y bigger than the part. If I changed this to 0.25mm the courtyards of adjacent parts would overlap.
Silk screen line width: the design rules on the KiCad homepage allows 0.1 .. 0.15mm. I agree that a uniform width is desirable and Notepad++ did the change in all files with a few clicks, also for the fab line width :-)
3d model offset and scale: well, I created the model following an outdated tutorial (from the recent KiCad homepage) but it works and displays fine. I will not re-create it in a different program just to make it 1:1. I could ommit the 3d model but it would be a pitty since it is pretty :-). If there is only a wrl (no STEP), maybe the checker in KiCad5 can set a mark that 3d collision checks cannot be done with this part (draw it in red or so). Maybe someone (or I myself) creates a better 3d model some other day but for now I'm tired of learning 3d-tools... Btw: for collision checks, the button in the model would have to be a union of the pressed and released button. Again not pretty, a lot of effort (since it is not documented in the datasheet) but truely useful.
Accuracy: Normally I would agree that the drawing has to be accurate. In this case, it is obvious that the switch can be aligned at multiples of 2.54mm and I preferred a simple-to-create drawing over a 100% exact one. Noone will notice the difference, I bet. Sometimes cheating makes live easier without doing any harm. Especially since the simplified dimensions are one or two tenth of a millimeter bigger this cannot cause problems.
I do not intend to make further changes.

poeschlr:
The reason for our rules is to have a library of constant quality. Users should be able to rely on some basic facts. 
One of them is that the pin 1 is at the origin for all tht components.
Another one is that the 3d models are scaled correctly and aligned to the footprint such that the footprint can use 1 for the scale and 0 for offset and rotation.

We can accept breakage of a few rules. (If the contributor can state a good technical reason for it.) But breaking nearly all of them and in addition having footprints that have the wrong measurements (both on the fab outline and for the drill sizes) is too much.

By the way it took me about 10 minutes to fix your footprint in such a way as to make it is usable. (The result is posted in my previous comment.)

PR #102, opened 2017-11-24
Add Omron-G5V-1 Relay

samt:
[Datasheet here](http://omronfs.omron.com/en_US/ecb/products/pdf/en-g5v_1.pdf).

screen shot 2017-11-27 at 8 34 27 am


SchrodingersGat:
Please add a 3D model reference. The 3D model does not actually have to exist yet, but if we place a reference for it, we do not need to come back later on and edit the footprint.

Also, please change the shape of pin-1 to rectangular! Thanks :)

samt:
@SchrodingersGat Added and updated screenshot above.

SchrodingersGat:
Nice work, thanks. 

Sorry but I should have also spotted this last time - please rotate the footprint 90 degrees clockwise, such that:

1. Pin 1 is located at the top-left - http://kicad-pcb.org/libraries/klc/F4.2/
1. Pin 1 is still at the origin (0,0)

Thanks :)

samt:
Not sure why the build failed, the travis logs do not appear to be an issue on my end (but correct me if I'm wrong). 

I've updated the screenshot and pushed the new version, should be all up to KLC.

poeschlr:
The log fails because we have a corrupted entry in the fp-lib-table. This is indeed not within your responsibility. (Sorry for the inconvenience.)

poeschlr:
It might be a good idea to add pin 1 markers on silk and fab layers. Here some inspiration: (green: suggestion for fab layer, light blue: suggestion for silk layer)
screenshot from 2017-11-28 02-37-05

SchrodingersGat:
Are you able to make the suggested changes @samt ?

samt:
@SchrodingersGat Yeah, sorry. I've been AFK for a bit due to moving and holidays. I'll be making the changes and pushing it up to my branch.

SchrodingersGat:
Thanks, please ping me when you've pushed the new changes.

poeschlr:
As oliver has other obligations i wil take over this pull request. @samt if you are finished please ping me instead of pinging oliver.

Shackmeister:
@samt are you still working on this?

Misca1234:
Here it is
So I guess there is no need for this push


bild

bild

Ratfink:
Yes, a footprint for this relay was merged in #254, so this PR is unneeded.

PR #103, opened 2017-11-24
added U.FL connector

Shackmeister:
Also drew a 3D model for it. (yes drew, not scripted)
u fl

ufl_3d


Shackmeister:
footprint anchor should be correct as is

SchrodingersGat:
Thanks, I think we can ignore that single Travis error in this case.

PR #104, opened 2017-11-24
Add all script generated molex slimstack footprints

poeschlr:
This pull request adds the slimstack footprints
(I did not change anything on purpose)

@evanshultz you added the 502429 and 502430 footprints. What is the difference between *10, *20 and *30? (Only difference i could see is a different datasheet link. But as far as i can tell it looks like they are completely equivalent otherwise.)

You also mirrored one of these two series. Should others also be mirrored?


poeschlr:
52991 series: Bug discovered by @evanshultz is fixed with this pull request. https://github.com/pointhi/kicad-footprint-generator/pull/66
(The calculation is correct)
http://www.molex.com/pdm_docs/sd/529910308_sd.pdf
screenshot from 2017-11-24 18-20-51

All other footprints are unchanged in their important parts (fab outline and pads)
Only slight differences are:
- slightly differently placed text fields (I used my automatic placement tool)
- slightly different silk-fab and silk-pad offset (now from central config file)

evanshultz:
@poeschlr 
 http://www.molex.com/molex/products/listview.jsp?query=502430&path=502430+inmeta:promotable%3Dyes+inmeta:CollectionName%3DImpulse+inmeta:category%3DPCB%2520Headers%26requiredfields=(engineeringnumber:502430%7Coldpartnumber:502430%7Cproductseries:502430)&offset=0&autoNav=0&sType=s&filter=&fs=&channel=products&encode=true discusses the differences between the different suffixes. I believe the actual part is the same, but *30 is with vacuum tape and *10 is without. *20 is only the 8-pin and it doesn't mention vacuum tape but I don't see what differentiates it from *10. I believe this only affects the machines used to place the connector on the PCB, but since they were unique PNs I included all of them. You can simplify if you want.

I'll wait to merge in case you want to purge this PR.

Thanks for including the 52991 dimension fix! I assume you'll upload the new scripts so I deleted my old PR to fix the script.

SchrodingersGat:
@poeschlr ok to merge?

poeschlr:
Yes sorry forgot to add a comment.
I now removed the footprints for versions that only differ by the use of the vacuum tape.

SchrodingersGat:
Great, thanks a lot :)

Issue #78, opened 2017-11-21
Diodes THT: Polarity marker on silk not ideal for most footprints and second (fab) reference interferes with other fab graphics


poeschlr:
This should be fixed with the generator scripts.

poeschlr:
Script location:
https://github.com/pointhi/kicad-footprint-generator

jkriege2:
Currently the silkscreen looks like this:
![2017-11-21 21_50_10-3d viewer d_do-35_sod27_p5 08mm_vertical_anodeup](https://user-images.githubusercontent.com/11967523/33096039-fadcdefe-cf05-11e7-969d-549eb496cf56.png)
![2017-11-21 21_49_59-3d viewer d_do-35_sod27_p2 54mm_vertical_anodeup](https://user-images.githubusercontent.com/11967523/33096040-fafba668-cf05-11e7-92f1-fbdca847185d.png)
![2017-11-21 21_49_41-3d viewer d_do-35_sod27_p7 62mm_horizontal](https://user-images.githubusercontent.com/11967523/33096041-fb1bc74a-cf05-11e7-9982-7e4fc0a8a920.png)

How do you want to change them?

Best,
JAN

poeschlr:
Examples where it does not look that nice:

Your last screenshot is a perfect example. If the view is obstructed by other devices next to the diode then it becomes very hard to see the intended direction of the diode. (The assumption is that the polarity mark is used to easily check the correct insertion of the diode when everything is assembled.)
Just imagine a high connector very near to one side and a large capacitor on the other. (Over voltage protection circuit for a DC input)

Other examples (of slightly different problems)
D_5W_P5.08mm_Vertical_KathodeUp (Interference between diode symbol and part outline)
![screenshot from 2017-11-21 22-07-45](https://user-images.githubusercontent.com/18327350/33096803-8ceb8f82-cf08-11e7-929e-4af06622648a.png)
Even worse if D_DO-201_P5.08mm_Vertical_AnodeUp is placed low down it might be very hard to see the polarity. (or if high components are next to the diode and it is impossible to look from the side.)
![screenshot from 2017-11-21 22-14-05](https://user-images.githubusercontent.com/18327350/33097140-7dce65a0-cf09-11e7-9157-7bd7466f0f70.png)

As for the how:
Depends on the exact footprint in question. In your last screenshot maybe elongate The top line to extend beyond the pad? Or you could add a dot, ...
Anything that is visible from directly above.
[...]

PR #106, opened 2017-11-24
new diode footprints (cathode marked with a K, which is better visible than the old marks)

jkriege2:
- K is now marked by a letter + removed the old diode drawing as it was invisible after assembly anyways
- script-based: https://github.com/pointhi/kicad-footprint-generator/pull/83
- based on this discussion: https://github.com/KiCad/kicad-footprints/issues/78
- also fixed 3D models: https://github.com/KiCad/kicad-packages3D-source/pull/64 and https://github.com/KiCad/kicad-packages3D/pull/161

2017-11-24 21_42_56-pcbnew d__temp_test kicad_pcb
2017-11-24 21_42_46-pcbnew d__temp_test kicad_pcb
2017-11-24 21_42_30-pcbnew d__temp_test kicad_pcb


poeschlr:
Could you also take a closer look at the positioning of the second reference designator? Currently it does interfere with other features of the fab layer. (I am assuming here that you adapted the scripts if you made this changes by hand we can merge as is.)

evanshultz:
It might be nicer to put a "K" or an "A" closest to the pin _without_ the diode body there. That may avoid putting the marker under a nearby  part. So for example, in the first screenshot above an "A" would go by pin 2 instead of a "K" by pin 1.

Another option could be to fill the circle if it's over the cathode pin and not fill the circle if it's over the anode pin. This isn't as clear, and maybe a bad idea, but it just popped into my head.

jkriege2:
Hi!

just making the circle filled is also not really visible after assembly, assuming the diode covers the circle, but marking A instead of K seems a great idea, so I stuck with that ...

Aso I did some other adjustments!

## horizontal diodes: %R-placement on F.Fab
2017-11-25 12_28_41-footprint editor no active library
2017-11-25 12_28_23-footprint editor no active library


## vertical diodes: %R-FFab-label inside or over SilkS-REF-label + A/K placement where space permits
2017-11-25 12_49_49-footprint editor no active library
2017-11-25 12_47_28-footprint editor no active library
2017-11-25 12_47_18-footprint editor no active library
2017-11-25 12_47_08-footprint editor no active library
2017-11-25 12_46_52-footprint editor no active library
2017-11-25 12_54_35-footprint editor no active library
2017-11-25 12_56_21-3d viewer
2017-11-25 12_56_06-3d viewer
2017-11-25 12_55_44-3d viewer
2017-11-25 12_54_55-footprint editor no active library
2017-11-25 12_54_35-footprint editor no active library

poeschlr:
Much better. Thanks for the quick fix jan.

PR #108, opened 2017-11-25
Add TI S-PVQFN-N40 (RHA) to Package_DFN_QFN.pretty

aimylios:
This PR adds the footprint for S-PVQFN-N40 (package code RHA) from Texas Instruments. It is required by some of the MSP430 devices I am currently working on. I based it on the information from the [MSP430G2x55 data sheet](http://www.ti.com/lit/ds/symlink/msp430g2755.pdf) (pp. 70-72) with only two exceptions:
- pads are elongated by 0,14 mm to accommodate for the rounded shape on the outside (which is not part of TI's recommendation, but semi-rounded pads are not supported in KiCad yet)
- a keep-out zone at the pin 1 marker is not included (TI says it's optional, and KiCad doesn't have native support for keep-out zones anyways)

texas_s-pvqfn-n40
texas_s-pvqfn-n40_3d
texas_s-pvqfn-n40_3d_m

There is already a similar footprint Texas_PVQFN-N40 that should also be named Texas_S-PVQFN-N40 (to avoid confusion with the R-PVQFN variants). It is based on the [DRV8308 data sheet](http://www.ti.com/lit/ds/symlink/drv8308.pdf) (pp. 56-58) and is not compatible to the one I made. Are there any guidelines how the naming should reflect these differences?

I think it might be a good idea to script all of the S-PVQFN-x footprints in the future to get a more uniform appearance.


aimylios:
Travis gives the following warnings, which (in my opinion) can be ignored:

**Pad '41' missing layer 'Paste'**
-> Separate pads are used for the Paste layer of the thermal pad.

**Attribute solder_mask_margin == 0.07 != 0!**
-> The data sheet explicitly recommends 0,07 mm for all pads except the thermal pad.

**3D model name is different from footprint name (found 'QFN-40-1EP_6x6mm_P0.5mm', expected 'Texas_S-PVQFN-N40')**
-> A generic QFN-40 model is used (as mandated by KLC 3.0.3 M1.2).

SchrodingersGat:
@aimylios thanks for the contribution - I agree with each of your points regarding the KLC. As much as we try, the scripts sometimes throw some incorrect warnings.

aimylios:
@SchrodingersGat, thanks for merging!

Do you maybe also have an answer to my question from the first post, how the naming can reflect the differences of footprints with an identical TI package ID?
I just had a closer look at the datasheet of the DRV8303, and the thermal pad apparently is the only difference (2,62x3,52mm instead of 4,15x4,15mm). It seems to me that this is an exception when comparing it to a random selection of other devices with an S-PVQFN-Nxx package (which all have a square thermal pad).
As a quick fix I suggest to rename the old _[Texas_PVQFN-N40](https://github.com/KiCad/kicad-footprints/blob/master/Package_DFN_QFN.pretty/Texas_PVQFN-N40.kicad_mod)_ to _Texas_S-PVQFN-N40_EP2.62x3.52mm_ and leave all others as they are (i.e. not adding EP suffixes for square thermal pads). I can provide a patch if this finds your approval.

**MSP430G2x55**
s-pvqfn-n40_msp430g2x55

**DRV8303**
s-pvqfn-n40_drv8308

SchrodingersGat:
Sorry I must have missed your first question. That's really strange, perhaps it is worth trying to contact ti for clarification. 

In any case your suggested fix is good. I would suggest adding the exposed pad size explicitly for both versions.

aimylios:
OK, sounds good! I'll open another PR to fix the naming within the next days.

SchrodingersGat:
Thanks @aimylios :)

PR #109, opened 2017-11-25
Add net ties

evanshultz:
Adding net tie footprints as discussed at https://github.com/KiCad/kicad-library/pull/1472.

This is for discussion purposes right now. There should not be merged as-is. I realize the descriptions are bad, line endings are not Unix, no connecting copper, etc. I would like to discuss the following (and any other points):
1. I chose `Net-Tie_{pin count}_{SMD|THT}_{pad diameter}`. Do we need to indicate pad shape?
1. THT pads are circular.
1. They are all virtual components.
1. They don't have F.Fab ref des.
1. All pads are covered with mask.
1. A zone will be added on front (SMD) or front and back (THT).
1. SMD pads have no paste which seems fine to me.

I'm planning to add 3- and 4-pin footprints in the same pad sizes once we decide on the above, giving 12 footprints total.

Image:
image


SchrodingersGat:
It was my understanding that these Net Tie footprints should have a copper connection between the two pads? Otherwise aren't they just essentially a jumper? How do you envision these being used?

jkriege2:
PS: new KLC-ckeck-scripts work better for virtual components as these (no more errors for missing F.Fab-drawing etc.)

evanshultz:
@SchrodingersGat 
As mentioned above, these footprints aren't finished. I specifically listed the lack of connecting copper.

I would like feedback on the enumerated list of items above so that I can finish off these footprints. Any comments there? Here are my reasons:
1. Seems OK to me. See below as well.
1. Circular allows the same thickness (resistance, inductance) or connecting copper but is smaller.
1. Yes.
1. Yes.
1. Seems fine since no components mounts here and it would avoid shorting.
1. As noted above.
1. Again, no components are mounted on these footprints.

evanshultz:
I've added copper and updated the descriptions for all net ties initially imported to avoid further delays. Please help to answer the questions above, and add any others you may have.

Is there a way to make a plated THT without any annular ring? In this case  we only need a plated barrel, but I believe an annular ring is required to get plating in the barrel. It's OK as it is, just using the minimum annular ring diameter, but I'm asking in case that drill hole has a place here or elsewhere.

@SchrodingersGat @jkriege2 
Can you take a look at the check scripts? They're not running here, so I've done something wrong with these footprints or maybe a recent change to the scripts broke something?

jkriege2:
Hi

1. OK for me
2. OK
3. Yes
4. OK
5. OK
6. OK (with zome you meant CrtYd?)
7. Having paste does not make sense if they are covered by varnish ... right?

SchrodingersGat:
> As mentioned above, these footprints aren't finished. I specifically listed the lack of connecting copper.

You're right, I must have read over that bit.

evanshultz:
@jkriege2 
6. I mean a copper zone.
7. Yes, that's what I thought.

Any further feedback? If not, I'll build out the other footprints and then advise when this can be reviewed and merged.

SchrodingersGat:
@jkriege2 fixed KLC bug - https://github.com/KiCad/kicad-library-utils/pull/187

SchrodingersGat:
Looks good to me :)

evanshultz:
@jkriege2 @SchrodingersGat Thanks! I'll add the other footprints soon.

Right now the check scripts generate a warning if a virtual component is missing a 3D model. Is that correct, or can we say that a virtual component has no 3D model?

SchrodingersGat:
Yeah for a virtual component it only generates a warning, not an error. You can ignore this :)

jkriege2:
Maybe we can even reformulate the warning to suggest to check whether this is correct, as we do in this warning for F6.1: `Footprint placement type set to 'virtual' - ensure this is correct!`

... i.e. we could warn: `No 3D model provided for virtual component (3D models are not required for virtuals). Check whether this is correct.`? Maybe this makes it easier for the user to see that this can be ignored!

evanshultz:
All new net ties added and ready for review and merge.

I went with a linear arrangement for 3-pin and a square arrangement for 4-pin. Image:
image

jkriege2:
Great work!

I was thinking a bit about the naming ... Seeing the names I somehow expect the given length to be the size of the whole thing ... maybe a naming scheme like this would be better:
  NetTie-N_[THT|SMD]_<WIDTH>x<HEIGHT>mm_Pad<PAD_DIAMETER>mm

Best,
JAN

evanshultz:
@jkriege2 I choose 2x the pad diameter as the spacing. That gives a small spacing for small ties, which are probably logic, and a bigger spacing for bigger ties which probably want higher current. I can change the name if that makes more sense.

I did not add copper on top and bottom for THT. Thanks! I will change that shortly...

evanshultz:
@jkriege2 I think all should be updated now. Footprints renamed and I also added bottom copper for THT net ties. Since that doesn't change the look of the footprints above, no new screenshot. Anything else?

SchrodingersGat:
Please add `NetTie` to the library table and we should be good! These look nice, great work :)

evanshultz:
@SchrodingersGat Thanks!

Unfortunately, I'm having trouble. I can push fp-lib-table from 'master' without issue, but whenever I add the NetTie lib entry I get a conflict. I tried pushing the file change from my PC and editing in the web interface just for kicks. No dice. What am I doing wrong?

poeschlr:
I had a similar problem today. Maybe my fix made it worse for everyone else. (Worst case scenario would be to copy your current files into a newly created branch that is in sync with the master.)

SchrodingersGat:
Thanks @evanshultz I fixed the conflicts and merged :0

PR #110, opened 2017-11-25
added fixed, pre-existing footprints for CPU-Fan headers

jkriege2:
- matching 3D models: https://github.com/KiCad/kicad-packages3D/pull/163, https://github.com/KiCad/kicad-packages3D/pull/163
- footprints check out OK for KLC ... but script crashes for some reason

2017-11-25 22_21_30-footprint editor no active library
2017-11-25 22_21_14-footprint editor no active library


poeschlr:
Add the connector lib to the fp-lib-table and the remaining travis error should go away.

Have these footprints existed in the old lib or do i need to check them in detail?

jkriege2:
They were pre-existing ... I will add to the fp-lib-table this evening.

JAN

evanshultz:
Do we need these connectors? They have no datasheets and don't appear to be specific connectors. We do have other headers which are very similar, if not identical, like Molex KK. Is there something specific or unique about these headers that warrants keeping them?

jkriege2:
These seem to be the connectors specified by Intel (see last pages of http://www.formfactors.org/developer%5Cspecs%5Crev1_2_public.pdf) for usage as CPU fan headers on PC mainboards ... so I think we should keep them.

Best,
JAN

evanshultz:
@jkriege2 Excellent. So these aren't generic connectors as page 17 gives the MPNs. Then I'd say we should make specific connectors instead of generic ones, and add comments about this CPU application in the description and/or keywords. What do you think?

jkriege2:
Hmmm ... I think these are more generic ... the PDF sais:
`... Wieson part number 2366C888-007 Molex 47053-1000, Foxconn HF27040-M1, Tyco 1470947-1 or equivalent ...`

... so all these manufacturer produce these pin-headers ... so shouldn't we leave them a s generic? Not sure about the name though.

jkriege2:
PS: fp-lib-table entry will be added by https://github.com/KiCad/kicad-footprints/pull/116

evanshultz:
@SchrodingersGat @poeschlr 
Anybody else want to comment? I think these connectors are as good as they're going to get, but I still lean away from adding such a generic connector. I'd rather rename these headers and make them (or just one) atomic and then let the user find the MPN in our library.

poeschlr:
Are these available in compatible form by many manufacturers? If so keep them generic. If not name them with Fan_man_mpn_1xN_Vertical.

In any case, _Vertical is missing. (At least in the screenshots)
This is required because i think we do not want to check every contribution if there is any connector similar to the contributed connector that has a different orientation.

jkriege2:
They are available from many manufacturer (at least for the 1x04 I listed several in teh description) and are very common on mainboards. Naming is already:
- `FanPinHeader_1x03_P2.54mm_Vertical`
- `FanPinHeader_1x04_P2.54mm_Vertical`
... so that should be OK and the screenshots are outdated ;-)

poeschlr:
In that case, thanks for contributing

PR #111, opened 2017-11-26
Moving several of the remainning footprints from connectors.pretty

jkriege2:
- with several KLC-fixes and moderinzations ...
- many still remain (see https://github.com/KiCad/kicad-footprints/blob/master/TODO.md)
- move of some 3D models: https://github.com/KiCad/kicad-packages3D-source/pull/67, https://github.com/KiCad/kicad-packages3D/pull/165
- one family is moved here: https://github.com/KiCad/kicad-footprints/pull/110
- remaining KL-errors are due to over-checking virtual components: This should be fixed by https://github.com/KiCad/kicad-library-utils/pull/182


poeschlr:
Is it on purpose that this deletes 3 footprints from Conn_Card?

jkriege2:
They were not deleted, but renamed (added the Prefix `microSDCardReceptacle_`) ... Github sometimes does not notice a rename, if (in addition) the contents of the files changes (3D model links!)

JAN

poeschlr:
That's a very long prefix.

In https://github.com/KiCad/kicad-library/issues/1447#issuecomment-335434885 i suggested:
 `SD_[Standard]{_[Size]}{_UHS_[UHS Type]}`
   * Standard: SC, HC, XC or IO (Specification of your component. Yes some of them are Backwards compatible, give the highest standard specified.)
   * Size: None = normal, Mini, Micro
   * For UHS enabled components the UHS type needs to be given.

jkriege2:
Hhmmm ...

`SD_HC_micro_...` seems a bit messy to read (IMHO) ... Why not go with:
- `SD[STANDARD][_UHC]`
- `microSD[STANDARD][_UHC]`

Then we  would e.g. get `microSDHC_Wuerth_...` ... I think it's more common to search for a `microSD` instead of `SDHC_micro` What do you think?

JAN

jkriege2:
Also: having normal-sized as normal would lead to `SD_HC_normal_UHS_...` ... where (at least) I would not know what `standard` stands for ...

Best,
JAN

SchrodingersGat:
Jan the word "normal" would not actually appear in the footprint names. So it would appear as `SD_HC_UHS`.

I agree with Rene that your longer names are too wordy - we had this discussion in particular at length for the new KLC - http://kicad-pcb.org/libraries/klc/F3.6/

Further notes:

* `SolderJumper` library name should be `Jumper` 
* Does the TRANSFO in the RF footprints mean **Transformer**? If so, it should be changed to **Transformer**. Capitalization should only be for company names (where required) or for acronyms.

jkriege2:
Good Morning ;-p

1. I will rename the SolderJumper.pretty ... should I also rename the footprints (I think they can also use an overhaul at some point) with a simpe prefix `Jumper_`, or should we differentiate inside the lib with different prefixes?
2. I think so (at least they match a symbol with transformer) ... but since these ar not from me and had/have very few documentation, I don't really know ... als the v1/v2 nomenclature is not that good ... I CAN do the rename, but I don't know enough about them to rework them properly.
3. About naming, I would at least move the [micro|nano] to the front (i.e. `[micro|nano][SD|HC|XC|...][_UHC]...`) to make the whole think more readable (as I said, I think microSD is typically used as a name, not seen as two separate things, so I think people will wonder where the microSD-footprints are, if they are hidden as `SD_HC_micro_...`. I think with all these naming things, we should also keep in mind, what people will EXPECT to see and how certain things are USED in comparison to what is the exact ontology of the thing. In the end this lib is thought to be used by humans ... Filtering is also fine with my proposal, as the microSD and SD have different pinouts!

Anything else? Note that this PR does not move ALL connectors, as for some I have no clue on where to put the, but the remainder should finally be handled in a separate PR (volunteer???)

Best,
JAN

PS: Just for info: I'm also currently working on a remake of the different types of TO-92 3D models and will also rework naming and footprints ... give me a few more days on that ;-)

evanshultz:
I agree that `microSD` is the common way to denote this type of connector. I suggested something similar in the like Rene posted and I still think it's more salient.

jkriege2:
Hi!

- I renamed to `microSD_HC_Hirose_DM3AT-SF-PEJM5` and the like now.  and removed the completely unspecified SDCardReceptable (wouldn't know hwere to buy a matching part???)
- I renamed SolderJumper.pretty to Jumper.pretty
- I added entries to fp-lib-table

poeschlr:
Two of the three barrel jack footprints are missing at least one pin 1 marker
All Conn_Audio connectors are missing pin 1 markers.
Jack_Horizontal in Conn_Audio has very large pads. Might be a good idea to change them to be elongated instead of circiular/square to increase pad to pad clearance.
Do we have any idea what BNC_Vertical represents?
Might it be a good idea to add a mask cutout between the pads for the solder jumpers?

We can merge as is and document my remarks above in an issue if you want.

jkriege2:
Hmmm ... wait until this evening. If I find some time, I will fiy at least a few of the topics above. The BNC_Vertical is a BNC-connector meant for mounting (with a nut) into sheet metal (see old 3D mdoel):
2017-11-28 07_13_55-3d viewer bnc

I didn't find a good naming for such a thing. Any ideas? Basically I think it is a good idea to have a footprint like that!

jkriege2:
Hi!

I fixed everything except the solder jumpers. I'm planning to rework that lib at some point (but not today). So can it be merged for now?

Best,
JAN

jkriege2:
PS: Should I still rename the BNC_Vertical? Ideas for a good name? @SchrodingersGat ?

poeschlr:
For the BNC Vertical, it would be a good idea to have a bit more information in the description field. One example part number might be a good start. (user can then go from there.) 
It should also state that this is a footprint for components normally intended for panel mounting.

jkriege2:
Remaing KLC-errors: roundrect-pads (will be available in v5) not allow: will be fixed by this PR for the scripts: https://github.com/KiCad/kicad-library-utils/pull/192

Everything else OK like that?

jkriege2:
PS: I renamed the BNC to `BNC_PanelMountable_Vertical` and added explanation to the description ... better like that?

SchrodingersGat:
Jan I think that `_PanelMount_` is superior to `_PanelMountable_` - it is more in keeping with the language convention. :)

jkriege2:
Yes, but wouldn't one expect with `_PanelMount_` something which is for a PCB close to the panel and the connector mounted in the panel, but also soldered to thePCB, whereas I chose `_PanelMountale_` to state the THE CONNECTOR is panel-mountable and is more or less misused ... but I'm really not sure whether this distinguation comes over clear by the name (it seems not ;-p

Should I rename?

poeschlr:
I think the name PanelMountable is ok. I can't think of a better way to communicate what this footprint is meant to represent.

An idea would be to add a hexagon on b.fab to signify the nut? This should at least notify users who already selected the footprint once they have it on the board. (or with the new component chooser/footprint browser it is already possible to see this in the preview.)

I think this can be merged as is. (@jkriege2 do you want to further play with this footprint or is it ok to merge now?)

jkriege2:
Fine with me ... if I should add the nut, wait until tomorrow with merging, otherwise just merge an add an issue (I could then also add footprints like this for SMA, SMB, ...) ... i.e. up to you!

JAN

poeschlr:
I will merge it for now and make an issue. Thanks for the time you put into this already.

jkriege2:
Thanks for review and merging ;-)

PR #112, opened 2017-11-26
moved/fixed/renamed pre-existing optoDevice-footprints

jkriege2:
- move 3D models: https://github.com/KiCad/kicad-packages3D/pull/166
- move 3D model sources: https://github.com/KiCad/kicad-packages3D-source/pull/68
- remaining KLC-errors. annular rings on very small Laser-Diode packages ... should be ignored!


poeschlr:
Should the lib name be Sensor_Optical? (Or are these non sensor devices?)

From a cursory look these look ok. Only thing i could see is that for some of them the second fab reference interferes with other fab drawings. (I would make an issue about it and mark it as a `good first issue`.)

jkriege2:
They are not all sensors ... e.g. laser diodes are not ... also some special opto-isolators are not ...

For now I would stick to optoDevice.pretty and maybe split later?

jkriege2:
Also naming in this lib could use a rework (prefixes and all), but since my time is a bit limited that will have to be done in a separate PR.

poeschlr:
Thanks for the clarification. (And for your continued work of curse.)

PR #114, opened 2017-11-26
added/fixed pre-existing footprints, that were not added when moving

jkriege2:
These were left out, when moving the major part of this lib here 

3D models: https://github.com/KiCad/kicad-packages3D/pull/168 and https://github.com/KiCad/kicad-packages3D-source/pull/70

PS: Should we split this directory into Crystal.pretty and Resonator.pretty (would need to be done in a separate PR, once this has been merged)?

2017-11-26 15_49_02-3d viewer
2017-11-26 15_48_44-3d viewer
2017-11-26 15_43_37-3d viewer
2017-11-26 15_43_20-3d viewer
2017-11-26 15_43_05-3d viewer
2017-11-26 15_42_25-3d viewer
2017-11-26 15_42_10-3d viewer


SchrodingersGat:
Thanks Jan!

Regarding your Crystal / Resonator question - are there any footprints which are "common" to both functions? Or are they completely separate footprints? Can you provide some examples?

jkriege2:
I think they are all separate ... all crystals that I know of are in those metal-cans, while the ceramic resonators are typically in molded packages (drop-style like tantal caps).

cu
JAN

SchrodingersGat:
Fair enough, sounds like the could be reasonably separated.

PR #115, opened 2017-11-26
Adds footprint for 3M 40-pin ZIF socket

RobertBaruch:
This is a footprint for 3M's 40-pin zero insertion force socket. Note that the courtyard includes an area for the lever. The silkscreen shows the lever (in the up position) for easy recognition and placement.

[Datasheet](http://multimedia.3m.com/mws/media/99720O/3mtm-textooltm-universal-zip-dip-socket-ts0392.pdf)

Note that in the image below, part of the courtyard for the lever is obscured by the Y axis line.

zipimage


SchrodingersGat:
Hi @RobertBaruch thanks for the contribution. We are organizing the libraries by function (not by manufacturer) so we need to address the naming scheme for this library.

The library name should also be "singular" instead of plural.

To meet both these requirements, I would suggest simply `Socket.pretty` as the library name.

Then, prefix the manufacturer name to the footprint i.e.

`3M_Textool_240-1288....`

poeschlr:
@SchrodingersGat do we want separate libs for such sockets or should these reside within the libs for the IC packages these sockets are designed for? (In this case Packages_DIP)

SchrodingersGat:
@poeschlr that's a fair idea. I'm happy to defer to this if people are happy. @evanshultz @jkriege2 care to chime in?

poeschlr:
About the footprint it self.
The drill size seems to be a bit strange to me (admittedly i think in metric). I would go with 0.95mm or even 1mm.

I would also make the pads oval with x dimension 2mm, y dimension 1.44mm
This would allow a larger annular "ring" in y direction (~0.5mm) but also a larger pad to pad (1.1mm) clearance. (This should allow for two 0.25mm traces with trace to trace and trace to pad clearance of 0.2mm)

screenshot from 2017-11-26 23-55-14

The body size (fab outline) is now quite right either. According to the datasheet it should be 22.9x65.8mm. Your footprint is 22.86x65.786mm.

screenshot from 2017-11-27 00-05-29


The silk outline should be moved towards the outside such that it is fully outside the body (fab outline)
Add a pin 1 marker on silk, visible after assembly.

RobertBaruch:
Thanks! I worked off the inch measurements, which is likely why the dimensions are off slightly in mm. I'll definitely fix up the pad dimension and shape. Thanks!

RobertBaruch:
Showing the majority of the changes:

zipimage

jkriege2:
Anou the lib ... that's hard ...

For the DIPs we have socketed variants (or sockets) in the same repo with added modifiers in the name ... but then I mostly expected those to be socketed ICs, not a mere sockets without an IC (which it could still be used to represent). The ZIF-sockets are used as test-sockets, i.e. you put in and remove diferent ICs for test, so they may not be permanently occupied ... so maybe a separate LIB `TestSockets.pretty`?

... but I'm not (yet) totally convinced of my arguments ;-)

SchrodingersGat:
After more consideration I am now leaning more towards separating these from the DIP packages. They are functionally different (e.g. difference between a hand and a glove!)

Options:

* `Socket.pretty`
* `Socket_DIP.pretty`

I would lean towards `Socket.pretty` and then break that into smaller libs if it ever gets too large...

jkriege2:
Fine with me ... should we also move all DIP-sockets there?
AND: Do we interpret them as populated or unpopulated? (This might be more for the sake of showing the correct 3D model ... so it's more or less like with the different LED colors ;-)

cu
JAN

evanshultz:
@jkriege2 Good question. I say unpopulated. There are 7-segment LEDs that fit in DIP packages, or it could be for 1/4W resistors, for example, so we can't even assume a plastic IC package (though that's surely most common).

jkriege2:
Yes, ... but then we maybe need two sets of footprints: One for unpopulated and one for sockets, populated with ICs ... what do you think?

jkriege2:
We could simply generate the sockets a second time with the script used to initially make them and give dem different names and a different repo?

SchrodingersGat:
They could be in the same library, but perhaps with a suffix e.g. `Populated` (or something)

evanshultz:
Yeah, that sounds better. And since they could be populated with different things, perhaps the suffix should be `IC`, `LED`, etc. `IC` would seem like the default, but it's certainly not the only thing one could stuff into the socket.

SchrodingersGat:
I think it would be handy to also draw the open-position lever on the `F.Fab` layer. Without it, it appears that the extra courtyard is there for no reason

RobertBaruch:
I agree, the extra courtyard is a bit bizarre. However, you can still put components in there, provided they have a maximum height of 4 mm. Putting the lever on the `F.Fab` layer would kind of get in the way of any components you want to put there. Maybe just a note saying what the extra courtyard is for?

SchrodingersGat:
Perhaps then this extra courtyard should actually be on the drawings later. If it is on courtyard it can cause DRC violations with any components placed "under" it...

RobertBaruch:
Hmm, I guess in that case I should the extra courtyard, and I think I should add the lever in the fab layer. If you really want to put components under there, you can always move the labels around yourself to make them more legible. Does that sound like a plan?

RobertBaruch:
It does look a lot nicer now.

zipimage

RobertBaruch:
Just saying, I appreciate the thorough review from all of you.

SchrodingersGat:
Thanks for the contribution @RobertBaruch  :)

PR #122, opened 2017-11-27
Update all variants of TI S-PVQFN-N40

aimylios:
As already discussed in PR #108 I renamed all existing variants of the S-PVQFN-N40 package and appended the size of the exposed pad to the part name. While doing this I recognised that the old footprint did not strictly follow the datasheet recommendations, and updated it accordingly (see pictures below).

s-pvqfn-n40
s-pvqfn-n40_thermalvias


evanshultz:
@aimylios 
Thanks for catching this issue and making the fixes!

Some review feedback:
- Include screenshots of the other footprints as well so it's all together and easy to see.
- Instead of oval pads, you can create pads just as TI requests. Overlap a circular and a rectangular pad.
- Set the mask and paste margins for the whole footprint instead of each pad.
- The outer 40 pads don't need their paste size defined on the pads.
- The 3D models could be updated since they are, in fact, unique physical packages. (optional)

aimylios:
@evanshultz, thank you for reviewing my patch. I have reworked the footprints based on your recommendations. While at it, I also added another footprint variant for the very same package. Please let me know if anything else needs to be done.

To be honest, I do not like these overlapping pads at all. I chose oval pads for a reason, as I consider the small drawback of having slightly longer pads (which will only be an issue with very dense layouts) to be negligible compared to the added complexity and ugliness of overlapping pad shapes (which show duplicated pad numbers, make your cursor snap in unwanted positions and cause weird effects on mask layers). But if they are required to get this PR merged, so be it. I'll keep a copy of my original footprint for personal use.

texas_s-pvqfn-n40

SchrodingersGat:
@aimylios thanks for this contribution, these look very good :)

PR #123, opened 2017-11-27
Add fuse footprints 

poeschlr:
newly generated smd fuses (standard sizes) plus renamed footprints from old repo.

For old footprints i added a second reference designator and i added 3d model references where they where missing.


poeschlr:
Script for generic fuse generator: https://github.com/pointhi/kicad-footprint-generator/pull/85

SchrodingersGat:
Nice work - looks good :)

PR #125, opened 2017-11-28
Renamed Conn_ to Connector_

SchrodingersGat:
We have only one type of footprint library that is abbreviated: `Conn_`

To prevent confusion in the future, let's fix this inconsistency now, and do not have abbreviations.

I have renamed all `Conn_` to `Connector_`


poeschlr:
I am ok with this. I will update my scripts such that they are up to date. (But i will not regenerate them.)
Luckily this is done in a central config file.

SchrodingersGat:
Thanks :)

PR #128, opened 2017-11-28
Add keepout area to DF13C connector footprints

poeschlr:
Ok i can not leave these footprints as they are.
The DF13C now also has the keepout area defined.

screenshot from 2017-11-28 19-28-34

I will also add a better FH12 here. (outline that more closely represents the connector)


poeschlr:
Now the FH12 connector footprints have a more detailed outline.
I also had the wrong name for the FH12 connectors.
screenshot from 2017-11-28 20-42-26

PR #129, opened 2017-11-28
add reverse mount 1206 SMD LED

descampsa:
Footprint from kingbright datasheet, except hole is circular instead of rectangular.
http://www.kingbright.com/attachments/file/psearch/000/00/00/KPTR-3216SYCK(Ver.15B).pdf

I also have a (very basic) 3d model, but i am not sure where i am supposed to put it? 
The kicad-packages3D repository does not have a LED_SMD directory.


poeschlr:
@descampsa wrote:
> The kicad-packages3D repository does not have a LED_SMD directory.

You are allowed to create it ;)

SchrodingersGat:
Nice footprint, thanks :)

PR #131, opened 2017-11-28
Rework TO-92

jkriege2:
This is an effort to remake all TO-92 (and variants) 3D models and to clean up naming a bit. I simplified the nomencature a bit:
  ``TO-92[L|S|Mini][-2][_Inline][_Narrow|_Wide][_Horizontal(1|2)]``

i.e.:
- L/S/Mini: package size modifier
- -2 pin-cpunt for 2-pin packages
- _Inline inline-pins vs. pins on circle: 
2017-11-28 23_13_52-pcbnew d__temp_test kicad_pcb
- Narrow/Wide: pin-distance: 
2017-11-28 23_14_09-pcbnew d__temp_test kicad_pcb
- Horizontal: Horizontal-mount, different sides up (1 or 2)

2017-12-06 19_20_22-pcbnew d__temp_test kicad_pcb
2017-12-06 19_20_05-3d viewer

What do you all think about this nomenclature and the look?


evanshultz:
The look is beautiful. About naming:
- Would it better to specify the pin number regardless? So use both `-2` and `-3`?
- I don't think it's good to use `_Narrow`. I've seen the rear pin stick out and also the front one. We could do like the TO220 and use `Staggered1` and `Staggered2` or something else suitable.
- Should pad shape be captured to allow easier hand soldering or pad shape variations?
- Some ref ds are rotated.

jkriege2:
Hi!
Thanks for the feedback:
1. Hmmm ... I think TO-92 is kind of a household name for the 3-pin variant ... I would like to drop the pin-count there ... but I'm not very passionate about that point. If we add the `-2`, we should also add it to TO-220 and the like, where it is also missing for the standard case (3-pins), I think.
2. `_Narrow`just encodes whether the pin pitch is 1.27mm (in one direction between any two pins), or 2.54mm (`_Wide`). Here's an illustration of the naming scheme: 
2017-11-29 07_00_13-pcbnew d__temp_test kicad_pcb

3. Abou the question where the center pin is bent towards: Currently I'm only doing the standard version (bends towards the front/rounded side of the package). There are `_Reversed`-variants in the old libs, which have the center pin pent towards the flat side of the package, but I didn't find any datasheet using them. Should we have them still (I would have opted for removal)? And if we want them: I'm not sure about the naming (`_Pin2Backwards`or something?). Ideas?
4. For now I omitted all the pad-variants (we currently have `_Rugged`and `_Oval`). I wanted to make the basics right and then devise one `_HandSoldering`-variant from there (where it makes sense).
5. Yes I noticed, these have already been fixed (old screenshots ;-)

Best,
JAN

jkriege2:
Hmmm .... here's a better overview:
2017-11-29 07_03_00-pcbnew d__temp_test kicad_pcb

Looking at it: Maybe we should drop the `_Narrow`altogether and see it as standard and then only have `_Wide` as modifier for pins with 2.54mm pitch. What do you think?

poeschlr:
Or better still include the pitch? So instead of _wide -> _P2.54mm

jkriege2:
But the pitch os not so well defined ... The question is what people think of as the "pitch" (I would expect the distance between pins, which is for pins on circle sqrt(x^2+y^2)=1.79!=1.27 for pins on circle, but would be 1.27 for inline) ... I think people won't have a natural definition for pitch for these packages, because the pins do not always forma line.

Best,
JAN

poeschlr:
You are right about that. Then i agree with making narrow the standard and using wide as the modifier.

SchrodingersGat:
I agree with this - normal is e.g. `TO-92-2` and wide is `TO-92-2_Wide`

jkriege2:
Hi!

so, some prgress here ... I added some new variants (TO-92Flat) and the Horizontal1/2-variants of the inline:

2017-11-30 21_57_54-pcbnew d__temp_test kicad_pcb
2017-11-30 21_57_42-3d viewer


Now I will be away for the weekend, so no progress before next week, But I would like to know what you think on these questions:

1. Do we also need the horizontal variants with pins on a circle
2017-11-30 22_01_43-footprint editor no active library
2. Do we need the horizontal variants for the TO-92L (I think no)
3. Do we need the reverse-type packages we had before (pin 2 sticks out to the back, not the front)? I think no, as they are not really available commerically ... at elast I haven't seen something like that
4. Where possible I will add ONE variant for hand-soldering. OK?
5. Am I missing any other variant?
6. Should I add SOT-33: http://www.icinform.ru/sites/default/files/iibase/corp_b/157_b.png (outdated I think)
7. Should I add SOT-25: http://www.icinform.ru/sites/default/files/iibase/corp_b/156_b.png (outdated I think)
8. Do we want some kind of PCB-edge SMD footprint, for mounting comparable to https://asset.re-in.de/isa/160267/c1/-/de/118028_LB_00_FB/H-Tronic-Temperaturgesteuerte-Luefter-Regelung-Baustein-12-V-DC-20-bis-50-C.jpg?y=155&align=center or https://files.elv.com/bilder/artikel/Produkte/5/523/52313/Internet/normalneu/52313_w01_temperaturschalte.jpg

Best,
JAN

poeschlr:
What is the difference between normal and handsolder version? 
Does it make sense to have non "handsolder" versions for THT? I tried to make all connectors such that the pad to pad clearance is high enough for a trace in between but still enough space to handsolder them. (In most cases i tried to have 0.5mm copper in y direction, in x direction at least the 0.15. This should be easily solder able by hand and it should be small enough for wafe soldering applications.)

Reverse type: Yes add it.

I would not bother adding outdated versions. If somebody comes along and contributes them i would still allow them to be merged.

About the edge mount. I would not bother for now. Again i would be prepared to merge it if somebody contributes them.

jkriege2:
## Handsolder
There were `_Oval`-variants, which had oval, enlarged pads (+ the rugged version, which I planned to remove), which hade REALLY HUGE pads ;-)

I would have added the oval/elongated-pad variants, as they are usually easier to solder, as you can more easily transfer heat from a (large-tip) iron to the pad ;-) ... I'll propose them and we can discuss.

PS: I also don't really have problems with SSOPs and TQFPs ... but maybe that's not the standard, especially not for hobbyists.

For reference: These are the xisting variants:
2017-12-03 20_01_41-pcbnew d__temp_test2_test2 kicad_pcb



## Reverse
About the reverse type: Any idea, where they are used? Just to be clear, this is the reverse type I'm talking about:
2017-12-03 20_05_18-pcbnew d__temp_test2_test2 kicad_pcb

jkriege2:
Here's my proposal for handsoldering-variants (where necessary

2017-12-06 07_45_15-pcbnew d__temp_test kicad_pcb

jkriege2:
@poeschlr @SchrodingersGat What do you think on the hand-soldering variants above?

jkriege2:
OK, here's the current status. Only few 3D models are still missing:
2017-12-06 08_27_56-pcbnew d__temp_test kicad_pcb
2017-12-06 08_27_24-3d viewer

KLC-status seems OK, up to rounded-rect pads (which are OK in KiCAD v5) and min. copper on a very tiny variants (which cannot be helped).

Comments so far?

poeschlr:
From the screenshots these look ok for me. (My current pc runs 4.0.7 so i can't look at the files. Will look at it later when i am back at university.)

To be honest the handsolder versions from your first post today look a bit excessive. >1mm elongation? I think half of that should be more than enough. (Assuming stepup did not totally screw up the rounded rectangles.)
screenshot from 2017-12-06 11-08-45

jkriege2:
2017-12-06 19_20_22-pcbnew d__temp_test kicad_pcb
2017-12-06 19_20_05-3d viewer

Hi @poeschlr @evanshultz @SchrodingersGat  !

I made the HandSoldering-pads smaller, added 3D mdoels (and some footprints) for the horizontal variants and added 45° variants, which can be stacked closer together than unrotated (+KiCAD cannot rotate a footprint easily by 45° correct?)

This should be ready for merge now!

3D models are here:

Best,
JAN

evanshultz:
> I made the HandSoldering-pads smaller, added 3D mdoels (and some footprints) for the horizontal variants and added 45° variants, which can be stacked closer together than unrotated (+KiCAD cannot rotate a footprint easily by 45° correct?)

What?? Works fine for me! See:
image

Am I missing something?

jkriege2:
Ahh ... didn't nkow that ... so I removed those variants (even better!)

Anything else?

Best,
JAN

jkriege2:
Oh, I forgot the reverse ones ... will add in a few minutes!

jkriege2:
OK, I also added the pin2-reversed-variants: 
2017-12-06 21_46_45-pcbnew d__temp_test kicad_pcb
2017-12-06 21_46_32-3d viewer

2017-12-06 21_43_19-3d viewer
2017-12-06 21_36_57-3d viewer
2017-12-06 21_34_51-3d viewer

evanshultz:
What about adding pin count on all footprints? Not much discussion above. At https://github.com/KiCad/TO_SOT_Packages_THT.pretty/pull/29 and the linked case we added pin count for all TO220 packages, even 3 pin.

I agree that the handsoldering pads looks excessively big. These are easy parts to solder even with normal pads. What is the criteria for choosing a non-standard (not from a datasheet) pad size?

Is there any reason to not use new pad shapes? Just wondering if we should concern ourselves with v4 compatibility. I don't think so, but maybe there is a reason.

Regarding the middle pin going forwards or backwards: Lead prepping machines can do whatever someone wants, but perhaps we should only stick to lead positions that can be purchased directly and documented in a datasheet. Otherwise, how can we know that footprint is useful? If a KiCad user wants some other custom lead prep, they can make a suitable footprint. I personally like having the options, but I'm wondering if the footprints will be used. Does that sound reasonable? What do you all think?

For the horizontal footprints, nothing in F.Fab indicates the position. Can the footprints help to show the user how to install the part? Even with a unique pin 1 shape we will still mark pin 1 on F.Fab for many footprints. Perhaps add extra lines to mark the sides of the "flat face"? I'll leave that an open question.

Lastly, is there a better suffix than `_HorizontalX`? It's so generic. The distance from the body to the pads is one unique and defining characteristic of the footprint, as is the orientation of the body (flat side up or down). I wonder if we should be more specific now instead of so general and non-descriptive. There was discussion about similar footprint naming for TO220 at the PR I linked earlier.

jkriege2:
Hi!

some answers to your questions above:

1. I would not add the pincount 3, as I think TO-92 is really a househld-name for the 3-pin variant
2. See the hand-soldering as an experiment ... I'm (as said above) not totally clear whether we need them, especially since the pads got a bit larger overall, which make them easier to solder. I was thinking about beginners with soldering that might habe problems getting small pads hot enough and to control their iron .... so there the asymmetrically elongated pads may help ... Note that I already reduced their size a bit
2017-12-07 07_12_57-footprint editor no active library
(@SchrodingersGat @poeschlr What do you think? If we all agree they are not really necessary, I can also remove them ... up to you all)
3. About new pad-shapes:??? I use new rounded-rects in several footprints (that's why KLC complains at the moment!)
4. What about one or both of these markings (silk lines withing the body to indicate the "edges" and line next to pin1 as marker ... als a small triangle as for many  connectors might be a good choice) for the horizontals If wer agree on something I can change them all!
2017-12-07 07_09_46-3d viewer
2017-12-07 07_09_30-footprint editor no active library
2017-12-07 07_08_16-3d viewer
2017-12-07 07_07_58-footprint editor no active library
2017-12-07 07_17_21-3d viewer
2017-12-07 07_17_07-footprint editor no active library

5. About naming ... happy to have proposals (for now I went with the old naming scheme here) ... What about `TO-92_W4.0mm_StaggerEven_Horizontal_FlatSideUp` instead of `TO-92_Horizontal2` and `TO-92_W4.0mm_StaggerEven_Horizontal_FlatSideDown` instead of `TO-92_Horizontal1`? Too lengthy?

Best,
JAN

jkriege2:
Ah, I missed your comment on the back-pointing pin2 ... I wanted to remove them from the lib (we had them before), but remade them on a request from @poeschlr ... I'm not really sure what we need them for, but since they are already done (with 3D models) and some see a need for them, we can as well keep them (they are only few anyways).

Best,
JAN

SchrodingersGat:
Maybe I missed a discussion on this, but `BackPointing` is a very strange way of stating the function. Are there some other options for this naming?

jkriege2:
THe old name was `_Reverse`, but that's IMHO even more obscure ... any ideas?

poeschlr:
About the hand solder variants, use the same suffix as we used in other footprints: `_HandSolder` not HandSoldering

jkriege2:
@poeschlr @SchrodingersGat @evanshultz What about my other questions above?

poeschlr:
For pin 1 marker maybe draw an L around the pin instead a simple line. But the triangle is also ok for me. (comes down to artistic preference.)

Rounded rect is ok to use. (These footprints are intended for v5.)

Handsolder version: As stated above, i don't think we need them. But i would prepared to keep them with the slightly reduced pads as you have them now. (Your argument about beginners is a valid one.)

The lines to indicate the edges are ok for me

Reverse type: If both of you guys think they are unnecessary i would be prepared to revert my above statement. (I would accept the decision that they get dropped.)

jkriege2:
Hi!

OK, I decided to use the small triangle as pin1-marker for hor. mounted packages, Also I renamed the horizontals:
2017-12-10 16_03_33-footprint editor no active library

Now only three questions remain:
1. do we keep the reverse-types
2. if we keep the reverse, how should we name them? `_Pin2BackPointing` is uggly, better proposals?
3. Do we need the pin1-marker on the vertically mounted devices?
4. naming OK?

Best,
JAN

SchrodingersGat:
Does anyone have a strong reason to keep this reversed type? I don't really mind either way but it would certainly be easier not to include them!

jkriege2:
OK, so I removed the reverse variants (nobody proposed any use for them so far!). What about the remaining two questions?

I'm more or less happy with this PR, so if nobody has any more request: merge?

SchrodingersGat:
I think we should still have a pin-1 marker. I'm happy with the naming, too. Anyone object to merging this? @evanshultz @poeschlr

poeschlr:
I am also in favor of keeping the pin 1 marker. It is ok to merge for me.

jkriege2:
So, should I add some kind of pin1-marker to the vertical footprints too (so far I only added triangle to horizontals)? ... This seems quite uncommon to me (never seen it anywhere else): The half-round-shape (of package and on silkscreen) prevents soldering it the wrong way + gives a hint on where the pin1 is after assembly (+ rectangular pad of pin1 ... )

What do you all think?

I'm happy to add e.g. triangles if we agree on that, could look like this:
2017-12-11 22_39_29-3d viewer
2017-12-11 22_38_46-3d viewer


JAN

jkriege2:
@evanshultz @poeschlr @SchrodingersGat any more input?

poeschlr:
I think we can merge as is. If we want to make any further modifications, we can do this in a separate PR.

Thanks for your work.

PR #136, opened 2017-11-29
Made naming of tantal caps easier to understand 

poeschlr:
Move letter code towards the back to avoid impression that these are manufacturer specific footprints

I noticed this shortcoming by this reported issue https://github.com/KiCad/kicad-packages3D/issues/176

If this is accepted we also need to update the KLC.

Edit: script PR https://github.com/pointhi/kicad-footprint-generator/pull/86

Edit2: If this is still not clear enough i could append the word Code somewhere.
Example: AVX-Code-J instead of AVX-J


SchrodingersGat:
Thanks :)

Issue #130, opened 2017-11-28
What about manufacturer specific pin headers and socket strips? (Samtec)


poeschlr:
Currently the samtec lib holds some pin header footprints. Should we transfer these specialized footprints over here? (They have slightly different drill sizes: 0.95mm compared to 1mm for the generic footprints)
Samtec creates quite a lot of slightly differing pin headers and socket strips. I am not sure it would be a good idea to include specialized footprints for all of them. (But i am also hesitant to simply "delete" contributions.)

evanshultz:
IMO, these are generic enough that we can use a "one size fits all" and eliminate vendor-specific footprints where possible. This should result in the least maintenance issues for librarians while still supporting all components a KiCad user would want to use. I share the guilt from removing someone's hard work from the library, but to me it seems the best way to proceed.

terjeio:
I have just remade the script for the pin sockets and when I did the research for them I found that dimensions can vary quite a bit between the vendors. My plan is to dig into this issue in more detail to see which variants is not covered by the existing ones and possibly add them. Bottom entry, guide pegs, different land patterns...

3D models is another issue, varying heigths, turned pins...

How much work should be put into this? I guess it depends on the level of accuracy and coverage the librarians are aiming for.
Also, if I am going to do this I will need some input on when a new footprint should be made. Eg. I have a SMD pin socket on my desk that requires a wider land pattern than the current KiCad version provides.

PR #137, opened 2017-11-29
Adding samtec edge connector footprints (MECF)

poeschlr:
The other footprints in the old Samtec repo are normal pin header footprints. For discussion if they should be included see: https://github.com/KiCad/kicad-footprints/issues/130

This PR will add both the MECF series edge connectors and the footprints for the compatible sockets.

# Edge connector

- should the connector pads include paste?
- should the pads be of type TYPE_CONNECT instead of TYPE_SMT?
- Currently we use a courtyard clearance of 0.25 is this ok? (Does a courtyard even make sense here?)


poeschlr:
# Compatible Sockets for Edge Connectors

Only change compared to the old connector is that now the NPTHs have the same pad size as drill size. (was 0 earlier which kills the 3d viewer of my version.)

For now i added the sockets to the same lib as the edge connectors. Should they be separated?

SchrodingersGat:
> 
>     should the connector pads include paste?
>     should the pads be of type TYPE_CONNECT instead of TYPE_SMT?
>     Currently we use a courtyard clearance of 0.25 is this ok? (Does a courtyard even make sense here?)

* I think these should **not** have paste - the socket grips exposed copper on the PCB, yes?
* What does TYPE_CONNECT actually mean? I have never understood this.

poeschlr:
I will remove paste.

I also do not really know why a special pad type for connector is necessary. But it seems to be there in kicad.

SchrodingersGat:
Check out this thread here - https://forum.kicad.info/t/usage-pad-type-connector/1087/3

So we should be using these pads for 

* Edge connectors
* Wire bonding pads

poeschlr:
Pads now have the type connect and don't include paste.

SchrodingersGat:
Nice work :)

Issue #134, opened 2017-11-29
Should we create a separate library that captures all FFC/FPC connectors?


poeschlr:
A lot of manufacturer seem to create such connectors. Currently we have footprints for connectors by at least Molex, TE and Hirose.

I would suggest a new lib called Connector_FFC-FPC and group them in there.
Naming of footprints would then be `[man]_[mpn]_{1xN or 2Rows_Npins}_[orientation]`

evanshultz:
Did we come up with a good naming convention if the pins are staggered and an uneven quantity?

poeschlr:
I don't know if it is good but currently we use 2Rows_NPins i think.

SchrodingersGat:
I think this is a good idea

PR #138, opened 2017-11-29
Adding TE Connectivity FFC/FPC connectors and moving other such connectors.

poeschlr:
Adding the script generated FFC/FPC TE connectivity connectors.

For now i created a separate lib for this type of connector as i suggested in issue https://github.com/KiCad/kicad-footprints/issues/134

If we decide to do it this way i also need to add the FFC connectors from Molex and Hirose.
If we decide against it, then i need to move these connectors into Connector_TE-Connectivity.

About the footprints: I did not change anything on purpose. (Maybe text positions and line widths changed.)

script PR: https://github.com/pointhi/kicad-footprint-generator/pull/88


poeschlr:
@SchrodingersGat we can archive the old TE-Connectivity repo now.

poeschlr:
This should now be ready to merge. 

I needed to regenerate the molex 502250 footprints because their pin layout was mirrored. (Because of that i converted the generator script to use kicadModTree)

screenshot from 2017-11-30 16-54-03

poeschlr:
Regarding part naming. Currently the cable pitch is used in the footprint name, not the pin pitch. Should i change this?

SchrodingersGat:
Hmm probably pin pitch is better

poeschlr:
I changed it to be the pin pitch

SchrodingersGat:
Thanks @poeschlr  :)

PR #139, opened 2017-11-29
Connector Wago footprints

poeschlr:
I scripted the Wago 734 connector footprints that did exist in the old repo. (In a very bad form)
Script PR: https://github.com/pointhi/kicad-footprint-generator/pull/89

datasheet: http://www.farnell.com/datasheets/2157639.pdf

I am not sure where pin 1 is. I trusted the old footprints with regard to that. (The datasheet does not specify a pin 1)
Pad to pad clearance needs to be 1.5mm to reach the voltage specifications.
screenshot from 2017-11-30 00-45-34
screenshot from 2017-11-30 00-46-24
screenshot from 2017-11-30 00-46-15


SchrodingersGat:
Thanks, these look much improved

PR #146, opened 2017-11-30
Terminal blocks

poeschlr:
Add unmodified footprints from TerminalBlocks.pretty (only fixed the 3d path)


SchrodingersGat:
Thanks @poeschlr this is an improvement on what's already there. I'm happy to merge this as is. Cheers

PR #148, opened 2017-12-01
Transfer footprint from Measurement_Scales

poeschlr:
Not sure what i should name the new lib. For now i used Calibration_Scales because i think the only use for these footprints is to calibrate PCB productions. (For home producers.)


poeschlr:
About travis report: I don't think these need a courtyard. (The footprints are not really intended to be used in a working design i think.

SchrodingersGat:
I can't immediately think of a better name. Thanks for bringing these across.

Issue #74, opened 2017-11-21
Some of the second references (on fab layer) for resistor THT footprints interfere with other graphical elements on fab


poeschlr:
Some oft he references need to either be scaled down or moved a bit.
Preferably done with the original generator scripts.
But can also be done by hand. (Only a few components are affected.)

Script can be found here: https://github.com/pointhi/kicad-footprint-generator

grob6000:
Looking at this now. Unfortunately http://kicad-pcb.org/libraries/klc/F5.2/ hasn't really considered the axial vertical THT resistor... Plan will be to change the line representing the bent pin to only reach the circumference of the body, and move the R% into the centre of the body. See what it looks like.

grob6000:
Having a bit of trouble - keeping this on pause for the moment.
Does anyone (maybe @pointhi, @poeschlr ?) know what the best practice is for the "Fields" in the footprint properties; i.e. "Doc" and "Keywords"? Seeing some differences between what's currently in the footprints and the output of the script and looks like I'm mucking it all up. :)

PR #151, opened 2017-12-04
Fix Fab layer errors in Resistor_THT

### LINKED ISSUE #76: Inductor THT: Missing 3d model references, second reference (fab) interferes with fab and some silk lines on top of pads.
This is ideally fixed within the generator scripts directly. (Fixing the inductor references would also fix some of the resistor references because they share the same generator scripts.)
So https://github.com/KiCad/kicad-footprints/issues/74 is related in part to this issue.

Script location: https://github.com/pointhi/kicad-footprint-generator

grob6000:
Script is updated in https://github.com/pointhi/kicad-footprint-generator/pull/93
Fixes the errors we saw with the Fab text, and model associations.
Doesn't perfectly comply with http://kicad-pcb.org/libraries/klc/F5.2/ for super-small components, where putting the secondary fab ref inside the body would result in unreasonably small text size (although down to 0.1mm is allowed, this seems non-sensical to me for THT parts with dimensions c. 2mm: how would somebody use this??). Elected to place the ref where the silkscreen reference goes.
Should address #74.
Separate PR for the inductor models (and some checking of course) will be required to fix #76, but should be possible using same updated script above.
The script has made some kind of subtle change to the silkscreen offset relative to the previous footprints. It doesn't seem to be a problem to my eye, and variation is very minor (0.06mm offset, larger shape). Let me know if this should be corrected / does not comply.
Image of example fixed footprints:
Image of fixed example footprints 4x


grob6000:
Oops included stuff that should not be.

grob6000:
Hooray! That should do it. 💯

Shackmeister:
looks good :D found the same error Travis did, but fixed now :) thank you so much for your work!!

PR #153, opened 2017-12-04
Connector usb1

krtkr:
Hi all!

I add a few USB packages, used at my project:
1. USB_B_TE_5787834_Vertical
1. USB_Micro-B_Tensility_54-00023_Vertical
1. USB_Micro-B_Tensility_54-00023_Vertical_CircularHoles


SchrodingersGat:
@krtkr nice these are great :) One minor complaint (very minor!) The F.Fab lines are a little wonky

selection_023

pointhi:
I would include the manufactur number in the first footprint as well.

poeschlr:
It seems to be TE-Connectivity 5787834
So the name should be `USB_B_TE_5787834_Vertical`

krtkr:
@poeschlr I have a little confusion regard naming here.

At new KLC for connectors you specify MPN. It seems that it consist of Manufacturer and Part number. What should be separator between Manufacturer and Part number? '_' or '-'?

For Connector_USB.pretty library both used. At my patch I use '-': Tensility-54-00023, you suggest '_': TE_5787834.

Maybe we should come up with something common for all footprints here?

poeschlr:
The separator between MAN and MPN is defined as "_"
See section `Connectors for Specific Functions` of http://kicad-pcb.org/libraries/klc/F3.6/

krtkr:
@poeschlr I guess _USB_Micro-B_Wuerth-614105150721_Vertical_CircularHoles_ should be named _USB_Micro-B_Wuerth_614105150721_Vertical_CircularHoles_ in this case?

poeschlr:
Yup exactly.

krtkr:
@poeschlr in this case someone should fix the KLC, because first example at F3.6 'Connectors for Specific Functions' is _USB_Micro-B_Wuerth-614105150721_Vertical_CircularHoles_.

poeschlr:
See PR for the webside: https://github.com/KiCad/kicad-website/pull/247

SchrodingersGat:
Thanks for the contributions :)

PR #164, opened 2017-12-07
Add footprint for 4P4C connector Molex 95501-6449

lorem-ipsum:
This adds a footprint for the Molex 95501-6449 (SMD variant). The library check is complaining about the center location. Currently the center is between the two mounting holes for the fixation clips. Regarding the center I am not sure whether to use the center of the component (including pads) or just the housing (without pads).

[Datasheet](http://www.molex.com/pdm_docs/sd/955016449_sd.pdf)
molex-95501-6449


lorem-ipsum:
I've just changed the name to 95501-6449 as Molex uses this name

lorem-ipsum:
I have started scripting the footprints variants. So far I have finished the THT variants, I will do the SMD variants next. I chaged the namings to nPmC_Molex_95501_{THT,SMD}

lorem-ipsum:
I've finished both versions. I don't really get where the center for the SMD components should be.
Datasheets: [SMD](http://www.molex.com/pdm_docs/sd/955016449_sd.pdf), [THT](http://www.molex.com/pdm_docs/sd/955012401_sd.pdf)

lorem-ipsum:
I now have a first version of a script reading a fairly generic definition for modular connectors from YAML and generating footprints (found [here](https://github.com/lorem-ipsum/kicad-footprint-generator/tree/connector-molex/scripts/Connector/Connector_RJ), currently only one series is defined in yaml). While making it I came around some general questions I couldn't find any answers to.
- Is there a standardised way to specify the edge of the pcb?
- Is there a way to mark the front of the connector? I've seen small arrows on some footprints in the USB library.
- Molex does only specify a drill size for THT connections. Is there a recommendation on the annular ring width? I just found the minimum of 0.15mm which appears very small.
- How should I handle duplicate footprints. Currently I am working on generating footprints for each part number. But some series have identical footprints. So far I coudn't figure out a pattern behind these.
- And most importantly: Is it of interest that I work my way through the modular connectors and add them to the library?

Schlumpf:
Hi @lorem-ipsum, maybe you find this article interesting for the annular ring question: http://www.pcb-3d.com/knowledge-base/pth-dimensions/

lorem-ipsum:
@Schlumpf thank you, this looks perfect.

poeschlr:
Sorry for the late answer.

> Is there a standardised way to specify the edge of the pcb?

If the connector relies on the edge being on a specific place, then draw on the edge cuts layer. If it is only a suggestion then you can draw it on the dwgs layer.
Maybe also place a clarification text on the comments layer.

> Is there a way to mark the front of the connector? I've seen small arrows on some footprints in the USB library.

There is no standard. A small arrow would be ok for me.

> How should I handle duplicate footprints. Currently I am working on generating footprints for each part number. But some series have identical footprints. So far I coudn't figure out a pattern behind these.

If the 3d model is also identical you can use wildcards in the mpn. But if the name would get too complicated it might be better to have duplicated footprints.

> And most importantly: Is it of interest that I work my way through the modular connectors and add them to the library?

We librarians don't know what parts the users will need. If you are interested in creating these they will be welcomed.
Maybe scripting might help here. Have a look at our python footprint generator framework: https://github.com/pointhi/kicad-footprint-generator

---

Notes from my side:
Move the silk layer outside of the fab line. (If the fab layer represents the body outline otherwise move the fab layer to the exact body outline.)

From the screenshot it seems your footprints are missing pin 1 markers. 
Here my suggestion:
33771614-ee333f3e-dc31-11e7-9103-5b4e4f7bdd61

lorem-ipsum:
@poeschlr Thank you for the detailed explanations. I already started a script for the kicad-footprint-generator which currently can be found [here](https://github.com/lorem-ipsum/kicad-footprint-generator/tree/connector-molex/scripts/Connector/Connector_RJ). It still has some issues which I need to fix and your suggestions will become part of that list.

Another minor question regarding that script. I added the Molex 95501 series which come with and without a small plastic extension to the side. Its only purpose is to align the connector with the housing. Should that be added to the Fab layer? Currently it looks like this:
screenshot_20171220_113512

poeschlr:
About courtyard: If you have pads to the side they of course need to be inside the courtyard area.

It is allowed to either make an easy rectangle or a polygon that hugs the outline more closely. (The later is of course a lot more work for the contributor. But it does create better results.)

poeschlr:
@lorem-ipsum is there some information required from the maintenance team or do you simply have too little time for finishing this right now?

lorem-ipsum:
I think I have all information, I just need to find some time to finish it.
I will try to get the existing footprints first into a state to get them commited and postpone any further features.

Misca1234:
Should you not place the script in here instead ?
https://github.com/lorem-ipsum/kicad-footprint-generator/tree/connector-molex/scripts/Connector/Connector_Molex

lorem-ipsum:
Finally came around to work again on the script. I think now everything should be fixed.
- Fixed courtyard to include all outside components
- Moved silk screen a bit to the outside
- Added pin one markers to silk screen and fab layer

The check script is complaining for some of them that the anchor is not centred. This is on purpose as it is placed on the pick & place position specified in the data sheet.
If everything is fine I will also create a PR for the script to the kicad-footprint-generator. It has become quite a monstrosity but it works. Regarding miscas comment: I have renamed the script into conn_modular as it is less about molex connectors, but about this general type of connectors.

Shackmeister:
Hi @lorem-ipsum 
 Sorry for the very long delay!
I would love to see this contribution make its way to the library, so I hope you are still up for finishing the last things :)

- it seems like Molex have stopped using the hyphens in the MPN

- Some of your silkscreen lines are in the solder mask clearance of the pads (0.2mm). Some are even crossing the pads.

- if you would like, it's now allowed to make closer courtyards.

- Perhaps "_Horizontal" should be added to the end of the name

SMD footprints:

- Mounting pads should now be called "MP"

THT footprints:

- A straight line would be fine for silk pin 1.

- A triangle would probably be better for pin 1 fab indicator.

- we now allow roundrects for pin 1 THT pads

antoniovazquezblanco:
Ping @lorem-ipsum

PR #166, opened 2017-12-07
added footprint for ESP-WROOM-02

sandhan26:
[ESP-WROOM-02 datasheet](http://espressif.com/sites/default/files/documentation/0c-esp-wroom-02_datasheet_en.pdf)

[3D Model here](https://github.com/KiCad/kicad-packages3D/pull/195)

ESP-WROOM-02 footprint


SchrodingersGat:
Please move this to the existing `RF_Module` library :)

sandhan26:
screenshot from 2017-12-08 19-34-05


How to enable Dwgs.User and Cmts.User layer in Kicad to add keep out areas for antenna, like below ?

screenshot from 2017-12-08 19-36-06

poeschlr:
Sadly one can not draw directly on the dwgs layer. But you can for example draw on B.Fab and then move the drawing to the dwgs layer within the line options dialog or using a text editor.

If you use a layer where nothing else is placed you can use a text editor to move everything on that layer to the dwgs layer using the replace functionality of that editor.

If you don't want to use a text editor. (or you don't trust yourself with it) there is another option:
You can draw one line, move it to the correct layer (line properties dialog) and then copy that line around using ctrl+d to reduce your workload. (changing the layer one line at a time would be more effort)

sandhan26:
screenshot from 2017-12-08 21-11-50

poeschlr:
@jhol wrote:
> the diagonal lines are not perfectly parallel. Is there any easy way?

You could place the grid origin at the top left corner and use a grid spacing that divides both the length and width of the rectangle. (User grid.)
In your case place origin at -9,-13.1 and set grid to 0.9

Or use a script.
I implemented the creation of a keepout marking in our python footprint script environment. You can simply make a footprint only existing of a keepout area of your required size and use a text editor to copy it into your footprint file.
https://github.com/poeschlr/kicad-footprint-generator/blob/master/scripts/tools/footprint_keepout_area.py

jhol:
:smile:

sandhan26:
@poeschlr Thats a very clever trick. Here is the result. Its awesome.

screenshot from 2017-12-09 07-44-18

sandhan26:
I have two more doubts.

1. Given below is the screenshot of that part of the datasheet, which has the dimensions.
    As you can see the antenna area height is 6.9 mm extending over to the first row of pads.
    In my case i used 6.3 height to avoid going over the pads. Is this ok, or it is recommended to
    follow the datasheet no matter what?

   screenshot from 2017-12-09 07-45-56

2. Secondly, what is the Anchor for? And why it isn't visible? Its very difficult placing it at exact 
    center of the footprint. Any comments, describing the proper use of Anchor will be very helpful.

poeschlr:
Regarding keepout area have a look at this document:
https://www.espressif.com/sites/default/files/documentation/esp-wroom-02_pcb_design_and_module_placement_guide_0.pdf

sandhan26:
@poeschlr You mean i should expand the outlines in the Courtyard layer

poeschlr:
The document i linked does show how the keepout area should look like. (The stuff on the dwgs.user layer) This was to answer your question 1.
But now i wonder if in such cases the crtyd layer should also include keepout area or if the keepout area can be larger than the courtyard. @SchrodingersGat, @jkriege2, @evanshultz what do you guys think about that.

---

For pad sizes you can look at IPC documents. 
Example: http://pcbget.ru/Files/Standarts/IPC_7351.pdf Page 13 table 3-9

Shackmeister:
hi @sandhan26 are you still planning on having a look at the last couple of issues so we can get this merged? :)

Ratfink:
It looks like the pad for pin 19 is on the wrong side.  The diagram you posted an image of [above](https://github.com/KiCad/kicad-footprints/pull/166#issuecomment-350416846) shows a back view, not a through-view (there's a photo of the back of the module in the PDF as well if you're uncertain), so it should be on the right side of the footprint, not the left as it is now.

sandhan26:
@Shackmeister and @Ratfink I am on it, will push the changes by today. Thanks :+1:

sandhan26:
screenshot from 2018-03-01 11-22-26

sandhan26:
@poeschlr @Shackmeister @Ratfink kindly check the footprint.
Also, i am getting unable to parse error, why?

poeschlr:
The 3d model offset is non zero. This should be fixed.
(It is also the reason why the testscript is unable to parse the file as newer versions of kicad replace 'at' with 'offset'. The later being in mm instead of inch.)

poeschlr:
Looks good. Thanks.

As you moved the part around and removed the 3d model offset it might be necessary to re-align the 3d model.

PR #167, opened 2017-12-07
Transfer HE 10 from Connectors into Connector_IDC

poeschlr:
Changes compared to the old version:
- rotated to have pin 1 at the top left corner
- added pin 1 markers on fab and silk
- added second reference
- added 3d model settings


poeschlr:
There are also footprints for GoldTek connectors in the old repo. These are also Locked IDC headers. Should we add the manufacturer information to the footprint names?

poeschlr:
@SchrodingersGat, @jkriege2 has this been forgotten or are there open questions?

SchrodingersGat:
I think that generic is fine for these, at least for now. Thanks for the contribution, sorry it was left unattended.

PR #168, opened 2017-12-07
More Footprints from the old Connectors lib

poeschlr:
Contains:
3M Card holder
FMC (Samtec)
Grove
JWT
And NMJ6 (Audio jack)


SchrodingersGat:
There is a 3D model error that should be fixed :)

poeschlr:
Now the only Travis error is that the samtec connector is not centered.

PR #172, opened 2017-12-08
FIX: UQFN-10_1.4x1.8mm_P0.4mm has wrong pin numbering.

adite:
Pin-Numbering of UQFN-10_1.4x1.8mm_Pitch0.4mm was wrong:
uqfn-10-wrong

Changed to anti-clockwise. According to document https://www.onsemi.com/pub/Collateral/488AT.PDF.
uqfn-10-good


poeschlr:
Thanks for this fix.

PR #173, opened 2017-12-08
Shift picoblade centroids to comply with KLC F6.2

robertlong13:
Shift picoblade centers to match vacuum pencil location in datasheet. This brings the footprints into compliance with KLC F6.2

Datasheet (53261): http://www.molex.com/pdm_docs/sd/532610571_sd.pdf
Screenshot (53261):
image

Datasheet (53398): http://www.molex.com/pdm_docs/sd/533980571_sd.pdf
Screenshot (53398):
image

Source PR: https://github.com/pointhi/kicad-footprint-generator/pull/94
3D model PR: https://github.com/KiCad/kicad-packages3D/pull/197


robertlong13:
The build check is complaining that the footprint anchor needs to be the center of the footprint, but these parts are an exception, since there is an anchor point specified in the datasheet.

robertlong13:
Crap, I got the location slightly wrong on the 53261. Don't merge yet.

robertlong13:
These are correct now. Here is a screenshot of the origin and the 3D model together for verification. Note that it looks a little low here, but that's because the cutout on the model below the centroid is slightly larger than it should be. This is the correct location.

image

SchrodingersGat:
Thanks for the contribution :)

Shackmeister:
Are these corrected through the scripts?

PR #177, opened 2017-12-11
add dual coil version of Schrack RT2 FormC

lorem-ipsum:
This adds a dual coil version of the existing Schrack RT2 FormC relays (only change is the added A3 pad). [Symbol-PR](https://github.com/KiCad/kicad-symbols/pull/33)
The check complains that A1 is not rectangular. Should I change that for all RT2-Relays?
rt2-dual-coil


poeschlr:
I fear you might need a bit more work on this.
screenshot from 2017-12-12 11-16-58

The pads for the main contacts need at least 3mm clearance (To reach the required isolation for 230VAC). For pin pitch of 5.04 i would suggest an oval pad with size=2x3mm.
screenshot from 2017-12-12 11-24-43

The coil pads can be drastically reduced in size. (Maybe also make them oval for easier soldering.)

lorem-ipsum:
I have reworked both Form C variants and the Form A version to use the exact dimensions specified in the datasheet and to use oval pads. I made the dual coil version and created the others by deleting the missing pads.
I did not rework `Relay_DPDT_Schrack_RT2_RM5mm`. It appears to be identically to the `_FormC` variant. If you don't object I would delete it to remove the duplication.
formc-dual

poeschlr:
If you delete it, make sure no symbol in the official lib does have that footprint in their footprint field.

poeschlr:
One small thing is still a bit wrong: (I assume the relay is symmetrical in y direction -> the distance between fab outline and pad in y direction should be equal at the top and bottom. More details see drawing)
screenshot from 2017-12-15 01-19-43

It might also be a good idea to change the pad shape for pad A1 to be a rectangle instead of oval.
Maybe also add some sort of "pin 1" marker.
The silk outline should be moved towards the outside such that it is fully outside the fab outline.
screenshot from 2017-12-15 01-27-36

lorem-ipsum:
I've updated the footprints with the rectangular pad 1 and the corner on the fabrication layer. I have separated silk and fab markings by 0.2mm.
screenshot

poeschlr:
Thanks for your contribution.

PR #179, opened 2017-12-14
Add Molex Micro-Fit 3.0 THT Vertical Headers

twam:
Datasheets:
http://www.molex.com/pdm_docs/sd/430450212_sd.pdf
http://www.molex.com/pdm_docs/sd/436500215_sd.pdf

Screenshots:
screen shot 2017-12-14 at 20 18 19
screen shot 2017-12-14 at 20 18 27

------------


poeschlr:
The single pin, single row version you add is not produced by Molex.
And you seem to multiply the number of pins in the MPN with 2 for the single row variant.

There are also some minor problems with the footprints:
screenshot from 2017-12-15 00-42-04

You could also elongate the pads for the single row connectors. (might make it easier to solder these.)

twam:
Thanks for the thorough review!

How did you do the nice drawing? Is there an (semi-)automatic way?

Update footprint screenshots:
screen shot 2017-12-16 at 16 06 31
screen shot 2017-12-16 at 16 06 20

poeschlr:
The drawings are made using kicad stepup to get the footprint into freecad and then i use the drawing dimensions workbench to measure stuff. (There is also the technical drawing workbench you could try.)

poeschlr:
Looks good thanks

PR #181, opened 2017-12-17
add footprint Mini-Ciruits CDxxx for RF transformers

lorem-ipsum:
This PR adds the footprints matching the symbol [PR](https://github.com/KiCad/kicad-symbols/pull/60) for RF transformers from Mini-Circuits
[Datasheet](https://ww2.minicircuits.com/case_style/CD541.pdf)
screenshot


poeschlr:
For reference, symbol PR: https://github.com/KiCad/kicad-symbols/pull/60#issuecomment-352940156

poeschlr:
I think the current naming convention would require the footprint name to include pin count, body size (length and width) and pitch. 
@SchrodingersGat, @jkriege2, @evanshultz what do you guys think. Do we want to include all of this?

It might be a good idea to include all 3 case variants. They have a different body height and need separated 3d models. If you do this also include the body height in the dimensions mentioned in the footprint name.

Example of a footprint name with all "fields" when all 3 variants are added:
`Transformer_Mini-Circuits_CD541-6_5.58x7.87x2.08mm_P2.54mm`

SchrodingersGat:
I'd be happy with either of the suggestions that Rene outlines above. The shorter one may be preferable as the Mini-Circuits P/N is enough to fully define the pad pitch, etc.

jkriege2:
I would also vote for the second variant

lorem-ipsum:
I have split the footprint into four, based on the second, shorter naming scheme.

poeschlr:
Looks good to me. Thanks for your contribution.

PR #183, opened 2017-12-18
Remove ThermalPad suffix and add -1EP descriptor to htssop footprints

poeschlr:
3 HTSSOP footprints did not follow the current naming convention.


jkriege2:
Hi!

a question on ThermalPads ... do we distinguish between simple bottom pads for connection reasons and thermal or cooling pads (often with vias) ... if these footprints have thermal vias, should we maybe add `_ThermalVias` (as there might be a variant without these?)

poeschlr:
You are right these should have the suffix _ThermalVias. Some of them should also specify that the mask area is reduced.

poeschlr:
@jkriege2 i added the ThermalVias suffix where required and added a version of these 3 footprints without thermal vias as well. 
In addition i split up the past layer for the exposed pad and ensured that the paste layer is designed in such a way that vias are always between paste areas.

PR #185, opened 2017-12-19
add Mini-Ciruits DL805

lorem-ipsum:
Add a footprint for the Mini-Circuits DL805 housing ([datasheet](https://ww2.minicircuits.com/case_style/DL805.pdf)), used by [this](https://ww2.minicircuits.com/pdfs/M3SWA-2-50DRA+.pdf) RF switch (symbol will follow).
screenshot_20171219_215727


poeschlr:
I am not really sure the button_switch lib is the correct lib for such a footprint. This lib holds mechanical switches.

This footprint seems to be similar to either LGA or DFN footprint so it might fit better there.
@SchrodingersGat, @jkriege2 or @evanshultz what is your opinion on that matter?

SchrodingersGat:
It does look like it's (pretty much) a DFN part.

Also, based on the image above, the pads are incorrectly ordered. It appears that you have used the "from below" image on the datasheet to draw the "from above" footprint.

jkriege2:
Yes, please correct the pin-ordering and move this to the DFN-lib.

lorem-ipsum:
I have moved the footprint and corrected the pin numbering. I also rotated the footprint to move pin number 1 to the top-left corner.
screenshot

poeschlr:
The pin 1 markers seem to be at the wrong place now. (At least in the screenshot.)

lorem-ipsum:
... Yeah, I should take more care. I moved the pin marker to actually be next to pin number 1.

poeschlr:
Sadly there is a small problem with the footprint. The outher dimension of the pads in y direction is too large:
screenshot from 2018-01-15 13-19-14

lorem-ipsum:
Oh. I will look into it and fix the issue.
May I ask which program you are using to check and generate these markings? I couldn't find anything like this inside of kicad, but it appears to be very useful.

poeschlr:
Everything is described in this FAQ post over at the forum https://forum.kicad.info/t/how-to-check-footprint-correctness/9279

lorem-ipsum:
Thanks for the link. I've updated the footprint (moved the pads to the inside).

poeschlr:
Looks good now. Thanks for this contribution.

For the future it might be better if you could add a separate commit instead of amending to your old one in the future. Then we librarians can check the changes made by your correction commits which makes it easier to check your fixes to the pull request. (If we don't want the history bloated we can still use squash merges.)

lorem-ipsum:
Thanks for your reviews and help.
I won't squash the commits in the future. I am kind of trained on squashing them to keep the history clean, but I see that it makes tracking changes easier.

PR #186, opened 2017-12-21
add Traco Power TCK-141 inductor footprint

matthijskooijman:
- Datasheet is here: https://www.tracopower.com/products/tck141.pdf
 - There is no matching symbol, the generic inductor symbol should suffice.
 - There is no  matching 3D model that I know of, I won't be providing one.

selection_012
selection_014


poeschlr:
Looks good. Thanks

PR #190, opened 2017-12-22
Add potentiometer footprints

evanshultz:
See https://github.com/evanshultz/kicad-footprints/pull/3 for screenshots, details on changes/fixes, and an outline of the script changes.

------------


evanshultz:
I think this is ready for merge. There are some slight off-center errors and warnings about non-polarized pins, but those seem to be unimportant.

Note that Bourns PRS11S has shifted to now be roughly around 0,0. This may cause the 3D model to be offset. Also, all 3D model params are at the normal values (for v5 library) which was not the same before.

jkriege2:
Hi!

looks good ... could you take a look at the remaining KLC-issues, I think 
- `Tags contain illegal character: (',')` should be easy to fix
... and is there a reason why the centering fails?

Best,
JAN

evanshultz:
@jkriege2 
1. The comma is because there is a comma in the PN. See the examples given.
1. I do not know why the centering is so wrong in some connectors, but I decided I didn't want to dig that deep into the script so it should be no worse than the existing footprints.

jkriege2:
Hi!

I think the reason is quite simple ... because they are off-center (from this PR):
2017-12-22 21_01_45-footprint editor no active library

I think this is caused by setting the origin to the center of mass of all the pins ... so the two pins on the right increase the weight on the right and move the pin there ... one should really calculate the surrounding rectangle of all pins (or something like that) and use its center as origin. the scripts do that by using a bounding rect (see line 670 in https://github.com/KiCad/kicad-library-utils/blob/master/pcb/kicad_mod.py)

jkriege2:
I think you can use the --fix option of the scripts to fix that error ...

evanshultz:
Yes, they are actually off center. At least, so says the check scripts.

The footprint generator script is causing this offset because it doesn't know what the check script is looking for. So... my question is if this is a problem? As this is the last footprint lib to be merged, and there are so many improvements already, is it OK to merge as is?

Without the generator script being changed, re-generating the footprint will move them off center again, according to the check script, so that doesn't seem like a true fix to use the --fix option of the check scripts.

jkriege2:
Currently I think the center does not make sense too much ... could you post the changed scripts, then I could have a look at them? Are they based on my old stuff or completely rewritten?

evanshultz:
[footprint_scripts_potentiometers.zip](https://github.com/KiCad/kicad-footprints/files/1583429/footprint_scripts_potentiometers.zip)

Based off of your script, hence why I didn't start over and fix anything fundamental if I didn't think I had to. Here they are zipped together, for brevity. The argparse interface needs a bit of work so I was planning to finish that once the footprints were merged.

jkriege2:
Hi!

OK, I found the bug in my scripts ... I think it was buggy from the start ;-)

In 

change the line 813 of `footprint_scripts_potentiometers.py`:
```
    if SMD_pads:
       # new version calculates bounding box and the center of that!
        offset = [-(max(padpos[0][1],padpos[1][1],padpos[2][1])+min(padpos[0][1],padpos[1][1],padpos[2][1]))/2, -(max(padpos[0][2],padpos[1][2],padpos[2][2])+min(padpos[0][2],padpos[1][2],padpos[2][2]))/2]
        # old version calculated CoM of pads:
        #offset = [-(pinxoffset+pinxoffset+rmx2+pinxoffset+rmx3)/3.0, -(pinyoffset+pinyoffset+rmy2+pinyoffset+rmy3)/3.0]
```

jkriege2:
Could fix that and I will merge!
Thanks,
JAN

evanshultz:
Done. It didn't center one of the Vishay pots, but otherwise worked to center them according to the check script.

poeschlr:
Might it be that the old poti footprints where centered around the axis of the control knob? (Would make sense to me.)

Don't just blindly use the fix options without checking if there was a reason for footprints diverting from the KLC.

jkriege2:
No they were not. the script simply calculated the cneter of mass between the pads, which pulls the center to one side when one side has more pads than the other. That was a bug in my scripts, so this fix is fine. Whether we want them centered around the screw is another question, but I think when I first made these it was decided to center them as anything else (e.g. THT is also centered at pin1!)

PR #192, opened 2017-12-22
Added Samtec FTSH range of connectors

MatthewStephenson:
This is the Samtec FTSH DV range of pin header connectors, as used for ARM Cortex SWD debug ports.

Datasheet: [http://suddendocs.samtec.com/catalog_english/ftsh_mt.pdf](http://suddendocs.samtec.com/catalog_english/ftsh_mt.pdf)

image

image

image

As always, comments and suggestions welcome.


poeschlr:
The version without housing (not shrouded) is simply a "standard" smd pin header.
Could you check if the generic smd pin head footprints fit these connectors?

We had a discussion about including these already. https://github.com/KiCad/kicad-footprints/issues/130
The argument against is that there are way too many components that do not differ in their required footprints.
The argument for it would be that we could then have specialized 3d models. (And the footprints could include the ordering information, which is not included in generic connector symbols nor generic pin header footprints.)

MatthewStephenson:
Well, I wish I'd known about that discussion before I started making a script to do all of these!

I put them side-by-side, and the pads are just a little shorter on the generic part, but I think they'll fit.

image

I'll wait for advice on whether I should close this PR or not.

MatthewStephenson:
Closing as it looks like the generic footprints are the way to go.

PR #195, opened 2017-12-27
Coilcraft xfl4020

johnfwhitmore:
Added the footprint for the Coilcraft XFL4020 Series of Inductors

Datasheet : https://www.coilcraft.com/pdfs/xfl4020.pdf

screenshot from 2017-12-27 15-01-29

Going to check Travis output here


johnfwhitmore:
I've never created a 3D file as yet. I'll look into it...

poeschlr:
It is not required to supply the 3d model it self. But the footprint needs to be setup to be able to add one without the need to touch it again.
This means: Fill out the model path `${KISYS3DMOD}/[lib name].3dshapes/[footprint name].wrl`, set the scale to 1, offset and rotation to 0

johnfwhitmore:
Added a reference to the 3D Model for the component.

Thanks for your help with that, know for the next time.

poeschlr:
Thanks for your contribution

PR #196, opened 2017-12-27
Added Bourns SRR1005 Series Inductor

johnfwhitmore:
Datasheet : https://www.bourns.com/docs/Product-Datasheets/SRR1005.pdf

screenshot from 2017-12-27 16-10-31


poeschlr:
Here the same applies as with your other pull request.
Setup the footprint for the 3d model.

johnfwhitmore:
Added a reference to the 3D Model for the component.

poeschlr:
The landing pattern is ok (the fab outline is also ok.)

Please ensure that the courtyard offset is correct. In y direction it should be relative to the diodes body, in x direction relative to the outer edges of the pads. The offset should be 0.25mm
(Rectangle size should be: 13.8x10.5mm)

Is there a reason why you have so little on silk?

johnfwhitmore:
>  Is there a reason why you have so little on silk?

Just getting the hang of this. I'll correct those problems.

poeschlr:
Thanks

PR #201, opened 2017-12-30
Fix swapped pad numbers 3 and 4

kacernator:
Diode bridge round have swapp pad numbers 3 and 4


poeschlr:
I can't find any datasheet that lists pin assignments. Is it possible that they are not standardized for this part? (There is a note in the datasheet stating that the terminal configuration is marked on the package.)

jkriege2:
How do they map to the existing diode bridge symbols? I created these footprints as generic, as it seems they are available from many manufacturers ... I hope I did them so they match up with one of the generic symbols ...

kacernator:
I can find any round diode bridge which matches current footprint. The AC pads should be across same as plus and minus should be across. Which I think is standard.
Here is datasheet: 
https://www.tme.eu/cz/Document/d97a8947eb7461daccec0981d4c63639/2w005g.pdf

jkriege2:
The symbol `D_Bridge_+A-A` matches your bridge with this footprint! Note that the datsheet shows a BOTTOM VIEW of the named pins on page 3 at the bottom (cf. https://cdn.shopify.com/s/files/1/1490/5112/products/02760268_00_1024x1024.jpg?v=1478044527)

jkriege2:
So can we close this?

kacernator:
I see where I made mistake. I choose the wrong symbol. So no changes to footprint is needed. We can close this.
Thank you very much

PR #203, opened 2017-12-31
Busses

DavidGriffith:
Adding S100 and PC 8-bit card edge footprints and clarify the AT and PCI busses.

------------


DavidGriffith:
S100 specs: http://www.s100computers.com/General%20Images/IEEE_696_1983.pdf
8-bit ISA specs: https://archive.org/details/bitsavers_intelbusSpep89_3342148/

DavidGriffith:
BUS_PC_M
bus_pc_m

jkriege2:
thanks for your contribution.

Can you please give an example for a specification against which we can compare yout footprint?

Also please add a courtyrad for `BUS_S100_M`

DavidGriffith:
What sort of specifications do you require?  The PDFs I referenced contains dimensions for 8-bit ISA and S100 card edges.

DavidGriffith:
BUS_S100_M
bus_s100_m

jkriege2:
uuups, sorry overread the spec PDFs ... will have a look at them ;-)

jkriege2:
Hi!

generally this looks good (as fas as I saw in the specs), but I would suggest some alterations (although the other edge connectors do not follow these yet):
- You put silkscreen, where actually the PCB edge should be cut out. Please add the PCB cuts on the layer  `Dwgs.User` and add a commenting text on `Cmts.User` saying e.g. `PCB edge cut` next to the cut-marking. Ideally these should be on Edge.Cuts, but that is currently not possible in the editor, so we will ahve to go with Dwgs.User. Then Silkscreen should only be where there will be PCB left (most probably above the pads). Maybe you can even hint with tiny horizontal lines of 1-2mm length, where the PCB continues
- Since this connector is not populated by any parts, you don't need the %R onf the F.Fab-layer

Best,
JAN

poeschlr:
I think the edge stuff could be directly put onto the edge cuts layer.
At least this is how it is done in for example `Samtec_MECF-50-0_-L-DV_2x50_P1.27mm_Polarized_Edge`.
Sadly this requires editing the footprint with a text editor.
If you do it this way make sure the end of the lines is on a quite coarse grid such that users can easily connect the rest of their edge cuts polygon to it.

If this is not how we want it to be done, then the script that generates the samtec connectors needs to be corrected and these footprints need to be re exported.

jkriege2:
Yes, I also see the `Dwgs.User` as a workaround for the missing capabilities of the editor ... So if you edit the text files in a text editor that would be the ideal way, but if you work with the footprint editor, please use the `Dwgs.User`-layer.

Best,
JAN

poeschlr:
@DavidGriffith Could you look into our suggestions?

DavidGriffith:
Yes.  I've been a bit busy lately with other stuff.

poeschlr:
Your changes look good right now.
Could you add pin 1 markers on silk at least? (Will help users trouble shoot their board)

DavidGriffith:
I just realized that I have not yet checked to make sure the solder mask layers make sense for card edges.

Shackmeister:
Hi @DavidGriffith 
Soldermask settings looks pretty fine, but I have a few requests.

- Courtyard clearance should be 0.5mm for connectors

- shouldn't the connector's be centered around the middle instead of pin 1?

- add a pin 1 marker on Fab layer.

- add the same fab drawing on B.Fab

- change the silkscreen pin 1 indicator to a small triangle or similar.

- on the connectors with one rounded end, it would be beneficial to use custom pad shapes. (or at least make the circles same pad number)

poeschlr:
The 3d model is not necessary. Footprint features are always rendered even without a model assigned.

---

Regarding reference i already stated above that i would like a refdes on the fab layer.

evanshultz:
@poeschlr 
Yep, but a commit above removed it so I wanted to surface your comment and prevent it from getting lost. :)

Shackmeister:
Hi @DavidGriffith 
Would it be possible for you to find some time to finish this PR? :)

DavidGriffith:
@Shackmeister, I think I applied all the requests you made except for pin-1 indicators on fab layers (how should they look?) and connectors with one rounded edge (I'm not clear on what you're asking for).  I'll need to go over it all again to verify the courtyard clearances.

DavidGriffith:
Okay, that's done.  How can I apply a square soldermask gap over the entirety of the gold fingers?

evanshultz:
@DavidGriffith 
Simply make a polygon only on the layer you want:
image

That requires two polygons, one for each layer. You could also make a single pad, only on the mask layers, and without a pin number:
image

Note that I haven't done this on these exact footprints so you should check Gerbers and make sure it's what you want, but one or both of these should work.

DavidGriffith:
Would you please explain how to get the line/polygon tool to draw a polygon?  I see lots of tutorials and such discussing polygons, but nothing actually shows HOW to get that tool to draw a polygon.

poeschlr:
You need kicad 5 to manipulate polygons. If you have kicad 5 then there is a separate tool for graphical polygons. (You then have the line tool and the graphical polygon tool)

I am just not really sure what you try to achieve to be honest. A single rectangle mask only pad should already do the trick. (or two if you want the "missing pin area" covered by soldermask) If you want mask between the pins then you do not need to do anything as the mask over the pins themselfes is already removed by activating the mask layer of the pads.

DavidGriffith:
I'm trying to leave the area where the edge connector is left clear of soldermask.  I don't understand what you mean by "no need to do anything" because the soldermask layer just masks off the pads (fingers) and the rest of the connector has mask applied.

poeschlr:
Read my comment again. It consists of more than one option. The no need to do anything part already mentioned that there will be soldermask left between the pads ;)

DavidGriffith:
For myself and any others interested, I found a better reference for S100 card edge specs than the grainy one I references earlier: http://www.edac.net/dat/files/161.pdf

evanshultz:
I also find it common to mask the entire edge where the fingers are, and not just between pins. This is the only way I can recall seeing it, in fact.

This library is for KiCad v5. If you're not using it, you should. You will find how to add a polygon in the Place menu, and adding a pad is nothing pad. A pad is probably easiest since it can do both sides at once.

DavidGriffith:
Ya, I was hanging on with V4 for a couple reasons.  One of the bigger annoyances for me is how v5 changed the way things are toggled in the 3D view.  I'm working on that right now (https://github.com/DavidGriffith/kicad-source-mirror branch "3d-viewer-mods").

For this pull, I would like for whoever might want to merge this pull request to please hold off until I declare my work complete.  I'm double-checking dimensions and redoing some stuff that made sense in V4 but not anymore for V5.

DavidGriffith:
I think the S100 card edge is complete.  I don't know if putting those strips along the bottom in F.Fab and B.Fab for beveling the card should be done in the footprint or completed board.  Please comment.

ProxyPlayerHD:
I'm sorry but has this ever been added to the program? I'm not 100% sure how it works.

I can find both the 8 and 16 bit ISA connector symbols but neither have a footprint (though the 16 bit one has a footprint "BUS_AT" but it's not assigned by default.)

And having these would obviously be useful for any IBM PC related project.

PR #204, opened 2017-12-31
Add Kingbright KRC011 Reflective Opto Sensor

MatthewStephenson:
This is the Kingbright KRC011 Subminiature Reflective Opto sensor.

Datasheet: http://www.kingbright.com/attachments/file/psearch/000/00/00/KRC011(Ver.15).pdf

image

Symbol PR: https://github.com/KiCad/kicad-symbols/pull/121

The automated check is failing on the "centre of footprint" part, but that is because it's taking the centroid of the pads, which are all on one side of the device. The actual body of the device is centred on the origin.

As always, comments and suggestions welcome.


jkriege2:
Looks good overall. A minor style issue: could you move the silkscreen closer to the outline? It should look something like this:
2018-01-01 12_01_54-footprint editor no active library
This way the footprint is more in line with what we already have in the lib.

Thanks,
JAN

jkriege2:
Also I think the pins are in the wrong order (or the package is pointing in the wrong direction). The following page show a photograph of the device with polarity marker (marking pin 1&2 according to DS):
https://www.mouser.de/ProductDetail/Kingbright/KRC011/?qs=%2fha2pyFaduin5ud%2fvcvjLzacMGbwk0PqVfg5KGgsny0=
Also the second drawing in the DS is a bottom view with the LED/PT pointing forward and the pad 1 on the left. So if you rotate that as it would be on the top of the PCB, that mean LED and PT point downward, when pin1 is on the left

MatthewStephenson:
Hi, thanks for the input. I've updated the silkscreen as suggested - I was initially going for the "diode" style of pin-1 marker but I can see that might not be appropriate for a four-pad device. Is it acceptable to have the 45-degree cut-corner style of marker, or would you like me to change it to the 90-degree style as in your suggestion?

image

Regarding the pin ordering, I'm struggling a bit to understand your interpretation of the datasheet diagrams. I agree with you that the wide end of the triangular polarity marker, on the bottom of the device, points towards the "Pin 1" end. However, I think the middle diagram in the datasheet is a *side* view rather than a bottom view, with the LED and optotransistor pointing towards the top of the page.  I think this is supported by the aspect ratio of the device in the diagram.

To make it clearer, I overlaid my proposed footprint on the image you found (please excuse my feeble photoshop skills), and I think the middle diagram in the datasheet is looking at the device from the direction of the green arrow. As far as I can tell, that would mean that if you tipped the device towards the camera, onto the footprint overlay in front of it, then pin 1 would land on pad 1.

krc011 3d

image

jkriege2:
Hmmm ... please make the 90°-style, as we do not want to have silkscreen below the mounted device. The cut-corner (45°) for the F.Fab is fine!

About the orientation .... Is this device meant to point upwards (then I understand you footprint), or sidewards (that's how I understood it) ...

MatthewStephenson:
I've updated the silkscreen as suggested.

I had intended this footprint to respresent the device pointing upwards, lying flat on the PCB - maybe I should have made this clearer. The device itself seems to have metal contacts on both the bottom and the side (mine haven't arrived yet but various pictures on the internet show this - examples below), so I think it can also be used either flat or side-looking. The datasheet gives few clues other than one bullet-point which says "side irradiance".

The footprint will allow it to be used sideways too without reordering the pads, just by standing it up on the pads instead of lying it flat. There will be a little wasted space behind the device as it's thinner than it is wide. If you want me to create a separate footprint for the "side-looking" orientation, I can.

image

image

jkriege2:
Hmmm ... maybe we should simply have both variants, please name the one, which is dieloocking as `Kingbright_KRC011_Horizontal` and the one pointing upwards `Kingbright_KRC011_Vertical` (V/H designate direction of light).

That would be great!
Thanks,
JAN

MatthewStephenson:
Updated as suggested with separate horizontal and vertical versions.

image

image

PR #207, opened 2018-01-01
made all ThermalVia-parts KLC-compliant 

jkriege2:
- added bottom pad
- removed mask/paste on bottom
- renamed two footprints not in sync with KLC naming conventions


poeschlr:
I would remove the mask completely from the tht pads. That way the mask area is controlled by the smd pads. (Makes it easier to adapt the footprint for smaller mask areas if needed)

jkriege2:
as requested.

poeschlr:
A lot of the tht pads have f.paste layers enabled.

jkriege2:
Which ones do you mean?
e.g. this one:
2018-01-02 09_20_35-footprint editor no active library

These were this way, so I refrained from completely redoing them, I more or less only did some alterations (bottom pad if required, mask-layers) ...

Or do you mean the ones that are completely covered in past (and should maybe have segmented past) ... also these I didn't red, but could if required ...
2018-01-02 09_22_36-footprint editor no active library

SchrodingersGat:
The ones shown in Jan's first example picture must remain that way as per datasheet requirements.

poeschlr:
HTSSOP-32 was named wrong and i needed a new version of HTSSOP-28 (reduced mask) This is why i created PR #208. It might be a good idea to either include the changes i made there in this PR or you revert the changes for HTSSOP-32 and HTSSOP-28. (otherwise there will be merge conflicts.)

jkriege2:
Then I'll revert HTSSOP-28/32 and leave those to you.!

jkriege2:
@SchrodingersGat @poeschlr Can this be merged then? It should make the footprints better, even if some are still not perfect!

poeschlr:
Yes can be merged. Thanks

PR #210, opened 2018-01-03
Add finder relay socket 96.12 for type 56.32

Ekran:
purrint566
purrint567

according datasheet http://gfinder.findernet.com/public/attachments/56/DE/S56DE.pdf

------------


poeschlr:
You advertise this PR as: includes 56.32 and 96.12 but it does only contain one footprint.

The silk outline should be outside of the body. (Because the Fab line should be the exact body outline, this would mean that you need to move the silk outside of the fab outline.)

Make pad A1square, add a 1mm chamfer to the top left corner and an L that hugs the top left corner on silk. ("Pad 1" marker)

Your screenshot is missing the courtyard rectangle. Please add a rectangle on the F.CrtYd layer with offset 0.25 relative to the fab outline.

Ekran:
The 96.12 is the socket for PCB. The 56.32 is mounted in this socket. I will do the required changes.

Ekran:
- silk outline is now outside the body

poeschlr:
The courtyard rectangle should be larger then thy component by 0.25 mm in all directions. (offset 0.25mm all around)

jkriege2:
Does the drawing on F-SilkS make sense? It will be covered/invisible after assembly, so probably we don't need it?

jkriege2:
... we have the solenoid on all existing relays though (switches are sometimes there, sometines not, especially since varying switch-connections might use the same package/footprint)

Ekran:
> Does the drawing on F-SilkS make sense? It will be covered/invisible after assembly, so probably we don't need it?
I use them in the printed documentations for PCBs. Ok, they can be moved from silk to fab.

poeschlr:
Thanks for your effort. The remaining fixes have been done by @Misca1234 in the now merged pull request #773

PR #212, opened 2018-01-03
Add reflexcouplers

jkriege2:
- several footprints for https://github.com/KiCad/kicad-symbols/pull/127
- 3D models: https://github.com/KiCad/kicad-packages3D-source/pull/74 and https://github.com/KiCad/kicad-packages3D/pull/208

2018-01-03 12_50_58-3d viewer
2018-01-03 12_51_14-3d viewer
------------


pointhi:
simple question: is the cutout big enough to be practical producable (because in production there would be no straight 90° edge, but instead some rounded edge due to the milling cutter)

jkriege2:
Hmmm ... do you knwo what cutter-diameters are typically used?

jkriege2:
also: Would you know whether they will cut ALONG the line (center of cutter on drawn line), or until the line (cutter outer edge touches drawn line-center)?

pointhi:
to my knowledge, they cut until the line (otherwise you would always need to know the tool diameter).

At the moment, I have no real numbers about the production tolerances to expect.

jkriege2:
Hi I found one manufacturer that uses 0.8mm tools for cutting ... I increased the cutout-size slightly, so the cutout will be big enough for 0.8mm and also above:

2018-01-03 13_57_33-footprint editor no active library

jkriege2:
@pointhi Better this way? Would you have time to merge this if there are no other problems?

Thanks,
JAN

pointhi:
at the moment, better a bit to big than to small.

@SchrodingersGat we should specify some rules in the future what pcb-manufacture-capabilites the library assumes.

SchrodingersGat:
AFAIK most fab houses operate on the assumption that the line you draw is the boundary of the cut line, not the center. i.e. you specify the *final product* and it is up to the fab house to decide how to adjust for tool diameter, etc.

(This is how I have operated with multiple fabs for many many projects. In my experience it is rare that you have to specify specific tool info).

jkriege2:
OK, then the symbols should be fine. Even if made with a 1mm cutting tool, the reflex coupler will fit in the cutout .... so can this be merged? Models and Symbols have already been merged!

PR #216, opened 2018-01-03
Add CTS 219 DIP switches

twam:
Screenshots:
screen shot 2018-01-03 at 21 13 14
screen shot 2018-01-03 at 21 13 23
Datasheet:
https://www.ctscorp.com/wp-content/uploads/219.pdf
Script:
https://github.com/pointhi/kicad-footprint-generator/pull/105

------------


jkriege2:
Hi!

note that there is a script there: https://github.com/pointhi/kicad-footprint-generator/blob/master/scripts/DIP_packages/make_DIP_footprints.py which generates a bit more detailes footprints for DIP-switches ... 
2018-01-03 21_54_53-footprint editor no active library

Please use/modify your script to use those tools, or maybe recreate the artwork from there, so all DIP-switches in the lib share a common design ...

Best,
JAN

twam:
I did not include those 'switches' on purpose, as the KLC states
> The fabrication layers are used to display the simplified mechanical outline of components on the PCB.

and
> Outline should be simplified and not display complex features

If I interpreted this wrongly, I can add those of course.

jkriege2:
I would add them to make recognizing these parts easier. Also having the position of the actual switching elemnt on the F.Fab layer can make e.g. front-plate design easier. Several switches e.g. have a circle to represent the actual button or show the lever etc. ... 

Looking a bit closer:
- wouldn't the series of switches names e.g. `SW_DIP_x8_W8.61mm_Slide_LowProfile` fit your switches in the Gull-Wing-Variant?
- But: for the J-leg variant there does not seem to be a footprint available yet ... maybe you could simple extend the above-mentioned family for the J-variants with W6.73mm?

twam:
You're right. The SW_DIP_x8_W8.61mm_Slide_LowProfile is somewhat similar. Do you know from which datasheet those values were extracted? Maybe it would help to put those original parts and also the CTS MPNs there as a comment.

Besides that the other DIP packages also look very inconsistent to me, e.g.:
screen shot 2018-01-03 at 22 33 27
screen shot 2018-01-03 at 22 33 32
* Some show the 'switches' in the Fab others in the Silk screen layer
* The KLC states, that there should be no Silk screen below SMD components
* Pin 1 identification is different
* Silk screen vertical lines are different
* Alignment of silk/fab is different

So it maybe worth to discuss what the ideal solution would look like and them align them all.

poeschlr:
You are right. Silk below SMD parts is undesirable and violates the KLC. It would also be nice to have a unified look within the lib. 

I agree with jan that it might be a good idea to add some graphic elements representing the switch in off position. 

About pin 1 marking: It should be a line along the full length of pad 1 if the pad extends beyond the body. (According to IPC) So your upper screenshot is definetly better than the lower one. It might even be a good idea to move the marker closer.
34541026-00aa6124-f0d7-11e7-93c6-e8c75ea6be0e


---

General note:
Not all footprints are fully KLC compliant. We made a lot of them compliant while transferring them over to this new lib but we are only 4 guys and we can only do so much. If you are motivated and have time you are welcome (but not required) to fix the remaining KLC errors in these footprints.

If you don't have the time to fix them, maybe you can at least create a new issue where you document your findings.

jkriege2:
I think the difference with silkscreen below might: The footprints for J-/Gull-pins were (I think) based on the basic DIP-packages, as usually these packages are used for the switches (at least the version with Pitch=2.54mm). The J/Gull-pins elevate the actual body quite far above the PCB, so silkscreen is not a problem (at least I think that was my rationale). For the SMDs with Pitch<2.54mm, the body is way closer to the PCB, so silkscreen starts to make problems (at least what I think COULD make sense ;-) ...

Anyways, it would be ideal to have one single script (or maybe 2) that generates all or mostly all DIP-switches, so the style can be changed at a single location. + the style should be as consistent as possible over the lib!

jkriege2:
Hi!

just a note: I'm starting a rework of the scripts now ... I could also try and add your switches. Would that be OK? I will post an example screenshot later, when I have something!

Best,
JAN

twam:
Good that you mentioned starting to work on it. I'd have done to same. I'm happy to check anything.

jkriege2:
Hi!

please have a look at these:
- https://github.com/KiCad/kicad-footprints/pull/218
- yours should be covered by SMDs with postfix `_LowProfile` and `_LowProfile_JPin`
- https://github.com/pointhi/kicad-footprint-generator/pull/106

OK for you? COmments please!

2018-01-04 12_35_44-pcbnew d__kicad_devproject_devproject kicad_pcb
2018-01-04 12_44_15-3d viewer

twam:
Closed in favor of #218.

PR #221, opened 2018-01-04
Fix SOT143/SOT143R footprints.

krtkr:
Hi all!

Not sure about this one. I believe I found a bug in the library. Here is what I found:

After soldering a board discovered that original footprints is too large and do not have any documentation referred.

Different manufacturers have different footprint recomendations for SOT143(R). One of the largest I found have vishay. Here are links to their documents:
http://www.vishay.com/docs/84030/sot143
http://www.vishay.com/docs/84029/sot143r.pdf

Update footprints according their documentation.

Here is a picture from my microscope:
sot-143-example

------------


krtkr:
By the way, should we rename SOT143R_Reverse to SOT143R? NXT and Vishay just call this package SOT143R:
https://www.nxp.com/docs/en/package-information/SOT143R.pdf
http://www.vishay.com/docs/84029/sot143r.pdf

poeschlr:
For pad sizes and positions you could look at the IPC standard. (Sadly not an open standard. But i found a mirror for the previous version of the standard on a russian site: http://pcbget.ru/Files/Standarts/IPC_7351.pdf)

krtkr:
Hey 
> For pad sizes and positions you could look at the IPC standard.

In the document SOT-143 is standardized at JEDEC TO-253 which is free for download (after registration) [here](https://www.jedec.org/standards-documents/docs/253-d).

Using [PCB Library Expert Lite](https://www.pcblibraries.com/) I got following results:

lxl_sot143_01

lxl_sot143_02

This differ from both current library (smaller) and vishay recommended.

Here is a table with notations from Library Expert to compare:

|Measure|Current|Library Expert|Vishay|
|-----------|----------|-----------------|---------|
|S|0.8|0.75|1.1|
|L|1.4|1.3|0.9|
|W|1|0.63|0.8|
|W1|1.2|1.03|1.2|

What do you think here?

evanshultz:
See https://github.com/KiCad/TO_SOT_Packages_SMD.pretty/pull/26. Jan and I worked on these set of footprints and found many variants available in the market. The good part starts at https://github.com/KiCad/TO_SOT_Packages_SMD.pretty/pull/26#issuecomment-271266480.

I believe all footprints in the lib now are used in the market and need to be retained, but perhaps they could be renamed if we find one is "standard".

Shackmeister:
@krtkr @evanshultz 
Is still open? Should these be changed according to IPC?

evanshultz:
First, it appears the footprint works from the first pic. That's good, so if I understand this issue is only about updating footprints to improve them.

You can see in IPC 7351 (for example http://cxem.net/comp/files/comp149_Footprint-Naming-Convention_-Surface-Mount-Components.pdf and https://www.cskl.de/fileadmin/csk/dokumente/produkte/pcbl/ipc_standard_pcb_library_expert_Land_Pattern_Naming_Convention.pdf) that the `_Reverse` part should go away since the `R` already indicates this.

We could use IPC recommendations, but the problem we run into is that the packages themselves are not standardized! Here are the very first two results from Google: https://www.centralsemi.com/PDFs/case/SOT-143PD.PDF and https://www.diodes.com/assets/Package-Files/SOT143.pdf. Pin 1-2 spacing doesn't match, pin 3-4 spacing doesn't match, etc. Pin width and distance from row-row is about the same, though. We could make a "super set" package that is the biggest of all these and then build a footprint around that which should work for all packages. However, this is a standard package given in IPC-SM-782 (which is old and superseded) while I can't find IPC 7352 or JEDEC TO-253 online. So we are probably best to use IPC-SM-782 to derive the footprint unless a newer standard can be found. Does that sound like the best method to y'all?

Lastly, the `Reverse` footprints don't have pin 1 in the top left so that could be changed as well.

Shackmeister:
Im not too familiar with the IPC standard, but it would be great to figure out if these footprints should be improved somehow :)

evanshultz:
JEDEC TO-236 is SOT-23 and there is a dimensioned package drawing available for free (with registration) from JEDEC. IPC SM-782 also has a SOT-23 package drawing.

If a manufacturer releases a package that differs greatly from the "reference" package dimensions then a custom footprint is in order. But IMO if a manufacturer simply suggests a different footprint for the same package we should err on the side of industry standards and still use IPC-compliant footprints.

PR #236, opened 2018-01-09
Added BLM 18650 Battery Holder

johnfwhitmore:
BLM 18650 Battery Holder - BLM-BH18650-PC2

I don't have a data sheet but a line http://www.blossom-ele.com/sdp/1074412/4/nd-5192049/107419/New_Products.html

screenshot from 2018-01-09 19-53-31


jkriege2:
please move your origin onto pin1 and make pin1 rectangular.

Also: Isn't the rest ring a bit small? Maybe use 3.5x3.5mm pads for you 2mm drill?

johnfwhitmore:
Working on moving the origin presently.

johnfwhitmore:
Travis is complaining about something but I'm not sure what exactly is failing now

jkriege2:
This is because the tag-field in your file is a single number only. I posted a fix for this behaviour, but maybe you could add some more details to the tag?

https://github.com/KiCad/kicad-library-utils/pull/201

poeschlr:
I think the tag should at least also contain the word battery. (or battery holder)

jkriege2:
Please set the correct 3D model name `${KISYS3DMOD}/ Battery.3dshapes/BLM-BH18650-PC2.wrl`

johnfwhitmore:
@jkriege2 Man you're quick. I got that error out of Jenkins and was just fixing it, when you commented. I'll get the hang of this eventually but at present it's taking me a few steps through Jenkins. I hope that's it for the moment.

jkriege2:
Thanks for your contribution!

jkriege2:
PS: Will significantly slow down, when I have to go back to work starting tomorrow ;-)

PR #245, opened 2018-01-11
Add SK6812MINI

fauxpark:
Modified the 5050 WS2812B footprint to create the 3535 NeoPixel Mini. This one is called SK6812 because Adafruit no longer sources their NeoPixels from Worldsemi and there appears to be no equivalent for this part. However, the 5050 sizes are interchangeable - perhaps it should be renamed and have "WS2812B" added to the keywords.

fp
3d

------------


poeschlr:
It seems PLCC led footprints are not as standardized as one would assume. for this reason please add the body size and pin pitch to the footprint name. In addition to this there should be a dash between plcc an the number of pins. This would result in the fp name: `LED_SK6812MINI_PLCC-4_3.5x3.5mm_P1.75`

plcc4

fauxpark:
Renamed and tweaked the clearances, should be good to go 👍

poeschlr:
The courtyard has always been relative to body dimensions and pads. It never was relative to the silkscreen.

About polarity marking. I will do a bit more research. Until now i assumed that the industry standard is to always mark the pad with the lowest "pin number".

fauxpark:
@poeschlr, any progress on this?

poeschlr:
@jkriege2, @evanshultz could you please state your preference how we should handle polarity markers if manufacturers mark pins different to pin 1 on their device?

[I asked over on the forum.](https://forum.kicad.info/t/are-polarity-markers-always-at-pin-1-lowest-pin-number-or-can-there-be-cases-where-they-mark-some-other-pin/9352) I got less answers than hoped. The best suggestion came form `dchisholm` i think. He suggested to always mark pin 1 no matter what. But allow additional mark of the pin marked on the component. (Example by placing a "3" near pin 3 in that case.)

jkriege2:
I would say the same ... mark pin1 in a standard way and add a marker which on F.Fab and F.SilkS which resembles somehow the style of marker on the package ... not sure whether that works in all cases though.

fauxpark:
FWIW I still think having two markings, or one on the "real" pin 1 seems unintuitive. For hand assembly at least, pin numbers are largely irrelevant -- only orientation matters -- so pin 3 is "effectively" "pin 1" for all someone with a soldering iron cares. They're gonna put the corner with the beveled edge on the pad with the silkscreen line, not the one with the "3" marked next to it. If you look at it as if the datasheet is incorrect... it kinda makes sense.

jkriege2:
The markings may also be useful AFTER assembly: You might want them for e.g. debugging to know on which pin you find a certain signal for measurement. then you typically will have the pin-number form the datasheet.

pointhi:
I still think Marking pin 1 as proposed in the standard style will likely result in a bug report in the future. Simply because its not obvious.

What about Markung pin 3 for visual inspection with the normal marker, and add a "1" to the actual pin. This should be more obvious and less error prone

jkriege2:
THat would be a good solution ... all for it!

evanshultz:
I'm with @pointhi on this. Why do we even have a pin 1 mark in the first place? Because most of the time the component has pin 1 marked. In this case, pin 3 is marked so I suggest moving the footprint's mark to pin 3 as well. I think the footprint mark should correspond to the component's mark (which is nearly always pin 1, but not here).

I'm ambivalent on having any mark on the actual pin 1.

fauxpark:
So here is what it looks like with both markers:
3d

Personally I would be left scratching my head for a bit upon seeing this, but I can see how one *might* infer that the reason for it is that the pinout is non-standard. If this is deemed mergeable then I will commit & push.

poeschlr:
I think this is a good solution for this problem. The fact that you would scratch your head upon seeing this is good in this case as it will result in you checking the datasheet of the part.

fauxpark:
Hm, Travis says it failed, probably because my kicad-footprints is out of date wrt kicad-library-utils, but the output shows no errors on the footprint itself. That's what I get for committing to master ;)

poeschlr:
The error is reported because i screwed up in a different pull request and added a new lib without it being in the fp-lib-table. This has been fixed now and is of no concern for your pull request.

Your contribution does look ok now. Thanks

PR #255, opened 2018-01-15
Add Cree RGB LED PLCC6 6x5mm SMD

ouinouin:
Hi , here a footprint SMD LED RGB PLCC6 6x5mm from Cree : 
I ll make 3D model if PR is validated.
http://www.farnell.com/datasheets/2003905.pdf

capture d ecran de 2018-01-15 01-24-46

Regards.


poeschlr:
There is another PLCC6 footprint in the same library. What is the difference between this contribution and the existing footprint?

If we decide to keep this contribution, please remove the suffix CCW. Counter clockwise is the default pin ordering direction.

ouinouin:
Hi, 
In fact I created this one since I was fooled by the existing one, I was naively thinking that PLCC 6 was a dimensional form factor standard , and it looks like it is absolutely not standard (or that LED maker advertise PLCC that is not PLCC : look at this led taken randomly on the internet : 
https://www.bivar.com/portals/0/products/SMP6-RGB.pdf page , you ll notice the package dimensions are not the same than the Cree RGB led.
another example : http://www.qt-brightek.com/datasheet/QBLP679E-RGB.pdf this LED has a space between pin axes of 1.6 mm, the Cree one has 2.1mm ...

maybe keeping a vanilla PLCC-6 package will fool other people, to my mind it would be good to add size info on the footprint naming of the already existing PLCC6 package.

Got if for removing CCW in the naming,  i ll update asap. 
Regards.

poeschlr:
I looked at your footprint. There are a few issues with it.

- The courtyard clearance should be 0.25mm all around. (clearance is measured either relative to pads or to the package. Whatever results in a larger courtyard area.)
- Please also add the pin pitch to the footprint name. And insert a dash "-" between part name and number of pins. So it should be called: `LED_RGB_Cree-PLCC-6_6x5mm_P2.1mm`.
- The pads might be a bit small. especially in x direction towards the inside. But also in y direction.

plcc6

ouinouin:
Hi, 
thanks for your clear comments, 
I put the courtyard clearance to .25 mm (was good on left and right side, adjusted up and down vs package and not vs pads .
I increased pads size taking into account pad tolerances AND pitch tolerances
I renamed the file to add the pitch at the end

PS : what soft did you use for your small explicative sketch , I like it :), is this freecad + some plugins? .
Regards.

poeschlr:
Sorry for my late response (We are short at least one lib maintainer.)

For the drawing i have a tutorial here: https://forum.kicad.info/t/how-to-check-footprint-correctness/9279

The footprint looks good now.

PR #257, opened 2018-01-15
added Coilcraft LPS4018 Inductor footprint

Shackmeister:
Datasheet:
https://www.coilcraft.com/misc/lps4018d.html

screenshot:
lps4018

------------


poeschlr:
You could use polygon pads here. (Would result in a better handling of the mask and paste layers.)
They are allowed for the v5 repo.

Shackmeister:
I know but these footprints are only made/fixed since I'm using them in a design at work :) so no nightly build yet :) if we can accept it as is now, Ill be happy to change it once I got V5 (or the nightly build gets changed so I can run stable and nightly simultaniously on ubuntu)

poeschlr:
I can merge it as is. There are other footprints that will benefit from polygon pads.

I will create an issue about that.

---

The only way i know of to run both is by having one of them in a virtual machine.

Shackmeister:
sounds good :)

would be great if it could work like Freecad and Freecad-daily :)
I talked to JS Reynaud about this, he said it's on his todolist

PR #258, opened 2018-01-16
added Texas_S-PWSON-N6 footprint

Shackmeister:
Datasheet: http://www.ti.com/lit/ds/symlink/tps717.pdf

screenshot:
s-pwson-n6

------------


poeschlr:
The exposed pad has the wrong size. I think you mixed up copper and paste dimensions.

Also the datasheet suggests thermal vias. So add a version that includes them as suggested in the datasheet.

screenshot from 2018-01-18 01-21-56

Shackmeister:
do we have a standard for the size of the Thermal Vias?

Shackmeister:
@poeschlr nice catch Mr René Eagleeye Poeschlr

Travis output should be ok

I used 0.6mm pad and 0.25 mm drill and only *.Cu layer
(found in another footprint)

pwson_thermalvias

poeschlr:
You also need a copper only pad on B.Cu for radiating heat.
More details see KLC
http://kicad-pcb.org/libraries/klc/F4.4/

Shackmeister:
Ill give it a try and then you'll probably have to correct me :)
If Paste is not ok to have on the Vias should I just split up the paste layer enough to avoid the vias?

Shackmeister:
first attempt
pwson_6_thermalvias

poeschlr:
Back layer should only be copper so this is ok.

Splitting paste does not really make much sense for this FP. (The datasheet does suggest the paste area as you made it.)

Shackmeister:
so it's all good? :)

poeschlr:
Thanks for the contribution

PR #263, opened 2018-01-16
added SW_SMD _RS-282G05A3-SM

fandres:
This is the model of the SMD switch RS-282G05A3-SM and similar, I am not sure if the name is correct, I do not know if there is any special notation for the names in kicad.

DataSheet: https://www.mouser.com/ds/2/60/RS-282G05A-SM_RT-1159762.pdf
Preview Model: 
screenshot_20180118_171204

------------


poeschlr:
The naming convention is part of the [kicad library convention](http://kicad-pcb.org/libraries/klc/) found on the kicad webside.
These rules do not yet have a detailed description for every possible component. One missing example are switches. But we can take other switches in this lib and the KLC as a guide.

I would say the correct name should include the manufacturer name: `SW_SPST_CK_RS282G05A3` (And yes this is missing for many switches. Man this lib is kind of a mess.)

poeschlr:
Please rename the footprint as by my request.
The fab outline is not quite right. Currently the body outline has the part width as 3.5mm but it should be 3.6mm as by the datasheet. (Currently the courtyard clearance is also wrong but if you fix the body outline it would be correct.)

screenshot from 2018-02-02 13-41-13

fandres:
I understand, as soon as I can fix it- Tnks

poeschlr:
If you have a courtyard clearance of 0.3 and move the fab line outwards by 0.05 you will result in the correct courtyard clearance of 0.25.
If you then also move the courtyard outwards by 0.05, you will again get the wrong clearance of 0.3mm
So move the courtyard back in.

fandres:
Excuse me, it was my confusion.

poeschlr:
Thanks for your contribution.

PR #267, opened 2018-01-18
BQ27441 footprint orientation fix

aurabindo:
As per the datasheet, Pin 1 is in Q2. Relevant section for BQ27441 :
bq27441_datasheet_orientation
Datasheet: www.ti.com/lit/ds/symlink/bq27441-g1.pdf

Existing footprint:
bq27441_old

Modified footprint:
bq27441_new

Kindly confirm if my understanding is correct, since I have not yet assembled any designs.


aurabindo:
How to generate the corresponding 3D footprint ?

poeschlr:
You are not required to supply the 3d model it self. But the footprint must be prepared to accept one. This means the 3d settings need to be filled out.

If you are asking because you are interested in this:
The best way is to script such 3d models. Normally we use the framework provided here: https://github.com/easyw/kicad-3d-models-in-freecad
Another option is to design the model in any CAD tool. For freecad there exists the kicad-stepup workbench which allows exporting 3d models for the use in kicad. (correctly scaled wrl file+step file)

---

About your contribution:

I don't think it is necessary to rotate the footprint within the library. As far as i understand IPC standards, the zero rotation for footprints is independent of the rotation of the component inside the pick and place machine. For footprint orientation there are normally two options. (A and B) The kicad library uses option "A" which states that pin 1 is always at the top left corner. (or on the left side if the pin is not located at a corner)

The person programming the pick and place machine knows what the zero orientation of the footprint is, they know the rotation of the footprint on the pcb is (from the pos file, this is relative to the zero orientation) and they know the orientation of the part within the assembly machine (tape). This is enough information for them to program the assembly machine.

aurabindo:
I making a design with the intention of using a pick n place assembly line, and this is exactly what I wanted to know. Just to confirm:
> The person programming the pick and place machine knows what the zero orientation of the footprint is

This would be the IPC type A which Kicad follows, and hence consequently the zero orientation of this particular footprint ?

evanshultz:
There is another step: The operator will also program the orientation of the IC within the reel (or this may be automated and scanned by the machine). Either way, by knowing the orientation of the part in the reel and the rotation of the footprint on the board that is all the required info. All KiCad footprints place pin in the upper left corner. There is some manual work to do be done by the operator, but KiCad is no worse than any other ECAD software in this respect.

So... as @poeschlr said this isn't necessary.

aurabindo:
Thanks for the info @evanshultz

PR #270, opened 2018-01-19
Add Texas_S-PVQFN-N20_EP2.7x2.7mm

stephf0716:
matching footprint(s) for DRV8662 package
http://www.ti.com/lit/ds/symlink/drv8662.pdf
image
image

Link to symbol PR: https://github.com/KiCad/kicad-symbols/pull/162


Shackmeister:
@poeschlr would you mind if we supply the TI specific packages both with and without ThermalVias?

poeschlr:
@Shackmeister look again at the contribution. @stephf0716 does supply both a version with and one without thermal vias.

Shackmeister:
Hmm wasn't fast enough to remove my comment :) noticed it a minute ago

poeschlr:
Thanks for this contribution. Everything looks good now.

PR #272, opened 2018-01-19
Fix: rename the Neutrik NMJ6HCD2 part

pushytoxin:
This is a socket for 6.3mm phone jacks, not 3.5mm


poeschlr:
Could you provide evidence for your claim?

pushytoxin:
Here's the manufacturer's product site, where it says it is a 1/4" jack (http://www.neutrik.com/en/audio/plugs-and-jacks/m-series/nmj6hcd2).
The drawing pointed to by the "datasheet" entry doesn't explicitly specify whether the plug diameter is 3.5/6.3mm, however the proportions definitely indicate the latter.

pushytoxin:
Please don't merge this yet, I think I found another issue. The part seems to be rotated (or mirrored).

pushytoxin:
The pins are definitely incompatible with the Connector/Audio-Jack-3 and Connector/Audio-Jack-3_3Switches symbols.

I have rearranged them so they properly match, however I'm not sure about the details.

Pin 1 used to be the sleeve connector, and it was a square pad.
Now Pin 3 is the sleeve connector. I believe, that Pin 1 should always be a square bracket, is that correct?

PR #275, opened 2018-01-20
rework several THT Relays (to make fit for new 3D models)

jkriege2:
- reworked several relays (rect pin1, pin1-marker on F.Fab/F.SilkS, orientation for 3D models, 3D model-links, some renames)
- added some missing relays (Finder_34.81, Finder_30.22, Finder_34.51, Finder_40.11, Finder_40.41, Panasonic_JW1)
- 3D models: https://github.com/KiCad/kicad-packages3D-source/pull/84 & https://github.com/KiCad/kicad-packages3D/pull/224
- symbols: https://github.com/KiCad/kicad-symbols/pull/195
- datasheets: 
  - https://gfinder.findernet.com/public/attachments/34/EN/S34USAEN.pdf
  - https://gfinder.findernet.com/public/attachments/30/EN/S30EN.pdf
  - https://www.finder-relais.net/de/finder-relais-serie-40.pdf
  - http://www3.panasonic.biz/ac/e_download/control/relay/power/catalog/mech_eng_jw.pdf?via=ok
- remaining KLC-errors/warnings:
  - Relay_SPDT_Schrack-RT1-16A-FormC_RM5mm: uses the 3D model from another footprint (thus the name-warning, but this is OK, just the pin-numbering changes in the footprint)
  - Relay_SPDT_Schrack-RT1-FormC_RM5mm checks out locally ... seems to be a bug in the scripts?

2018-01-20 16_35_34-3d viewer
2018-01-20 16_35_24-3d viewer
2018-01-20 16_34_01-pcbnew d__kicad_devproject_devproject kicad_pcb

------------


Ekran:
Hello, I would suggest to show the common, NO and NC pin like here: https://github.com/KiCad/kicad-footprints/pull/210

jkriege2:
This drawing on silk screen will be covered by the package, so is not visible after assembly ... so what's the point of it?

poeschlr:
We could add them to B.Silk. That way they are visible for trouble shooting without the need to look at the documentation. Maybe in addition also include them on fab for people who don't want to afford bottom silk.

evanshultz:
I would merge without any drawings or pin labels, just mentioning that personally I have found it helpful when troubleshooting to have that info on the footprint.

I have created unique footprints for each type of relay so the footprint does truly match the installed part. Not sure if this would be part of having an atomic component or not. Lastly, relays are generally simpler than the list you sent above and also, at least in my experience, and there are few in my library.

Ekran:
> Then what about relays in the same package, just with different pin-connections for different switch configs? 

This is the case only for relay with DIL housing, but not the most other because the inductor has pins named A1 A2, the switch 11, 12, 14 and 21, 22, 24 and so on

> I would merge without any drawings or pin labels,

+1

> just mentioning that personally I have found it helpful when troubleshooting to have that info on the footprint.

+1

poeschlr:
I think relays are one of the components where we might want to have true atomic parts. (Footprint and symbol specialized for one component. Both named with the manufacturer and manufacturer part number. 1:1 connection.)
Just because there are not really generic relays (or too many different types to find a naming convention that is still readable but conveys all the information necessary.)
If the relay footprints are then meant for only one specific part we can include graphics as well.

jkriege2:
I would be against a requirement/fixed rule, if we require something like that I would like to see that as a proposal, where it makes sense!

Since I redid several of the relays in the past weeks ... I fear a fixed rule would overwhelm us/the contributor with work? 

I would try and limit the number of footprints as good as possible (often the pin-numbering of the Manufacurer requires different variants anyways, then why not add the drawing), but I wouldn't force such a scheme ... just starting with the DIP-relays that will end up in a real mess, I think ...

jkriege2:
@poeschlr @evanshultz 

So to sum this discussion up: We don't require those drawing, but also are not against them if they are there.

Since I won't really want to put much more time into this PR, I will leave them out for now and we can possibly add them for these footprints ata later time. OK?


Would one of you then have time for a review of this PR (+symbols+3D packages)?

Best,
JAN

poeschlr:
Finder 30.22 does confuse me a bit. Where did you get your strange dimensions for the pad placement?
And the bottom courtyard clearance is not quite right.
screenshot from 2018-01-29 00-33-54

34.51 horizontal also has a wrong courtyard clearance on the bottom side.
screenshot from 2018-01-29 00-40-44

34.51 vertical looks good.
40.11 looks good

The datasheet you linked for the Panasonic JW1 Form B does not contain a form B. Only Form A and Form C. I assume this one should be called Form C.
If it is form c, two measurements are slightly wrong.
screenshot from 2018-01-29 00-56-57
The same measurements are also wrong for the Form A variant.

---

I did not check the footprints you only reworked. They don't have an obvious problem.

jkriege2:
- Relay_DPDT_Finder_30.22: fixed
- 34.51 actually that should be OK ... the pads to the right are larger (as are the pins), so the courtyard there should be fine!
- JW1's fixed (including 3D models)

poeschlr:
Looks good. Thanks

PR #279, opened 2018-01-21
Fix #235: FSDH321L and FSDL321L are missing an 8LSOP package

### LINKED ISSUE #235: FSDH321L and FSDL321L are missing an 8LSOP package
Datasheet: https://www.fairchildsemi.com/datasheets/FS/FSDH321.pdf

Components FSDH321L and FSDL321L available at the symbols repo (Converter_ACDC) are missing a footprint.

stephf0716:
Fix #235: FSDH321L and FSDL321L are missing an 8LSOP package
https://www.fairchildsemi.com/package-drawings/ML/MLSOP08A.pdf

image
image


jkriege2:
Hi!

I think you have two options for the name here (footprint looks good ;-)

1. See it as manufacturer-specific footprint, then the name should be `Fairchild_8LSOP`
2. You can add this to the standard range of SMD-mounted DIP-packages, then the name should be `SMDIP-8_W9.53mm_Clearance6.7mm` ... you could even have a look at the existing `SMDIP-8_W9.53mm`, if it might simply be used (although it has another clearance of 7.45mm):
2018-01-21 08_57_08-footprint editor no active library

In both cases the package should go to `Package_DIP`. Personally I would tend to option 1 ;-)

Also have a look at the existing `PowerIntegrations_SMD-8` in `Package_DIP`, which is already very close to what you want (clearance is 6.6mm only):
2018-01-21 08_59_39-footprint editor no active library

Maybe you can also modify the latter to match you symbol and retain the style used there?

stephf0716:
I will opt for `Fairchild_8LSOP` for the naming.

Interesting, I wasn't aware of SMDIPs, I will locate it in `Package_DIP` as you mentioned. 

I could modify the latter symbol, but doesn't it violate [KLC F5.1](http://kicad-pcb.org/libraries/klc/#F5.1)?

> 3. For SMD footprints, silkscreen must be fully visible after boards assembly (no silkscreen allowed under component)

jkriege2:
You mean about silkscreen below the package ... yes in principle it does, but then these are actualla THT-components with bent pins and mounted as SMD ... so we kept the style from the well-known DIPs ... but then that#s always matter of debate ;-)

jkriege2:
KLC are meant more as guidelines that we try to adhere to as best as we can. Whenever there is a reason, an exception can be made (but just on a case-by-case basis.).

stephf0716:
should be ok to merge and close #279

antoniovazquezblanco:
Thank you @stephf0716!

stephf0716:
image

Why does it say keyword field is empty? The tags field is not empty as shown in the following: 
image

jkriege2:
looks like a script-bug to me ... just ignore it ;-)

jkriege2:
looks like a script-bug to me ... just ignore it ;-)

jkriege2:
Thanks,
JAN

PR #280, opened 2018-01-21
add DFN-14-1EP_4.5x3mm_P0.65mm footprint

menschel:
This patch adds DFN-14-1EP_4.5x3mm_P0.65mm footprint as described in 
https://ww1.microchip.com/downloads/en/DeviceDoc/20005688A.pdf 
on page 87

dfn-14-1ep_4 5x3mm_p0 65mm

Checking footprint 'DFN-14-1EP_4.5x3mm_P0.65mm' - No errors

There is no 3D model yet!

------------


poeschlr:
Please add the ep size as a suffix to the footprint name. (`_EP[size-x]x[size-y]mm`)

The datasheet also suggests the addition of thermal vias. You might want to add a separate footprint with these. More details see: http://kicad-pcb.org/libraries/klc/F4.4/

For large pads like the exposed pad of this footprint, the solder paste layer should be split up into smaller sections. See http://kicad-pcb.org/libraries/klc/F6.3/
The paste coverage should be between 50% and 80%.

menschel:
The file was renamed and the solder paste area changed to 80% by design
rule.

Unfortunately the check script did not raise an error towards that issue.

I can make the _ThermalVias footprint but KLC conflicts with the
datasheet in terms of filling the vias.

Is there a "right way" to make filled vias in KiCAD?

poeschlr:
If by filled vias you mean [plugged vias](http://www.epectec.com/pcb/plug-via-process.html) then i fear there is no native kicad support for them.

I would add them in the usual way similar to other footprints and let the user communicate the fact that these should be plugged in their personal preferred way.
(Via plugging is normally done to limit loss of solder paste. This can also be achieved by ensuring that paste is only applied to areas where no via is located.)

jkriege2:
Sidenote (so this appears over there): This footprint is intended for https://github.com/KiCad/kicad-symbols/pull/201

menschel:
The requested footprint variant with thermal vias has been added.
Paste coverage calculates to 53percent.

Sidenote: check_kicad_mod.py finds false positives when Mask and Paste
Fields are not identical.

'DFN-14-1EP_4.5x3mm_P0.65mm_EP1.65x4.25mm_ThermalVias':
  Violating F6.3
    Pad requirements for SMD footprints
    Pad(s) potentially missing layers
     - Pad '' missing layer 'Mask'
     - Pad '' missing layer 'Mask'
     - Pad '' missing layer 'Mask'
     - Pad '' missing layer 'Mask'
     - Pad '' missing layer 'Mask'
     - Pad '' missing layer 'Mask'
     - Pad '' missing layer 'Mask'
     - Pad '' missing layer 'Mask'
     - Pad '' missing layer 'Mask'
     - Pad '15' missing layer 'Paste'
     - Pad '15' missing layer 'Paste'

evanshultz:
Body dimensions are XxY, so for this part it should be `3x4.5mm`.

evanshultz:
The issue with missing layers is a known bug. We improved the pad design we use but the script hasn't yet been updated to check it properly. You can ignore this error/warning.

menschel:
fixed the name

evanshultz:
Thanks for making these updates! There is still more work to be done:
- You changed the file names, but not the footprint name or 3D model name within the footprint.
- Please add a link to a datasheet with this package in the footprint description.
- Shrink the F.Fab text slightly to fit within the part body outline.
- Silk lines should be 0.12mm (yours are 0.15mm).
- Fab lines should be 0.1mm (yours are 0.15mm).
- The courtyard should be 0.25mm outside the body or pads for this type of part.
- Your paste pads have the F.Cu layer turned on; set the copper layer to None.
image

These rules, and a lot more, can be found at http://kicad-pcb.org/libraries/klc/. There are check scripts that verify some subset of these rules, and the results are visible at the Details link next to the "Travis CI" text above. Please ask if you have more questions.

menschel:
Done.

evanshultz:
Thanks! Just one more thing: the courtyard still isn't quite right. It should be on a 0.01mm grid and 0.25mm from the part body or pad edges. The X dimension should be 2.18mm (you have 2.3mm) and the Y dimension 2.5mm (you have 2.8mm). If you can fix that then I will merge.

X:
image

Y:
image

menschel:
Done.

evanshultz:
Great! Thank you!

I see this was your first contribution to KiCad. Thanks for sticking with the comments above and getting everything updated. And also for making comments about check script issues (which have at least a preliminary fix now). I hope to see more contributions from you in the future!

PR #281, opened 2018-01-21
Add Fiber Optic Tx/Rx AFBR-1624Z AFBR-2624Z

vk6677:
Add Fiber Optic Tx/Rx AFBR-1624Z AFBR-2624Z


jkriege2:
Hi!

please provide screenshots and sign the license agreement for your contribution. Also please have a look at the output of our automatic (continuous integration) check scripts. They check several aspects of the KiCAD library conventions that we require any contribution to adhere to. You can find the current KLC version there: http://kicad-pcb.org/libraries/klc/

Best,
JAN

Shackmeister:
hi @vk6677 I had a look  at you footprint and have a few comments

- a datasheet is missing from the description (makes it pretty hard to check it for me as well)

- The upper part of the fab outline doesnt seem to fit very well with the model you provided. in case the dimensions arent provided in the datasheet, I would suggest to use freecad to measure them.

- is there a specific reason why all the arc are split into multiple smaller arcs?

- Courtyard can be a more complicated shape to fit better, and should be 0,5mm from the fab outline

- since pin 1 should be in top left corner, the fooprint should be flipped 90 deg CW 

Please let me know if you need some help with some of it :)

PR #285, opened 2018-01-21
Add EP size to fp names. Fix paste definition.

poeschlr:
In the end this PR will add the EP size definition to all footprints that have an external pad.
I use this also to fix the paste definitions of the footprints that i touch.

When this is finished, i will also make a PR for the packages3d repo and the symbol repo to get them in line with this PR.


poeschlr:
@jkriege2 could it be that travis is not working for this PR? (It says everything is ok but i think there should be more output in the report. I also think it should report some errors.)

poeschlr:
This should be ready now. Remaining travis errors are from the THT pads used for thermal vias footprints. (We might need to add an exception to that rule check.)

I checked the footprints multiple times and i am sure they are better now then they were. I don't suggest that there is no error left in them. (I hope i did not introduce new errors however.)

evanshultz:
OK. I only did a very cursory check. We do need to update the scripts to ignore thermal pad errors because there are several causes where bogus errors are created.

poeschlr:
I know. I think one way to reduce false positives is by turning off this error if the footprint name contains _ThermalVias. Maybe in addition check if a smd pad on top and bottom with the same pad number exist. (A perfect check would also test the position of the THT pads against these pads but i think that might be too much work.)

evanshultz:
I agree. For robustness, I think checking for a via with matching pin number inside a copper pad is best.

Also, we should not generate an error if a paste opening if embedded inside a copper pad. The paste opening should not have a pin number, either.

While we're talking about scripts, should the check scripts be a pinned repo at https://github.com/KiCad? Should the footprint generator scripts be migrated to https://github.com/KiCad as well, and pinned there too?

evanshultz:
@poeschlr 
Can you please confirm how you want thermal via mask handled? Just to be absolutely clear. Here is `HTSSOP-16-1EP_4.4x5mm_P0.65mm_EP3.4x5mm_Mask3x3mm_ThermalVias`:
image

The mask is opened under the EP.

So in the official KiCad library, mask should be over vias by default. Mask is only removed over vias indirectly if the mask is removed for another reason.

I'm wanting to make sure the latest comment at https://github.com/KiCad/kicad-symbols/pull/289 is answered properly.

awygle:
Note that Cypress specifically requests a top-side via mask cap for this footprint in the datasheet:

image
image

poeschlr:
If you want to build a footprint exactly as in a datasheet prefix the footprint name with the manufacturer name.

poeschlr:
Yes the bottom pad and the vias should not have mask active unless there is a good reason given for it.

The top pad has mask selected unless the datasheet asks for a mask cutout of a specific size. In such cases it makes sense to create a mask only pad that handles this.
(Even if the mask cutout can be achieved by setting the mask clearance as it reduces math required. If it can not be achieved with clearance settings then it is always necessary to have a separate mask pad.)

(If in doubt, `HTSSOP-16-1EP_4.4x5mm_P0.65mm_EP3.4x5mm_Mask2.46x2.31mm_ThermalVias` is a footprint that has all possible options in one.)

PR #287, opened 2018-01-21
added TE 1-776087 23POS AMPSEAL

kleini:
Datasheet: http://www.te.com/usa-en/product-776087-4.html
The 3D model is already added and positioned. I will create a separate pull request to the kicad-packages3D repository.

te_1-776087

Replaces pull request 286 due to wrong branch on my side according to contributions guide.

------------


Shackmeister:
Hi @kleini have you taken a look at the Traivis output? :) these errors will need to get fixed before it can be merged

kleini:
Sorry, I don't have enough knowledge about PCBs and all their names and definitions. What is RefDes, F.Fab, courtyard B.Cu, B.Mask and so on? I have no idea, what I need to fix.

poeschlr:
RefDes is the reference designator [More info see Wiki](https://en.wikipedia.org/wiki/Reference_designator) Kicad by default adds one to the footprint. To get a second reference designator on the kicad layer "F.Fab" you need to add a user text field with content %R

Everything else you mentioned are technical layers in kicad. Every layer can be exported to one gerber file.
- The F.Cu and B.Cu layers are the copper layers (F = front, B = back), 
  - Pads can also define that they should be on all copper layers (=*.Cu) but they can not be placed on a specific inner copper layers.
- F.Mask and B.Mask define [soldermask](https://en.wikipedia.org/wiki/Solder_mask) cutouts (It is the negative of the resulting film that covers your board.)
- F.Fab and B.Fab are documentation layers. These are intended to be used for communicating with board assembly houses and for user documentation.
- F.CrtYd and B.CrtYd are used to define a courtyard area. (The courtyard defines where no other component should be placed.)
  - This area takes as reference the bounding box around pads and component body (Whichever results in a larger area)
  - Depending on component there is a clearance between that reference shape and the courtyard shape. (For connectors the clearance has to be 0.5mm)
  - In addition the courtyard is drawn to the nearest 0.01mm grid.

---

Travis tries to tell you that through hole pads should have all copper layers enabled plus top and bottom mask.

It also requests that you put the exact component outline on F.Fab layer plus a copy of the reference also on fab.

From the screenshot it seems you have not defined a courtyard area.

You also do not seem to have a pin 1 marker on the silk layer. (This marker should be made such that it is visible after component placement -> outside the area covered by the component)

---

About other kicad layers:
- F.Paste and B.Paste define the area that will be covered with [solder paste](https://en.wikipedia.org/wiki/Solder_paste) (In datasheets often called stencil) This is not used for through hole components like your connector. (Only SMD components can be [reflow soldered](https://en.wikipedia.org/wiki/Reflow_soldering))
- F.Adhes and B.Adhes are layers to define adhesive (=glue) areas. For automatically manufacturing boards this becomes necessary if you have large components on the bottom side while reflow soldering.
- Dwgs.User and Cmts.User are used for user drawings and comments. (In the official lib we use these layers to communicate between the footprint designer and the footprint user. An example is to communicate copper keepout areas as these are not directly supported by kicad.)
- Eco1 and Eco2 are layers with no specific defined purpose. (The user can use them for whatever they want. We do not use them in the official lib)
- Margin layer: I do not really know the purpose of this layer.
- Edge.cuts: This layer is used to communicate with the manufacturer what the final board shape should look like.
  - It is not supported in the footprint editor. (Limited support will come in version 6 of kicad)
  - It would be useful for defining edge connectors or components that need a cotout

kleini:
Many thanks for the very good explanation. "I'll do my very best."

poeschlr:
I used my explanations above as a basis for a new FAQ entry on the kicad forum: https://forum.kicad.info/t/what-is-the-meaning-of-the-layers-in-pcb-new-and-in-the-footprint-editor/9688

Shackmeister:
hi @kleini are you planning to continue with this PR? :)

kleini:
all open issues are fixed

poeschlr:
Sorry it took so long. I finally found time to look at this in more detail. I used [this mirror of the datasheet](http://www.te.com/commerce/DocumentDelivery/DDEController?Action=srchrtrv&DocNm=776087&DocType=Customer+Drawing&DocLang=English) to check the footprint as there was a problem with the TE website at the time i checked the footprint.

In general the pad positions look ok. However the drill size for the stabilizing pins should be 2.85mm instead of 1.75mm

---

Now the really bad part. You should definetly make the outline on fab a bit more detailed. You should also add a mark where the board edge should be. (or where it is not to be)
Add a fab outline also for the connector part that is outside of the board. (to show the user of the footprint which side of the connector is the one where the connection takes place.)

Add a silk outline (only on the part that is on the board.
Add a pin 1 marker on silk

The courtyard outline should have a clearance of 0.5mm all around.

---

For reference the dimensioned drawing i made of this footprint.
screenshot from 2018-05-08 15-59-04

kleini:
Maybe we should give up on this connector trying to add it to the library. KiCad does not work any more after upgrading to Bionic Beaver.

poeschlr:
KiCad has a few problems on ubuntu 18.04 it seems. Head over to the bugtracker or to the forum for tips how to get it back to work. (My suggestion formed by my 10 years of experience using fedora is not to upgrade any linux operating system on the first few days of its release. Especially if you use specialized software.)

---

The thing with the "not even able to make a pad square" part is what surprises me a bit. The pad shape is set in the same dialog that you use to add a pad. Getting to this dialog after the pin has been placed is done like in every other software. (using the left click context menu) There is even a shortcut (e for edit)

For the connector outline you can use dxf import or freecad stepup if you much rather use mechanical CAD tools for that part.

kleini:
Sorry, in my edit dialogue was no option to make pad square. There are just different sizes of holes and pads available. Both the routing part and the part editor die with a segfault. I still have one system, that is still on Ubuntu 17.10. Will use that to make the necessary changes. Do you have a hint for a similar part where all these necessary outlines exist, so I can get some advice how to do that?

poeschlr:
Any sub-d connector should give you an inspiration.

The edit dialog for pads should look like this:
screenshot from 2018-05-09 14-13-09

Shackmeister:
Hi @kleini are you planning on working on this PR? :)

kleini:
Hi, I am still working on this. I added already all necessary outlines. Still open are the rectangular pads. Stay tuned.

Shackmeister:
Sounds great :) thank you for the response.
And please don't hesitate to write here if you need some help :)

Shackmeister:
Hi @kleini 
footprint is looking pretty good, but I still have a few remarks :)

- silkscreen width should be 0.12mm instead of 0.1mm :)

- instead of a triangle on the silkscreen, a line is typically used. (see image 1)

- F7.5 only says the annular ring must be **at least* 0.15mm, so the pads should definitely be bigger than they are now. Bonus info, remember the courtyard is from the fabline or the pad, which ever is further out.

- Stabilizing pads should have "MP" as the pinnumbers, short for "Mounting Pads"

- please add a triangle on the fablayer indicating where the pin 1

- fab REF** (%R) should be in the center of the footprint when possible

- A line should be added on Dwgs.user together with the text "PCB edge here" to indicate where the Edgecut should be.

- the courtyard line going through the footprint should be removed

kleini:
- I copied the silkscreen width from some D-Sub connector. Then it is wrong there. Corrected to 0.12mm.
- Triangle is copied from same D-Sub connector. Fixed.
- Regarding F7.5: The connector has two mounting pins placed through the board. Those pins are made out of plastic and they can not be soldered. Therefore a pad does not make sense here at all, but the whole should be there in the board. These plastik pins are inside the fabline and inside the courtyard. I still have no solution to modell that without violating F7.5.
- Named those holes for mounting pins "MP".
- Created a triangle to mark pin 1 on the fablayer.
- fab REF** can not be in the center of the footprint, because there are pads. Tried to position it as near to the middle as possible.
- Removed courtyard line going through the footprint. This was the indicator for the PCB Edge, stolen from the D-Sub 25 connector. Replaced that with a line on fab and placed the text "PCB Edge".

You should mention somewhere in your documentation a perfectly designed footprint, which can be used to learn from.

Shackmeister:
Sorry didn't understand it was plastik.
You should do the following :
- remove the "MP" pin number. 
- change the pad type to "NPTH" which stands for "Non Plated Through Hole" which is the type you should use for these holes. 
- Fab REF** can be placed over pads, shouldn't matter when only fab layer is plotted (great for placement drawing)

kleini:
- Got NPTH through checks.
- Fab REF** is placed in center over pads.
Done from my side. Please check.

Shackmeister:
last note for the drawing, please center the value designator.

I believe the Name needs some tweaking.
TE_AMPSEAL_776087-x_3Rows_23_P4.0mm_Horizontal
or
TE_AMPSEAL_1-776087-x_3Rows_23_P4.0mm_Horizontal

If I understand it correctly, the "-1" is depending on gold pins.
Shouls we go with one of them or should we have an "x-" option?
@poeschlr @evanshultz @Ratfink

kleini:
Will rename to TE_AMPSEAL_776087-x_3Rows_23_P4.0mm_Horizontal. The leading "1-776087" vs. "776087" indicates contacts in Gold while the other one has tin. The trailing "-x" indicates the color of the housing
- -1 black
- -2 natural
- -4 gray
- -5 blue
Regarding the above explanation, how should the naming be?

Shackmeister:
Since all the other TE footprints has "1-" in the MPN, and nobody has made any noise. I suggest we add it, and then merge it.
Thanks for the patience and your contribution to kicad :)

thomasonw:
One suggesting: it would be helpful to indicate somewhere definitively where the front mounting surface of the connector is (the face that mates up with the the panel).  Much like the PCB edge is show, but in this case showing where the panel inside face would expect to be.

Shackmeister:
Hi @kleini 
The (hopefully) final changes needed :)

- add the "1-" in the MPN

- the 2 horizontal lines at position Y 15,8mm, should be moved +0,75mm (17,3 in total) since thats where the actual part ends (note the small bumps). Please replace these to lines with one going thrught the component

- Add a text like the "PCB edge" in dwgs.user. with something like "Panel edge" (@poeschlr @evanshultz @Ratfink suggestions for better name?)

kleini:
@Shackmeister 
- Names have everywhere the leading "1-".
- I moved the lines to 16,55mm as there ends the plastic part of the connector. The additional 0,75mm of the gasket needs to be compressed to create a fully waterproof sealing.
- I created the additional test "Panel edge"

kleini:
Sorry, wrongly pressed the "Comment and close" button.

Shackmeister:
@kleini 
Since there hasn't been any objections, Im gonna say thank you for your contribution and merge the footprint :)

PR #289, opened 2018-01-22
Courtyard fixes for keyswitches and IDC headers

fauxpark:
@poeschlr alerted me to the fact that I've been doing courtyard lines wrong. I figured I should go back through my previous contributions and redo them so they're offset from the fab outline instead of the silkscreen.


poeschlr:
I checked a few of these at random, they seem ok so i will merge as is. Thanks

PR #292, opened 2018-01-22
Add Microsemi_QFN-40-32-2EP_6x8mm_P0.5mm

evanshultz:
I just realized that https://github.com/KiCad/Housings_DFN_QFN.pretty/pull/74 didn't get merged. See that PR for an image.

Datasheet: https://www.microsemi.com/document-portal/doc_download/131677-pd70224-data-sheet

Image:
https://user-images.githubusercontent.com/1936989/32519326-3e3328c0-c3c1-11e7-999c-6e7c61c73e98.png

Notes:
- I didn't add any paste pads on the bottom layer.
- I used the pin numbering scheme @SchrodingersGat mentioned when pins are removed, and I compressed the pin numbers used to avoid any skipped pin numbers.
- Schematic symbol PR is at KiCad/kicad-library#1812.
- Footprint is used by symbol PD70224 at https://github.com/KiCad/kicad-symbols/blob/master/Power_Management.lib.
------------

- [ x ] Provide a URL to a datasheet for the footprint(s) you are contributing
- [ x ] An example screenshot image is very helpful 
- [ x ] If there are matching symbol or 3D model PRs, provide link(s) as appropriate
- [ x ] Check the output of the Travis automated check scripts - fix any errors as required


poeschlr:
The current way to define such footprints is as follows:
- Have one large smd copper pad on top that optionally defines mask as well. (If a specific mask cutout size is required this should be done via a special mask only pad that has no pin number.)
- The vias are defined as copper only with the same pin number as the pad.
- If vias are added, a separate copper only pad on the bottom should also be added. (No mask and definitely no paste on bottom)
- Paste should be broken up into smaller sections by using special paste pads. The paste pads should not have a pin number. paste coverage should be between 50% and 80% of the mask free copper area. 

This allows easy modification of the footprint by the user. (Example: They can change where the vias are, change their size, ... without the danger of changing the copper pad size.)

evanshultz:
@poeschlr 
Sorry. It appears the official Microsemi URL isn't working.

Please see pages 15-16 of http://www.mouser.com/ds/2/523/ds_pd70224_rev1_3-1013855.pdf. I have built the footprint exactly to the datasheet recommendations and named it as a manufacturer-specific footprint. (I could have made a mistake, of course, but I believe I did it correctly.)

evanshultz:
F.Cu:
image

B.Cu:
image

No F.Paste (showing F.Mask):
image

Top and bottom have exposed copper where indicated, with small squares of paste on top that will spread during reflow.  No mask over the vias.

Have I done something wrong? It's a complicated footprint, so perhaps my mind isn't reading the datasheet correctly.

evanshultz:
And no F.Mask (showing F.Paste):
image

evanshultz:
In the last screenshot the lower pair of paste pads don't seem to show up, but they are there. 1.68x2.95mm.

evanshultz:
@poeschlr Is this good to merge now?

poeschlr:
The bottom "paste" pads have the wrong layer. They are currently mask pads.

The courtyard clearance is currently 0.295. Should be 0.25 (or because of rounding 0.245)
The rest of the landing pattern looks good.

evanshultz:
@poeschlr 
Sorry, I don't quite get it about the pads. I see copper and mask on the bottom, which is correct. I don't see any paste on the bottom. Can you help me understand? Quickly looking at the web interface for GH, here are the only pads with any bottom layers:
image

My courtyard was wrong. Fixed now. Sorry about that. Odd that Travis didn't catch it...

poeschlr:
Not bottom of pcb. "bottom" of screen. (The two paste pads that are not visible in your screenshot.)

evanshultz:
@poeschlr 
Ahh! Thank you. Fixed now.

poeschlr:
Thanks

PR #295, opened 2018-01-23
Added APV21_SSOP footprint for photodiode output optocoupler

skumfidus:
Datasheet: https://www.panasonic-electric-works.com/cps/rde/xbcr/pew_eu_en/technical_information_photomos_en.pdf
Screenshot:
screenshot from 2018-01-23 17-17-41

------------


poeschlr:
Please have a look at the [kicad library convention](http://kicad-pcb.org/libraries/klc/).

Especially the section about footprint naming.

skumfidus:
Would "Panasonic_SSOP_4_APV21.kicad_mod" be more correct?

poeschlr:
Well is this footprint that special to this one component?
Maybe a generic name like `SSOP-4_[body_size x]x[body_size y]mm_P[pitch]mm_Clearance[clearance between the pads in y direction]mm` would be better.

In that case the footprint should be moved to Package_SSOP.

skumfidus:
I guess that is an option, but it's not really an SSOP in the conventional sense, Panasonic just calls it that, and i do believe that this footprint is specific to two Panasonic parts.

Shackmeister:
@poeschlr would you go with a generic or specific name on this?

Shackmeister:
@poeschlr can you please have a look at the naming?

Shackmeister:
@evanshultz can you perhaps chime in with the naming for this?

evanshultz:
Hmm... this one is tricky and annoying.

It's in a general sense a generic footprint, but the nomenclature is odd. I'd therefore go with manufacturer-specific. That should at least make it clear this is intended for a non-generic part and also help a use avoid using the footprint where it turns out badly.

Panasonic lists a fair number of different specific MPNs for this footprint, so it's probably best to avoid that and use the series name. And I think adding dimensions to make it `Panasonic_APV-AQY-SSOP-4_4.45x2.65mm_P1.27mm` or `Panasonic_PhotoMOS-4_4.45x2.65mm_P1.27mm` would be best.

@Shackmeister 
I'll let you consider that. I'm not entirely sure the best method right now. Maybe you have better ideas?

evanshultz:
Another reason to make this vendor-specific is that the package doesn't appear to be a normal DIP. Right? So they 3D model will be unique, and that drives a unique footprint name.

Shackmeister:
@poeschlr @evanshultz @jkriege2 @Ratfink 
would anyone object against this name?
Panasonic_APV_AQY_SSOP4_4.45x2.65mmP1.27mm

Shackmeister:
APV and AQY are the parts using this footprint. We could use the generic name, but the part itself isnt a gullwing as you would expect with an SSOP. That's why I believe a manf specific name would make more sense

evanshultz:
I would think `APV-AQY`, with a hyphen, is best so there is no underscore within the "MPN" section. We typically put an underscore after the manufacturer name and then only again before body dimensions but not within that section. Right? That is what I wrote above unless I'm getting cross-eyed.

Shackmeister:
Panasonic_APV-AQY_SSOP4_4.45x2.65mmP1.27mm
or
SSOP-4_4.45x2.65mm_P1.27mm_Clearance3.45mm

I dont know which one makes best sense, but hopefully we can get it resolved soon

poeschlr:
Lets take the manufacturer specific one. Meaning `Panasonic_APV-AQY_SSOP-4_4.45x2.65mm_P1.27mm`

Shackmeister:
@poeschlr Thanks for the decision

@skumfidus Can you change the name accordingly so we can get this merged? :)

poeschlr:
Open points are finished in https://github.com/KiCad/kicad-footprints/pull/877

PR #277, opened 2018-01-20
Add Max30102 heart rate sensor

aurabindo:
Signed-off-by: Aurabindo J <mail@aurabindo.in>

Add MAX30102 Heart Rate Sensor. I'm not sure if Sensor lib is the right location for this.
Datasheet: https://datasheets.maximintegrated.com/en/ds/MAX30102.pdf
max30

- [x] Provide a URL to a datasheet for the footprint(s) you are contributing
- [x] An example screenshot image is very helpful 
- [ ] If there are matching symbol or 3D model PRs, provide link(s) as appropriate
- [ ] Check the output of the Travis automated check scripts - fix any errors as required

PR #300, opened 2018-01-24
Add MLF-8 footprint from Micrel

aurabindo:
Signed-off-by: Aurabindo J <mail@aurabindo.in>

Add Footprint used by MLF8 MIC2290 from Micrel
Datasheet: http://ww1.microchip.com/downloads/en/DeviceDoc/mic2290.pdf
mic2290

------------


poeschlr:
Footprints that include thermal vias need the suffix _ThermalVias.
For such footprints a copper only pad on bottom is required that covers the thermal vias.

The Pad suffix is not meant to give the dimension for the exposed pad. Use the EP suffix.
This means your footprint name should be: `Micrel_MLF-8-1EP_2x2mm_P0.5mm_EP0.8x1.3mm_ThermalVias`

Exposed pads are too large to have one single paste area. So it is required to split it into multiple smaller sections. (Remove the Paste layer from the large smd pad and add paste only pads without pin numbers to define the paste layer.)
Have a look at other similar footprints for inspiration. (Make sure you look at the current head of the master branch as we merged the new footprints just a few days ago.)
For the paste stuff [DFN-6-1EP_2x2mm_P0.65mm_EP1x1.6mm](https://github.com/KiCad/kicad-footprints/blob/master/Package_DFN_QFN.pretty/DFN-6-1EP_2x2mm_P0.65mm_EP1x1.6mm.kicad_mod) has a similarly sized exposed pad. For thermal vias handling, [MSOP-16-1EP_3x4mm_P0.5mm_EP1.65x2.85mm_ThermalVias.kicad_mod](https://github.com/KiCad/kicad-footprints/blob/master/Package_SSOP.pretty/MSOP-12-16-1EP_3x4mm_P0.5mm_EP1.65x2.85mm_ThermalVias.kicad_mod) is the most similar to yours. (But it has a much larger exposed pad -> more vias.)

aurabindo:
From the footprints you listed for reference, I noticed that the large SMD EP pads have a big paste layer as well. Nevertheless, I have removed the paste layer as you suggested.

 mic2290_3d
mic2990_3

mic2290

Are the KLC errors excepted for Thermal Vias or should I fix them as well ?

poeschlr:
What footprint i linked has large pad sizes? Are you sure you checked to current master?

aurabindo:
> What footprint i linked has large pad sizes? 

I assumed (and meant) the relatively larger pad in the symbol, which is the exposed pad. Such pads have both F.Paste and F.Mask layer.

This is just a 2x2mm footprint. So shall I give the EP both F.Mask and F.Paste ?

poeschlr:
Might it be that you did not check the latest version? The latest version of the DFN footprint i linked looks like this: (green -> paste, red -> copper) It has two separate paste pads on top of one larger copper + mask pad.


Your paste pads should not have copper enabled. (only paste)
screenshot from 2018-01-26 13-04-29


The copper landing pattern looks good.
But the courtyard is not correct. there should be 0.25mm clearance all around. (measured from either pads or body depending on what gives the larger area. In your case: in x direction it is relative to the pads and in y direction relative to the body)

screenshot from 2018-01-26 13-11-15

aurabindo:
You're right, thanks for explaining in detail. I had pulled the master, but the branch I was using, was based before this change, so I've been seeing the old one.

I've autogenerated the courtyard.

aurabindo:
@poeschlr  Does this need any more fixes ? I was hoping the pad related violation is excepted for thermal vias. Is this not the case ?

poeschlr:
Travis is not always right. (We have ideas how to make it better but there is just not enough manpower to do everything.)

I seem to have missed the fact that your silkscreen looks very different to all other footprints in this repo. Could you maybe make it more like this:
screenshot from 2018-01-28 18-42-50

aurabindo:
all
@poeschlr Done. Please take a look.

There does not seem to have a standardization for the footprint silkscreen. In #277, closed box was preferred.

poeschlr:
Maybe @jkriege2 can comment why he suggested the enclosing rectangle style in that pull request.

My guess is that it is because your other pull request is for a different lib. We normally try to have a similar style for all footprints (or symbols) within one lib. This does not necessary mean that we have one single style across all libs. (We try to get to that point, but well ...)

poeschlr:
Thanks for contributing. And sorry that you got different suggestions for different pull requests.

jkriege2:
Yes, I think I scanned through the lib the footprint was originally posted to (I think LGA or BGA) and suggested a layout matching that ... maybe we should really try and get an overview of standard package-types+standard-layouts into the KLC as a suggestion?

PR #303, opened 2018-01-25
HLGA-6 footprint added

Resaj:
Added footprint HLGA-6 for sensor audio.
Datasheet: http://www.st.com/content/ccc/resource/technical/document/datasheet/group3/27/62/48/98/44/54/4d/36/DM00303211/files/DM00303211.pdf/jcr:content/translations/en.DM00303211.pdf

PR packages3D: https://github.com/KiCad/kicad-packages3D/pull/231
PR packages3D-source: https://github.com/KiCad/kicad-packages3D-source/pull/88

imagen

imagen


Resaj:
I've tried to modify the anchor but I think it's perfectly centered on the footprint.
Travis calculates the center as (0.12855, -0.12855)mm. Can it be an error?

poeschlr:
Travis does use the bounding box of the pads to calculate the center. In your case pin 6 does increase it in y direction.
I would set the footprint "anchor" such that pins 2 and 4 are at y=0. (Travis is not always right.)

poeschlr:
The fab outline should represent the exact body size. From your 3d model screenshot it seems the silk outline is larger then the body. From that i assume the fab ouline must be moved inwards.
It should also be a closed outline with a chamfer as the pin 1 marker. (Have a look at other similar footprints as inspiration)

The footprint name should include body size and pin pitch. As this is a special footprint also include the manufacturer as a prefix. @jkriege2, @evanshultz the name also include the part number for such specialized devises? (As there could be a similar audio device that does use a slightly different similarly named footprint by the same manufacturer.)

Are there no special requirements for the mask and paste layers?

Resaj:
I supose you refer to set footprint anchor with y=0 for pins 2 a 5. Is it right? That's how it is now.

I drew silk and fab with 0.2mm clearance because [KLC specification](http://kicad-pcb.org/libraries/klc/F5.1/). I've just adjusted again, as you say.

I've added a chanfler on fab layer. Not in silk layer because it would be on top of the pad 1. I could draw a different type of marker on silk layer, like a point or displaced corner.

Footprint name changed to include manufacturer, body size and pin pitch.
There are audio sensors from other manufacturers ([Knowles](https://www.mouser.com/ds/2/218/SPM1423HM4H-B-876897.pdf) and [DB Unlimited](https://www.mouser.com/ds/2/683/MM042602-4-611169.pdf)) that uses the same package but with different name. Could I keep it in mind?

I didn't see any special requirements for the mask and paste layers in the [datasheet](http://www.st.com/content/ccc/resource/technical/document/datasheet/group3/27/62/48/98/44/54/4d/36/DM00303211/files/DM00303211.pdf/jcr:content/translations/en.DM00303211.pdf).

imagen

imagen

poeschlr:
KLC specifies clearance for silk. (clearance to the pads is defined as 0.2mm. There is no exact definition with respect to the body. Only that the full silk with must be outside the body dimensions.)
The fab outline represents the exact nominal body dimensions.
Fab outline is a different manner. Have a look at any other footprint. You will find that silk is always outside of the fab outline.

Resaj:
Thanks for the clarification. I've just modified the layers. Is it right now?

imagen

poeschlr:
The screenshot looks better now.

poeschlr:
Please set the 3d model scale to 1,1,1 and the offset and rotation to 0,0,0
(Use for example kicad stepup to export correctly scaled and positioned 3d models.)

The courtyard clearance should be 0.25 all around. (currently you have 0.29 in y and 0.27 in x direction)
The pad diameter should be 0.91 not 0.9
The oval pad is not quite right.

lga

As a help for determining the dimensions for the oval pad have a look at this sketch:
screenshot from 2018-01-28 14-44-06

poeschlr:
Man i seem to miss everything lately.

The 3d model reference has the wrong name. The 3d model should have the same name as the footprint.

Also the oval pad is now even more wrong than before. (now the second center is 0.1mm in both directions instead of 0.2)
screenshot from 2018-01-28 19-57-04

Resaj:
No problem. I've changed the 3D part name and I'm going to change it in the PRs of the other repos.
Sorry, the grid confused me and I thinked the measures were all OK.

poeschlr:
Thanks for your contribution.

PR #304, opened 2018-01-26
Added Dinkle DT55 Terminal block

crackerizer:
Added Dinkle DT55 Terminal block with 4 terminal footprint, https://www.dinkle.com/en/terminal/DT-55-B01W-XX
------------


poeschlr:
I don't think it is a good idea to make a separate lib for just one footprint. Might be better to add it to the TherminalBlock lib.

The silk outline should be outside the body, whereas the fab outline represents the exact body outline. (means the silk outline will result in being outside the fab outline)
Add pin 1 markers on fab and silk. (The silk pin 1 marker must be visible after assembly)
The fab artwork is a bit complex. Maybe make it a bit simpler. (Remove a few of the horizontal lines)
For connectors the courtyard clearance should be 0.5.
The pads are quite large. Maybe reduce their size by a bit. (Especially in y direction.)

crackerizer:
I will added all pin count after this one got merged to make sure I get it right.
I will fix the footprint to all requirements. Do I have to add pin 1 marker to this footprint? Because it is symmetric. I added rectangle pin 1 because it does not pass the Travis CI build.

jkriege2:
Hi!

a note on the side: All the other terminal blocks were generated by scripts that you find e.g. here:
https://github.com/pointhi/kicad-footprint-generator/tree/master/scripts/TerminalBlock_RND

Do you use a comparable approach, or did you hand-draw them? If you scripted them it would be great to have the code posted in the repo I gave above (also that will make adding variants also from other manufacturers, way easier).

Best,
JAN

jkriege2:
About pin1-marker: We have them on all (or at least the great majority) of terminal blocks. Just orient your design on the already existing footprints, e.g. 
2018-01-27 11_01_30-footprint editor no active library

crackerizer:
I did handwrite them. The move exact function makes things much easier to create these footprints by hand.

crackerizer:
Fixed

poeschlr:
The footprint looks good now. Thanks for your contribution. (If you want to add the other footprints you can open a new pull request for them.)

crackerizer:
Thanks, I will start adding the rest of this footprint.

PR #306, opened 2018-01-26
Add TO-277B

strombom:
TO-277B

Drawn according to page 3 in Littlefuse datasheet: https://media.digikey.com/pdf/Data%20Sheets/Littelfuse%20PDFs/DST2050S.pdf

image

image

image

3D Model pull request: https://github.com/KiCad/kicad-packages3D/pull/238

------------


poeschlr:
For paste pads, only the paste layer should be enabled. (Set copper to none, deselect mask and most importantly remove the pad number.)

We normally do not include the pins in the fab outline.
As polarity marker on fab a chamfer is more common.
The fab outline should represent the nominal body size not the maximum.

strombom:
Thank you for the comments, updated.

image

strombom:
Changed the paste fill pattern to 60%

image

strombom:
Question: Should only QFN etc have a paste fill of 50-70% or does this apply also to packages such as this diode TO-277B?

poeschlr:
Currently your footprint has the wrong courtyard clearance in x direction. (0.3 on the left and 0.31 on the right side. Should both be 0.25 or 0.245 if the grid requirement requires rounding)

---

I think the 50-70% is valid for all parts except if the manufacturer gives a different suggestion.

The reason for doing it that way is to reduce to allow for easier outgasing and for reducing the danger of the component floating on the solder paste of the large pad which can result in missing or weaker connections for smaller pads. Both of these things apply for all packages where one pad is significantly larger than other pads. (The outgasing issue even applies to all larger pads.)

https://www.allaboutcircuits.com/technical-articles/stencil-design-for-exposed-pad-packages/

strombom:
I discovered that the body in the 3D model had wrong lenght, updated here: https://github.com/KiCad/kicad-packages3D/pull/238

Courtyard, fab outline and silkscreen all updated.

image

image

image

poeschlr:
You used the maximum dimensions here instead of nominal. (for the body outline)

strombom:
I'm not sure I can find where the body outline (F.Fab?) is not nominal dimensions. It's 3.98 x 5.38 mm as the datasheet states?

image

image

poeschlr:
you are right. (got the reading from the wrong column in the datasheet.)

Thanks for your contribution.

PR #311, opened 2018-01-28
Add hexdip

jkriege2:
- add HVMDIP/HEXDIP package for MOSFETs from Vishay, see https://www.vishay.com/docs/91361/hexdip.pdf
- basically this is a DIP-4 with pins1+2 connected
- source: https://github.com/KiCad/kicad-packages3D-source/pull/91
- exports: https://github.com/KiCad/kicad-packages3D/pull/243

2018-01-28 20_54_07-footprint editor no active library
2018-01-28 20_53_58-3d viewer

------------


poeschlr:
The courtyard clearance for this is not quite right.
Also the body with does not agree to the one you used for the fab drawing.
dip

It might be a good idea to increase the drill size slightly. (If the leads are at their maximum measurements, there is not much tolerance left for the drill.)
screenshot from 2018-01-28 21-40-43

jkriege2:
Hmmm ... interesting. I just took teh existing DIP-4 footprint and added a line on F.SilkS and fused the pads 1+2. This means we should rework the whole lib of DIPs, as they should have the same issues.

I will do this symbol here and think about modifying the scripts for the DIPs.

jkriege2:
Hi!

- OK, the holes are now 0.9mm, which leaves 0.15mm above the required diameter  estimated from the max. measures for the pins in the DS: sqrt(0.6*0.6+0.45*0.45)=0.75
- reworked courtyard (auto-generated by our scripts now)
- reworked body size on F.Fab to 6.29x5.0mm as in datasheet

OK to merge now?

poeschlr:
Something went wrong with your courtyard fix. There is now a 0.45mm clearance on the bottom side.

jkriege2:
fixed now ... seems like our scripts have problems with pads where there is an offset of the drill

poeschlr:
Thanks

PR #313, opened 2018-01-29
Add SOT-1210_LFPAK33

strombom:
SOT-1210, LFPAK33

https://assets.nexperia.com/documents/outline-drawing/SOT1210.pdf

image

image

image

3D Model pull request: https://github.com/KiCad/kicad-packages3D/pull/244

------------


poeschlr:
The travis error comes from a different pull request. There it was forgotten to add the sensor_audio lib to the fp-lib-table. You can ignore this error here.

evanshultz:
@poeschlr Can you answer a question about this footprint? "SOT-1210" is a package name only Nexperia uses. See the datasheet linked above:
image

But it's not a manufacturer-specific footprint. Should this footprint instead use the industry name of `LFPAK33`? Unfortunately, I don't see a standards body like JEDEC giving a name for this package. I don't find other vendors using this package right now, though, only Nexperia.

Similarly, in our repo we have a footprint for [SOT1333-1](https://github.com/KiCad/kicad-footprints/blob/master/Package_TO_SOT_SMD.pretty/SOT-1333-1.kicad_mod) (https://www.nxp.com/docs/en/package-information/SOT1333-1.pdf). That package does have an industry name (`XSON9`) and a JEDEC name (`MO-252`) but neither of those terms are used in our footprint name. Should that footprint be renamed to use a name that is not manufacturer-specific? That goes for [SOT1334-1](https://github.com/KiCad/kicad-footprints/blob/master/Package_TO_SOT_SMD.pretty/SOT-1334-1.kicad_mod) too.

poeschlr:
If a footprint is for a package that has a JEDEC name than we should probably use that name. (I think the original naming convention had this specified but that got dropped to make it more easy to read.)

If there is no standard name than we can stick to manufacturer names. (This might mean we would need multiple footprints for different manufacturer naming conventions unless one name is so common that it is a quasi standard.)
Maybe include alternative names in the description field. (Alias names for footprints would be nice here.)

evanshultz:
@strombom 
OK then, we agree.

1. Call this package `LFPAK33` and put both `LFPAK33` and `SOT-1210` in the description and keywords.
1. Your F.Fab outline is 3mm tall (vertically) and should be 2.6mm.
1. Shrink the F.Fab ref des text slightly to fit inside the body.
1. It is customary to put the silk 0.12mm outside the F.Fab lines; yours is bigger.
1. Your copper is centered horizontally on the F.Fab outline. How did you decide on the placement? For example, see that L and L1 are different. I don't see this relative location clearly specified in the doc.
1. Since the mask (solder resist) dimension is specified, you should include this in your footprint. You can use the global setting for mask oversize since the doc says 0.05mm everywhere.
1. Pads 1-4 and the protrusions from pad 5 (the parts sticking out on the right side) do not have stencil openings defined in your footprint, but the doc shows them.
1. This is optional, but the pin 1 mark is quite large. You could use a smaller notch.

Item 5 may require some investigation. Or maybe I've misread the datasheet. Otherwise, thanks for your contribution so far and I look forward to merging this footprint.

strombom:
1. OK
2. Current model is vert: 3,3mm. hor: 2.6mm. I believe it's correct.
3. OK
4. OK
5. The pins are not symmetric. In order to center the land pattern according to KLC the body center of the component was moved 0,2 mm to the right. This may be the cause of confusion in point 2 above.
6. OK
7. OK. Changed from KLC recommended value of 1mm to 0,5mm.

image

image

image

evanshultz:
Thanks for making updates.
1. Please add the other package names to the description and keywords.
2. My mistake. It's correct.
4. Can you please snuggle the silk closer to the fab lines as mentioned above? Also, and sorry that I just noticed this now, but silk should go ~0.3mm from copper and yours is farther away.
5. I'm talking about the relative location of the copper/mask/paste to the body. On page 3, there is no dimension linking the copper and body outline vertically. Your copper is centered to meet KLC, which is good, but how did you know to offset the body 0.2mm? Is that dimension present and I'm not seeing it?

strombom:
1. OK
2. OK
3. OK
4. I calculated center from H/2 on the body drawing and assumed this was centered with 3.9/2 on the land pattern drawing if that makes sense?

image

evanshultz:
3. Most of the silk is now 0.1mm from the body, except the vertical line by pin 1 that is 0.15mm. Can you use 0.12mm everywhere? This will avoid silk under the part, match the existing footprints, and meet KLC.
4. That's exactly my question! I do not see any dimension that tells me the copper is centered on the body. Perhaps it's offset? The drawing does not show this. Can you ask Nexperia?

Everything else looks good. :)

strombom:
3. Let's see if I got it right this time.
4. I have sent a question, will post the response.

image

evanshultz:
3. Yep. Looks good.
4. Let us know when you get an answer. Thanks!

I just noticed that the X position of the silk ref des, the fab ref des, and the package name aren't 0. This doesn't stop me from merging, but there's no reason not to make them 0.

strombom:
Ok centered the texts

evanshultz:
Thanks! Any answer from NXP?

evanshultz:
I talked to Nexperia and the body and copper should be centered on each other. So I believe you need to change the X locations of the left and right sides of the body (F.Fab lines) to +/-1.3mm, since the copper is already centered horizontally on your footprint.

evanshultz:
@strombom 
What I am telling you is that Nexperia says the drawings on page 2 and 3 should be centered both vertically and horizontally. If we deviate from the manufacturer's recommendations, we must have a good reason (and probably also provide a canonical footprint too).

I see no reason that our footprint should deviate, so please center the body and I can merge.

In addition, look at the alignment of the 3D model to the copper above. One row of pins is at the edge of the pads and another is well into the middle of the pads. It looks to me like the body _should_ be moved!

strombom:
image

I don't know if we agree or disagree, because I agree fully with this statement: "the drawings on page 2 and 3 should be centered both vertically and horizontally". I would say that conforms to option A below, which is the version that is currently checked in.

Option A:
Measurement H (the entire part including pins) centered over footprint

image

Option B:
Measurement D (plastic body) centered over footprint

image

Issue #117, opened 2017-11-27
Nets on wing pads of connectors


Shackmeister:
Not sure if this belong here or on the forum, let me know if its inappropriate here.

How are people handling connecting the wing pads on connectors to different nets (typically GND)?
usually I just edit the pads and add the net I want, but this is lost when updating the netlist, which usually makes me forget to change some of the 20ish pads. Unless somebody has a good way, we should maybe consider making a standard padnumber for these. This way the user could use a symbol with or without the wings depending on the nescessity.
Thoughts?

poeschlr:
I am currently not understanding what you mean with wing pads. Do you mean the mounting pads like the two larger pads of the picoblade connectors? 
These do not need to be connected to a net because they are just for mounting purposes.

Connectors that have shielding connected to such pins need a pad number of course.

Shackmeister:
Yes the larger pads, I typically connect them to GND to have a stronger mount. Sometimes I connect them to other nets to void splitting planes, easier routing etc.

poeschlr:
Adding these pin(s) would at least double the amount of generic symbols needed. (And it will add a further complication to users. They now need to answer the question: does my connector come with mounting pads? Even if they don't plan on connecting them.)

poeschlr:
@SchrodingersGat @jkriege2 @evanshultz care to comment?

evanshultz:
**Oops! I forgot to submit this comment earlier! Sorry.**

In the existing (old) KiCad libs I recall seeing they were universally pin 0 (zero).

There are several situations where a net connection to the mounting tabs is important:
- Mounting strength (mentioned above)
- Shielding (mentioned above, though this may require a specific symbol anyway)
- Chassis ground connection to circuits
- HV / ESD protection
- Connection to touchable surfaces for safety during plug/unplug events

Adding symbols with a pin 0 for mounting pads is good, but I agree that the overhead in the quantity of symbols and footprints that would be added is substantial.

Another option might be to have only one set of symbols, all with pin 0, and the pad pin can be left floating if desired. A corresponding footprint would need a numbered pad 0 which would be floating or connected depending on the symbol's connection. So we eliminate any duplication of symbols or footprints and it's up to the user to make the connection or not.

jkriege2:
At least for D-Sub connectors there are symbols with a specific PAD-pin (optional, i.e. variant). the same is done for crystals!

jkriege2:
[...]

PR #316, opened 2018-01-29
Add Bourns SRR1260

strombom:
Bourns SRR1260 series

Datasheet: http://www.bourns.com/docs/Product-Datasheets/SRR1260.pdf

image

image

Footprint pull request: https://github.com/KiCad/kicad-packages3D/pull/246

------------


jkriege2:
Please add the datasheet link to the description. Rest is fine and I will merge when that's done!

Thanks for ll your work!
JAN

jkriege2:
Is there a reason for the rounded pads?

strombom:
Mostly that I think it looks nicer, and a similar inductor I made has rounded land pattern:
http://www.bourns.com/docs/product-datasheets/srp2313aa.pdf

Also a large PCBA company I worked with always insisted on rounded pads but I'm not sure of the reason.

Should I go back to square pads since the datasheet has square pads?

strombom:
Maybe not related to such large components as this, here is a recommendation on page 9:

http://www.ocipcdc.org/archive/What_is_New_in_IPC-7351C_03_11_2015.pdf

They also state that maximum radius is 0.25mm so my rounded corners are wrong either way.

poeschlr:
I would like to keep rounded rectangles to a minimum until we have made our kicad version 4 backport of the new lib. (Will happen soonish. Depends a bit on the decision we will reach in https://github.com/KiCad/kicad-footprints/issues/117 as i would like to include that change if we decide for it.)
As IPC does suggest the use of rounded rectangles we might even consider to switch over to them in the future. (I seem to remember that the main argument for rounded rectangles is that the paste layer stencils need a radius. And there is an argument that solder flow is better with rounded rectangle pads.)

High voltage devices are a different story. There rounded rectangles are required to ensure their voltage rating. So if your component is a high voltage device than go ahead.

strombom:
No, not using it for high voltage, thank you.

evanshultz:
Corners also radiate more than curves. This can be important with high edge rates (voltage or current) when meeting EMC.

jkriege2:
For now I would stick to the datasheet recommendation (I think that's best we can do until we have a proper rule). So I would remove the rounded corners from the pads here.

strombom:
The rounded corners were removed in the latest commit.

jkriege2:
Thanks,
JAN

PR #318, opened 2018-01-29
Add LFCSP-24 footprint based on Analog Devices dimensions

jcyrax:
This is for symbols PR https://github.com/KiCad/kicad-symbols/pull/244
I prefixed with Analog, but maybe it can be a general part?

Part datasheet
[http://www.analog.com/media/en/technical-documentation/data-sheets/ADL5801.pdf](http://www.analog.com/media/en/technical-documentation/data-sheets/ADL5801.pdf)
Footprint datasheet
[http://www.analog.com/media/en/package-pcb-resources/package/56702234806764cp_24_3.pdf](http://www.analog.com/media/en/package-pcb-resources/package/56702234806764cp_24_3.pdf)

Full name: Analog_LFCSP-24-1EP_4x4mm_P0.5mm_EP2.5x2.5mm
footprint_fixed

3d_view

*Edit: Footprint screenshot after fix commit.
**Edit broken links

------------


jcyrax:
I have warning:
Violating F9.2
    Footprint properties should be left to default values.
    Attribute clearance == outline != 0!

But I have everything set to default. Maybe because of pad shape.

jcyrax:
I had wrong pin number for exposed pad. Was 26 changed to 25 as needed. Also edited first post to have corrected screenshot.

evanshultz:
@jcyrax 
Did you review the footprint information ADI releases? The BXL file can be found at http://www.analog.com/en/design-center/packaging-quality-symbols-footprints/symbols-and-footprints/ADL5801.html. For many of their footprints, this is the best and most accurate (and sometimes only!) way to get complete footprint information.

jcyrax:
I didn't because it was not text format and that software is windows so I went with datasheets and document from Analog (http://www.analog.com/media/en/technical-documentation/application-notes/AN-772.pdf). I based on existing footprint and changed.
I will try to run it and check information from BXL file.

jcyrax:
@evanshultz ok, I downloaded Ultra Librarian Reader (2 GB disk space!) and it is very bad. You can't measure anything in there. It can export to KiCAD so I did.
So I have footprint from BXL generated presumably by Analog Devices. However if you compare this footprint by rules/notes in http://www.analog.com/media/en/technical-documentation/application-notes/AN-772.pdf it violates a lot of them.
bxl_footprint
For instance Fig.5 has pads with rounded edge. Pad X width for packages with 0.5mm pitch is specified 0.28mm, they have 0.3048mm. This is just some things, I did quick follow trough the formulas and I think everything is different.
Also I have to mention that my original footprint is wrong as well.

What do you think is correct steps here? For me I can tell I won't be using BXL footprint for my designs.

evanshultz:
@jcyrax 
Unfortunately, ADI only supplies footprints in this BXL format. I dislike being forced to use Ultra Librarian to view the default footprint as well, but we are both suffering.

The PDF you linked originally is the package drawing, and not the footprint. So the "footprint" you submitted isn't actually the footprint anyway. :)

In the past ADI told me that BXL is the way to go and I should use that to create footprints. But this one is a huge mess. Let me see if I can get some resolution. Hang on, and sorry for the delay. Unfortunately this footprint isn't straightforward...

jcyrax:
> The PDF you linked originally is the package drawing, and not the footprint. So the "footprint" you submitted isn't actually the footprint anyway. :)
I know that. It was just for reference. My original footprint isn't just copy of device metalization. I always add some zone for heel and toe etc.

evanshultz:
For comparison...

BXL:
image

[ADL5801 eval board](http://www.analog.com/media/en/technical-documentation/evaluation-documentation/ADL5801_a02307b_Gerber.zip) (green = Cu, cyan = silk, orange = paste):
image

I believe the BXL is the most "correct" and it's also unique to this particular package. Can you please align your footprint with the BXL?

jcyrax:
Why are you in favor of BXL? Because its different and that would suggests its made different on purpose? Don't you think that BXL footprint can be just an error?

evanshultz:
BXL is specific to this footprint, while AN-772 is a general (but good!) guideline for this generic package style. Also, the other LFCSP footprint appear to follow the style ADI provides in BXL format. Lastly, ADI said to use the BXL design.

jcyrax:
But why did you post eval board? Is eval board BXL? It seems different looking at gerbers?

evanshultz:
Simply because it was another place to look at, but it didn't corroborate with any other source of information from ADI on this part's footprint. :(

jcyrax:
image
I changed pads as per BXL.
I also changed solder paste as per BXL.
image

Question: should this footprint stay with Analog_ prefix as is now? I saw few footprints from Analog BXL in library that didn't have any prefix. If so I need to change it for my symbol.

evanshultz:
@jcyrax 
Thanks for updating this footprint. Review notes:
1. Please fill out the description as the other LFCSP footprints.
1. The EP size in the footprint name needs to be corrected.
1. The footprint name can mention "VQ", as with one other footprint in our lib.
1. These VQ footprints have a bevel on each corner of the package drawing.
1. F.Fab lines are 0.1mm thick.
1. F.SilkS lines are 0.12mm thick.
1. Remove the zone connection property from pads 1-24.
1. I will merge without this, but I prefer to flip the X and Y size of each pad rather than rotate them.
1. Remove paste from pad 25 and all smaller paste squares. This is covered in other PRs.

@poeschlr @jkriege2 
Please check this PR. There are obvious KLC violations, like line thickness, but Travis says all checks have passed. What's going on?

evanshultz:
@jcyrax 
I started reviewing before your last comment.

Yes, please remove the "Analog_" prefix. This is part of making the footprint just slightly generic, and in the case of the paste layer not sticking perfectly to the BXL info.

jcyrax:
@evanshultz  I think I fixed all of them.
Only question about 2. So footprint name EP field is actual footprint pad size and not component package EP size? I got impression its similar to body dimensions field, as size for component.

evanshultz:
LFCSP-32-1EP_5x5mm_P0.5mm_EP3.25x3.25mm refers to http://www.analog.com/media/en/package-pcb-resources/package/pkg_pdf/lfcspcp/cp-32/CP_32_27.pdf. The EP size is of the pad, not the package. LFCSP-72-1EP_10x10mm_P0.5mm_EP6.15x6.15mm refers to https://www.intersil.com/content/dam/Intersil/documents/l72_/l72.10x10c.pdf and also gives the pad size for the EP dimensions.

@poeschlr 
In addition to the question above, can you also help answer the EP size question? The PR at https://github.com/KiCad/kicad-footprints/pull/285 doesn't specify how EP size was determined, and http://kicad-pcb.org/libraries/klc/F2.2/ seems to indicate to me that the size of the pad on the component should be used. And that also sounds correct to me, because that will be the 3D model size which should use the component's thermal pad. However, it contradicts the footprints above which use the copper pad size in the footprint name rather than the component's thermal pad size.

evanshultz:
@poeschlr @jkriege2 
I think something is wrong with our check scripts. See https://github.com/jcyrax/kicad-footprints/blob/fc6cbfb6e9fc0ce6f4a7d88a2fbf6913c8683ee0/Package_CSP.pretty/Analog_LFCSP-24-1EP_4x4mm_P0.5mm_EP2.5x2.5mm.kicad_mod. There are obviously 0.15mm F.Fab lines and other problems. But when I run the newest check scripts from https://github.com/KiCad/kicad-library-utils, I don't see those errors. Am I doing something wrong or are the scripts broken?

poeschlr:
@evanshultz 
The pcb pad size is used in the footprint name because the fp name describes the footprint not the package it is made for.
The pcb pad in the model name is used because that was the information freely available (from the footprint renaming). It also fits the KLC rule that states the footprint and 3d model name should be equal.

---

More details:
For most footprints this is a bit larger then the pad on the package.

The usage in the 3d model names could be misleading if you take the name as describing the model it self.
(In our case we could say the name of the model represents for which footprint it is intended.)
It was already too much work to fix all footprint names. In reality i just couldn't care less that the 3d models now have a slightly wrong name. 

It took me a week to fixup all footprints with exposed pads. Replacing it with the package pad size in the 3d model name would require checking every datasheet. (or where existing, the 3d models) What makes this task even harder is that a lot of the datasheets links listed in the footprint description are not valid any more or they might even be missing. 
I would guess this would be at least one week of work maybe even a month. (It just isn't worth the effort.)

evanshultz:
@poeschlr 
Got it. Sounds like an issue, but not a high priority right now.

Also, are you able to look into the check scripts mentioned above? Running them locally on a version with 0.15mm line widths on F.Fab and F.SilkS doesn't trigger an error either. I'll try to check into it today.

@jcyrax 
Looks good. Thank you! And also thanks for a good PR by including links and screenshots above.

poeschlr:
I have very little experience with our KLC test scripts. I am also a bit short on time right now. I would suggest you open an issue about that if you don't have time to look into it.

PR #324, opened 2018-01-30
added Texas_R-PVQFN-N20 footprint

Shackmeister:
Datasheet: http://www.ti.com/lit/ds/symlink/tps61088.pdf
Screenshot:
texas_r-pvqfn-n20
texas_3d

------------


poeschlr:
Courtyard clearance on left and right side is 0.24mm instead of 0.25.

As your pads are rounded on both sides i would suggest to ensure that for the outer side you make sure that the suggested pad is completely inside the oval pad. (as if you add on the rounded side to the pads suggested in the datasheet.)

Shackmeister:
@Rene_Poschl can you take a look at this?

Shackmeister:
@evanshultz can you take this one as well? :)

evanshultz:
@Shackmeister 
Could you help me by taking a look at https://github.com/KiCad/kicad-footprints/pull/384 and seeing if any similar nonsense is happening here?

Using v5 and polygon pads would make pad 21 a lot easier...

- Fix the description, including the URL which is for LT.
- You used oval pads while the datasheet shows half oval. For this vendor-specific footprint I'd like to see it exactly match the recommendation.
- There are no vias; can you create a version with them?

For dimensions checks I'll need more time. Sorry.

evanshultz:
@Shackmeister 
I'm so sorry! I totally forgot about this until last night when I woke up at 3am. So weird!
- You didn't make any commits since my comment above so those items are mostly still valid. Except...
- While http://www.ti.com/lit/an/slua271a/slua271a.pdf shows only half-oval pads for QFN packages, this datasheet shows rounded rectangles; you can do it in the recent nightlies and it's OK for this repo.
- Where did you get the name `R-PVQFN-N20`? I see `RHL0020A` in the datasheet but not this name. And `N20` doesn't appear in the datasheet either.
- For a manufacturer-specific footprint, do we need the EP size? Did you see an otherwise identical footprint but with a different EP size.
- Furthermore, on our other footprints the EP size is the copper pad size and not the exposed pad on the package like you've named it here.
- Your ref des on silk is rather far from the courtyard.
- Your silk is under the part body. 0.11mm offset from the fab lines will work.
- Your pads are 0.85x0.28mm; the datasheet is 0.6x0.24mm. I think this is due to not matching the pad shape in the datasheet, but I again question why you don't just match it when KiCad can do it.
- Your footprint is missing the vias.
- Here are dimensional checks showing mismatches:
image
- Note above that the courtyard clearance all looks OK to me. 0.25mm everywhere. I can't verify what Rene mentioned above about only 0.24mm on the left and right sides.
-Your paste is wildly different than the datasheet (page 32):
image

evanshultz:
@Shackmeister 
Are you able to make the fixes for this footprint? The number of PRs here and for symbols is getting out of hand! :)

Shackmeister:
@evanshultz I totally agree :)
ATM Im only using stable, since I use it professional :)

Ill look into it today or tomorrow :)

Shackmeister:
@evanshultz I think I will close this for now.
Im having some trouble figuring out where I got the name (pretty damn sure it's not random, but I do prefer stuff to be well documented)
also I dont have V5 for the pad shapes.

Ill be back with this later :)

evanshultz:
Alright. But it can be done just fine in v4, no need to wait for the new pad shapes if you want to get this merged now.

PR #327, opened 2018-01-30
Remove pin numbers from paste pads

strombom:
TO-252-2 had pin numbers and copper its four paste pads.

image

------------


evanshultz:
Thanks! There are surely many older footprints that also need this fix as well.

strombom:
Ok I can do a few more

PR #329, opened 2018-01-30
Remove pin number and copper from paste pins

strombom:
Remove pin number and copper from paste pins. Removed paste layer from large copper pads that have paste pins.

TO-252-2_Rectifier was missing paste pins.

SOT-669 needs double checking.

------------


strombom:
Not sure about the script errors on Infineon_PG-HSOF-8-1_ThermalVias, does it recognise the circular SMD pads as THT pads?

poeschlr:
The thermal vias footprint has tht pads. These "vias" should keep their copper layers and their pin number. (You removed the pin numbers from them. This will result in DRC problems if the footprint is used.)

So please put the pin numbers back for these pads.

poeschlr:
Thanks

PR #343, opened 2018-02-03
PLCC NeoPixel tidyups

fauxpark:
Renamed and fixed up the clearances & KLC issues for the other NeoPixel footprints. The WS2812 thankfully does not need the "1" marker as with the other two.

ws2812
ws2812b
3d

------------


strombom:
Just a question, why is the notch in the silkscreen at pin 3 and not at pin 1?

image

fauxpark:
@strombom: It's a whole thing, see #245 ;)

strombom:
Oh interesting, I wonder how the manufacturer was thinking :)

Is there a reason that your pads on WS2812 are 1.6 mm wide while the datasheet has 1.5 mm? Also I think they should be placed at x=2.45 mm?

fauxpark:
Hm, I guess I was too trusting that the original author of these footprints only got the silkscreen and courtyard lines wrong.

strombom:
The courtyard should be 0.25 mm from the pads in x direction, so that would be 3.45 mm instead om 3.55 which it is now.

Also I think the silkscreen can be 0.25 mm from the pads in x direction and 0.12 mm from the Fab.

Something like this:
image

jkriege2:
Hi!

I'm not sure about the footprint for SK6812MINI ... from images on the internet (and the datasheet linked in the footprint), the device has rather large solder-pads on the bottom of the package: http://www.opscoled.com/showshop.asp?/76.html
I think the copper pads on the PCB should cover that area ... unfortunately I didn't find any reference to a proposed land-pattern. Didi you already use these LEDs in a project? 

the other two footprints look fine to me and I will merge as soon as we find a solution for the MINI.

Best,
JAN

fauxpark:
Yeah, the pads are kinda funky looking. [Adafruit's EAGLE library](https://github.com/adafruit/Adafruit-Eagle-Library) contains this part, but I don't have EAGLE to see what it looks like. It's just a big XML file though, and from what I can see, the [SMD pads on the "LED3535" package](https://github.com/adafruit/Adafruit-Eagle-Library/commit/4abf84c3856535c0ab6ae96ec4be784d7dcc6b3e#diff-5038edab6df44c3629c71165144cdfb4R33819) have the same X/Y positions and Y size at least. But the X size of 0.85 is throwing me off, apparently it looks like this:

fp085

I did have some boards made with this footprint, but of course the symbol pinout was wrong, and they really don't like heat. It will be at least a week until my new boards even get fabbed due to Chinese New Year.

fauxpark:
Someone took a screenshot of it: https://forums.adafruit.com/viewtopic.php?f=47&t=108295
And according to [this thread](https://forums.adafruit.com/viewtopic.php?f=47&t=90836) it appears the pads should have a Size X of 1mm, and a Size Y of 0.85mm.

fp
3d

jkriege2:
Hmmm ... I found an old Eagle try-out version and scanned some Eagle libs on the net ... fallout is: 
2018-02-16 07_11_05-control panel - eagle 6 5 0 light
2018-02-16 07_00_34-control panel - eagle 6 5 0 light
2018-02-16 07_00_08-control panel - eagle 6 5 0 light

So I think people don't really care for the pads on bottom. But since photos of the device show them ... could you maybe come up with pads that extend far enough inside to cover those areas? If that's over the top for some user, they might complain, but at least it's safe and prevents people from e.g. having un-varnished vias there, which could be connected by accident to that pad ...

fauxpark:
Perhaps go back to 0.85mm height but keep the 1.6mm width?

jkriege2:
OK, I snooped around a bit more ... and all footpritns that I could find look like the oe you proposed ... smaller or larger pads are present. I would therefore propose to use the 1.6x0.85mm² pads. In the end we then stick to the footprint also proposed by AdaFruit, who sell this. If someone uses them, he will have to take care anyways. Could you do the changes, so I can merge?

cu
JAN

jkriege2:
Also, as I just spotted these: Please rename the footprints, so there is an UNDERSCORE before `PLCC`, not a minus, since that's the default separator according to KLC.

Thanks,
JAN

jkriege2:
nce work,

Thanks,
JAN

Issue #301, opened 2018-01-24
Connector_RJ/RJ12_E Problems


openelectron:
I cant seem to find a connector to fit Connector_RJ.pretty/RJ12_E.kicad_mod.

To me it looks like the mount pin holes are too small. Also, the pins seem to be in the wrong orientation as the current part could only be fitted to the opposite side of the board to the silkscreen.

Before I do a PR to fix this does anyone know why this part is this way or know a manufacturer part number for it?.

poeschlr:
As this connector does not include a datasheet link and is not named after a manufacturer part number it might be hard to find what part this is intended for. My suggestion would be to remove such generic parts and replace them with parts that include ordering information. (Naming should then follow the rule "Connectors for specific Functions" in the [Connector naming convention section of the KLC](http://kicad-pcb.org/libraries/klc/F3.6/))

In this case `RJ12_E_[Manufacturer]_[part number]`

mickaelv:
Hi @openelectron , Did you find a part matching with the RJ12e footprint ? 
I think that a wrong footprint too. Not the good orientation and fixing hole are not correct.

openelectron:
@mickaelv  I never found a part. I added a new connector in #347, however it still needs some work before it can be merged.

Pwuts:
@poeschlr The only RJ12 connector in the library is now the one created in #347, but a connector that looks a lot like RJ12_E can be found [all](https://nl.aliexpress.com/item/32767485682.html) [over](https://nl.aliexpress.com/item/32948988244.html) [AliExpress](https://nl.aliexpress.com/item/32826457723.html). It doesn't seem to have a clear brand or type (every reseller gives their own name as brand name) but it is widely available and can easily be found.

What would be your suggestion for a type name in this case? I have made a footprint to its specifications and would like to send in a pull request but I have no idea what to call it.

PR #347, opened 2018-02-06
Amphenol RJ12

openelectron:
Added connector AMPHENOL Modular Jack - 54601

Datasheet:
https://cdn.amphenol-icc.com/media/wysiwyg/files/drawing/c-bmj-0082.pdf

screenshot_2018-02-06_12-50-40

------------


openelectron:
partially fixes issue #301 but did not remove the old footprint

Ratfink:
Thanks for the contribution!  I have a few notes, though.

* It looks like pins 2, 4, and 6 are in the wrong positions.  The datasheet shows them evenly staggered between pins 1, 3, and 5, but you've got them shifted to the right a bit.
* The proper courtyard clearance for connectors is 0.5 mm, not 0.25 mm.
* This isn't an especially high-density part, so a silkscreen line width of 0.1 mm is a bit thin.  0.12 mm is the standard value.
* Though not required by KLC for through-hole parts, moving the silkscreen outline to be just outside of the part's shape would improve visibility.

Also, I notice there's a second footprint in here that you didn't mention in the first post.  A datasheet link and screenshot for that part would help in reviewing it.

openelectron:
@Ratfink, I have made the changes you requested, except the last one. When the connector is fitted in the board the silkscreen is visible already.

Ratfink:
Thanks for making these changes!

* Please include the datasheet link at the end of the footprint description.
* You left the lower vertical silkscreen lines at 0.10 mm thickness, please change them to 0.12 mm as well.
* Please move the value field down a bit more so it's completely outside of the body outline.

openelectron:
@Ratfink I wasn't clear where you wanted the datasheet link. I think its in the right place...

everything else done as requested.

Ratfink:
Looks good, and yep, that's the right place for the datasheet link.

Sorry, but I edited the comment above with two more changes, which it looks like you didn't see before you pushed what you did.  If you could address those points, I think this will be ready to merge.

openelectron:
@Ratfink I think that is it..

openelectron:
@Ratfink  Now I see what you mean (I think) ....

Ratfink:
Thanks for the contribution!

PR #353, opened 2018-02-07
Added buzzer CST-931RP-A

Resaj:
Datasheet: http://www.cui.com/product/resource/cst-931rp-a.pdf

I don't attach any 3D model because it can be found on [traceparts](https://www.tracepartsonline.net/(S(ztalwz2eaq2bpj4jvra0lrz2))/PartDetails.aspx?PartFamilyID=10-11022016-090258&PartID=10-11022016-090258&sk_Reference=CST-931RP-A&SrchRsltType=4&SrchRsltId=1).

imagen
imagen


Ratfink:
Thanks for the contribution!

I'm sorry this wasn't reviewed sooner.  I have a few suggestions:
* The footprint name should be `MagneticBuzzer_CUI_CST-931RP-A`.  There's no need to list the dimensions because this is a footprint for one specific part, and the `MagneticBuzzer` prefix will help footprint filter matching.  And of course please change the 3D model filename to match the new footprint name.
* The F.Fab outline should be 0.10 mm thick.
* The silkscreen circle should be entirely outside of the corresponding portion of the part's outline.  A radius of 4.62 mm or so should be fine.
* Please make the courtyard outline stay entirely outside of the F.Fab outline.  Right now the sound emission hole extends outside the courtyard; I'd like to have that inside the courtyard, especially since the datasheet doesn't list a clearance between the PCB and the bottom of that feature.
* I suggest rounding the pad size up to 1.6 mm.

Resaj:
Thanks for the revision. I've just changed the name and the details of the component.

Ratfink:
Thanks!

Issue #336, opened 2018-02-01
LFCSP-8-1EP_2x3mm_P0.5mm wrong name


jcyrax:
I think it should be names LFCSP-8-1EP_3x2mm_P0.5mm, but maybe I don't quite understand.

Are the body dimensions x/y always from orientation with pin 1 in upper left?

This is how it looks now:
![image](https://user-images.githubusercontent.com/16546947/35694266-7041b2bc-0789-11e8-886f-62b7e0ad2f09.png)

poeschlr:
If this footprint has indeed that name in the current head than it is wrong.

Not only are the body dimensions wrong but the name should also include the size of the exposed pad.

Ratfink:
Fixed by #354.

PR #354, opened 2018-02-07
Fix LFCSP-8-1EP footprint name and silkscreen offset

jcyrax:
Issue described here: https://github.com/KiCad/kicad-footprints/issues/336
Footprint name had size dimensions switched (x/y)
The same for 3D model name (there is no actual model, so no problems with that)
Silkscreen offset from exposed copper was ~0.1mm. Changed to > 0.2mm as per KLC F5.1.2.
Also added exposed pad dimensions to footprint name.
image
image
This is new offset
image

------------


Ratfink:
Thanks for the contribution!

Sorry this didn't get reviewed sooner.  Thanks for fixing this bug.  There's just one more thing I see wrong with this footprint: if you could make the F.Fab reference designator smaller so that it fits entirely inside the outline, that would be wonderful.

If you fix that one little problem and make a PR to change the corresponding symbol (I believe there's just one) to use the new name, I'd be glad to merge these changes.

jcyrax:
Fixed
image

I think everything should be ready to complete this.

Ratfink:
Thanks!

PR #356, opened 2018-02-08
Added APFA3010 footprint

Resaj:
[Datasheet](http://www.kingbrightusa.com/images/catalog/SPEC/APFA3010LSEEZGKQBKC.pdf)
The 3D model can be found in [kingbrightusa.com](http://www.kingbrightusa.com/product.asp?catalog_name=LED&product_id=APFA3010LSEEZGKQBKC) but the STEP is not a solid nor unique part component.

imagen
imagen


Resaj:
I've reviewed the fails, but I can't attach a 3D model and it seems the anchor shouldn't be centered in this case.

jkriege2:
- Please prepend the 3D-model path with `${KISYS3DMOD}/`
- Is there a (specified) symbol for this footprint? If not, it might be a good idea to create one ...
- for the offset-problem: @poeschlr Should we make an exception here?

Resaj:
I've just created the APFA3010 symbol. You can find it in this [PR](https://github.com/KiCad/kicad-symbols/pull/327).

jkriege2:
@poeschlr @evanshultz What should we do with the symbol origin here? Leave it where it is (I would), or move it where the scripts want it (slightly lower in the CoM of the pads)?

poeschlr:
I think centered around the mounting pad would be ok.

The script does use the bounding rectangle around pads. That is why it does suggest a slightly different placement.

Does the datasheet for this device specify some origin for pick and place machines?

jkriege2:
not really ...

jkriege2:
OK, so I'll merge!

@Resaj Thans for your contribution!

JAN

PR #358, opened 2018-02-09
Dip 4 8 8 clearence

Misca1234:
This push consist of SMD-4 and SMD-8 with 2.54 mm pitch and 8 mm clearence,
They are needed for 

VOR1142B4
https://www.vishay.com/docs/84299/vor1142b4.pdf

and 

LH1522AAC
https://www.vishay.com/docs/83821/lh1522ab.pdf

3D models are here
https://github.com/KiCad/kicad-packages3D/pull/260

bild

bild

bild

bild


Misca1234:
I renamedthe foot print, the expected dimension is in XxY 
I did them in YxX of some reason

Misca1234:
wow, that displacement was hard to find

jkriege2:
From the screenshots you show above, the 3D models (actually their pin-pitch) are too small ...

Misca1234:
Well, tjhey should slope, but the script does not allow it 
I made it a little wider, some thing like this 

bild

evanshultz:
Sorry, I do not.

poeschlr:
It is for sure possible. You would need to play the stuff starting at [line 342 of the generator script](https://github.com/easyw/kicad-3d-models-in-freecad/blob/98462aac5fb234e6a7ac1657f664bab8ac4b7ac8/cadquery/FCAD_script_generator/GW_QFP_SOIC_SSOP_TSSOP_SOT/main_generator.py#L342)

Find the lines that are currently horizontal and give them a angle. But this will also require the three point archs to be adjusted.

Misca1234:
The strange miss alignment was due to an error in the foot print itself

I had copied the pad size and distances fro mthe 8 pin variant, all look good now

bild

bild

bild

bild

Shackmeister:
@Misca1234 
I took a look at the footprints
Fab lines are 0.12 and Silk lines are 0.1

Besides that I believe they are ready to merge :)

Do you need some assistance with the 3D models?
the open top is typically if the edge chamfer is to high

Misca1234:
I changed the line with

Regarding the 3D models, well, the script is built up in such fashion that the leg construction need to be reworked  from basic, which I don't intent do right now.

Shackmeister:
Hi @Misca1234 
SSO-8_6.7x9.8mm_P2.54mm_Clearance8mm
Still as the same problem with Silk having 0.1mm width and fab having 0.12mm width.

@poeschlr should we perhaps let Travis give a warning if silk is not 0.12mm and fab is 0.1mm?
I know it's allowed, but I only see it used very rarely

Ratfink:
@Shackmeister Looking at KLC's wording closely, I think that warnings for non-default silk and fab line widths would be OK.  File an issue against [KiCad/kicad-library-utils](https://github.com/KiCad/kicad-library-utils/issues) and I'll try to get to it soon. 🙂

evanshultz:
Or you could just refer to https://github.com/KiCad/kicad-library-utils/issues/217.

Shackmeister:
Hi @Misca1234 
a friendly reminder about the fab and silk width being swapped ;)

Misca1234:
The F.Fab is 0.1 and the F.Silk is 0.12 or is I missing some line ?


Now when poeschlr  got his script running we should put an effort in to overhaul our foot print libraries and convert all these handmades fp's into script based ones.

We could even make a homepage which users can specify the foot print and then generates the foot print, check for duplicates or suggest already existing ones.

Misca1234:
Fixed the width in SSO-8_6.7x9.8mm_P2.54mm_Clearance8mm

Shackmeister:
@Misca1234 
Thank you for the continuing contributions! :)

PR #367, opened 2018-02-12
Add LFCSP-16-1EP_4x4mm_P0.65mm_EP2.1x2.1mm

evanshultz:
As promised at https://github.com/KiCad/kicad-symbols/pull/288, here is the missing footprint.

Datasheet: http://www.analog.com/media/en/technical-documentation/data-sheets/ADG633.pdf

Lamely, ADI doesn't give all footprint dimensions in the datasheet or other PDF docs. Here is the image of Ultra Librarian which shows their recommended footprint:
image

Image:
image

------------

- [ x ] Provide a URL to a datasheet for the footprint(s) you are contributing
- [ x ] An example screenshot image is very helpful 
- [ x ] If there are matching symbol or 3D model PRs, provide link(s) as appropriate


Ratfink:
Thanks for the good work!  I noticed an inconsistency though: the EP on your footprint is smaller than the one in the drawing you provided.  I also notice that ADI's name for the footprint in the datasheet doesn't match the one in the titlebar in the drawing you provided (CP_16_5 vs CP-16-23) - might it be the wrong drawing?

evanshultz:
Yup, I used the dimensions of the slug and not the copper pad. Fixed. Thanks for catching that!

It is interesting that the footprint numbers don't match. I got the drawing above from ADI's website, as they use the Ultra Librarian software to share footprint designs and don't give a PDF like any reasonable company. The ADG633 footprints are at http://www.analog.com/en/design-center/packaging-quality-symbols-footprints/symbols-and-footprints/ADG633.html. Let me ask ADI to confirm.

evanshultz:
ADI provided an updated BXL file for the YCP package which looks like this:
image

Now the package name matches. It looks basically the same as above so no new image.

evanshultz:
I'm fine to merge. The silk is at least definitive enough, without a clear distance away from pads (we have a minimum), and the ref des will fit 4+ chars (e.g. R999) which is quite a few.

PR #368, opened 2018-02-13
Connector_Harwin: Add the M20-89 family

antoniovazquezblanco:
Info: https://www.harwin.com/products/M20-8900705/

imagen

No symbol PR.
No 3D PR.
Generation script PR: https://github.com/pointhi/kicad-footprint-generator/pull/113

------------


antoniovazquezblanco:
Travis doesn't seem to be happy with the script generated models.

Is this a parser thing or should I notify this to the script guys?

evanshultz:
We are the script guys! :)

Many other footprints are script-generated, so I don't think that's the sole issue. Travis says `Error: Trouble with nesting of brackets`, so let's look at `Harwin_M20-89010xx_1x10_P2.54mm_Horizontal.kicad_mod` and see if we can figure it out:
1. First, run the scripts on the footprint locally. I get the same error.
2. I see `(0.1\")` in the description. You seem to be trying to write the inch character but the escape character is being written too. In KiCad the entire description, if there are spaces, is captured in quotes. You cannot use a quote mark within the description. Changing this to `(0.1 inch)` fixes the issue for me. But let's continue since we're here...
3. I also see `M20-890{pincount:d}xx` in the description. I suspect you're not using the `format` method of a string properly.
4. You do not need `, generated with kicad-footprint-generator` in the description.
5. Please use other recent connectors as a guide for the description. The order and terms you use don't seem to be as succinct and useful as many of the recent connectors. See Harwin_Gecko-G125-FVX3405L0X_2x17_P1.25mm_Vertical.kicad_mod and Molex_SlimStack_54722-0304_2x15_P0.50mm_Vertical.kicad_mod, and others, for inspiration.
6. There are two spaces here in the keywords: `Harwin  top`.

I didn't review the symbol itself, just investigated how to fix your issue.

Also, did you write this script yourself or use the framework at https://github.com/pointhi/kicad-footprint-generator? If you started with a recent script using that framework, you will have a good reference for how previously accepted footprints have been generated.

Lastly, this wasn't hard to find and fix. You should look at the resulting footprint files for correctness. You should load them into KiCad and check that there are no problems. In this case the footprints can be placed onto a board and opened in the footprint editor with no issues (and it evens renders the inch character correctly!). But simply looking at the text file made the problem appear quickly.

poeschlr:
About point 4: I added that to all my script generated footprints as somebody suggested that in some issue half a year ago. The idea is to communicate to future lib maintainers and to the user that there is a script for this specific footprint.

antoniovazquezblanco:
Fixed some of the points avobe. Did not work on point 5. I am using the connector yaml file in the connectors script directory. Should I add a file only for this series?

Thanks

pointhi:
@evanshultz 

> I see (0.1\") in the description. You seem to be trying to write the inch character but the escape character is being written too. In KiCad the entire description, if there are spaces, is captured in quotes. You cannot use a quote mark within the description. Changing this to (0.1 inch) fixes the issue for me. But let's continue since we're here...

I think this is probably an error with the checker script on travis. Escaping should work correctly with my generator script (https://github.com/pointhi/kicad-footprint-generator/pull/102)

antoniovazquezblanco:
I am switching to an organization repo. I have fixed the center issue and re pushed to https://github.com/KiCad/kicad-footprints/pull/428

Sorry to switch to another PR

PR #288, opened 2018-01-21
added Alps EC11 and EC12 rotary encoder footprints

jkriege2:
- made to match generic symbols from device.lib (see also https://github.com/KiCad/kicad-symbols/pull/206)
2018-01-21 17_59_24-symbol library editor - device d__kicad_kicad-symbols_device lib
- variants with and without switch
- for now 20mm shaft lengtrh variants (given height for future 3D models, variants with other shaft lengths can come in future)
- variants with circular holes for mounting-pins
- mounting pins are assigned pin0 to allow connecting them to a shield explicitly with symbols from https://github.com/KiCad/kicad-symbols/pull/206
- datasheets:
  - http://www.alps.com/prod/info/E/HTML/Encoder/Incremental/EC11/EC11E15204A3.html
  - http://www.alps.com/prod/info/E/HTML/Encoder/Incremental/EC12E/EC12E1240405.html 
  - http://cdn-reichelt.de/documents/datenblatt/F100/402097STEC12E08.PDF
- cross marks axis center on F.Fab and F.Silks

2018-01-21 21_39_36-pcbnew d__kicad_devproject_devproject kicad_pcb
2018-01-21 21_39_24-pcbnew d__kicad_devproject_devproject kicad_pcb
2018-01-21 21_39_48-3d viewer

## check against 3D models from ALPS: EC11
These images show that the ALPS models fit the footprints:
2018-01-22 19_41_26-3d viewer
2018-01-22 19_41_12-3d viewer
2018-01-22 19_40_20-3d viewer
2018-01-22 19_39_53-3d viewer

## check against 3D models from ALPS: EC12
These images show that the ALPS models fit the footprints:
2018-01-22 19_46_43-3d viewer
2018-01-22 19_45_52-3d viewer
2018-01-22 19_45_43-3d viewer

------------

Thanks for creating a pull request to contribute to the KiCad libraries! To speed up integration of your PR, please check the following items:

- [x] Provide a URL to a datasheet for the footprint(s) you are contributing
- [x] An example screenshot image is very helpful 
- [x] Check the output of the Travis automated check scripts - fix any errors as required

PR #370, opened 2018-02-13
Add PEC11R

strombom:
Bourns PEC11R rotary encoders
http://www.bourns.com/docs/Product-Datasheets/PEC11R.pdf

image

image

------------


strombom:
I think the mounting holes are too small, is there a guideline for the sizing of such holes for the Kicad library when the database doesn't have any information?

evanshultz:
By "database" do you mean datasheet? I don't see a footprint given on page 3, but those parts would seem to match with the page 2 footprints. What exactly is the problem you see?

strombom:
Yes I meant datasheet :)

I have downloaded many different encoder and potentiometer datasheets, they all have square mounting holes, but I'm not sure it's a good idea or even possible to specify a square mounting hole in a footprint.

If the hole is to be oval I think it needs to be a bit larger, but I would only guess that it should be maybe 0,5 mm longer than the length of the square hole in the datasheet. Also I guess the annular ring would need to be quite a bit larger for mechanical stability, and that would be a question of taste, and I don't know what other Kicad users would want. :)

poeschlr:
One thing i can say with certainty. There is no way to manufacture a true square (or rectangular) hole. There will always be a radius on the corners.

Sadly kicad does not support rounded rectangle holes. So i fear the only option would be a oval hole.

Are the mounting pads intended to be bend over to lock the part to the pcb or is it intended to be soldered?
If it is intended to be bend over then the straight edge should really be as long as specified for the rectangle. 
Otherwise you just need a large enough hole such that the pin can fit. (But one crucial measurement is missing from the datasheet to truly determine that.)

strombom:
I'm fairlly certain that the mounting pins should be soldered, at minimum I would like to have the option to do so.

I looked at the Amps RK09L in the library, it has an annular ring of 1 mm, so I updated this footprint to 1 mm.

For the drill measurement i checked the 3D model from Bourns and measured the pin dimension to 2,5 x 1,3 mm. (2.5**2+1.3**2)**0.5 = 2.8 mm. So I added a small margin and selected 3 mm for the drill.

image

image

jkriege2:
Also have a look at my PR: https://github.com/KiCad/kicad-footprints/pull/288 (I made one version with circular holes and one with oval holes) ... and the discussion started here: https://github.com/KiCad/kicad-symbols/pull/220#issuecomment-365507581 (for pin-numbering)

strombom:
Oh I didn't see those, very nice, I guess I will wait until those are merged and then make a new pull request with an updated footprint that matches yours.

PR #371, opened 2018-02-14
Add Potentiometer_TT_P0915N

strombom:
TT Electronics P0915N
http://www.ttelectronics.com/sites/default/files/download-files/Datasheet_PanelPot_P09xSeries.pdf

image

------------


strombom:
Regarding travis output:
  Violating F7.2
    For through-hole components, footprint anchor is set on pad 1
    Pad '1' not located at origin
     - Set origin to location of Pad '1'

For a potentiometer, the center of the shaft must be the most important dimension, so it should be set at the origin right?

strombom:
Does this need to be rotated?

One could argue that pin 1 is actually on top left ;)

Misca1234:
Is P0915N really the right name, P0915N indicates it should be 15 gangs?

Should the name not be P091N ?

herostrat:
Yes you are right Misca,

I think the correct name would be the same as of their datasheet:
P09x

@strombom do you want to finish this PR up? :)

herostrat:
I just checked the datasheet again to review the dimensions in freecad and noticed something:

The x5N is their specification for the top adjust ones?
image

You wanted to model them stromborn, didn't you?
I missed it because they do not specify that there can be a one row version (although there can be _up_ to two gangs, I know)

Looking up the products in google showed that you were correct before with the name 

P0915N

Sorry for the inconvenience...

---

For the check:
Original:
image

image

I think it is good to go.

strombom:
Here are a couple of pictures of the component and the footprint:

image image image
image

As for modelling, I don't really know how to use FreeCAD, I could do it in OnShape but that's not open source either :/

herostrat:
Thanks for the update and the images!

I meant FreeCAD for the dimension verfication, not the 3D modeling.
The second image I posted was generated with FreeCAD and the imported footprint.
So I can easy check the dimensions.

Misca1234:
@herostrat 
If you checked the dimension and all good, you can just merge it

herostrat:
I think they are all good.
And the picture he provides shows a reallife example too:)

I have no write access, we will have to wait for the maintainers

Misca1234:
the maintainers have arrived

herostrat:
Thanks a lot :D

PR #364, opened 2018-02-10
Added footprint for panasonic caps

matyro:
https://industrial.panasonic.com/cdbs/www-data/pdf/RDE0000/DME0000COL92.pdf

Created multiple panasonic smd caps with "c_elec_round.py" script.
B
grafik

grafik

C
grafik

grafik

D
grafik

grafik

E
grafik

grafik

F
grafik

grafik

G
grafik

grafik

H
grafik

grafik

J
grafik

grafik

K
grafik

grafik

PR #372, opened 2018-02-14
Brand spanking new SMD Radial Capacitors

Shackmeister:
Hi guys please be thorough when reviewing this, and dont hesitate to suggest improvements.
Ill fix 3D models after
capacitor_smd

------------


poeschlr:
Just out of curiosity: How did you determine the pad sizes? 

The ipc document i used for the other two terminal smd devices has a table for these type of devices as well. You could then use the same sort of calculation and we would get footprints that would be aligned with IPC standards. I don't want to link it here directly as i assume it is kind of illegal but the link is in the ipc.yaml file in my two terminal smd devices generator.

Clarification: It is not mandatory but if you have the time it might be a good idea.

---

Could you list what you improved? (I don't have the time to make the comparison my self)

Shackmeister:
I havent changed the parameters used (but might consider the IPC stuff in the future)

improvements:

- aligned "style" in the scripted ones to the ones in the repo
- changed the script to use the properties from "config_KLCv3.0"
- added auto sizing of fab REF**
- added proper alignment of silkscreen and courtyard
- added autoplacement and sizing of the polarity marker to have it visible after placement
probably a few more 

actually I just wanted to fix the polarity marker issue, before fixing the 3D models.
But since the script was obsolete I redid most of it

jkriege2:
Hi!

Just looking at the uploaded files, I see manufacturer names in the descriptions of the CP_ELEC ... so I think they should either be named as the CP_Panasonic. An alternative would be to call them `CP_Elec_D<dd.d>mm_h<ll.l>mm_P<pp.p>mm`. This way the naming is somewhat aligned with the naming of the THT-CPs.

Here are some datasheets listing sizes (and land patterns):
- http://www.nichicon.co.jp/english/products/pdfs/e-p_ref.pdf
- http://www.nichicon.co.jp/english/products/pdfs/e-ch_ref.pdf
- https://de.tdk.eu/inf/20/30/db/alu_x/B41121.pdf
- https://www.vishay.com/docs/28415/180cps.pdf

Shackmeister:
The manufacturer names was originally in another PR, but made with the old script. I just added them in here to avoid redoing it later.
Im not sure if we should add more info in the name, add manf or just stick with the generics.
Suggestions are very welcome.

poeschlr:
The naming syntax suggested by jan sounds reasonable. (`CP_Elec_D<dd.d>mm_h<ll.l>mm_P<pp.p>mm`)

However i am not sure if pitch makes sense for these. (Do datasheets mention it? How would we define pitch here? Is it center to center for the pad on the pcb or center to center for the "lead" on the capacitor?)
So maybe it would make sense to have `CP_Elec_D<dd.d>mm_h<ll.l>mm` as the naming. (This is the same information as now but closer alignment to the current KLC.)

---

If we really want to give additional information i would guess the lead to lead clearance plus lead length might be better choices (Dimensions E and A from this datasheet)
screenshot from 2018-02-20 11-02-20
Maybe even the body size (B and C)

So a fully specified name could be: `CP_Elec_<bb.b>x<cc.c>mm_D<dd.d>mm_h<ll.l>mm_Clearance<ee.e>mm_Lead<aa.a>mm`
But i think this all is not really necessary as these are normally quite standardized.

Shackmeister:
I agree these footprints are generally quote standardized. I have always just taken whatever Diameter seemed right, never had issues at pick and place production or handsoldering for that matter..

Most of these footprints are simply corrections of the existing ones. The Panasonic ones were added here:
https://github.com/KiCad/kicad-footprints/pull/364
I simply decided to included them here to avoid redoing too much. 
@matyro perhaps you can enlighten us about why you made the Panasonic footprints?

evanshultz:
I've come across packages that vary only B or C dimension, but are otherwise the same. Why don't we add that now so we've covered in the future?

IPC-7351B, for this type of part, shows `Capacitors, Aluminum Electrolytic: CAPAE + Base Body Size X Height `. I suppose "base body size" includes both dimensions?

[Digikey ](https://www.digikey.com/products/en/capacitors/aluminum-electrolytic-capacitors/58)also shows varying only one size for these caps:
image

Shackmeister:
@evanshultz if B and C are the main variants I believe adding them in the name would be great. I honestly don't know if adding pitch, pad size or similar would make sense to add.. None the less, I really hope we can come to a decision soon since I still need to fix the 3D models and have a lot of other things I would like to fix before V5 :)

evanshultz:
I think B and C are important and would define the footprints well. I haven't seen any issues with pitch and pad size using various vendors. It's fine with me if a few datasheets are searched and we choose the largest reasonable envelope of pad size and position for all caps.

matyro:
@Shackmeister
I was in need of the H, J and K footprint (Diameter: 12.5mm, 16mm, 18mm) and checked the other footprints briefly and saw some (small) differents. Because I used the script it was only a couple of minuts addition time needed to create them all and match them to the footprints in the datasheet name.

I wanted to add the vibration proof footprints as well in the near futur before I heared of your rework.

Page 9 for the footprint information of SMD Panasonic (Electrolyte) caps:
https://industrial.panasonic.com/cdbs/www-data/pdf/RDE0000/DME0000COL92.pdf

Shackmeister:
I can kinda see the nice thing to have the manufacturer naming (H, J, K whatever) included in the library. not sure if they should be seperate or spliced in the naming.
I believe the Tantalum SMD library is practically made up by manufacturers naming schemes

poeschlr:
Yes tantulum caps include the letter-code naming schemes. The main naming is the metric size code. (Letter codes are to the back and they include which manufacturers standard they represent as they sadly are not always compatible. Including the letter-codes there does not mean that each footprint is manufacturer specific. It is just a separate way to identify the size.)

matyro:
OT:
I think alias or symlinks for some footprints would be a nice feature as well...

Shackmeister:
Should I perhaps leave out the Panasonic caps for now? I would like to get the other ones corrected and fix the 3D models accordingly soon.
Im personally not against the manf specific footprints in general, they might be even more important if we want to add the vibrationproof footprints

poeschlr:
This seems to have been forgotten.

First a bit of detail about why the tantals inclued "manufacturer" size codes: This is mainly because the old lib has it as the main identifier. We where not prepared to get rid of it completely but we at least moved it back towards the end of the naming and added from which "standard" the letter comes from.

So i doubt that this would apply here.

---

If the panasonic caps are new additions then pleas move them to a separate PR. We should really start to use git a bit more properly and avoid mixing different tasks in the same PR. (This would avoid so much trouble as PRs that fix stuff can be merged faster then new additions.)

Shackmeister:
removed the Panasonic caps from the PR

Shackmeister:
HI guess

I would like to get this merged before V5, should we stick to the current naming scheme?

Ratfink:
[KLC]: http://kicad-pcb.org/libraries/klc/ "KiCad Library Convention"
[G1.1]: http://kicad-pcb.org/libraries/klc/G1.1/ "Only standard characters are used for naming libraries and components"
[G1.2]: http://kicad-pcb.org/libraries/klc/G1.2/ "Each library is limited to 250 items"
[G1.3]: http://kicad-pcb.org/libraries/klc/G1.3/ "Libraries are organized by functionality"
[G1.4]: http://kicad-pcb.org/libraries/klc/G1.4/ "English language should be used throughout libraries"
[G1.5]: http://kicad-pcb.org/libraries/klc/G1.5/ "Plural naming is to be avoided"
[G1.6]: http://kicad-pcb.org/libraries/klc/G1.6/ "Capitalization conventions"
[G1.7]: http://kicad-pcb.org/libraries/klc/G1.7/ "Library files use Unix style line endings"
[G1.8]: http://kicad-pcb.org/libraries/klc/G1.8/ "Components can only contain features supported in latest stable release"
[G1.9]: http://kicad-pcb.org/libraries/klc/G1.9/ "Dimensional units"
[G2.1]: http://kicad-pcb.org/libraries/klc/G2.1/ "Defining generic and atomic parts"
[S1.1]: http://kicad-pcb.org/libraries/klc/S1.1/ "Symbol libraries should be categorized by function"
[S2.1]: http://kicad-pcb.org/libraries/klc/S2.1/ "General symbol naming guidelines"
[S2.2]: http://kicad-pcb.org/libraries/klc/S2.2/ "Non-functional variations in part number should be replaced with wildcard"
[S2.3]: http://kicad-pcb.org/libraries/klc/S2.3/ "Where atomic parts are available in multiple footprint options, a separate symbol must be drawn for each footprint"
[S3.1]: http://kicad-pcb.org/libraries/klc/S3.1/ "Origin is centered on the middle of the symbol"
[S3.2]: http://kicad-pcb.org/libraries/klc/S3.2/ "Text fields should use a common text size of 50mils"
[S3.3]: http://kicad-pcb.org/libraries/klc/S3.3/ "Symbol outline and fill requirements"
[S3.4]: http://kicad-pcb.org/libraries/klc/S3.4/ "Symbols with complex functionality may incorporate simple functional diagrams"
[S3.5]: http://kicad-pcb.org/libraries/klc/S3.5/ "Pin connection points must be placed outside of the symbol"
[S3.6]: http://kicad-pcb.org/libraries/klc/S3.6/ "Pin name position offset"
[S3.7]: http://kicad-pcb.org/libraries/klc/S3.7/ "Pin numbering for exposed pads"
[S3.8]: http://kicad-pcb.org/libraries/klc/S3.8/ "Multi unit symbols"
[S4.1]: http://kicad-pcb.org/libraries/klc/S4.1/ "General pin requirements"
[S4.2]: http://kicad-pcb.org/libraries/klc/S4.2/ "Pins should be grouped by function"
[S4.3]: http://kicad-pcb.org/libraries/klc/S4.3/ "Rules for pin stacking"
[S4.4]: http://kicad-pcb.org/libraries/klc/S4.4/ "Pin electrical type"
[S4.5]: http://kicad-pcb.org/libraries/klc/S4.5/ "Pins not connected on the footprint may be omitted from the symbol"
[S4.6]: http://kicad-pcb.org/libraries/klc/S4.6/ "Hidden pins"
[S4.7]: http://kicad-pcb.org/libraries/klc/S4.7/ "Active low pins should be designated using a bar above the symbol name"
[S5.1]: http://kicad-pcb.org/libraries/klc/S5.1/ "Symbols with a default footprint link to a valid footprint file"
[S5.2]: http://kicad-pcb.org/libraries/klc/S5.2/ "Footprint filters should match all appropriate footprints"
[S6.1]: http://kicad-pcb.org/libraries/klc/S6.1/ "Component Reference Designator (RefDes)"
[S6.2]: http://kicad-pcb.org/libraries/klc/S6.2/ "Component fields must be filled appropriately"
[S6.3]: http://kicad-pcb.org/libraries/klc/S6.3/ "Component metadata is added to symbol (and all aliases)"
[S7.1]: http://kicad-pcb.org/libraries/klc/S7.1/ "Power flag symbols"
[S7.2]: http://kicad-pcb.org/libraries/klc/S7.2/ "Graphical symbols"
[F1.1]: http://kicad-pcb.org/libraries/klc/F1.1/ "Footprint libraries are categorized by function"
[F1.2]: http://kicad-pcb.org/libraries/klc/F1.2/ "Connector footprint libraries"
[F2.1]: http://kicad-pcb.org/libraries/klc/F2.1/ "General footprint naming conventions"
[F2.2]: http://kicad-pcb.org/libraries/klc/F2.2/ "Footprint naming field prefixes"
[F2.3]: http://kicad-pcb.org/libraries/klc/F2.3/ "Manufacturer specific version of generic footprints"
[F2.4]: http://kicad-pcb.org/libraries/klc/F2.4/ "Footprint naming for non-standard pin numbering"
[F2.5]: http://kicad-pcb.org/libraries/klc/F2.5/ "Footprint naming conventions for specific components"
[F3.1]: http://kicad-pcb.org/libraries/klc/F3.1/ "SMD chip package naming conventions"
[F3.2]: http://kicad-pcb.org/libraries/klc/F3.2/ "Resistor naming conventions"
[F3.3]: http://kicad-pcb.org/libraries/klc/F3.3/ "Capacitor naming conventions"
[F3.4]: http://kicad-pcb.org/libraries/klc/F3.4/ "SMD IC package naming conventions"
[F3.5]: http://kicad-pcb.org/libraries/klc/F3.5/ "THT IC package naming conventions"
[F3.6]: http://kicad-pcb.org/libraries/klc/F3.6/ "Connector naming conventions"
[F3.7]: http://kicad-pcb.org/libraries/klc/F3.7/ "Fuse naming conventions"
[F4.1]: http://kicad-pcb.org/libraries/klc/F4.1/ "Datasheet recommendations take priority"
[F4.2]: http://kicad-pcb.org/libraries/klc/F4.2/ "Pin 1 should be located at the top left"
[F4.3]: http://kicad-pcb.org/libraries/klc/F4.3/ "Connected copper elements have the same pad number"
[F4.4]: http://kicad-pcb.org/libraries/klc/F4.4/ "Thermal pads"
[F4.5]: http://kicad-pcb.org/libraries/klc/F4.5/ "Specifying footprint keepout areas"
[F4.6]: http://kicad-pcb.org/libraries/klc/F4.6/ "Local clearance and settings should be set to zero"
[F5.1]: http://kicad-pcb.org/libraries/klc/F5.1/ "Silkscreen layer requirements"
[F5.2]: http://kicad-pcb.org/libraries/klc/F5.2/ "Fabrication layer requirements"
[F5.3]: http://kicad-pcb.org/libraries/klc/F5.3/ "Courtyard layer requirements"
[F6.1]: http://kicad-pcb.org/libraries/klc/F6.1/ "Footprint placement type must be set to surface mount"
[F6.2]: http://kicad-pcb.org/libraries/klc/F6.2/ "Footprint anchor should be placed in the middle of the component body"
[F6.3]: http://kicad-pcb.org/libraries/klc/F6.3/ "Pad requirements for SMD footprints"
[F7.1]: http://kicad-pcb.org/libraries/klc/F7.1/ "Footprint placement type must be set to Through Hole"
[F7.2]: http://kicad-pcb.org/libraries/klc/F7.2/ "Footprint anchor should placed at the location of Pin-1"
[F7.3]: http://kicad-pcb.org/libraries/klc/F7.3/ "Pin 1 should be rectangular, and other pads circular or oval"
[F7.4]: http://kicad-pcb.org/libraries/klc/F7.4/ "Pad requirements for THT footprints"
[F7.5]: http://kicad-pcb.org/libraries/klc/F7.5/ "Minimum annular ring width"
[F7.6]: http://kicad-pcb.org/libraries/klc/F7.6/ "Minimum hole diameter"
[F8.1]: http://kicad-pcb.org/libraries/klc/F8.1/ "Virtual components"
[F9.1]: http://kicad-pcb.org/libraries/klc/F9.1/ "Footprint meta-data is filled in as appropriate"
[F9.2]: http://kicad-pcb.org/libraries/klc/F9.2/ "Footprint properties are as default values unless otherwise required in datasheet"
[F9.3]: http://kicad-pcb.org/libraries/klc/F9.3/ "Footprint 3D model requirements"
[M1.1]: http://kicad-pcb.org/libraries/klc/M1.1/ "Models should be created by contributor"
[M1.2]: http://kicad-pcb.org/libraries/klc/M1.2/ "Manufacturer models should not be duplicated"
[M1.3]: http://kicad-pcb.org/libraries/klc/M1.3/ "Source files for 3D models should be supplied"
[M2.1]: http://kicad-pcb.org/libraries/klc/M2.1/ "Permitted 3D model file types"
[M2.2]: http://kicad-pcb.org/libraries/klc/M2.2/ "Model alignment and scaling"

It looks like the current naming you have here is `CP_Elec_[diameter]x[height]`.  This breaks [F2.2], which says that a `[…]x[…]` field is for X and Y dimensions, not D and H.  I suggest using explicit D and H fields, e.g. `CP_Elec_D16mm_H22mm`, which would be more consistent with both KLC and the naming used by our THT electrolytic capacitor footprints.

To be fully consistent with KLC though, you'd need to specify body X and Y dimensions, e.g. `CP_Elec_16.6x16.6mm_D16mm_H22mm`.  This would allow different body sizes for the same diameter, which may be important.  I looked closer at the parts listed from the DigiKey screenshot Evan gave, and found that for some where the body is listed as non-square, they're counting the ends of the leads as one of the dimensions and the body *is* square.  Others are actually non-square.  How should we measure these, then?

* [listed as 5.3x5.8mm, correct body size, lead ends 6.5mm](http://www.illinoiscapacitor.com/pdf/seriesDocuments/ATB%20series.pdf)
* [listed as 5.3x6.5mm, body actually 5.3x5.3mm, lead ends 6.5mm](https://content.kemet.com/datasheets/KEM_A4002_EEV.pdf#page=2)

poeschlr:
@Ratfink The [sice x]x[size y]mm body size "field" can be replaced with other fields if they hold the same information. (See for example the through hole capacitors and resistors.)
I don't think it is necessary to include it here.

I had already suggested a naming convention that would fulfill KLC without needing additional data.: `CP_Elec_D<dd.d>mm_H<ll.l>mm`

poeschlr:
As rc2 is near and this does address a bug (wrong polarity marking) i will merge it as it now stands.

I think we can still do some minor renamings before v5 but we should only do it if we have good reasons for it.

PR #375, opened 2018-02-15
Add TE-Connectivity DDR2 SODIMM connector

bobc:
For TE DDR2 SODIMM connector

[datasheet](http://www.te.com/commerce/DocumentDelivery/DDEController?Action=showdoc&DocId=Customer+Drawing%7F1473149%7FK3%7Fpdf%7FEnglish%7FENG_CD_1473149_K3.pdf%7F1473149-1)

image

image

------------

- [n/a] If there are matching symbol or 3D model PRs, provide link(s) as appropriate


jkriege2:
Is there a symbol for these connectors?

bobc:
I'm planning to submit symbols for the Raspberry Pi Compute modules which use SODIMM for factor, but I haven't quite finalised that yet.

I can also create a symbol for generic 200 pin SODIMM, which lib would that go in?

poeschlr:
Either connector* or memory. 

In memory if the symbol will have function specific pin names. (This would only make sense if the interface is specialized.) But even then one could argue that it should still go into the connector lib as it currently also holds the sd card symbols.


*) The connector_specialized lib will be renamed to connector (specialized suffix does not really fit that lib and we will drop it)

bobc:
I think that the SODIMM socket is like a DIP socket. There are numbered pins assigned, but the functions depend on what module you plug in to it.

I think a SODIMM socket symbol would be a generic 2x100 connector style with a filter like *SODIMM?200*.

poeschlr:
Yup in that case put it in to the connector lib. (If somebody later makes specialized symbols for connecting specific memory modules then we can put them into the memory lib.)

poeschlr:
I checked the footprint against the [3d model downloaded from TE](http://www.te.com/usa-en/product-1473149-4.html). The pad placement looks ok.

The fab outline looks quite different (and a lot larger) then the 3d model. 
screenshot from 2018-02-15 13-09-50

Is there some part missing from the 3d model or is there maybe some accessory that would make the part larger?

bobc:
Just FYI, the footprint was originally drawn by @Joan_sparky (https://forum.kicad.info/t/raspberry-pi-compute-module-component-library/5480) I fixed up issues flagged by KLC script.

I can adjust the courtyard, should it be +0.5mm for connectors?

Shackmeister:
@bobc the courtyard should be +0.5mm

herostrat:
I added a generic symbol for SODIMM 200 as proposed here.
https://github.com/KiCad/kicad-symbols/pull/638

Please let me know if thats ok for you.

herostrat:
@bobc Is there any way I can help you so that it can be merged? :)

bobc:
Sure, redraw the courtyard to +0.5mm, I just don't have time.

Misca1234:
@herostrat 
Do you want to take over this?
that is, you copy bobc's work to your own branch and then we close this one ?

herostrat:
I created a PR for the updated footprint under:
https://github.com/KiCad/kicad-footprints/pull/1158

I think it can be closed now (or when 1158 got merged)

Misca1234:
Thanks

PR #376, opened 2018-02-15
Add Dwm1001 footprint

ouinouin:
Hi, here is the footprint for the DWM1001 RF module from decawave , the documentation with footprint is available [here] (https://www.decawave.com/sites/default/files/dwm1001_datasheet.pdf).
a PR for the schematic lib available [here](https://github.com/KiCad/kicad-symbols/pull/306).
here is the footprint view : 
capture d ecran de 2018-02-15 14-34-26

3d model already made, will be pushed while I ll get more Bandwidth


jkriege2:
Hi!

and thanks for your contribution! I just had a look at your footprint:
1. Please rename to the `MAN_MPN`-standard, i.e. `DecaWave_DWM1001` in this case.
1. the courtyard offset at top and bottom edge is much larger than the required 0.25mm
1. the F.Fab-REFDES (%R) should be placed at (0,0)
1. the footprint name (on F.Fab) should be moved above outside the footprint
1. The F.SilkS-REFDES could be moved closer to the outline of the device
1. Silkscreen-lines could be moved closer to the F.Fab outline (usually we use an offset of 0.1-0.2mm there)
1. add some kind of pin1-marker on F.Fab (often we use a trinagle/corner in the outline or a beveled corner, e.g. 
2018-02-17 16_14_42-footprint editor no active library
2018-01-03 13_51_32-footprint editor no active library
1. Finally please mark the keepout zone on Dwgs.User (see http://kicad-pcb.org/libraries/klc/F4.5/ ... you will have to draw on another layer and then go to the line-properties and set them to the desired layer)
2018-02-18 12_33_14-footprint editor no active library

Thanks,
JAN

ouinouin:
Hi, I think I implemented all your corrections propositions :

jkriege2:
Looks mostly good to me!  Only one minor thing left: Please add some kind of pin1-marker on F.FilkS

Thanks,
JAN

ouinouin:
Hi, @jkriege2 I finally had time (whah it took me two monthes to make a 5 min job...) to make a pin designator on pin 1 for the DWM1001 footprint. to close this last point.
capture d ecran de 2018-05-05 10-27-38

Regards and thanks for comments.

Ratfink:
Oops, I didn't even think to look for an open PR for this footprint.  We're currently working on fixing all the broken footprint associations for the symbol library (https://github.com/KiCad/kicad-symbols/issues/520), and this part was one of the problems, so I created a footprint for it last week (https://github.com/KiCad/kicad-footprints/pull/523) which has already been merged.  Unless there are any problems with the footprint that's already in the library that are fixed in this version of the symbol, I feel like this PR should be closed as a duplicate.

ouinouin:
I just saw that, fair enough, I was a bit long to finish to polish my footprint (thanks to librarians comments to improve overall footprint quality). if the footprint quality is equivalent or higher, this PR should effectively be closed.

PR #381, opened 2018-02-17
5-pin SAW filter 1.4x1.1mm

calebns:
I couldn't find a place where to put the footprint, so I propose to create a new "Filter" library. Please let me know if that's OK or if I should move it somewhere else.

1.4x1.1mm:
https://www.mouser.com/ds/2/40/AVX_140127%20Connectivity-1158389.pdf
image

1.1x0.9mm:
https://www.mouser.com/ds/2/842/B8813-1092540.pdf

image

------------


jkriege2:
Hi!

thanks for the contribution. I found a few issues with these footprints:
1. Could you please add a pin1-marker to F.Fab (see screenshot below
1. Maybe use a pin1-marker on F.SilkS which is a bit closer to other components (we often use double-stroke corners) in the lib, e.g.:
2018-02-17 16_14_42-footprint editor no active library
1. About the naming I'm not sure ... these are not standardized, but they seem to be used by more than one manuacturer ... @evanshultz @poeschlr Do you have proposals for the naming? Maybe `Filter_1411-5_1.4x1.1mm` and `Filter_1109-5_1.1x0.9mm`??? That does cover the pad-count, but not the layout ... Using a manufacturer MPN does not work either, because the packages are used by multiple manufacturers and for different types of filters ...

calebns:
Is it OK now?

image

jkriege2:
looks goot, but please replace the minus `-` before the package size `1.4x1.1mm`/`1.1x0.9mm` by an underscore `_`

Thanks,
JAN

calebns:
Updated.

jkriege2:
Thanks!

JAN

PR #382, opened 2018-02-17
Remove 3D models from SMD HC49 crystals (inexistent)

diegoherranz:
Remove inexistent models.
Not found in https://github.com/KiCad/kicad-packages3D/tree/master/Crystal.3dshapes


jkriege2:
We require 3D model links in every footprint, even if the 3D model does not (yet) exist. This way we don't need to change the footprint, when someone provides the 3D model. (see http://kicad-pcb.org/libraries/klc/F9.3/).

SO I'm closing this!

JAN

diegoherranz:
Sorry for the noise. As soon as Travis failed, I saw that on the KLC.

PR #383, opened 2018-02-17
Use same model for HandSoldering version of SMD HC49 crystal.

diegoherranz:
HandSoldering or not, the 3D should be the same.
Note: that model is missing at the moment.


jkriege2:
Thanks,
JAN

PR #384, opened 2018-02-17
S-PWSON-N10 with 4EPFingers footprints added

Resaj:
Packages added for TPS6306x.
Datasheet: http://www.ti.com/lit/ds/symlink/tps63060.pdf

imagen
imagen


aimylios:
I think you don't need to add so much information to the name, Texas_S-PWSON-N10 should be sufficient (unless there is more than one variant with the same name). This is the convention used by most of the other manufacturer specific footprints.

Confusingly, there is already [a file with that name](https://github.com/KiCad/kicad-footprints/blob/master/Package_SON.pretty/Texas_S-PWSON-N10.kicad_mod) in the Package_SON repository. But this is a typo, it should be called Texas_S-PVSON-N10 instead. You could fix this as part of your pull request and then drop the suffix from your new footprints name.

poeschlr:
@aimylios have a look at the discussion over at the symbol lib: https://github.com/KiCad/kicad-symbols/pull/316

This package does not seem to be as standardized as one might expect from such a long manufacturer identification number

aimylios:
@poeschlr Thanks for the hint, I missed that!

Nevertheless, I think my comment is valid. The existing footprints Texas_S-PVSON-N10 and Texas_S-PWSON-N10 are duplicates. One has been made for the the [LM5165](http://www.ti.com/lit/ds/symlink/lm5165.pdf), the other for the [TPS61200](http://www.ti.com/lit/ds/symlink/tps61200.pdf). To my understanding, both of these parts use exactly the same package, therefore the current Texas_S-PWSON-N10 is not actually an S-PWSON-N10 and should be removed or at least renamed. After that, the footprint submitted in this PR can be (correctly) named Texas_S-PWSON-N10 without any suffix.

texas_s-pwson_s-pvson

Resaj:
Thanks for the advises.
I'll substitute the current S-PWSON-N10 by my new designs with simplified names.

jkriege2:
@aimylios @Resaj @poeschlr Sorry, just renaming does not work ... If you look at the TPS61200 datasheet ... these devices are available in a VSON and a WSON variant (screenshots fomr the very same datasheet): 
2018-02-18 08_03_31-tps61200-1 pdf - adobe acrobat reader dc
2018-02-18 08_03_44-tps61200-1 pdf - adobe acrobat reader dc

The major difference between the two seems to be the hieght VSON is 0.8-1mm high, WSON is max. 0.8mm high! The rest seems to be equal ... Therefore we really do need the two footprint variants, as they should refer to different 3D models. Now the landpattern of the 4-fingered WSON looks like this: 
2018-02-18 08_12_28-tps63060 pdf - adobe acrobat reader dc
Here the EP is significantly different (size, 4 fingers ...).

Therefore we really need the two WSON-variants ... ! maybe the naming can also contain the EP-size like `_EP1.7x2.15mm`, although the question would be WHICH size of the EP ... over the fingers, the inner part only ... so I proposed the `_4EPFingers` as footprint-name modifier nomenclature ... In the end people will SEE the footprint in KiCAD, when they select it, so they know what they get and can compare to the datasheet.

To wrap this up, please rename back, so we can proceed!

aimylios:
@jkriege2 I'm confused. The TPS61200 is indeed available in two different package variants. Although the generic term "WSON" is used for the second variant, it is (as far as I can see) never called S-PWSON-N10, but DSC0010J. Therefore I still see no reason to have these complicated names. But I don't care that much, so I'll be quiet now and let you do your job. :-)

jkriege2:
Hmmm ... I'm really confused by TI by now too ... I though maybe we can use the package drawing names (DSC...) for naming, but it seems this is also not possible, because DSC is used for the 4-fingered and 2-fingered EP of a WSON (see screenshots above). DRC is used for the 2-fingered VSON-version, which equals the 2-fingered WSON, but with a different package height ... AAAaahhrrg!

So I think we should really stick to the `_EP8Fingers` (NOTE: I changed to 8Fingers, as 4 was the number of fingers on ONE side and it's better to have the total count ... 4 would be ambiguous, as 2 fingers on one side is 4 fingers).

Any more comments, or does anybody have any idea how to clear this up in a different way?

Resaj:
I think it would be better if we include DRC and DSC in the names. They appears in the datasheet and maybe people search for "standardized" name, like S-PVSON-N10 or DRC, for example.

I attach a summary:
img_20180218_132058

Each of this names includes "Texas_" as prefix and has a variant with "_ThermalPads" as sufix.

Then, we would have:
Texas_S-PVSON-N10_DRC
Texas_S-PVSON-N10_DRC_ThermalPads
Texas_S-PWSON-N10_DSC
Texas_S-PWSON-N10_DSC_ThermalPads
Texas_WSON_DSC0010J
Texas_WSON_DSC0010J_ThermalPads

Another idea:
Texas_DRC-S-PVSON-N10
Texas_DRC_S-PVSON-N10_ThermalPads
Texas_DSC_S-PWSON-N10
Texas_DSC_S-PWSON-N10_ThermalPads
Texas_DSC0010J_WSON
Texas_DSC0010J_WSON_ThermalPads

Maybe, someone can think "WSON" is very short and confusing, even the number of pins (10) is included in DSC0010J. So, we can change "WSON" by "WSON-10". For example:
Texas_WSON-10_DSC0010J
Texas_WSON-10_DSC0010J_ThermalPads
or
Texas_DSC0010J_WSON-10
Texas_DSC0010J_WSON-10_ThermalPads

Please, tell me the names we will use and, if it's necessary, I'll change all names in symbols, footprints, packages3D and packages3D-source.

jkriege2:
I would not include the DRC/DSC ... but ensure that they are in the tags and/or description. I would proposed this naming scheme:
- Texas_S-PVSON-N10
- Texas_S-PVSON-N10_ThermalPads
- Texas_S-PWSON-N10
- Texas_S-PWSON-N10_ThermalPads
- Texas_DSC0010J
- Texas_DSC0010J_ThermalPads

@evanshultz @poeschlr What do you think?

jkriege2:
@Resaj could you do the naming scheme I proposed?

Best,
JAN

Resaj:
I've renamed and changed the footprints basing the names you proposed. As we told, the DSC0010J is equal to S-PVSON-N10, even thought the datasheet specifies a little differences of footprint for each ones. So I decided to modify the existing S-PVSON-N10 footprint and make a copy for DSC0010J, with different description.

jkriege2:
Thanks!

JAN

PR #385, opened 2018-02-18
Updated the foot print

Misca1234:
LEM_LV25-P

http://www.lem.com/docs/products/lv_25-p.pdf

Made the following modifications 
Centered the REF**
Moved out the symbol name
Changed the holes more align with the data sheet
Redraw the lines

bild

bild


jkriege2:
Note: The silkscreen pin1-marker should be visible AFTER assembly, so please move it outside the silkscreen-outline

Also the pads 1 and 3 seem excessively large ... is tehre a technical reason? If not, I think 2mm diameter is sufficient for their 1mm holes.

Rest looks good, thanks for cleaning this up ;-)

JAN

Misca1234:
It is a current sense device so I thought some extra copper would make sure to keep the resistance low and the possibility to attach broad copper lines to it but I will make them smaller

Misca1234:
bild

jkriege2:
Hhmmm ... I see the argument ... the manufacturer does not say anything about the land-pattern ... in the end, the resistance should be mainly made up by the copper traces, not the copper around the pin, and I think a designer will have to ensure that's correct anyways, so smaller should be OK, or?

Misca1234:
I made them smaller and moved the pin 1 marker

jkriege2:
About the new pin1-marker ... why not simply a double-corner, as usually done, e.g.
2018-01-20 17_47_51-library browser package_csp d__kicad_kicad-footprints_package_csp pretty
2018-01-20 16_34_01-pcbnew d__kicad_devproject_devproject kicad_pcb
2017-11-12 10_29_35-footprint editor no active library
2017-10-21 15_13_38-footprint editor no active library

Misca1234:
Like this 

bild

jkriege2:
Gteat!

Thanks,
JAN

Misca1234:
I will re push this, the right side was not correct either

PR #386, opened 2018-02-18
moved right side

Misca1234:
Complement push to 
https://github.com/KiCad/kicad-footprints/pull/385

the right side was not at correct distance

bild


poeschlr:
As you don't have a mail address visible in your account and i do not know of a alternative communication channel i hijack this thread:

You are quite active lately (again). Would you be willing to help out with the lib maintenance? (We are clearly short on manpower.)

Misca1234:
@poeschlr 
Oh, i did not notice your post, but yo ugot in touch with me on a different thread

PR #388, opened 2018-02-19
Adafruit displays

Shackmeister:
3D model has also been made and will be coming shortly

Datasheet:
https://learn.adafruit.com/monochrome-oled-breakouts/downloads

Screenshot:
adafruit_ssd1306

------------


Shackmeister:
does anybody know what that error message is about?

poeschlr:
We might want to add a stacktrace output on error to our scripts. (Not sure what the problem is)

Shackmeister:
The Stepup parser can't handle when there are " inside the desc, I believe they are written as /" when they are not supposed to be parsed

evanshultz:
Usually I see the escape character being a backslash, so `"` would turn into `\"` to result in a quote mark after parsing.

poeschlr:
Is there some better documentation out there that contains all measurements or do i really need to open the eagle files for that?

You currently have a cortyard clearance of 0.5mm i think this does not fall under one of the exceptions that would need such a large clearance. So reduce it down to 0.25mm

For reference the drawing i created from the footprint:
screenshot from 2018-05-08 18-18-18

Shackmeister:
In the link I provided you can find an image with the dimensions, they are unfortunately given in inches :(
lcds___displays_fabprint

poeschlr:
Looks good, sorry for forgetting about that for so long.

Shackmeister:
Thanks :) Ill get around to PR The 3D model soon (hopefully) :)

PR #392, opened 2018-02-20
Added LED_Rectangular_W7.62mm_H4.55mm_P5.08mm_R3

MartinStej:
Added rectangular LED (7.62x7.62mm) used mainly at automotive
Datasheet can be found at https://www.gme.cz/data/attachments/dsh.511-795.1.pdf

led_1

Pull request for 3D model is at [github](https://github.com/KiCad/kicad-packages3D/pull/274)
Known errors:
**Violating F7.3** - Since cathode is connected to 2 pins number "1", this warning appeared. But it should be OK
**Violating F9.3** - Not actually sure why.


MartinStej:
Hi,
 thanks for comment. The 3D model offset is here due to footprint anchor (respective to pin 1). I updated footprint and 3D model.

led_footprint
led_2

Thank you
  Martin

jkriege2:
Also please make a standard pin1-marker. We usually have a double-corner on silkscreen, instead of a cut-away corner, e.g.: 
2018-01-27 11_01_30-footprint editor no active library

MartinStej:
Thanks for explanation. Fixed.
I was also thinking about silk around pins, but I'm not quite sure about that.

jkriege2:
Thanks,
JAN

MartinStej:
Thank you for patience :+1:

PR #398, opened 2018-02-27
Add L_Pulse_PA4320

strombom:
Pulse_PA4320
http://productfinder.pulseeng.com/products/datasheets/P787.pdf

image

3D Model: https://github.com/KiCad/kicad-packages3D/pull/264

------------


poeschlr:
For the adheisive i would place a adheisive only pad instead of drawing many overlapping circles. (Circular smd pad of the required size with: No pin number, no copper, only f.adheisive ticked)

This would also be the first component in the lib to have anything on the adheisive layer.
So could you give a reasoning why this component needs adheisive. (You seem to know more about that then i do.)
For me it would be enough if you tell us that you had problems with this component in production and adheisive fixed it. Better would be a general document from which we could get rules for when adheisive should be added. (Could then be added to the KLC)

---

Otherwise the footprint looks good.

strombom:
Many of the Wurth inductors in the library has adhesive, that's what I copied. I wouldn't mind removing the adhesive, should I?

poeschlr:
I think changing it to a pad instead of overlapping circles might be a good idea. At least for new contributions.

evanshultz:
I've used adhesive for parts (especially bottom-side parts) that were heavy (like this one) and wouldn't stay on the board through double-sided reflow without glue. Or some factories that always glue everything. Or in a few hi-rel applications.

poeschlr:
@strombom do you have time to fix the issue regarding the adhesive layer?

Misca1234:
Hi

This one is repushed at 
https://github.com/KiCad/kicad-footprints/pull/792

poeschlr:
Replacement pull request has been merged.

PR #400, opened 2018-03-02
QFN56 8x8mm, P0.5, EP4.8x5.5mm

calebns:
QFN56RGQ

http://www.ti.com/lit/an/scea032/scea032.pdf

image

------------


evanshultz:
Review notes.
- F.Fab lines should be 0.1mm wide; F.FilkS should be 0.12mm wide.
- Silk should not be under the part, so make the silk offset from fab lines 0.11-0.12mm.
- For best compatibility with square QFN parts, can you make the pin 1 mark like the other 3 corners but missing the vertical line by pin 1?
- The paste should be shrunk on the outer pads. It's by an unequal amount, but surely shrinking paste globally by 0.05mm would be OK.
- Pad 57 has a custom paste margin but no paste layer. You can remove this setting.
- One paste pad is 1.14x2.226; it should not have the 6.
- All paste pads are at a X location of +/-1.2mm and they should be +/-1.26mm.
- Some paste pads are at a Y location of +/-0.7mm and they should be +/-0.72mm.
- The other paste pads are at a Y location of +/-2mm and they should be +/-2.16mm.
- The center pad is the wrong height.
- Courtyard should be at 4.43mm (0.25mm clearance from body and copper).

Shackmeister:
Hi @calebns 
Would have some time to fix up the changes, so we can get the footprint merged? :)

poeschlr:
I might script this footprint if this is ok for @calebns

calebns:
If you could script it that would be great. I don't have a time to continue work on it at this moment.

poeschlr:
scripted alternative has been merged

PR #401, opened 2018-03-02
ST_SOT-23-6L added

Resaj:
Footprint for ESDA6V1BC6.
[Datasheet](http://www.st.com/content/ccc/resource/technical/document/datasheet/21/07/21/e3/a8/df/42/a2/CD00001906.pdf/files/CD00001906.pdf/jcr:content/translations/en.CD00001906.pdf)
imagen


jkriege2:
Is this different from the existing SOT-23-6 (since you made this manufacturer-specific)?

Resaj:
Yes, the footprints are different. I've reviewed both footprints and I think it's better to include the SOT-23-6L of ST, even thought the measures between SOT-23-6 and SOT-23-6L are very similar.

jkriege2:
Except the pad size (existing: 1.06x0.65, new: 1.2x0.6) I don't really see the difference ... can you elaborate it a bit?

Resaj:
That's the difference. I don't know why ST added a variant with "L" sufix...
But now people can may be confuse about the names.

If you prefer, we can suppress this new footprint and use SOT-23-6 instead of SOT-23-6L.

jkriege2:
I think that the difference is so minor that we can see it as "noise" of how different manufacturers do the footprint for that package. I think that ST added the L just to indicate 6 is the number of "leads" ... not to really set it apart from a "standard" SOT-23-6.

So I would simply use the standard footprint.

@evanshultz @poeschlr What do you think on that?

evanshultz:
Yep. Insignificant. No need to make another footprint.

Resaj:
Ok, thanks for the advices.
I'll close the PR and change the footprint reference to SOT-23-6 in ESDA6V1BC6 symbol.

PR #403, opened 2018-03-04
Hdsp led

Misca1234:
This is the complementary push of the HDSP 7 segments display as mentioned in 
https://github.com/KiCad/kicad-symbols/pull/311

HDSP-A401
HDSP-A151 alias HDSP-7501 
HDSP-7401
HDSP-7801
HDSP-A403
HDSP-A153 alias HDSP-7503
HDSP-7403
HDSP-7803
HDSP-7507
HDSP-7807
HDSP-7508
HDSP-7808

https://docs.broadcom.com/docs/AV02-2553EN

Symbol push
https://github.com/KiCad/kicad-symbols/pull/346

foot print push
https://github.com/KiCad/kicad-footprints/pull/403

3D model push
https://github.com/KiCad/kicad-packages3D/pull/276

3D model source push
https://github.com/KiCad/kicad-packages3D-source/pull/109

bild

bild

bild

bild

bild

bild


jkriege2:
I think you're missing some kind of pin1-marker on F.SilkS

Best,
JAN

Misca1234:
Ok, I add it, i just copied the old one

Misca1234:
They all look like this now 

bild

jkriege2:
Generally the footprints look good now. But I spotted some problem with them:
1. All have a second pad (without pad number) overlayn to pad 2 @ (0, 2.54)mm
2. The hole diameter seems fine, but why did you choose such a strange non-rounded number 0.70104, instead of simply 0.7? Do you edit in inch? 
3. The pad size seems a bit large. I think 2*diameter should be fine, which would be 1.4mm instead of 1.9, which it is now. 

for 2&3: Maybe you can use the same pads as the standard DIP-packages, which would hole-diameter=0.8mm and pad-diameter=1.6mm

Could you fix those?

Thanks,
JAN

Misca1234:
they are copies of the orignal HDSP footprint, which, i guess, was made in some strange non metric system, let me clean up them

Misca1234:
They do look like this now 

bild


The double pad is removed

jkriege2:
THanks & nice work!

JAN

PR #404, opened 2018-03-04
added DO-201AE footprints

jkriege2:
- dimensions: http://www.farnell.com/datasheets/529758.pdf
- e.g. used for 1.5KE TVS-diodes (see https://github.com/KiCad/kicad-symbols/pull/348)
- script-generated: https://github.com/pointhi/kicad-footprint-generator/pull/116
- 3D models: https://github.com/KiCad/kicad-packages3D/pull/277 & https://github.com/KiCad/kicad-packages3D-source/pull/110

2018-03-04 12_10_52-pcbnew d__kicad_devproject_devproject kicad_pcb
2018-03-04 12_10_19-3d viewer

------------


evanshultz:
@poeschlr @jkriege2 
Anybody know why Travis isn't finishing on this PR? I've refreshed the tab several times and the last commit was hours ago.

poeschlr:
I logged into travis. For some reason it might not have noticed the last commit. (So it seems there was a communication problem between the travis server and github.)

evanshultz:
Just a couple things (applies to all footprints):
- There is a double comma in the description; perhaps a field isn't getting populated in your script?
- Do we want the math in the description or just the result?
- Maybe better to write length and diameter separately instead of together?
- There's a double space in the keywords
- With all the info in the description, should the keywords be much shorter and not the same info again?
- I believe "cathode" is the English term; "kathode" is used in Germanic European languages

jkriege2:
Hi!

about the description/tags ... I can see to a clean-up.

About the K/C-question: I decided for K, as we use that as common pin-name for a cathode, also to separate it from a Common-pin, present in some multi-diode-symbols ... 

Also: Note that I just extended the existing scripts, so these symbols look like (and have the description of) all other THT-diodes. So maybe we can also merge this and then resolve the other issues for all THT-diodes together?

cu
JAN

evanshultz:
@jkriege2 
Sure! We can address the other items later.

Since "C" is the ref des prefix for capacitors, but it's also for common cathode, I see a problem. I still favor "cathode" with a "c" since that's the correct English word, but more feedback can be gathered later. Thank you!

evanshultz:
@jkriege2 
Oh! Will you please make an issue to capture the necessary changes for all footprints? Then I'll merge.

PR #389, opened 2018-02-19
added Wuerth_61400826021USB connector and fixed minor issues

Shackmeister:
changed all "manf-mpn" to "manf_mpn"
removed all generic footprints

Datasheet: http://katalog.we-online.de/em/datasheet/61400826021.pdf
screenshot:
wuerth_usb

------------

Thanks for creating a pull request to contribute to the KiCad libraries! To speed up integration of your PR, please check the following items:

- [ ] Provide a URL to a datasheet for the footprint(s) you are contributing
- [ ] An example screenshot image is very helpful 
- [ ] If there are matching symbol or 3D model PRs, provide link(s) as appropriate
- [ ] Check the output of the Travis automated check scripts - fix any errors as required

PR #406, opened 2018-03-04
Added Broadcom_AFBR-16xxZ

Misca1234:
Foot print for


Misca1234:
Not sure if the suffix vertical, horizontal and tilted should be used or if the names in the data sheet should be used?
Instead of 
Broadcom_AFBR-16xxZ_Vertical.kicad_mod
it should be 
Broadcom_AFBR-x624Z_AFBR-1629Z_AFBR-2529Z

Shackmeister:
looks like we have some duplicate work
https://github.com/KiCad/kicad-footprints/pull/281

yours does however look closer to merge

poeschlr:
I am a bit unsure of the use of circles in the courtyard layer. It is hard to get a closed "polygon" that way. (Not yet a problem but when the developers add a courtyard violation check it might be.) Have a look at https://github.com/KiCad/kicad-footprints/pull/389 for more details.

Misca1234:
Yes, I remade the symbol, the original push one seems to have been abandoned so I did a takeover.
Shall I removed the circles in the courtyard layer and use straight lines instead ?

poeschlr:
@Misca1234 yes i think approximating the arcs with straight lines has a higher chance of getting a closed polygon for the newly added courtyard check. (Unless we are sure we fulfill the 0.01mm tolerance that is required by the polygon closing algorithm.)

Shackmeister:
Hi @Misca1234 can you please fix the courtyard? it's not closed at pretty much all the arc's

Misca1234:
Yes, give me a day or so

Misca1234:
I will change the courtyard but I was thinking 

Because I am remaking a abounded push, should we use the suffix vertical, horizontal and tilted or should we use the names in the data sheet ?
Instead of, current name 
Broadcom_AFBR-16xxZ_Vertical.kicad_mod

should we instead use ?
Broadcom_AFBR-x624Z_AFBR-1629Z_AFBR-2529Z

poeschlr:
I think the current name is more readable and especially easier to understand. 

I am not strictly against the later but there i would still include Vertical. To make it easier to read it might be better to have 3 footprints in that case.

Misca1234:
Something like this ?

bild

bild


bild

bild

bild

Ratfink:
You seem to have git set up wrong and have been pushing lots of branches directly to the KiCad repository rather than your own fork.  Please fix this, it's created a lot of clutter.

Misca1234:
jeee, i pushed to upstream instead of origin

poeschlr:
A tip: you can make upstream pull only by setting the url for push to something nonsensical. (I would suggest doing this everywhere where you have the official repo added as you have direct write access!)

Example: `git remote set-url upstream --push (no-push)`

Misca1234:
That is a good idea, I am looking into kicad-footprint-generator at the moment and could start there

Shackmeister:
@Misca1234 how is the Git stuff coming along? :) the footprints seems very nice and ready for merge

Misca1234:
Well, I am gitting away with origin nowdays, but to keep you guys on your toes I will now and then use upstream

> the footprints seems very nice and ready for merge

I think there will be a slight increase in the email traffic if I merge my own pushes

PR #409, opened 2018-03-05
Move DIN41612 type B family

evanshultz:
Original PR for old lib is at https://github.com/KiCad/Connectors_IEC_DIN.pretty/pull/7. This covered the KLC violation for origin.

While moving, I made the following changes:
- Renamed according to KLC rules (I hope)
- Updated description ("family" -> "type", "angled" -> "horizontal", etc.)

Since they look the same as the old footprints I didn't include a screenshot.

I plan to update the script to capture the changes above and generate more DIN41612 footprints soon if I have time. The other DIN41612 footprints, for type C, are very old and are best left behind and replaced with new script-generated footprints later (I think).

@poeschlr @jkriege2 
I forgot the mention that I looked at the TODO docs in both the symbol and footprint libraries. It seems this is the only thing left. Any other empty checkboxes have some issue.

------------


evanshultz:
@poeschlr @jkriege2 @Ratfink @Misca1234 
Can someone review and merge this? It would be nice to include this in v5, if possible, since I believe it's the only thing that's currently missing from the old repo. And also if any issues with this are resolved it will hopefully inform the script for adding more connectors in this family.

Ratfink:
I'll do a real review of this tomorrow.  I have just one note until then: it looks like this PR adds a new library, so please add it to the fp-lib-table as well.

evanshultz:
@Ratfink 
Thanks! Good catch, I just fixed that.

I'd rather get this merged than make any marginal changes so that least no footprints are missing in v5 (assuming we can still make it). So if there are improvements but no real problems, perhaps list them here for me or go ahead and create a case and someone (me?) can address it in a future PR. Thanks!

Ratfink:
Overall, this looks good.  One thing I wonder before this can be merged: should `DIN41612_3B_2x10_Horizontal` be called `DIN41612_B3_2x10_Horizontal`?

Other than that, I notice that the `_Vertical` parts seem to be missing their fab outlines.  Given that this PR is mainly moving old footprints to the new repository though, I don't think this has to be fixed before merging.

evanshultz:
Good catch! Renamed that one footprint and ready to merge. The script is now merged so the updates can begin!

PR #414, opened 2018-03-06
QFN-32 for Nordic chips

jcyrax:
Footprint for nRF52810-QCxx symbol.
http://infocenter.nordicsemi.com/pdf/nRF52810_PS_v1.1.pdf
image
image
image

The source I used is demo board gerbers, but with larger EP.
My EP is 3.55x3.55 I got using IPC calculation formula from here:  https://github.com/KiCad/kicad-symbols/pull/340#issuecomment-370405440
Nordic gerbers use 3.5x3.5mm EP.
EP of package is 3.5 (+/-) 0.1

Nordic comments on footprints: https://devzone.nordicsemi.com/f/nordic-q-a/19860/qfn-land-pattern-in-reference-design/77244#77244

------------


poeschlr:
The footprint is missing its polarity marker on the fab layer. I would suggest to chamfer the top left corner of the fab rectangle.

Use paste only pads for the paste layer instead of graphical polygons. (A paste only pad is a smd pad without a pin number, no copper, only paste selected.) See any other footprint with an exposed pad.

The exposed pad might be a bit small. Last time we looked into it we found that the maximum size of the "lead" on the package is a good approximation for the pad size suggested by IPC standards. (In this case 3.6x3.6mm)

jcyrax:
I see that you created QFN-32 for Nordic in https://github.com/KiCad/kicad-footprints/pull/518
Pads are a bit different, but I guess this PR is not needed then?

PR #416, opened 2018-03-07
Rename MLP14D→WQFN-14, add a WQFN-16

Ratfink:
This is the footprint companion PR to https://github.com/KiCad/kicad-symbols/pull/359.

ON Semiconductor renamed Fairchild's MLP14D package to WQFN-14, and that's totally okay of them.  This PR updates the name and documentation, and fixes up the footprint for modern KLC.
Datasheet: http://www.onsemi.com/pub/Collateral/510BR.PDF
screenshot from 2018-03-07 15-54-51

This PR also adds the WQFN-16 footprint, needed for the FUSB307B.
Datasheet: http://www.onsemi.com/pub/Collateral/510BS.PDF
screenshot from 2018-03-07 15-55-05

------------


poeschlr:
I didn't know about the standard clearance either ;)
My scripts use `fab_line_with/2+silk_line_width/2` which results in a clearance of 0.11 (So if we add a standard clearance it should allow some range as otherwise most connector footprints will be non compliant again.)

I would use the statement that it is allowed between what i used (fab and silk lines are touching) up to the outer silk line edge does touch the inner edge of the crtyd lines.

evanshultz:
Thanks!

I used the word "standard" since I saw 0.12mm used in scripts and hand-made footprints everywhere when I started. And this gives some gap to and ensure silk isn't under the body of the part. Note that registration between layers varies so if we have the lines touch there's no guarantee that they won't actually overlap in the real board. But even 0.12mm doesn't guarantee that. 0.11mm does ideally avoid overlap and you, Rene, re-made all (or at least many) connector footprints by script since I last submitted a connector script and I haven't seen your calculation.

Eh. I didn't say the old Fairchild footprint name was unnecessary, just noted it was there.

Keywords are kinda mystifying. I see all different stuff in there and don't have a feel for what is required or even common.

So annoying when the lines get re-ordered by making a small change and then saving a footprint...

PR #418, opened 2018-03-10
Add SK6805

fauxpark:
This is the "NeoPixel Nano", another addressable RGB LED that I must have missed.

Datasheet: https://cdn-shop.adafruit.com/product-files/3484/3484_Datasheet.pdf
Symbol PR: KiCad/kicad-symbols#365

fp

------------


Ratfink:
Thanks for the contribution!

I have a few suggestions to make this footprint better.  It looks like this footprint is based on some other similar LED footprints which have a few problems of their own, so it inherited those problems.

* The silkscreen needs work.  The lines are a bit far from the part body, and the pin 1 mark is odd.  It looks like the line by pin 3 is supposed to be a pin 1 mark, so please move that to pin 1 and remove the "1" text.  See [F5.1](http://kicad-pcb.org/libraries/klc/F5.1/) for reference.
* Similarly, the fab layer outline needs to be fixed.  The cut-off corner is the right size, but in the wrong corner.  Please move it to the upper left, and make the rectangular outline not continue past the diagonal line.  Also, the fab layer refdes should be moved to the center of the footprint and shrunk to fit within the fab layer drawing.  See [F5.2](http://kicad-pcb.org/libraries/klc/F5.2/) for reference.

poeschlr:
The polarity mark for these has been discussed in some older pull request. (See: https://github.com/KiCad/kicad-footprints/pull/343)

The reasoning for doing it this way is that the manufacturer seems to mark pin 3 on their package. This is why we have a 1 next to the real pin 1 such that the user is alerted to the fact that something fishy is going on here. (We even asked at the user forum back then. The feedback was inconclusive. Nobody was really opposed to this solution.)

Ratfink:
Yes, I just saw that after submitting the comment above.  I'm working on updated recommendations as we speak.

Ratfink:
Okay, updated recommendation time!

* The marked pin on this package is pin 2 according to the datasheet, so please move the marks from pin 3 to pin 2.
* Please shrink the refdes on the fab layer and center it in the part body, as per [F5.2](http://kicad-pcb.org/libraries/klc/F5.2/).

fauxpark:
Done. I also did point 2 for the other related footprints I previously added.

Ratfink:
Looks good, thanks for fixing the other footprints as well!

The refdes mark on the SK6805 footprint is still at pin 3.  If you move that to pin 2, I think this will be ready to merge.

fauxpark:
Argh, thought I had that.

Ratfink:
Thanks for the contribution!

PR #419, opened 2018-03-10
Added WQFN 24 pin pacakge

Duckle29:
I added a footprint for WQFN packages with 24 pins and a thermal pad with vias

------------

Datasheet http://www.ti.com/lit/ds/symlink/bq25601.pdf

image


Duckle29:
I'm Mikkel Jeppesen, I don't know why there's two contributors on my commits.

As for the travis, I really think the thermal vias in the center pad shouldn't have the mask on them, as the front is exposed by the larger square bad, and having the mask on the back, limits solder wicking when reflow-soldered.

Ratfink:
Thanks for the contribution!  It looks like you had the wrong email address set on two of your commits, which confused the CLA assistant.  We see that you've signed the agreement, so we can ignore its complaint safely. 🙂

I see a few problems with this footprint as it is currently.  I know this looks like a lot, but you're really off to a good start!

* There is currently quite a bit of silkscreen under the chip body, which is against KLC rule [F5.1](http://kicad-pcb.org/libraries/klc/F5.1/).  Please move the silkscreen marks outside of the outline on the fab layer.  Also, please use a pin-1 mark like the other QFN packages have (a missing line next to pin 1 - not the whole corner missing, just the line next to pin 1).  Further, it would be good to make the silkscreen lines 0.12 mm wide instead of 0.1 mm wide, as 0.12 mm is generally the preferred line width for silkscreen.
* There are a few problems with the fab layer outline as well.  See KLC rule [F5.2](http://kicad-pcb.org/libraries/klc/F5.2/).  Please mark pin 1 on the fab layer outline as the rule describes (right now the corner bevel is too small).  Also, please make your second reference designator a bit smaller - right now it's running into the outline.
* The courtyard is too close to the pads (see [F5.3](http://kicad-pcb.org/libraries/klc/F5.3/).  It should have a clearance of 0.25 mm to the pads and the part body, but right now it's only 0.2 mm away from the pads.
* Please center the soldermask reference designator over the part.
* The pins are numbered wrong, I believe.  The datasheet's package drawing shows a view of the bottom of the package with the pins numbered.  It looks like you interpreted that as a through-view and numbered the pins on the footprint clockwise instead of counter-clockwise.
* The footprint name is wrong for the EP size.  It should be `EP2.7x2.7mm`, but now it says `EP2.6x2.6mm`.
* You're right about the thermal via Travis errors not being a problem.  Please fix the other warnings it gave, though.

One other thing: it would be great if you'd add a non-ThermalVias version of this footprint.  The datasheet says that its thermal vias are just a suggestion, so some people might want to place them differently.

Duckle29:
@Ratfink I did the required changes, it's odd that it's complaining about pad 21 not having mask, and pad 21 being tht, it's not. I even checked the file in a text editor.

Ratfink:
Looking good!  A few more things:

* Sorry I didn't notice this before, but the recommended thermal vias in the datasheet are 0.3 mm diameter, you have 0.4 mm.
* Please keep the height and width of the fab layer refdes the same as each other, and keep the thickness 15% of that value.  So, since you changed the width to 0.8 mm (which looks like a good size), please change the height to 0.8 mm and the thickness to 0.12 mm.
* The silkscreen still needs a bit of work.
  * Please set the silkscreen lines' width to 0.12 mm.
  * The lines could be moved a bit closer to the chip body too, just as long as there's never any silkscreen under the chip when it's installed properly.  Given the tolerances on this chip's dimensions, I'd suggest placing the lines 2.2 mm away from the axes rather than the 2.3 mm you have now.
  * The lines should be extended a bit closer towards the pads.  Once the widths are updated as suggested above, I believe you could extend them another 0.05 mm without exceeding the clearance in F5.1.
* The value label of the non-ThermalVias footprint doesn't match the filename.  Please fix that.
* Please set the 3D model filenames according to [F9.3](http://kicad-pcb.org/libraries/klc/F9.3/).  The one you chose is close to correct, but not quite.  It would be preferable to use a 3D model filename that matches the footprint filename so a proper model can be added without having to change the footprint again.

As for the complaints about pad 21, those are for another footprint.  It looks like KiCad decided to reorder some attributes in a couple footprints you didn't modify.  We can safely ignore those errors, just like the ones for the ThermalVias in your footprint.

Duckle29:
Cool man. yeah the 0.4mm vias was because I initially was just making the footprint for personal use, and 0.4 is usually the smallest the cheap pcb houses will accept. I forgot about it when I decided to make it KLC compliant.

As for the 3D model. Should I copy the QFN model and name it WQFN instead, or should I model a new one based on the datasheet? I don't really know how different it'll look.

Thanks for spending the time making sure the libraries are stock libraries are of good quality :)

Ratfink:
The 3D models for QFN packages are generated by [a script](https://github.com/easyw/kicad-3d-models-in-freecad/tree/master/cadquery/FCAD_script_generator/QFN_packages), so it would be best to add dimensions for this part to that script, then submit the generated model to https://github.com/KiCad/kicad-packages3D.  We can merge the footprint before the 3D model though, as long as you have the correct filename set (they should be `${KISYS3DMOD}/Package_DFN_QFN.3dshapes/WQFN-24-1EP_4x4mm_P0.5mm_EP2.7x2.7mm.wrl`).

It looks like you only made the changes I suggested above to the ThermalVias footprint.  Please update the non-ThermalVias one to match.

Also, the value fields are still wrong: the ThermalVias one says "EP2.6x2.6mm", which is the wrong size; and the non-ThermalVias one says ThermalVias.

It looks like you can ignore the Travis warning about "Attribute clearance == outline != 0!".  I believe that's a bug in the script.

You're welcome for taking the time to review this, thanks for putting up with how picky I am! 😉

Ratfink:
Almost ready!  Please just fix the 3D model filenames (they should say `EP2.7x2.7mm` where the currently say `EP2.6x2.6mm`) and I'll merge this.

Duckle29:
Done :) Thanks for the help

Ratfink:
Thanks for the contribution!

PR #420, opened 2018-03-10
Set Travis to use Python 2.7

Ratfink:
The default language is Ruby, which causes Travis to waste about four seconds setting up a Ruby environment we don't need every time CI is run. This PR changes the language to the one we actually use, Python 2.7.

Like https://github.com/KiCad/kicad-symbols/pull/370, but with a cleaner commit history because I already figured out what needs to be done.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,6 @@
+language: python
+python:
+  - "2.7"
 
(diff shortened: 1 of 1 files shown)


poeschlr:
thanks

PR #421, opened 2018-03-10
KiCad logo on copper and on solder mask. Height 5mm.

4x1md:
KiCad logo on copper and on solder mask without "DESIGNED WITH" caption.

Attached a photo of a PCB with KiCad logo on copper covered with solder paste.

kicad_symbol_copper


4x1md:
The same logo on a green PCB.

Misca1234:
Cool, thx
The copper one is nice, diffucult to remove without making marks

PR #422, opened 2018-03-11
added Mounting pads numbers in connectors

Shackmeister:
as discussed here: https://github.com/KiCad/kicad-footprints/issues/117
The only connectors I believe are missing is Hirose DF13C @poeschlr is it due to the keepout zone it's not in your "Connector_SMD_single_row_plus_mounting_pad" script?
mp

------------


Shackmeister:
fixed it for DF13C, please let me know if I missed any

poeschlr:
I thought about adding keepout support to the single row generator. It was a lot easier to make a separate generator for them. (Adding it to the large generator would have required a lot of work as it does not make sense to have a keepout generation specific to one series in it. I would have needed to add some very generic support which was just not worth it for only one series of connectors.)

poeschlr:
Could you link the pull request for the script?

poeschlr:
Some 3d models will also require renaming. (molex picoblade for example)

Shackmeister:
1: Makes sense
2: I typically wait for some feedback before making the PR, but maybe you would have some feedback for the scripting it self. Ill do it in a bit or tomorrow
3: Ill ofcourse also do the renaming of the 3D models if we agree on the naming

poeschlr:
For such a large change it is easier and faster to look at the script instead of at the 300+ changed files ;)

Shackmeister:
PR for script is here: https://github.com/pointhi/kicad-footprint-generator/pull/118

poeschlr:
Looks good to me. (Both the script and the files. I made a random look at some of them and all seem only to change stuff that is ok. For some reason the text positions changed but this might be because i updated some central part of the script after originally committing the footprints.)

I am not really conferrable merging it my self though as i am quite invested in it. 
Maybe @evanshultz, @jkriege2 or @Ratfink can take a quick look at both our pull requests. (If not i will merge this sometime next week after i take another look.)

evanshultz:
I took a quick look at the footprint change and scripts and it looks OK to me as well. I could be missing something without really using the scripts, though.

I only see Travis errors because footprints are not centered. Is this violation on purpose? https://www.molex.com/pdm_docs/sd/430451010_sd.pdf is a datasheet for the Molex Micro-Fit 3.0 horizontal connector. I don't see why this one couldn't be vertically centered. Perhaps because the center is on the plastic part and the leads sticking out make it unbalanced?

One question: `conn_hirose_df13c_smd_top.py` and `conn_molex_micro-fit-3.0_smd_side_dual_row.py` and `conn_molex_micro-fit-3.0_smd_top_dual_row.py` all contain `mounting_pad = "-1MP"`. So does `smd_single_row_plus_mounting_pad.py`. Would it be better to build that string up instead of hard-coding it? `conn_config_KLCv3.yaml` contains `mounting_pad_number: "MP"` and then the mounting pad quantity could be calculated. I think it's fine to merge, just wondering. Maybe it makes it easier if we ever had one with multiple different mounting pads? Or maybe useless now and can done done later if ever needed.

All told, I'm fine to merge.

poeschlr:
If you think the footprints are ok, don't forget to also merge the symbols that go with this:
https://github.com/KiCad/kicad-symbols/pull/372

Issue #46, opened 2017-11-16
Testing and travis ideas


antoniovazquezblanco:
I had a look to the kicad library utils repo the last weekend. I have some ideas that I am willing to contribute to but I don't know what you think about them.

1. Testing now requires cloning more than one repo (travis clones both this repo and the utils one). I think this could be avoided by moving the testing code to this repo. The reused code between the different testing benches for PCB, SCH and SCHLIB is minimal, thus being a good option. This would reduce check time. Keeping the test code in the same repo would also make easier point number 2.

2. Using relative paths in scripts would ensure that scripts can be run both localy and in travis environment. Currently some scripts can only be used by travis. My proposal is to include a batch and a shell script that the user can click or execute in a terminal that will test the last modified components (I was thinking on testing the whole lib if in the master branch and the modified components if in another branch or something like that).

3. Testing now requires sudo and there seems to be no obvious reason for that. Sudo makes travis testing significantly slower. Dropping sudo needs would improve check times a lot.

4. I was thinking in including a coverage report. A simple script that would tell in which percentage the whole lib complies with KLC and/or another script that would test how many 3D models are present and how many missing footprints looking at the symbol lib.

Please let me know what you think about the previous ideas and if you would like to see them implemented.

Thank you!

SchrodingersGat:
1. I don't want to separate out the scripts that are in kicad-library-utils
   a. There is common code that would have to be duplicated
   a. It's a good central place to store the scripts
   a. I think it would clutter the repo which should be reserved for only library files (as far as possible)

2. If you can come up with a scheme that allows this, that's fine, but I don't want to break our current Travis functionality.

3. I didn't know that. That can be dropped from the .travis.yml file I think

4. This is a good idea, I have toyed with this idea before. It would be great to add this output to the https://kicad.github.io website, too! Green indicator for 100% pass, for e.g.

Shackmeister:
+1000 for 4.

antoniovazquezblanco:
Discarding 1 and 2. Opening issues where it belongs for 3 and 4.

PR #424, opened 2018-03-12
Remove sudo from .travis.yml

Ratfink:
There were [claims](https://github.com/KiCad/kicad-footprints/issues/46) in the past that the `sudo` in the Travis CI script was needed for some unknown reason.  I just did some testing on my own fork of kicad-footprints and found that it made absolutely no functional difference whether the `check_all.sh` script was run with `sudo` or not.  Not using `sudo` also gave a decent speedup (see [below](https://github.com/KiCad/kicad-footprints/pull/424#issuecomment-372709337)).  This PR removes it, running the `check_all.sh` script as a normal user.

Fixes https://github.com/KiCad/kicad-library-utils/issues/177.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,5 +6,5 @@ before_install:
 - git clone --depth 1 https://github.com/KiCad/kicad-library-utils /home/travis/build/kicad-library-utils
 
 script:
-- sudo bash /home/travis/build/kicad-library-utils/pcb/travis/check_all.sh $TRAVIS_BUILD_DIR -v
+- bash /home/travis/build/kicad-library-utils/pcb/travis/check_all.sh $TRAVIS_BUILD_DIR -v
 - python /home/travis/build/kicad-library-utils/check_lib_table.py $TRAVIS_BUILD_DIR/*.pretty --table $TRAVIS_BUILD_DIR/fp-lib-table


antoniovazquezblanco:
Thanks @Ratfink 

Please let me know if checks run faster after this change :)

Ratfink:
Comparing two Travis builds with and without this patch, it looks like there's a decent difference in execution time.  To be fair, with this patch also has the language set to Python instead of Ruby, which saves about 4-5 seconds, but it still looks like there's a decent speedup beyond that.
[With](https://travis-ci.org/Ratfink/kicad-footprints/builds/352502561) (29 s)
[Without](https://travis-ci.org/KiCad/kicad-footprints/builds/350630179) (45 s)

antoniovazquezblanco:
I thought so. Giving root permissions means that travis has to setup a limited access environment which seems to take quite some time (~10s).

Thanks for the feedback!

Ratfink:
Could someone please take a look at this PR?  By my testing, it speeds up CI significantly and hurts nothing.

evanshultz:
I don't feel I'm the right person to do that. I would suggest @SchrodingersGat, if he's available, and/or @poeschlr. Sorry to not be helpful. :(

FYI, I didn't see that anyone claimed sudo was required in your link above. I only see Oliver saying that he wasn't aware that sudo isn't required. Probably there is more discussion elsewhere.

SchrodingersGat:
Y'all better not break anything if I merge this... ;)

PR #428, opened 2018-03-13
Connector_Harwing: Generate M20-89 connectors

antoniovazquezblanco:
Re push of https://github.com/KiCad/kicad-footprints/pull/368

Datasheet: https://cdn.harwin.com/pdfs/M20-890.pdf

imagen


evanshultz:
Please add a screenshot and datasheet link for ease of review and to keep all relevant info together. From the old PR I found the product page, and there the datasheet is at https://cdn.harwin.com/pdfs/M20-890.pdf.

Also, it's "Harwin" and not "Harwing". :)

antoniovazquezblanco:
Sorry, I was being lazy haha.

> Also, it's "Harwin" and not "Harwing". :)

Keeps happening to me 🤦‍♂️

evanshultz:
Script is at https://github.com/pointhi/kicad-footprint-generator/pull/113.

Review notes:
- The family is M20-890xx, with xx being the pin count. You're missing a zero for pin counts <10. The image above should be `Harwin_M20-89005xx_1x05_P2.54mm_Horizontal`. This is line 20 of your script. I didn't test this but I believe you can change `'M20-890{pincount:d}xx'` to `'M20-890{pincount:02g}xx'.format(pincount=pincount)`.
- Same issue in description.
- Keywords say "top entry" but it's not.
- You're not generating the 20 pin version because the upper argument to range() is not inclusive on line 187 of your script.
- I wonder if it would be better to put the pin 1 mark inside the pin instead of outside? Would it be less likely to have components between the pins? Or for that matter, the pin 1 mark could be smaller since the header can't really be put in backwards.
- Please adopt the style of pin headers in the library, merged at https://github.com/KiCad/Pin_Headers.pretty/pull/24#issuecomment-312755357, which differs in the pin 1 silk mark under the pin, among other things.
- Compared with our current pin headers, you have pin 1 on the opposite side. I feel it would be best to orient them in the same way. I don't see any pin mark in the Harwin documentation so apparently we're free to select pin 1 as we like?
- Your pin length from the plastic back to the pads is 4.2mm, not 4.7mm.
- How did you decide on the pad placement with respect to the pins? I assume you centered it on purpose? The flat part of the lead touching the PCB is 1.8mm long, so the 2.5mm long pad should be centered on it? So the pad centers should be 3.8mm behind the plastic (currently they are 3.3mm)?
- Your silk is only 0.06mm offset from the fab lines. Can you copy the offset from the PR above (should be 0.11-0.12mm).

Here is the dimensional check:
image

evanshultz:
Forgot to check courtyard clearance... most of them are a bit bigger than the normal 0.5mm for connectors:
image

evanshultz:
@antoniovazquezblanco 
Will you be updating your script and submitting new footprints?

antoniovazquezblanco:
Yes, but not very soon

evanshultz:
@antoniovazquezblanco 
Will you be able to get back to this soon?

antoniovazquezblanco:
I will try to do something about it this week.

antoniovazquezblanco:
imagen

imagen

evanshultz:
- Reading the datasheet again, it appears the MPN is M20-890xx05, where xx is the pin count (03 to 20). Then should these be named (for example) M20-8901905? Are there other options for the last two characters but 05?
- Around pin 1 there are a lot of silk lines which end up as just dots (delete the horizontal lines to see).
- The vertical line at the end of pin 1 is doubled.
- Instead of all the lines for pin 1 you could consider a pad with no number and only on the silk layer. I honestly haven't tried if the generator will allow this, but it works if I create it manually.
- The courtyard clearance is too big everywhere.

Looking good now! Dimensions drawing:
image

evanshultz:
By the way, the script doesn't check fab lines and pads for centering so you can't believe those errors. But, SMD footprints should be centered and your pad 1 is at (0,0). Please move the footprint center for all footprints so it's at the geometric center of the fab lines + pads. Sorry I didn't catch that earlier.

And thanks for continuing on with this PR!

antoniovazquezblanco:
> Reading the datasheet again, it appears the MPN is M20-890xx05, where xx is the pin count (03 to 20). Then should these be named (for example) M20-8901905? Are there other options for the last two characters but 05?

The last two digits determine the contact finish. Other values than 05 exist. https://cdn.harwin.com/pdfs/C001XX_M20_Series_Connectors.pdf

imagen

imagen

evanshultz:
The footprints are centered vertically but not horizontally. The pad is 8.025mm on the left but the fab lines on the right only go to 5.525mm. I'll wait for a fix to that before doing any more reviewing.

evanshultz:
Looks good. Thank you!

antoniovazquezblanco:
Sorry for all the reviewing needed and thank you so so much!

PR #430, opened 2018-03-14
Minor footprint fixes (3D models and footprint name)

evanshultz:
When the footprints were moved to this repo I made a queue of issues to be fixed later. Now is later.
- Correct "Pitch" to "P".
- Add missing 3D models.
- Rename a couple footprints which seemed to be manufacturer-specific and had unhelpful suffixes. I also made a few corrections to those footprints to match the TI datasheet at http://www.ti.com/lit/ds/symlink/tps5430.pdf. (Note that TPS5430DDA exists in Regulator_Switching.lib, but it's using a different footprint. And that footprint's prefix is `Texas`, not `TI`. Sigh.)

------------


evanshultz:
Unless there are objections, I'd prefer to merge this as-is.

poeschlr:
The texas_hsop-8 footprints have strange EP pad stackup. Please use the same as we use for all other such packages.

Especially problematic is the SMD pad with "All copper layers".

evanshultz:
@poeschlr 
Do you have time to finish reviewing this? Anybody else is welcome as well. https://github.com/KiCad/kicad-symbols/pull/512 could use a footprint that is getting changed in this PR.

poeschlr:
It is strange that the texas footprints have only one paste pad but as this is the suggestion from texas i will merge it as is. Thanks for your work.

PR #431, opened 2018-03-15
Added Package_SO:TSOP-28_11.8x8mm_P0.55mm

daab4848:
**Pull Request Details** 
-----------------------------------------------------------------
**Part Name**: TSOP-28_11.8x8mm_P0.55mm
**Part Manufacturer**: Microchip
**Datasheet URL**: http://ww1.microchip.com/downloads/en/devicedoc/doc0807.pdf
**Current Footprint Library**: Package_SO
**Associated Symbol(s) and/or PR(s)**: https://github.com/KiCad/kicad-symbols/pull/366

**Footprint Screenshot**:
image

**Additional Comments and/or Questions**
- @poeschlr @evanshultz @Ratfink Please advise the TSOP-28_11.8x8mm_P0.55mm footprint. The Package_LLC library is currently non-existent so the PLCC-32 footprint is delayed. Should I create a Package_LLC library or would you guys do that?

**Edit: Package_LLC library does exist. I just wasn't looking at the version 5 of the footprint library...whoopsies!**
------------

Thanks for creating a pull request to contribute to the KiCad libraries! To speed up the integration of your PR, please check the following items:


evanshultz:
How did you determine the copper? The datasheet you linked only shows a package drawing but nothing about the copper.

For the package question above, that you figured out, it's called LCC, not LLC. :)

daab4848:
@evanshultz 

image

If by the copper you mean the length of the package's pads, then I just calculated it by taking the difference between the end-to-end width of the package's pads and the width of the package. The worse case scenario is that the pad lengths have to be 2mm, calculated by 13.7mm-11.7mm, and then I added 0.1mm just in case.

evanshultz:
Hmmm... I would like something better than that. Imagine if there's a soldering problem and the user says "what footprint is being used?" and they see the KiCad footprint is not built from any official recommendation. I would at least want to be able to say the footprint matches a document.

It seems this TSOP 11.8x8mm package is common for memory. Searching online I see lots of datasheets that use this package, but few footprint recommendations. Could you try to find a document giving a recommended footprint, then update your footprint to match the document?

Based on page 2 of http://www2.elo.utfsm.cl/~lsb/elo311/datos/packages.pdf, TSOP-I has pins on the short sides and TSOP-II has pins on the long sides. It appears our footprints are already named accordingly, so could you please also name your footprint this way? Unfortunately, that means the symbol will now need to be updated to match with this footprint. That can be a second step.

evanshultz:
Or, show the IPC calculations to support pad length, pad width, pad spacing, etc. for this package. Finding a recommended footprint is easier but doing it from scratch is fine too.

daab4848:
@evanshultz 
image

I couldn't find a direct recommended footprint, but I did find an smd ipc calculator (see above)

evanshultz:
Thanks! Can you post a new screenshot of the footprint?

daab4848:
@evanshultz Sorry for the wait. Footprint is down below: 
image

evanshultz:
Thanks!
- Update the description and tags, making sure to put the Microchip datasheet in the description and also noting that you used the IPC calculator to determine the footprint pads.
- Can you make the pad width exactly 1.87mm? It seems pedantic but it's the IPC value.
- And for a pad row-to-row distance of 12.54mm, each row should be at +/-6.27mm. Not 6.85mm.
- The above will affect the courtyard placement.
- Is the 3D model different than the footprint name? Looks like it only shows 5 pins instead of 28.
- Since is seems having pins on the short side is `TSOP-I`, can you change this footprint (and 3D model) name to match?

Once that's all done, I'll do another review with dimensions.

daab4848:
@evanshultz 
image

*Fixed the copper pad sizes and changed the row-to-row distance. However, the fab outline must stay where it is because the datasheet shows a nominal package width of 11.8mm. I will fix the courtyard placement once I've gotten a confirmation that the current changes are as intended.
*Removed the 3D model that has the 5 pins. I actually took a TSOP 5 pin footprint and modified it to 28 pins.

evanshultz:
@daab4848 
Thanks for making changes. We still have a bit of work to do.
1. Let's put more than just the Microchip datasheet in the description. Something like `TSOP I, 28 pins, 18.8x8mm body, 0.55mm pitch`.
1. Noting that you used the IPC calculator to determine the footprint pads in the description not on the fab layer. Perhaps `IPC-calculated pads`.
1. Add keywords similar to https://github.com/KiCad/kicad-footprints/blob/master/Package_SO.pretty/TSOP-I-56_18.4x14mm_P0.5mm.kicad_mod.
1. There should be a 3D model reference, just with the same name as this footprint.
1. Since is seems having pins on the short side is TSOP-I, can you change this footprint (and 3D model) name to match? So it would be `TSOP-I-28_11.8x8mm_P0.55mm`.
1. Trim down the pin 1 mark so it's only as long as the pin 1 pad.

Here are dimensions. Fab lines were already fine. All is OK except for courtyard clearance:
image

daab4848:
@evanshultz 

Here's what the footprint looks like currently:
image

>Let's put more than just the Microchip datasheet in the description. Something like TSOP I, 28 pins, 18.8x8mm body, 0.55mm pitch.

*Done

>Noting that you used the IPC calculator to determine the footprint pads in the description not on the fab layer. Perhaps IPC-calculated pads.

*Fixed

>Add keywords similar to https://github.com/KiCad/kicad-footprints/blob/master/Package_SO.pretty/TSOP-I-56_18.4x14mm_P0.5mm.kicad_mod.

*Added to keywords

>There should be a 3D model reference, just with the same name as this footprint.

*The current version of the 3D library does not have any 3D model for the TSOP I, 28 pins.

>Since is seems having pins on the short side is TSOP-I, can you change this footprint (and 3D model) name to match? So it would be TSOP-I-28_11.8x8mm_P0.55mm.

*Fixed except for 3D model since it is currently non-existent.

>Trim down the pin 1 mark so it's only as long as the pin 1 pad.

*Done.

evanshultz:
Those changes look good, but there are still a couple things to clean up before I can merge:
1. The file name and F.Fab text are now `TSOP-I-28_11.8x8mm_P0.55mm`, but the footprint name doesn't match. This is line 1 of the footprint file. Travis caught this.
1. Add a 3D model even if it's not present. If/when it is added, the footprint won't need any changes. Travis caught this.
1. The text you added on F.Fab still exists, but just as an empty string. Totally remove it. This is best done in a text editor.
1. Remove ` Microchip` from the keywords.

daab4848:
@evanshultz 
Fixed the remaining issues.

evanshultz:
Thanks for more fixes! Very close, but not quite there.
1. Remove the `Package_SO:` from the F.Fab text.
1. The empty string of text was removed totally, but in your last commit it came back.

daab4848:
@evanshultz 
My latest commit already accommodates these latest fixes and travis has given my latest commit passing remarks.  Below is the latest version of the footprint as well as the first few lines of the file opened in a text editor, and both of these confirm that my latest commit has these changes already. Could you verify and advise?

image

image

evanshultz:
Ah! You pushed more commits after I started looking. Sometimes the commits take more time than the messages, which trigger me to look since I got an email. Sorry for the confusion.

Thank you very much for pursuing this footprint until it's ready to merge!

PR #434, opened 2018-03-16
Add Atmel MCU QFN footprints

### LINKED ISSUE #417: Add missing footprints for Atmel MCUs
Follows KiCad/kicad-symbols#340.

There are a number of missing QFN footprints for some Atmel MCUs. These should be based on [Atmel Appnote 8826](http://ww1.microchip.com/downloads/en/AppNotes/Atmel-8826-SEEPROM-PCB-Mounting-Guidelines-Surface-Mount-Packages-ApplicationNote.pdf).

- [ ] Package_DFN_QFN:DFN-10-1EP_3x3mm_P0.5mm_EP1.75x2.7mm (1)
- [ ] Package_DFN_QFN:QFN-20-1EP_3x3mm_P0.45mm_EP1.6x1.6mm (4)
- [ ] Package_DFN_QFN:QFN-20-1EP_4x4mm_P0.5mm_EP2.25x2.25mm (9)
- [ ] Package_DFN_QFN:QFN-28-1EP_4x4mm_P0.45mm_EP2.25x2.25mm (2)
- [ ] Package_DFN_QFN:QFN-32-1EP_4x4mm_P0.4mm_EP2.9x2.9mm (1)
- [ ] Package_DFN_QFN:QFN-32-1EP_5x5mm_P0.5mm_EP3.8x3.8mm (13)
- [ ] Package_DFN_QFN:QFN-32-1EP_7x7mm_P0.65mm_EP4.65x4.65mm (2)
- [ ] Package_DFN_QFN:QFN-44-1EP_7x7mm_P0.5mm_EP5.2x5.2mm (13)
- [ ] Package_DFN_QFN:QFN-44-1EP_7x7mm_P0.5mm_EP5.4x5.4mm (2)
- [ ] Package_DFN_QFN:QFN-64-1EP_9x9mm_P0.5mm_EP7.25x7.25mm (22)

fauxpark:
Resolves #417, see that issue and the linked PR for more info and the "datasheet".

Example:
fp
3d

------------


evanshultz:
Did you make these by hand of use a generator? Perhaps https://github.com/pointhi/kicad-footprint-generator/tree/master/scripts/Housings_DFN_QFN?

fauxpark:
I did these by hand, adapting the closest existing footprints, but I can have a look at the generator if you'd prefer. It doesn't look like it supports oval pads though, and the 3D model directory it spits out is still the old one - in fact a lot of the script directories haven't been renamed yet it looks like.

evanshultz:
By hand is fine. I just wanted to know what to check. :)

fauxpark:
@evanshultz any movement on this?

evanshultz:
@fauxpark 
Not yet. I intend to get to this and the other footprint PRs soon. Just a couple more things first. Then, with all missing footprints merged, back to Atmel symbols!

fauxpark:
Fab and silk line widths fixed. DFN silk adjusted. Silk clearances are now all 0.2mm from fab lines (you stated 0.11 to 0.12mm but KLC says minimum 0.2mm).

Dimensions for the DFN EP and the QFN32 4x4mm footprint are from the datasheets of the symbols using them: [DFN](http://ww1.microchip.com/downloads/en/DeviceDoc/doc2535.pdf#page=165), [QFN](ww1.microchip.com/downloads/en/DeviceDoc/Atmel-8153-8-and-16-bit-AVR-Microcontroller-XMEGA-E-ATxmega8E5-ATxmega16E5-ATxmega32E5_Datasheet.pdf#page=70). This is also why the QFN32 has rect pads.

evanshultz:
[KLC F5.1.2](http://kicad-pcb.org/libraries/klc/F5.1/) says 0.2mm minimum from silk to pads (and the newest IPC recommendation is silk line width or greater). Regarding offset from fab lines, silk only needs to avoid being under the part body (KLC F5.1.3) so 0.11-0.12m offset from fab lines is correct.

fauxpark:
Hmm... that seems not so well defined, per F5.1.3 0.2mm would still be correct too. It only says you can't put silkscreen underneath. Anyway, it's 0.12mm now.

evanshultz:
Thanks for more changes!

Yes, somewhere there was a discussion about adding a clear rule. 0.11-0.12mm is used almost everywhere in the library.

DFN:
1. The copper pad will need to be slightly bigger than the exposed pad of the IC. There was a discussion about this where @poeschlr went through it, but of course I can't find that PR now. However, it's not such a big deal and I could merge as is since there's no official decision on this topic.
1. Could you trim the pin 1 mark on silk back to the length of the pin 1 pad (a little shorter). This will avoid sticking out and possibly getting merged with other silk, keeping the pin 1 silk more clear.
1. The silk is floating rather far from fab lines:
image
1. Some silk is still farther from fab lines than others, like `QFN-32-1EP_4x4mm_P0.4mm_EP2.9x2.9mm` is 0.15mm, but I can merge. It's close enough. :)
1. http://ww1.microchip.com/downloads/en/DeviceDoc/Atmel-8153-8-and-16-bit-AVR-Microcontroller-XMEGA-E-ATxmega8E5-ATxmega16E5-ATxmega32E5_Datasheet.pdf#page=70 is only a package drawing, not a footprint, so can you switch to oval pads on that QFN? Since Atmel specifies oval pads in their app note I suspect they would recommend it everywhere.

After another quick glance if you can fix 2, 3, and 5 I can merge. Anything else is an optional bonus.

fauxpark:
1. It already is, He mentioned [here](https://github.com/KiCad/kicad-symbols/pull/340#issuecomment-370216365) to simply use the max size from the package drawing.
 2. I have dragged it back 0.06mm so that the edge lines up with the pad clearance. Any further and it might be difficult to tell which line is longer. The other corners are already quite close to the edge of the pads as it is.
 3. The silk is 0.2mm from the EP clearance. Any closer and they look uncomfortably close to the edge of the soldermask:
lines
 4. The silk looks fine to me:
silk
I went through the rest of them and they definitely all are 0.12mm.
 5. Hm, the 3x3 QFN20 is in the Atmel doc thus has oval pads in the footprint, but has rect pads in every package drawing I can find trawling through the datasheets. So I have changed the QFN32 to oval.

evanshultz:
Thanks!

Just a couple things on the DFN:
1. The left side of pin 1 is at -1.85mm (1.45mm + 0.85mm / 2)so you can use that for the silk line too. It's -2.015mm now.
2. The silk lines can be moved down to 0.11-0.12mm outside the fab lines (+/-1.61-1.62mm). The copper pad is only +/-1.35mm vertically, which leaves plenty of room between copper and silk. I haven't ever seen the distance between copper and mask to be important. And even at +/-1.61mm they just touch.

fauxpark:
I've made the changes you suggest, but I still think it is tricky to tell which way around the IC goes with such a small difference in length.

3d

I am not worried about the copper-silk clearance, as you say, there is plenty of room. My concern is the silk-mask clearance. See the image in my previous post for what I mean - the edge of the silk is basically on the edge of the solder mask. But if you say it is fine I will just point at you if someone complains 😉

evanshultz:
Thanks!

Once the part is placed on a board, it's pin 1 mark should clarify the position. And the bevel edge on the fab layer will make it clear there. But I see what you're saying. I'd like to keep the silk within the bounds of the pads, but what about bringing the other vertices of the silk down to +/-1.5mm? That will make the difference at least a bit greater than now.

fauxpark:
That does look a lot better. Should be good to go now.

evanshultz:
Thanks so much! You have made some huge improvements to the Atmel parts, both symbols and footprints. Sorry I've been a bit more busy lately so we can't get your good work merged as quickly.

PR #435, opened 2018-03-16
WSON-10: Remove mask over thermal vias

evanshultz:
The 3 vias on this footprint created a KLC violation. I noted this when the footprints were moved to this repo and am now fixing it. The actual PCB wouldn't change due to this, since a big central pad with mask already exists, but it fixes the KLC violation and doesn't hurt. See page 32 of the datasheet at http://www.ti.com/lit/ds/symlink/tps62177.pdf.

------------


Ratfink:
I don't know which KLC rule this was violating (please share!), but the change surely can't hurt.  Thanks!

evanshultz:
Thanks!

Beats me. I just noted
```
thermal vias missing mask:
'WSON-10-1EP_2x3mm_P0.5mm_ThermalVias'
```

poeschlr:
Hm that might be a false positive. (I have not looked at the footprints)
The via pads should not have mask enabled. If travis still flags them then you can ignore this error. (It might be possible to do something similar with this error message like @Ratfink did with the paste messages.)

poeschlr:
Ok i should have looked before commenting. I fear this really was a false positive and this "fix" should not have been merged.

The reason is that we normally want the mask controlled by other pads. (Either by the large pad that also does the copper part of the EP or by a separate mask only pad if the footprint requires tighter control over where mask cutout is.)

evanshultz:
??? The big pin 11 pad already had mask removed on the top and not the bottom. This doesn't change the actual footprint at all when placed on a board. (Whether we want soldermask placed on the bottom and not the top is a different question.)

I looked in the SON folder at the first two I found and they both had mask removed on the vias: https://github.com/KiCad/kicad-footprints/blob/master/Package_SON.pretty/Texas_DSC0010J_ThermalVias.kicad_mod and https://github.com/KiCad/kicad-footprints/blob/master/Package_SON.pretty/Texas_S-PVSON-N10_ThermalVias.kicad_mod. Now I checked the third one (https://github.com/KiCad/kicad-footprints/blob/master/Package_SON.pretty/Texas_S-PVSON-N8_ThermalVias.kicad_mod) and it doesn't have mask on the vias.

I'm not sure I see the problem. The big pad covers these vias already. If the vias should be open, and solder paste loss isn't too bad, the then the user can modify the bottom side. If the user wants tented or plugged vias I've usually handled that with a note on the fab drawing and KiCad can't note it otherwise.

So if we want a default rule about how thermal vias are handled in library parts, that's fine. And a check for the via design chosen would also be good. Right now the library is not consistent, but this PR doesn't change this particular footprint and it doesn't make the library more or less homogenous.

poeschlr:
It does not make a difference for this footprint.
As there are footprints where some of the thermal vias are outside the masked areas, it would make a difference for them.

Having one rule that works for all footprints would make it easier to make a travis check for it. (It also makes describing the rule easier) 
In this case that one rule is that we do not add mask to any of the tht pads used for vias. (I don't quite remember who came up with that particular way of doing it. It might have been jan and oliver.)

I am sorry that we did not communicate this better. (I just looked at the KLC and we seem to have forgotten that part. I added the stuff about separating paste pads but i seem to have forgotten to add the details about thermal via footprints.)

---

Some time ago i made all footprints comply with that (sadly unwritten) rule set. This was the PR in question https://github.com/KiCad/kicad-footprints/pull/285

PR #437, opened 2018-03-18
Maxim WLP-4 package

calebns:
Package  Maxim WLP-4 (0.83x0.83mm)
https://pdfserv.maximintegrated.com/land_patterns/90-0366.PDF
for
https://datasheets.maximintegrated.com/en/ds/MAX2679-MAX2679B.pdf

Purposely violated KLC and made courtyard with 0.28mm clearance (instead of 1mm for BGA) because the part is so small that it does not make any sense to leave such a large gap.
Also, did not provided size nor pitch in the file name to comply with previous Maxim* footprints inside the library.

image

------------


evanshultz:
A couple quick things:
- Why does it not make sense to have such a large gap? What is your criteria for courtyard distance?
- I think we do need more info in this footprint name. Is this footprint really MAXIM-specific? https://datasheets.maximintegrated.com/en/ds/MAX40002-MAX40005.pdf mentions a WLP-4 but in a different size. And https://www.ablicinc.com/en/doc/datasheet/temperature_sensor/S58LM20A_E.pdf is also a WLP-4 in a slightly different size.

evanshultz:
In addition to the above:
- Add the RF amplifier datasheet and keep the land pattern doc too. That includes the part body dimensions which the doc you attached does not.
- Your part is 0.84x0.84mm, 0.01mm too big.
- A good offset for silk over the fab lines is 0.11-0.12mm.
- Perhaps the pin 1 mark should remove silk above pad A1 too? Or there are a few other variations of pin 1 silk mark I see on similar footprints in the lib.

calebns:
Changed:
- name to be generic
- size to 0.84x0.84mm  (it will work for 0.83 maxim and 0.84 ablic)
- silk offset to 0.12mm
- removed mark over A1

The rationale behind 0.28mm courtyard is that the package is small and designed to be used in dense electronics. I just see no reason why a part similar to 0402 should use different clearance. But if you thinks I'm wrong, I will modify.

image

evanshultz:
Thanks for making the updates! Everything not mentioned below looks good!

How did you find https://pdfserv.maximintegrated.com/land_patterns/90-0366.PDF for this package?

The datasheet shows https://pdfserv.maximintegrated.com/package_dwgs/21-100107.PDF for the package drawing and https://www.maximintegrated.com/en/app-notes/index.mvp/id/1891 for the land pattern. But that package drawing shows 0.27mm ball diameter and the website only goes down to 0.3mm ball diameter. For 0.3mm balls it recommends 0.18mm pad diameter which is much smaller than the 0.25mm pad diameter in the link you included above. Can you help me sort that out? Do you find other information?

I have seen the courtyard be driven by device package, not size. BGA parts need more clearance for inspection and rework than other parts regardless of how big or small the BGA package is. KLC currently says 1mm and so I'd like to merge parts that meet KLC. I know there are other important factors that drive courtyard size, and that 1mm is not right for all applications (or maybe even most applications!) but I'd rather not have a conversation about updating KLC hold up this footprint. It's up to you, but if you adjust courtyard clearance to 1mm, meeting the current KLC, I can merge this now. And a separate discussion to suggest changes to KLC for BGA courtyard clearance is welcome.

calebns:
image

image


The website https://www.maximintegrated.com/en/app-notes/index.mvp/id/1891 shows that a pad diameter should be 0.25mm, (pitch is not a ball diameter!).

Btw. looking at the datasheet again I found that this part has 0.4mm pitch, not 0.5mm... That's good that this patch was not merged!
I've fixed it and added both WLP-4 packages (0.9x0.9mm P0.5  and 0.84x0.84mm P0.4).

calebns:
Right! I forgot to move the pads away while doing P0.5 version, thanks!

Both parts should already include proper datasheet and/or link to Maxim website. Let me know if something needs to be changed.

evanshultz:
I believe both packages have a ball pitch of 0.4mm. Please check.

For the 0.9mm square package, you have a PDF land pattern doc. Change that to the website. Use the PDF datasheet (which links to a package drawing) or the PDF package drawing, and then the website for the land pattern.

calebns:
OK, now  see your point :)
Should be ready, let me know if anything is still missing.

Replaced previous patch witch:
- both packages now are P0.4
- 0.9x0.9 has a link to package drawind and to the website
- 0.83x0.83 has a link to datasheet and to the website

evanshultz:
Thanks! Those changes look nice.

Should your "0.9mm" actually be "0.86mm" as shown in the MAX4281 datasheet? Similarly, for MAX2679 should the footprint be 0.83mm? I'm only asking now because, upon further thought, why shouldn't our footprints exactly match the components? The 3D models are also tied to a footprint, so if either the footprint or 3D model are unique between one of these MAXIM parts and any other part then a unique footprint would be warranted. What do you think?

calebns:
Renamed parts to be 0.83 and 0.86, also moved courtyard, fab and silk lines accordingly.

evanshultz:
Seems you're OK with exactly matching the dimensions of the parts. Two things:
- I think the packages MAXIM calls out are 0.83mm and 0.86mm. You have 0.84 instead of 0.83.
- Both footprints have the layout URL, but the 0.86 footprint also has a package drawing. Could you use the datasheet instead of the package drawing for both?

calebns:
Could you point out where did I left 0.84? I think I replaced them everywhere along with all fab/silk lines, but I may be missing sth.

I've updated both footprints with a package drawing in pdf and a link to landing pattern to be consistent. No part datasheet here, as imo. it is not a good to mix part-specific datasheets with generic footprints.

evanshultz:
That's fine. Thanks!

Issue #436, opened 2018-03-17
How to name footprints missing a single pin?


evanshultz:
@poeschlr @jkriege2 @Ratfink @Misca1234 
Started from a conversation at https://github.com/KiCad/kicad-symbols/pull/322.

Some parts are missing a pin. In some cases the footprint for the part will be manufacturer-specific. But in other cases it's not, or it doesn't make sense to do that because it would result in multiple footprints which are otherwise identical.

Let me take the common DIP8 package as an example and show some datasheets with a missing pin:
- http://www.onsemi.com/pub/Collateral/NCP1076A-D.PDF: either pin 3 or pin 6
- https://ac-dc.power.com/system/files_force/product-docs/tinyswitch-iii_family_datasheet.pdf: pin 3
- https://www.mouser.com/ds/2/328/lnk520-5820.pdf: pin 6
- http://www.st.com/content/ccc/resource/technical/document/datasheet/cd/d4/93/15/03/e1/40/81/CD00279018.pdf/files/CD00279018.pdf/_jcr_content/translations/en.CD00279018.pdf: pin 6
- https://www.infineon.com/dgdl/Infineon-ICE3AR0680VJZ-DS-v02_01-en.pdf?fileId=db3a304341e0aed00141ffe31ad53057 pin 6
- http://www.onsemi.com/pub/Collateral/NCP1216-D.PDF: pin 7
- http://rohmfs.rohm.com/en/products/databook/datasheet/ic/power/isolated_converter/bm2pxx2-e.pdf pin 6
- http://www.onsemi.com/pub/Collateral/NCP1337-D.PDF: pin 7
- https://www.diodes.com/assets/Datasheets/AP3983BCD.pdf: pin 7
- https://www.dialog-semiconductor.com/sites/default/files/iw1818_product_summary.pdf: pin 6
- https://www.diodes.com/assets/Datasheets/AP3983BCD.pdf: pin 7

We would need at least three, and possibly up to six, footprints for the parts above. Three with pin numbers up to 8 but missing pin 3, or 6, or 7, and another three with pin numbers up to 7 and missing the same three pins.

For packages with hidden or deleted pads according to IPC 7351, footprint names would be `DIP-8-7_W7.62mm` or `DIP-7-8_W7.62mm`. We would then need a way to capture the missing pins as there is no way to differentiate which pin is missing.

I see two ways this could be done, using one example:
[...]

PR #442, opened 2018-03-19
Add Atmel MCU BGA footprints

fauxpark:
Some of the ATmega and ATtiny parts come in BGA packages. In preparation for adding those symbols, here are the footprints they use.

Datasheets:
[BGA-15_3.0x3.0mm_Layout4x4_P0.65mm](http://ww1.microchip.com/downloads/en/DeviceDoc/atmel-8235-8-bit-avr-microcontroller-attiny20_datasheet.pdf#page=211)
[BGA-32_4.0x4.0mm_Layout6x6_P0.5mm](http://ww1.microchip.com/downloads/en/DeviceDoc/atmel-8271-8-bit-avr-microcontroller-atmega48a-48pa-88a-88pa-168a-168pa-328-328p_datasheet.pdf#page=25)
[BGA-49_5.0x5.0mm_Layout7x7_P0.65mm](http://ww1.microchip.com/downloads/en/DeviceDoc/Atmel-8272-8-bit-AVR-microcontroller-ATmega164A_PA-324A_PA-644A_PA-1284_P_datasheet.pdf#page=647)
[BGA-100_7.0x7.0mm_Layout10x10_P0.65mm](http://ww1.microchip.com/downloads/en/DeviceDoc/Atmel-8067-8-and-16-bit-AVR-Microcontrollers-ATxmega64A1-ATxmega128A1_Datasheet.pdf#page=72)
[BGA-100_9.0x9.0mm_Layout10x10_P0.8mm](http://ww1.microchip.com/downloads/en/DeviceDoc/Atmel-2549-8-bit-AVR-Microcontroller-ATmega640-1280-1281-2560-2561_datasheet.pdf#page=413)
The pad sizes should conform to the BGA landing pattern from [Atmel Appnote 8826](http://ww1.microchip.com/downloads/en/AppNotes/Atmel-8826-SEEPROM-PCB-Mounting-Guidelines-Surface-Mount-Packages-ApplicationNote.pdf#page=8).

fp
3d

------------


evanshultz:
- Can you place ref des and footprint name outside the courtyard?
- Make the fab refdes as big as possible, up to 1mm. `BGA-15_3.0x3.0mm_Layout4x4_P0.65mm` is 0.5mm and bigger text can fit within the fab lines.
- I wonder if removing silk by the corner of pin 1 (or A1 here) is better than double silk lines? It is mentioned at https://www.pcblibraries.com/forum/ipc7351c-pin-1-polarity-marking_topic1530.html. Would you mind trying that out? Page 13 of http://www.ocipcdc.org/archive/What_is_New_in_IPC-7351C_03_11_2015.pdf shows something a bit different but puts silk under the body (in this case the body doesn't touch the PCB, which I think is OK, so I would vote for an update to KLC).
- Page 8 of http://ww1.microchip.com/downloads/en/AppNotes/Atmel-8826-SEEPROM-PCB-Mounting-Guidelines-Surface-Mount-Packages-ApplicationNote.pdf#page=8 covers pad diameter. `BGA-15_3.0x3.0mm_Layout4x4_P0.65mm` has 0.25mm pads. http://ww1.microchip.com/downloads/en/DeviceDoc/atmel-8235-8-bit-avr-microcontroller-attiny20_datasheet.pdf#page=211 shows dimensions `b` and `b1` and it's not clear to me which one should be used in the calculation. You mentioned your pads conform to that doc so how did you figure out pad diameters?
- http://ww1.microchip.com/downloads/en/DeviceDoc/atmel-8235-8-bit-avr-microcontroller-attiny20_datasheet.pdf#page=21 calls the package `UFBGA` and I think it would be best to use the exact type of BGA package. http://www.statschippac.com/~/media/Files/Package%20Datasheets/FBGA.ashx shows that the height is different. Plus these balls are sunk into the package. If nothing else, it will ensure the 3D models can be unique between the various BGA packages.

@poeschlr @jkriege2 @Ratfink @Misca1234 
Do we have a convention for BGAs that are missing pads? I don't see anything at http://kicad-pcb.org/libraries/klc/F3.4/. Similar to my question at https://github.com/KiCad/kicad-footprints/issues/436, but it gets messy if we use `-Nxx`, where `xx` is the ball number, because a missing ball could theoretically be start with an `N`. Also, some BGA are missing many, many balls. If we don't already I'll merge these (no worse than what we have already) and open an issue about this topic.

Ratfink:
I don't think we have any convention for that, and you're right that listing all the balls missing from the grid could get ridiculous, since so many BGAs are missing a lot of balls from their grids.  Personally I'm inclined to not worry about it for BGAs until we actually get a naming conflict.

fauxpark:
- Moved refdes and name out 0.75mm
 - Enlarged fab refdes
 - Something like this maybe?
fp
3d
In at least the BGA-32 and -49 footprints, following the fab line chamfer will place it too close to the pads. I think shortening the left silk line is the best option.
 - I used `b` to determine the pad size. It states nominal 0.3mm, so per the Atmel doc the pad should be 0.25mm. `b1` indicates "the solderable surface" according to note 2 in the datasheet, and also happens to be 0.25mm.
 - Renamed the footprints accordingly

evanshultz:
What about cutting out the top and left left silk lines? Since pin A1 is in a corner, there isn't a single side that needs the pin 1 mark but instead the entire corner should get it. I wasn't suggesting putting silk under the part, just noting that the latest IPC doc showed it. Cutting away silk on both corner lines seems good.

`TFBGA-100_9.0x9.0mm_Layout10x10_P0.8mm` looks like it should be `CBGA` from http://ww1.microchip.com/downloads/en/DeviceDoc/Atmel-2549-8-bit-AVR-Microcontroller-ATmega640-1280-1281-2560-2561_datasheet.pdf#page=413.

Everything else looks good to me! Thank you!

fauxpark:
I was going for something similar to the QFN pin 1 markers, where it just has the horizontal silk in the top left. Since there is only one correct orientation where A1 is closest to the missing silk, I don't think it really matters whether the top line is also shortened other than to save some miniscule amount of ink.

TFBGA and CGBA appear to be the same thing - the datasheet and the document you linked specify the same package height. [Digikey](https://www.digikey.com/product-detail/en/microchip-technology/ATMEGA1280-16CU/ATMEGA1280-16CU-ND/1245826) seems to agree. The [Wikipedia page](https://en.wikipedia.org/wiki/Ball_grid_array#Variants) says that CGBA is a ceramic package, however I'm pretty sure these are made of plastic.

evanshultz:
Good enough for me, then. Thank you!

fauxpark:
You merged without me changing the pin 1 markers - should I wait until the referenced issue is resolved?

evanshultz:
No. This is fine. There are lots of different pin 1 silk styles in BGA parts. Once a conclusion is reach they can all be fixed at once. Thank again!

PR #443, opened 2018-03-20
Change TI_ prefix to Texas_ for X2SON footprint

evanshultz:
Just what it says. Check all the other repos, including this one, to see Texas_ is used everywhere else.

Noted by https://github.com/KiCad/kicad-symbols/pull/386.

------------


Ratfink:
Good catch!  `grep` says that this footprint isn't used by anything in kicad-symbols, so I suppose there's no harm in merging this.

PR #447, opened 2018-03-21
Add TI GAN FET footprints

evanshultz:
Datasheets:
- http://www.ti.com/lit/ds/symlink/lmg5200.pdf (MOF...)
- http://www.ti.com/lit/ds/snosd10c/snosd10c.pdf (RWH...)

Image:
image

------------


evanshultz:
Here is the best dimensional drawing I know how to make:
image
image

The paste layers don't appear to be imported by KSU. And I don't know why the outer pads on the RWH footprint are rendered as circles; perhaps KSU doesn't know about rounded rectangle pads?

Because I'm lazy, I found a few incorrect dimensions above but continued with dimensioning. I'm fixing the errors next.

evanshultz:
That should do it. On the RWH footprint not all via X positions are shown, so the middle row is 1.225mm, the center of the big pad 33, and the dimensions on the right I mirrored from the left dimensions.

Ratfink:
I'll review this, then once the footprints are done, I'll try to get crackin' on the symbol PR.

#### Texas_MOF0009A

* I believe this should be in Package_LGA.
* Drawing for the package and footprint only: http://www.ti.com/lit/ml/mpsi063a/mpsi063a.pdf
* Most parts of the symbol seem to be shifted 0.0025 mm in the Y direction from where they should be.  It looks like the pads, stencil openings, F.Fab outline, and silkscreen outline need to be changed and the rest needs to stay the same, but don't take my word for it.
* After fixing the positions of everything, please double-check the courtyard.  The top line is just barely not horizontal, and the bottom line is too far from the body by 0.01 mm.

#### Texas_RWH0032A

* I couldn't find a drawing of only the package and footprint. 🙁
* Are the dimensions in the drawing on page 26 supposed to be of the metal, or the soldermask opening?  I assume it's the mask opening (like in the drawing for MOF0009A), in which case all the pad sizes are too small.
* By note 5 on page 26, this should be a _ThermalVias variant.
* The via positions aren't quite what's shown in the datasheet (X spacing is too big)
* Following https://github.com/KiCad/kicad-footprints/pull/435#issuecomment-373879818, the vias should have the mask layers turned off.
* The paste is missing from pads 11, 17, and 27.

evanshultz:
I think KiCad v5 saves every footprint in the target repo, not just the one footprint that's open. I changed line endings file-by-file when I finished them, but they were apparently all changed back to Windows.

Sorry for the noise. I did something dumb.

Ratfink:
Looks good, thanks for the fixes.  The reason I was unsure about the classification of MOF0009A was because it doesn't look like a regular QFN, as the leads are only on the bottom, with no metal on the sides.  This seems more like other packages we've filed under Package_LGA, but you're right that calling it that might be the wrong thing to do.  And for the record, QFM stands for Quad Flat Module, according to the package drawing.

I see only two problems now.  The solder paste for Texas_RWH0032A* needs to be fixed: most pads have more paste than exposed copper (I think only the EP has correct paste right now).  Also, the non-ThermalVias version shouldn't have the bottom copper pad.  Once those issues are fixed, I think this will be ready.

evanshultz:
Thanks for checking again and catching the paste, now that the metal sizes were changed.

I think all the paste should be fixed now. I also removed the bottom copper from the no vias version.

Our scripts probably need a bit of work, based on the output from these footprints, if you have some time once things settle down a bit.

Ratfink:
It looks like you've made correct stencil openings for every pad except 16 and 28, which have slightly off-center stencil opening pads.  But most of the copper pads still have the F.Paste layer turned on, so there's still excess paste.

evanshultz:
Good catches. I think that should do it.

Ratfink:
Yes, everything looks good now.  I'll wait to see how #529 and the corresponding mailing list discussion shake out before merging this though, as a couple further changes may or may not be required.

evanshultz:
@Ratfink 
I think that is correct. These footprints all open in KiCad and look like they're OK, but if you can confirm this is the right this to change before merging it would be great. Thanks!

Ratfink:
I believe the current practice is to also override the solder paste ratio clearance with a value of -0.0001% for mask opening pads.

Also, the pad-local solder paste clearance you set on the paste pads for Texas_MOF0009A made the paste pads too big, because they were sized appropriately for the footprint-local paste clearance.  You could just remove the pad paste clearance for those without any problems, change the pad setting to explicitly set the same setting as the footprint, or keep the pad setting you have and resize the pads.  Do set the paste ratio clearance for those pads though.

I think, anyway!  This new policy is kinda confusing to me and I'm not clear on where this should be done and where it shouldn't, but @poeschlr seems to want it.  I'd like to see an explanation in KLC very soon to end my confusion.

poeschlr:
@Ratfink i would suggest a small experiment. (One learns a lot more with simple experiments then by reading a set of rules.)

Make a test footpirnt.
Give it one pad with mask and paste layer active. Padsize 1x1mm. Set the soldermask clearance to -0.1mm and the paste clearance to -0.2mm Keep the solderpaste ratio at 0 (will get important below)

Add a copper only pad some distance away with also 1x1mm. On top of it add a mask only pad with 0.8x0.8mm and a paste pad with 0.6x0.6mm (In the footprint editor the two pads should now look the same.)

---

Now comes the fun part. Add that footprint to a project and play with the project soldermask clearance, paste clearance and solderpaste ratio. You should notice that the soldermask clearance and paste clearance setting only affect the pad that consists of the 3 pads on top of each other. But changing the project solderpaste ratio will still affect the other pad as well. (We kept the pad at 0 -> tells kicad take the setting of the project.)

---

Another note: Never set the clearances in both the footprint and the pad. They get added to each other. (so in the above example if you set the solder paste clearance of the footprint also to some value you will notice that it affects both pads.) At least if i remember my experiment correctly. I am too tired to repeat it right now.

---

I can provide the following set of rules.

For mask only pads, set the soldermask clearance of the pad to 1e-6
for paste only pads set the solder paste clearance to 1e-6 and the solder past ratio to 1e-4 (both of these values result in 1e-6 in the file it self as the ratio is stored in 1/100%)

If you set either the solder paste clearance or the solder paste ratio of a pad, set the other to the value listed above. (It is ok to set only soldermask clearance.) This is also true if you set it in the footprint instead of the pad settings.

Repeat the experiment above while using these rules. (The result should be that the both pads keep the same measurements as originally defined.)

---

Also ignore spelling and grammar errors, it is 4 in the morning for me. I should really go to sleep.

stambaughw:
Rene,

Before you merge this commit, a solution has been suggested that doesn't require adding the tolerance and ratios to prevent non-copper pads from having the inherited values applied to them but I would like to discuss it with you and the developers group before I implement it.

Cheers,

Wayne

poeschlr:
This is good news. If a good solution for this problem can be found in kicad 5 i am all for it.

stambaughw:
@Rene,

I will let you know as soon as I have everything sorted out.

Cheers,

Wayne

Ratfink:
The patch was merged into KiCad master in https://github.com/KiCad/kicad-source-mirror/commit/93c7f65d96130c7165d412945668293f571cf2e9.  @evanshultz you should be able to undo the paste margin workaround now.

evanshultz:
@Ratfink 
I missed this one, but the RC2 freeze made me check any outstanding PRs from me. If this is good, can you merge this and also the associated symbols?

Ratfink:
I'm out of town right now and can't really review anything until mid-next-week, so I think another librarian had better finish this to make sure it's done before RC2.

evanshultz:
@poeschlr @jkriege2 @Misca1234 @Shackmeister 
Can anybody else take a look to get this done before the weekend is out so this is in v5? Also note that https://github.com/KiCad/kicad-symbols/pull/410 is waiting on these footprints.

evanshultz:
@Ratfink 
If you're around are you able to take a look at this?

Ratfink:
Texas_MOF0009A needs its paste opening pads to be shrunken slightly, because the clearance values no longer apply.  Texas_RWH0032A looks okay as-is.

evanshultz:
It looks to me like pins 2-7 have the mask and paste both 0.05 smaller than the copper and 0.5x0.7mm. And for the other pads I see they are all 0.7mm tall and the right width. So I'm not seeing the problem. Can you please point out the issue you found?

evanshultz:
Oh. Right. I installed the RC2 nightly and now see it. Thanks for clarifying. Should be fixed now.

Ratfink:
Yep, everything looks good now!  Thanks for all the work on these footprints.

PR #450, opened 2018-03-22
Fixed broken install prefix setting

metacollin:
This does not work for macOS users.  It will try to install to /usr/local/share/kicad (which does not exist) and fails without installing anything.

This is caused by how CMAKE_INSTALL_PREFIX is being handled.  Without using FORCE because the project ( ) command, whose job is to setup the basic variables needed, will have already set CMAKE_INSTALL_PREFIX.  Some older versions of cmake with an incomplete implementation of project don't set the install prefix, allowing this to accidentally work.  That is not the intended behavior however, nor does it work with recent versions of cmake.

This commit fixes the problem by using the canonical way to handle the CMAKE_INSTALL_PREFIX according to the cmake developers: https://public.kitware.com/pipermail/cmake/2010-December/041135.html

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,8 +7,11 @@ cmake_minimum_required( VERSION 2.6.1 FATAL_ERROR )
 # Locations for install targets.
 #================================================
 if( APPLE )
-    # Like all variables, CMAKE_INSTALL_PREFIX can be over-ridden on the command line.
-    set( CMAKE_INSTALL_PREFIX "/Library/Application Support/kicad/" CACHE PATH "" )
+    # The project command has already set CMAKE_INSTALL_PREFIX, so FORCE must be used.
+    # This check makes sure CMAKE_INSTALL_PREFIX can still be overridden from the command line.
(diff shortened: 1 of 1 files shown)


poeschlr:
The same change for the symbol repo is done in https://github.com/KiCad/kicad-symbols/pull/409

@metacollin does the [3d model repo](https://github.com/KiCad/kicad-packages3D) also require something similar?

poeschlr:
I asked the developers for input on this one. (As i don't know enough about cmake)
https://lists.launchpad.net/kicad-developers/msg35113.html

antoniovazquezblanco:
What is the status on this issue @metacollin?
It seems you have been trying to solve two different issues and one have been fixed by the following line:
https://github.com/KiCad/kicad-footprints/blob/43bf2cf785ec4ca392b659dd7c65278cd74495cc/CMakeLists.txt#L46
But I do not know enough to check wether the other issue is gone.

Thank you!

poeschlr:
The devs where not happy with the proposed solution so i would guess this can be closed.

PR #455, opened 2018-03-26
Add F.Fab pin 1 mark to LFCSP-16

Ratfink:
I merged #367, then immediately noticed that the footprint had no pin 1 mark on the F.Fab layer.  D'oh!  This fixes the problem by adding the pin 1 mark.

screenshot from 2018-03-26 16-24-59

------------


evanshultz:
My fault. Thanks for fixing!

PR #456, opened 2018-03-27
Add ALPS SKRK series button footprint

xl0:
Data: http://www.alps.com/prod/info/E/HTML/Tact/SurfaceMount/SKRK/SKRKAHE020.html .

skrk

No 3D model


Ratfink:
Thanks for the contribution!  There are a few things that should be changed before this can be merged.

* First, please make sure the fabrication layer outline has the same dimensions as the part.  Right now the main rectangle is 0.1 mm too big in both the X and Y dimensions.  See [F5.2](http://kicad-pcb.org/libraries/klc/F5.2/).
* Also, please make the silkscreen meet [F5.1](http://kicad-pcb.org/libraries/klc/F5.1/).  In particular, the lines should be 0.12 mm wide, should not be within 0.2 mm of any exposed copper (there should be gaps for the pads), and there must be no silkscreen under the part (remove the silkscreen circle in the center).  The silkscreen outline should be slightly outside of the fabrication layer outline (an offset of 0.11-0.12 mm is preferred).
* Please center the fabrication layer reference designator on the origin.  The size seems okay, but if you add a fabrication layer circle to show the switch's stem, the fabrication layer ref.des should be shrunken to not overlap any lines.
* The courtyard clearance at the top and bottom is too large.  It should be 0.25 mm from exposed copper and the part body.
* Please add the 3D model filename, as per [F9.3](http://kicad-pcb.org/libraries/klc/F9.3/).  Although there's no 3D model provided (and that's fine!), setting the filename allows us to add a 3D model in the future without having to change the footprint again.
* I'd suggest adding the keyword "SPST" to make searching for footprints a bit easier.

evanshultz:
May I also suggest putting ref des above the footprint and the footprint name below (swap their locations)? I don't know this is in KLC, but it matches all but a few very old footprints I find.

xl0:
Thanks! Will do.

Ratfink:
@xl0 Just checking in: do you think you'll be able to work on this (and your other two PRs) any time soon?

xl0:
Hey,

my bad, time's a bit short. I will get back to this soon. Got many more footprints and symbols to push.

Ratfink:
@xl0 Sorry to keep bugging you, but it's been two more weeks and still nothing's happened on these PRs.  Is there a chance you'll be able to finish them up soon?

Misca1234:
@xl0 
Any news on this one?

poeschlr:
I took this over and finished the remaining issues in https://github.com/KiCad/kicad-footprints/pull/1247

Thanks @xl0 for your contribution.

PR #457, opened 2018-03-27
Improve SKQG footprints

Ratfink:
Since there are two different types of ALPS SKQG switch (with and without stem) that require different 3D models, we need to have two different footprints.  This commit makes that split.  I don't have 3D models yet, but I plan to work on them before too long.

The outline has been cleaned up as well, after measuring the 3D model from ALPS for dimensions missing in the drawing.  The keep-out zones have been added, and the courtyard has been made a bit tighter.

It looks like ALPS doesn't have datasheet PDFs for these switches anymore.  Here are links to HTML datasheets:
* [Without stem](http://www.alps.com/prod/info/E/HTML/Tact/SurfaceMount/SKQG/SKQGAEE010.html)
* [With stem](http://www.alps.com/prod/info/E/HTML/Tact/SurfaceMount/SKQG/SKQGAFE010.html)

screenshot from 2018-03-27 17-36-54
screenshot from 2018-03-27 17-36-52
screenshot from 2018-03-27 16-57-58

------------


jkriege2:
nice work!

thanks,
JAN

PR #467, opened 2018-04-05
Footprints for 2xAA battery holders from MPD.

ppelleti:
Footprints for [BC12AAPC](http://www.memoryprotectiondevices.com/datasheets/BC12AAPC-datasheet.pdf) and [BC2AAPC](https://media.digikey.com/pdf/Data%20Sheets/Memory%20Protection%20PDFs/BC2AAPC.pdf).

bc12aapc

bc2aapc

------------


Ratfink:
Thanks for the contribution!  I see just one issue with this PR as it is: for `BC2AAPC`, please use the datasheet URL from the manufacturer: http://www.memoryprotectiondevices.com/datasheets/BC2AAPC-datasheet.pdf

I notice that you used a 0.5 mm clearance around these parts.  KLC doesn't seem to explicitly say what clearance to use for battery holders, but quite a few already in the library use 0.5 mm, and I suppose they could be considered a type of connector (which would mean the clearance should be 0.5 mm), so I think that it's fine the way it is.

ppelleti:
OK, I've changed the URL.

Ratfink:
Thanks!

PR #468, opened 2018-04-05
fixed Lumberg USB footprint

Shackmeister:
Just wanted to fix the name (Lumber -> Lumberg)
turned out there was a few other things which needed fixing.
http://downloads.lumberg.com/datenblaetter/en/2486_01.pdf
lumberg_usb

------------


evanshultz:
I see a couple of things that need fixed:
- Update the datasheet URL
- The ref des on F.Fab can be rotated so it's horizontal by default.
- The silk line on the right side isn't as long as the left side.
- Your mounting holes are 0.9mm diameter.
- Your big pad spacing is 5.5mm.

Extremely (extremely!) messy dimensioned drawing:
image

FYI, the datasheet URL in the footprint is http://cdn-reichelt.de/documents/datenblatt/C160/DS_2486_01.pdf, and it has rather different dimensions in some cases. Perhaps since you're looking at a different datasheet that's why you thought it was built wrong? Since that URL was 10 years old and yours is new, I'll go with it.

Shackmeister:
@evanshultz thanks for the review, they should all be fixed now

evanshultz:
Thank you!

Issue #466, opened 2018-04-04
Resistor 2512 footprint dimensions


Palmitoxico:
I've previously discussed this problem in the kicad-devel mailing list, but it is more appropriate to post it here.

I've verified that the current footprint for the 2512 SMD resistor (imperial) doesn't follow the recommended IPC dimensions.

I'm reading the IPC-SM-782 standard section 8.1 subsection 5.0 and here is a table for comparison:

Dim(mm) | IPC | Kicad
------------ | ------------- | ----------
Z | 7.4 | 7.5
G | 3.8 | 4.8
X | 3.2 | 3.4

The G dimension is too large. I can barely solder the resistors in my board (I should have checked it before sending the gerbers).

Thanks,
Augusto Fraga Giachero.

sethhillbrand:
Just a quick note that it looks like more than just the 2512 are off IPC specs.

Palmitoxico:
Yes, I just spoted this after trying to solder the resistors by hand. Not a very pleasant surprise : (

I couldn't find where the scripts for generating the footprints are located in this repository. Any hints?

evanshultz:
The SMD chip generator scripts are at https://github.com/pointhi/kicad-footprint-generator/tree/master/scripts/SMD_chip_package_rlc-etc.

poeschlr:
The footprints in this repo use the formulas of IPC-7315. (Which is newer then IPC-SM-782)
As far as i can tell the IPC-7315 reduced the pad sizes for all devices.

By the way your measurements are wrong. (or you measured the wrong footprint) The current footprint has G=4.84, Z=7.18, X=3.4

I might have made some mistake for parts where i had the spacing between the contacts given. There i used the maximum spacing instead of using the one where the Root mean square of the different tolerances is used. With the later it would result in G=4.33

sethhillbrand:
@poeschlr it looks like the heel/toe/side data were copied from table 3-4 for J-leads instead of table 3-5 larger chips (and used row order rather than labels).

Palmitoxico:
@poeschlr I forgot to mention that I measured the handsoldering variant, that's why the Z dimension is higher.

evanshultz:
For posterity, IPC 7315 above is really IPC 7351.

Palmitoxico:
@evanshultz I was about to post it after spending some time searching for this standard.

By the way, I found the IPC-7351 much more harder to understand than the IPC-SM-782.

evanshultz:
[...]

Issue #439, opened 2018-03-19
KLC: Silkscreen-to-pad clearance?


evanshultz:
https://github.com/KiCad/kicad-footprints/pull/354 fixed a footprint by pushing silk lines farther from the pads.

[KLC F5.1.2](http://kicad-pcb.org/libraries/klc/F5.1/) requires at least 0.2mm clearance. IPC 7351C is [recommending the silk line width](http://www.ocipcdc.org/archive/What_is_New_in_IPC-7351C_03_11_2015.pdf) clearance from pads, which would be a minimum of 0.12mm (KLC allows some deviation but I believe almost all footprints are 0.12mm silk line width).

Should KLC be updated? Move the minimum from 0.2mm to 0.12mm in accordance with IPC 7351C?

Ratfink:
The main thing that PR was fixing was the reversed X and Y dimensions in the footprint name.  But you're right, it also moved the silkscreen lines, as they were very close to the pads.  They would have been too close by the rule you're suggesting as well; the current rule just made them need a bit extra clearance.

Here's the discussion from the old kicad-library repository that resulted in this KLC rule in the first place: https://github.com/KiCad/kicad-library/issues/1422

I notice that you mentioned the same thing in the old discussion also, but that was mostly ignored after 0.2 mm was suggested.  I'm certainly not opposed to revising the rule to allow the reduced clearance recommended by IPC, just so long as it's done in a backwards-compatible way.

poeschlr:
The current silk-pad clearance did not come from IPC recommendations. We choose 0.2mm as we could then use a "feature" of kicad 4.x to easily check it. (Kicad 4.x and at least older nightlies shows the mask clearance in the opengl canvas of footprint editor as 0.2mm by default.) And because this is compatible with the fab i use.

If we want to update the KLC i would suggest to (for now) add either a range or document that old footprints used 0.2mm.
If we go with the allowed range we could specify some relation between part size and clearance. As for larger parts i think it is possible to have a larger clearance while still having well defined silk without resulting in a much larger covered area.

PR #469, opened 2018-04-05
Fix pad sizes for two terminal smd devices (issue #466)

poeschlr:
The wrong IPC tables have been used for devices >=0603_1605Metric
For devices where the dimension S instead of T was given,
Smax was used instead fo Smax(RMS).

The example footprint given in #466:
Dimension G of the 2512 (6332Metric) footprint is now 4.54mm which is approximately Smin - 2*Heel = 4.45 - 2*(-0.05). The small difference comes from other considered tolerances.

Script pull request: https://github.com/pointhi/kicad-footprint-generator/pull/119


poeschlr:
@evanshultz @Ratfink @jkriege2 @Shackmeister 

Can one of you take a look at this? The current smd two terminal footprints are not usable and this PR does fix that.

Ratfink:
Overall, this looks great to me!  My one concern is that it looks like overly-conservative silkscreen-to-pad clearance caused 0805 parts to lose their silkscreen entirely, but the lines would be very short unless we go with the suggested clearance in #439, so this might be okay.

Ratfink:
Actually, @Shackmeister brought up a good point at https://github.com/KiCad/kicad-packages3D/pull/280: some of the LED footprints we have don't seem to have any real parts that fit them.  I'd be in favor of removing those, as I mentioned in that thread.

poeschlr:
If i remember correctly then i checked it with either farnell or RS and i could find some parts that would fit them. I don't have time to check it again. And even if we decide to remove them this should be done in a separate PR anyways. (One topic per PR)

poeschlr:
Guys our current footprints are not usable for production. This PR (hopefully) fixes the problem. Could one of you guys look at it and decide if it can be merged?

Please only look at the size of the pads. Other discussions should be made in a separate issue.

Shackmeister:
pad sizes looks good to me, Ill merge it
But a few more eyes on this wouldnt hurt as it's some pretty important parts

a small script enhancement which could be added in the future, flip the Fab REF** is the package is wider than long

PR #471, opened 2018-04-05
Add Footprint for 10/100BASE-T magnetics by Pulse

Schlumpf:
For devices: H1100NL, H1101NL, H1102NL, H1121NL, H1183NL, H1199NL, HX1188NL, HX1198NL and H1302NL.
https://www.mouser.com/pdfdocs/Pulse_1198.pdf
image
I'll add symbols in a separate PR.

------------


evanshultz:
Thanks for adding this footprint!

Your drawing is from 2009 (Mouser archives and hosts their own datasheets) and I find the Pulse website has a 2015 drawing. There are differences. Can you use the datasheet at https://productfinder.pulseeng.com/doc_type/WEB301/doc_num/H1100NL/doc_part/H1100NL.pdf?

- Update datasheet link.
- Rotate F.Fab refdes.
- Shrink pad width as shown in the newer datasheet mentioned above. (And then adjust courtyard to match.)

Drawing:
image

Schlumpf:
Done :)

evanshultz:
Thanks! All these changes look good.

Before I merge, this footprint has many `X` in the name. And the actual Pulse PNs have `X` (some of them). I wonder if we should use a different footprint name. Perhaps `x` instead and leave the `X` to cover the Pulse MPN? For example, for http://suddendocs.samtec.com/prints/hle-1xx-02-xxx-dv-xx-xx-xx-mkt.pdf we use `x` in https://github.com/KiCad/kicad-footprints/tree/master/Connector_Samtec.pretty. Since there are a number of different Pulse MPNs that can use this footprint, what was your idea for the footprint name?

Schlumpf:
The name came from the 2009 datasheet (linked on initial post). The last datasheet, that covers more than one device, I found. On page you can see the name in the mechanical drawing. I skipped the special character `/` . If you lilke I could replace the capital `X` by a lower `x`?

Shackmeister:
@schlumpf can you please have a look at the renaming? It seems like the final step before merge :)

poeschlr:
PR finished in https://github.com/KiCad/kicad-footprints/pull/875

PR #481, opened 2018-04-10
added LFPAK56 footprint

daviid5:
added LFPAK56 package, used for the PSMN... mosfet in this PR: https://github.com/KiCad/kicad-symbols/pull/445

lfpak56

Datasheet:
https://assets.nexperia.com/documents/outline-drawing/SOT669.pdf

------------


Ratfink:
Thank you for creating this footprint, but I believe it's redundant.  We already have `Package_TO_SOT_SMD:SOT-669_LFPAK`, which could perhaps use some updates (the datasheet link is broken, and it could use the LFPAK56 keyword) but is for the same package.  If you'd like to fix some of those issues that would be welcome, but I'm closing this PR because we don't need a second copy.

daviid5:
@Ratfink 
Should I change the name of the "SOT-669_LFPAK" to "LFPAK56" or just adding "LFPAK56" to the keyword list?
FYI: there is also a LFPAK33 package
My concern is, that nobody (including me) will find out that the "SOT-669_LFPAK" package is the same as "LFPAK56" ;)

Ratfink:
Following #313, I'd say it should be called simply `LFPAK56`, with the other names from the drawing added to the keywords.  Of course, the two _ThermalVias variants of this footprint would need to be renamed as well.  This change will require another PR at https://github.com/KiCad/kicad-packages3D to rename the 3D models, and it would be good to make another PR to update the script over at https://github.com/easyw/kicad-3d-models-in-freecad that made the model in the first place.  No symbols need to be changed because apparently no symbols already in the library use this footprint.

daviid5:
I now just updated the link to the PDF
In the PDF https://assets.nexperia.com/documents/outline-drawing/SOT669.pdf its also called SOT669, so I think its ok after all, maybe I should have read the datasheet more precise ... :D

evanshultz:
SOT-669 is a NXP/Nexperia name. I also vote for eliminating it in favor of LKPAK56.

Renesas is the only other vendor shipping LKPAK parts according to Digikey, and their drawing is at https://www.renesas.com/en-in/package-image/pdf/mountpad/fig1041e.pdf.

https://assets.nexperia.com/documents/outline-drawing/SOT669.pdf indicates PowerSO8 is an industry name, but https://assets.nexperia.com/documents/leaflet/75016838.pdf says using that name would not be a wise choice because it doesn't mean one specific thing amongst different manufacturers.

daviid5:
But as far as I understand this PDF
https://assets.nexperia.com/documents/leaflet/75016838.pdf

there is no "real" name for this package type, SOT669 and LFPAK56 or both names created by Nexperia, right?

Ratfink:
The decision to go with only the LFPAK* names for these was made in #313, where the LFPAK33 footprint was added.  We typically try to avoid NXP-/Nexperia-specific SOT* names because many of their packages also have industry standard names.

daviid5:
ok, so I now reamed it and created a new PR #482

PR #483, opened 2018-04-12
Add MSOP-8 package w/ large thermal pad for Micrel MIC5355

awygle:
Add MSOP-8 package w/ large thermal pad for Micrel MIC5355

Symbol PR: https://github.com/KiCad/kicad-symbols/pull/472
Datasheet: http://ww1.microchip.com/downloads/en/DeviceDoc/mic5355_6.pdf
Image:
image

image

image

image

------------


evanshultz:
Thank for adding the footprint!
- Please improve the description based on similar parts in the library. The datasheet link is fine, but include a description of the footprint
- The 3D model name must match the footprint name.
- The extents of the exposed pad under the IC are 2.36mm tall, so make the center pad at least that tall. Perhaps something like 2.8mm?
- Add thermal pad copper on the bottom side, same as top side.
- The outer pads are the wrong size and the wrong spacing. I assume you coped the existing MSOP footprint, but please change it to match the Micrel datasheet.
- Fix the courtyard after that is all done.
- Fab lines should be 0.1mm wide and silk lines 0.12mm.
- Silk should be offset 0.11-0.12mm from fab lines.
- 0.3mm vias are OK, but make the annular ring 0.6mm.

awygle:
I made the above modifications and also added the two DFN footprints per the request on the symbol PR.

evanshultz:
General:
- Footprint with vias can use the non-via 3D model (the package will be the same).
- Set silk offset from fab lines to 0.11-0.12mm.

MSOP:
- Fix pad X locations and all courtyard locations.
- When the pads move out, lengthen the pin 1 silk mark too. And move it down to -1.41mm in the X location so it matches the line on the right side of the footprint.
- The fab ref des can be 0.7mm square with a 0.1mm line width and still fit in the fab line outline.
image
- Make the paste 3 horizontal strips, not over the vias, which have a total coverage of the main thermal pad of around 70% (it can vary a bit so see what you can do).

MLF:
- In Microchip parlance, the footprint is called MLF so I would expect to use that name. There may be a general height difference between MLF and DFN. I found https://amkor.com/packaging/leadframe/microleadframe/, which seems ambiguous to me, though. So I'm not sure.
- Eliminate the vertical and horizontal silk lines by pin 1 to make the pin 1 mark.
- Avoid paste pads over the vias, if possible.
- Vertical courtyard is too big.
image

@Ratfink @poeschlr @jkriege2 
Any thoughts about the MLF vs DFN comment above? Use MLF since that's what this datasheet says?

Ratfink:
I'd be in favor of calling the package MLF if that's what the manufacturer calls it.

awygle:
I believe I've made all of the changes requested above.

evanshultz:
Thanks for making changes! Very close.

Both:
- Set silk offset from fab lines to 0.11-0.12mm. It's just a bit off (smaller and bigger) now.

MLF:
- Remove `(ML)` and `[DFN]` from the description.
- Paste coverage is only ~50%. Can you get it to at least 60%?
- Make the vertical silk lines a bit longer so they'll print more clearly. They can be 0.12-0.2mm from pads.

evanshultz:
Paste coverage is now 61%. Looks good. Thanks!

PR #485, opened 2018-04-13
Add TO252/263 SMD diode footprints

evanshultz:
For https://github.com/KiCad/kicad-symbols/pull/442.

Datasheets:
- http://www.onsemi.com/pub/Collateral/MBRD620CT-D.PDF
- https://www.wolfspeed.com/media/downloads/87/CSD01060.pdf
- https://www.wolfspeed.com/media/downloads/137/C3D06060G.pdf

Image:
image

One footprint in the TO_SOT_SMD lib was for a diode, so I moved it to the diode lib and named it as a common cathode footprint so it's easier to identify from the footprint name. The pin numbers are somewhat arbitrary because we have symbols for all different pinouts. It was not being used by any symbols so this should be safe.

------------


evanshultz:
I believe all Travis errors/warnings are OK (custom paste pads and using a generic 3D model).

jkriege2:
Hi!

I think som (more) time back, we decided to use the standard-TO-XYZ-packages for diodes and not double them in deth Diodes_SMD/THT-directories (In the very early lib, we e.g. had TO-220 for ICs+for transistors + in variants for diodes ... and I think that really makes it hard to manage the footprints). Therefore I would propose:
1. add version TO-252-2_TabPin1 (and if required TO-252-2_TabPin2) to Package_TO_SOT_SMD.pretty
2. the same for TO-263-2
3. The TO-252-2_Rectifier does not really make sense (it just has a bit larger pads) + nobody uses it, so I would rename it either to TO-252-2_LargePads or remove it alltogether.
4. FInally don't add any TO-XYZ footprints to Diodes_SMD/THT

Best,
JAN

evanshultz:
Thanks for looking! I didn't know or recall that. Sorry. Should be updated now.

I made SOT-23 diodes a while back. They have a diode symbol in F.Fab and different pinout from the regular SOT-23 but are otherwise the same. Should that be changed too?

jkriege2:
Hhmmm ... I think the question is here: do we want to have that logic in symbol (pin-numbers), or in the footprint (pad-numbers) ... and do we want to SOT-23 to comatible with the generic diode symbol ... I would say we should also remove them (but that will mean we'll possibly have to modify several symbols too...

cu
JAN

jkriege2:
Otherwise: nice work!

Thanks,
JAN

evanshultz:
I recall reading in KLC that the symbols should not be duplicated for pinout variations (or something to that affect) and this should instead be handled by having various footprints. That matches what you wrote above. Of course I can't find that KLC clause right now.

PR #486, opened 2018-04-14
Fix naming of "Hopref" footprints.

ppelleti:
The RFM9XW modules are made by the same company that makes the RFM69HW modules.  However, currently their footprints start with different manufacturer prefixes:

hopref

1. First off, the name of the company is "hoperf", not "hopref".

2. Secondly, the capitalization should be harmonized.  The company's [web site](http://www.hoperf.com/) puts the name in all caps, so "HOPERF" seems like the way to go.

3. Finally, the `HOPERF-RFM69HW` footprint uses a dash between the manufacturer name and part number, while `Hopref_RFM9XW_SMD` and `Hopref_RFM9XW_THT` use an underscore between the manufacturer name and part number.  I can't find anything in the KLC about which to use, and existing footprints seem to use a random mixture.  So, I went with the dash.

------------


jkriege2:
Hi!

actually there is a KLC-rule for that: http://kicad-pcb.org/libraries/klc/G1.3/

So please use the underscore ... a dash is only used to e.g. separate the pin-count, e.g. in `TO-220-5`.

Best,
JAN

ppelleti:
Thanks!  Somehow I missed that in the KLC.  (And the existing footprints do not set a very good example.)

That means the `HOPERF-RFM69HW` footprint needs to be renamed, as well.  I've gone ahead and renamed all three, so now they should be consistent.

PR #487, opened 2018-04-14
Led 5mm horz

thomasonw:
I am adding a footprint for the LED 5mm Horz with 3 pins.  It is based off of the 2-pin version already in place.

This is my 1st footprint submission, and if successful I will add the other matching offsets for the rest of the 5mm Horz versions.

Also:  I am unable to create 3D images, so have left the 3D field blank.  I do not know if this will be a problem or not.

image


thomasonw:
I have addressed the initial Travis feedback (Wonderful tool) - except, as noted, my inability to create 3D images.

thomasonw:
OK, have made the edits.  Thanks for the coaching.   Script sounds like a great idea, am not a python guy - but will look to see if I can make something of it w/o mucking it up too much...

jkriege2:
2018-04-15 07_46_20-footprint editor no active library

thomasonw:
OK, thank you for your patience.  I have made those corrections.

jkriege2:
Thanks,
JAN

Issue #426, opened 2018-03-12
What is the recommended folder for footprints and symbols?


arildj78:
I'm new to KiCad and have downloaded a nightly build to be able to do circuit simulation. One problem when entering this project is that I can't find the documentation on where to put symbols, footprints and models.

It is suggested to clone the different projects to get the latest change, but could you include some folder names for Win10 in the Read.me?

poeschlr:
There is no requirement. You can point kicad to look for libs in any location you like.

for footprints i wrote a description over at the forum https://forum.kicad.info/t/how-can-i-install-a-specific-version-of-the-footprint-library/9269

For symbols it is similar but instead of copying the fp-lib-table file you need to copy the sym-lib-table file.
Or use the symbol library wizard/manager found in eeschema->(setup or preferences menu. I don't know where they put it in the current nightly version.)

Ratfink:
Closing this, as it's been inactive for nearly two months, and as Rene pointed out, there is no specific recommended path for the library.

PR #489, opened 2018-04-14
Create QFN-24-1EP_4x4mm_P0.5mm_EP2.7x2.7mm.kicad_mod

robkam:
Copy of WQFN-24-1EP_4x4mm_P0.5mm_EP2.7x2.7mm.kicad_mod 

For AS33xxF symbols in PR [Add further packages for AS33xx and V33xx #426](https://github.com/KiCad/kicad-symbols/pull/426 )

http://www.alfarzpp.lv/eng/sc/AS3320.pdf for AS3320F
http://www.alfarzpp.lv/eng/sc/AS3330.pdf for AS3330F
http://www.alfarzpp.lv/eng/sc/AS3340.pdf for AS3345F

capture


poeschlr:
You state this is a copy of another footprint. Could you elaborate what the differences are  between this and the original?

robkam:
The only difference is this is QFN instead of WQFN, otherwise they're identical. Copied from [Package_DFN_QFN.pretty/WQFN-24-1EP_4x4mm_P0.5mm_EP2.7x2.7mm.kicad_mod ](https://github.com/KiCad/kicad-footprints/blob/master/Package_DFN_QFN.pretty/WQFN-24-1EP_4x4mm_P0.5mm_EP2.7x2.7mm.kicad_mod)

poeschlr:
Hm looks like the difference is in the height of the part -> different 3d models required -> so this should be ok then.


To be honest i think the exposed pad is on the small side. You chose the same size as the nominal "lead" on the package.

I think the last time this came up we looked into what IPC would recommend and we found that the maximum lead size is a good approximation. (IPC has no direct recommendation so we used the table from a lead style similar to an exposed pad.) 
In your case this would mean a pad size of 2.8x2.8mm.

robkam:
qfn-24-1ep_4x4mm_p0 5mm_ep2 8x2 8mm

evanshultz:
Your seed footprint has loads of custom pads which isn't necessary (that part specifically showed semi-circular pads but there isn't a reference footprint for the ALFA package that I find. Does ALFA have a reference footprint?

robkam:
I've not heard back from Alfa regarding whether they have a reference footprint for the AS3320F, AS3330F or AS3345F parts. On the data sheets for these parts the bottom view does have semi-circular pads.

robkam:
There's a reference footprint for QFN-24L at http://www.topline.tv/drawings/pdf/QFN/SLP24M.5.pdf

robkam:
With footprint wizard:

http://www.topline.tv/drawings/pdf/QFN/SLP24M.5.pdf

qfn-24l

poeschlr:
Your choice for the pad sizes and position is a bit strange. How did you derive it?
The heel fillet is not only non existent but extremely negative. (Even before accounting for tolerances it is at least -0.2)

---

screenshot from 2018-05-06 12-12-30

robkam:
@poeschlr I used the QFN wizard

poeschlr:
What qfn wizard?

robkam:
@poeschlr 
qfn wizard

poeschlr:
That wizard does not calculations to determine how large your pads should be and where to place them. Meaning you must do that on your own. So the question remains. How did you determine the pad placement and size.

(As you might already guess i am convinced it is wrong. You need to convince me of the opposite such that i would merge it as is.)

poeschlr:
Additional info: i looked at the code of that particular wizard. It does not come with enough degrees of freedom to make a good footprint. In my mind at least this one wizard should be removed from kicad. 

It always centers the pads around the body size. Meaning there is no way to control the heel and toe fillet independently. (you can only control it via the size of the pad. Half of it will be the toe fillet. The heel fillet will be approximately length of pad / 2 - size of pad/2)

robkam:
I've given up on the wizard, latest version is instead a modified copy of [Package_DFN_QFN.pretty/WQFN-24-1EP_4x4mm_P0.5mm_EP2.7x2.7mm.kicad_mod](https://github.com/KiCad/kicad-footprints/blob/master/Package_DFN_QFN.pretty/WQFN-24-1EP_4x4mm_P0.5mm_EP2.7x2.7mm.kicad_mod)

Data sheet for reference footprint at http://www.topline.tv/drawings/pdf/QFN/SLP24M.5.pdf

qfn-24-4x4x0 9-p5-1ep-2 8x2 8

robkam:
please merge

evanshultz:
- Your link is a package drawing, not a footprint. You can find some very similar ones (check out the pad width and length) at http://www.analog.com/media/en/package-pcb-resources/package/pkg_pdf/ltc-legacy-qfn/(UF24)%20QFN%2005-08-1697%20Rev%20B.pdf, https://www.idt.com/document/zssc3018-technical-brief-qfn24-package-specifications, http://www.ftdichip.com/Support/Documents/TechnicalNotes/TN_166%20FTDI%20Example%20IC%20PCB%20Footprints.pdf, http://www.aosmd.com/pdfs/datasheet/AOZ1110QI.pdf, etc. Those may have a different thermal pad size but you can find info on pads 1-24 from them.
- Can you push the silk to 0.11 outside the fab lines? It's a bit far away right now.
- Pads 1 through 24 mention a custom shape but they're not. They should be "rect" if you want this shape.
- Four 1.1mm square paste pads gives ~62% coverage on a 2.8mm square copper pad. That's a bit low. Shoot for ~75%.

robkam:
capture

evanshultz:
- You changed the link to another package drawing. Please use a footprint drawing.
- You changed the paste pads from 1.1mm square, which I said above was too small, to 1.05mm square.

robkam:
I can't find a footprint drawing to link to.

poeschlr:
@evanshultz paste coverage 65% is right in the middle of the suggested ranges i could find when i researched it. (The suggestions vary between 50% to 80%) 75% would be ok but a bit high for my liking. (If you go that high you might need to have varied thickness stencils to avoid floating of your parts.)

poeschlr:
@robkam the clearance between the pads and the exposed pad should be at least 0.2mm according to IPC. (I might add this to the KLC.)

poeschlr:
My advice regarding exposed pad size was wrong. See https://github.com/KiCad/kicad-footprints/issues/678

So this should use 2.7x2.7mm ep size. Sorry about the confusion caused by this.

If you want i can also add this to my generator script which would make future maintenance a lot easier.

robkam:
Please do add to your generator script.

poeschlr:
Done: https://github.com/KiCad/kicad-footprints/pull/700

evanshultz:
Closing as https://github.com/KiCad/kicad-footprints/pull/700 is now merged.

PR #493, opened 2018-04-15
Dtf13 12 px

thomasonw:
This submission is for DTF13-12Px connector as used in the transportation sector.

[http://www.te.com/usa-en/product-DTF13-12PA-G003.html](http://www.te.com/usa-en/product-DTF13-12PA-G003.html)
 
image


poeschlr:
Having a circle on the courtyard layer is not a good idea as it makes it hard to get a closed polygon. KiCad 5 will offer a couryard check. I am not sure it will work well with open polygons.

So i would suggest a simplification of the courtyard area by approximating the circular part via a rectangle or part of an octagon:
screenshot from 2018-04-17 12-22-03

The courtyard clearance should be 0.5mm all around. Tolerance: +/- 0.05mm to result on the 0.01mm grid. (you use a mixture of measurements between 0.2 and 0.4mm)

---

For non plated through holes (NPTH) the drill and pad diameter must be equal. (Affects your two holes for the stabilizing pins.)

How did you determine the drill size and annular ring? 
The drill size seems ok as you result in 0.16mm remaining drill size tolerance for the largest possible pin diameter.
The annular ring seems a bit excessive to be honest. (>0.8mm) maybe reduce it down to ~0.5mm which would increase the pad to pad clearance. (Increased voltage tolerance or more place for traces between the pads.)

The drill for the stabilizing pin is given with 1.96. You used 2.375 which is a lot larger. I would suggest to either use the value given in the datasheet or use 2mm.


---

For reference the current dimensioned sketch of the footprint:
screenshot from 2018-04-17 12-40-50

poeschlr:
Also fix the 3d path settings. you point to the Connector**s**.3dshapes lib instead of the singular Connector.3dshapes

Ratfink:
It looks from the datasheet that the mounting tabs are the same size as one another, but the one on the left is bigger in this footprint.  Their radii are only given implicitly in TE's drawing, from the width of the component in the upper left and the hole spacing in the lower right.

Also, it looks like you have the grid of pins slightly off-center, but that should be centered according to the drawing.

thomasonw:
@poeschlr   Hello, and thank you for the feedback.  To your questions:

1) How did I determine drill size?  I have not been able to locate a suggested PCB layout for this connector, so in this case I took the stated pin size of 0.0625” + the 0.0020” tolerance and rounded it up with soldering margin.  The resulting hole size if VERY close to 1.75mm, so I will edit the footprint to use that and a matching 3mm ring size.  (Noted on other connectors as a common paring).

2) Locating hole:  Will make NPTH pad and drill size the same.  Size was taken from the speced #2 mounting screw for the component and the standard machinist hole size for clearance of a #2 screw.    The resulting value is: 2.3749mm, given most of KiCAD seems to be sized in metric will set this at 2.4mm drill size. The 1.96mm hole size as shown on the connector is the self-taping size needed to allow a #2 screw to take hold in the plastic ears.

3) Courtyard:  Will change to box.  Now here is a question:  Hand-drawing the court yards is a PITA – does anyone have some coaching on how to do this?  Is there any way to allow simple and accurate courtyard creation?  Hopefully I have drawn these courtyards better.

4) ConnectorS.  vs. connector.... Thought I had fixed that (Travis noted it), perhaps it did not get pushed up.


@Ratfink  Thank you for sleuthing out the radius of the mounting ears, that approach obviously escaped me.  Will adjust the outlines to that value.  Not sure I understand your comment about grid of pins being off-center, I did align the center-line around pin 1 for KiCAD vs. the CL shown in the drawing..  But is there something else I am missing?


And on pin1:  I realized the pins were backwards:  I believe the drawing is showing the bottom view of the connector, and hence needed to mirror the pins.  Can folks confirm this is the correct interpretation of the drawing?

thomasonw:
@poeschlr   Hello, and thank you for the feedback.  To your questions:

1) How did I determine drill size?  I have not been able to locate a suggested PCB layout for this connector, so in this case I took the stated pin size of 0.0625” + the 0.0020” tolerance and rounded it up with soldering margin.  The resulting hole size if VERY close to 1.75mm, so I will edit the footprint to use that and a matching 3mm ring size.  (Noted on other connectors as a common paring).

2) Locating hole:  Will make NPTH pad and drill size the same.  Size was taken from the speced #2 mounting screw for the component and the standard machinist hole size for clearance of a #2 screw.    The resulting value is: 2.3749mm, given most of KiCAD seems to be sized in metric will set this at 2.4mm drill size. The 1.96mm hole size as shown on the connector is the self-taping size needed to allow a #2 screw to take hold in the plastic ears.

3) Courtyard:  Will change to box.  Now here is a question:  Hand-drawing the court yards is a PITA – does anyone have some coaching on how to do this?  Is there any way to allow simple and accurate courtyard creation?  Hopefully I have drawn these courtyards better.

4) ConnectorS.  vs. connector.... Thought I had fixed that (Travis noted it), perhaps it did not get pushed up.


@Ratfink  Thank you for sleuthing out the radius of the mounting ears, that approach obviously escaped me.  Will adjust the outlines to that value.  Not sure I understand your comment about grid of pins being off-center, I did align the center-line around pin 1 for KiCAD vs. the CL shown in the drawing..  But is there something else I am missing?


And on pin1:  I realized the pins were backwards:  I believe the drawing is showing the bottom view of the connector, and hence needed to mirror the pins.  Can folks confirm this is the correct interpretation of the drawing?
 


Thank you both for the review of this.

Ratfink:
The problem with the pin positions is that they're supposed to be centered relative to the part body, but right now you have them visibly shifted slightly to the left.

Also, having now consumed a less unreasonably small amount of coffee, I notice that the way of getting the mounting ear radius I suggested above is not quite right, because the width given is for the latches, not the ears.  So I just checked the .dxf file from TE, and the radius appears to be 3.175 mm (1/8 in.).

As for pin numbering, it looks like from left to right, the top row should be 1-6, and the bottom row should be 12-7.

thomasonw:
Forest for the trees – yes, is obvious something is now centered.  Tks for looking at the DXF file, I have no ability to do that.  One question, if you can dig this out – I am not able to determine how ‘long’ the latch  ears are.  I can see they are 0.675” from the front to the far end of the latch ears, but no idea what the dimension is for the begging of them.   Would you be able to pull that dimension out?

 

Tks for the pins, seems we are reading them in the same order.

Ratfink:
The latches are 0.580 in. long in the DXF, and the back edge is 0.765 in. from the front of the connector (you transposed two digits, so I'm being overly careful).

You can load DXF files and measure them like I'm doing if you install [FreeCAD](https://freecadweb.org/), which is available for Linux, OS X, and Windows.  It's a pretty advanced program, and I can't possibly claim to be very experienced with it, but for just taking measurements from DXFs it's not hard to figure out.  I just use the Draft workbench's [dimension tool](https://freecadweb.org/wiki/Draft_Dimension).

thomasonw:
@Ratfink   Thanks for the dim.  Have updated the file to match that.   Will have to take some time and learn the basics of FreeCAD sometime - on my list of things to do. . . .  Tks for the help on this, think it should be close now.

poeschlr:
The right mount pad still has a difference between drillsize and pad size.

---

Your courtyard clearances are still all over the place.

screenshot from 2018-04-17 22-47-41

thomasonw:
Adjusted.   This time I used the CL of the outline, adding 0.5mm and using 'IRS' rounding rules to get to 0.01 grid.   Is there truly no other way to create these courtyards outside calculate each individual lines end points?

poeschlr:
Looks good now. Thanks.

---

Regarding your question about automatic drawing the courtyard:

We have a pyhton script in the kicad-library-utils repo that can draw a simple rectangle automatically. But i don't think there is something out there that could draw such an elaborate outline automatically.

You could use for example freecad plus kicad-stepup toolchain to design such complex footprints. More details see: https://forum.kicad.info/t/kicad-stepup-the-sketcher-for-footprint-generation/9582

PR #465, opened 2018-04-02
3M Textool 16-pin zero insertion force socket

ppelleti:
[Datasheet](http://multimedia.3m.com/mws/media/494546O/3mtm-dip-sockets-100-2-54-mm-ts0365.pdf).

Based on existing footprint `3M_Textool_240-1288-00-0602J_2x20_P2.54mm`.

zif16

------------

Thanks for creating a pull request to contribute to the KiCad libraries! To speed up integration of your PR, please check the following items:

- [ ] Provide a URL to a datasheet for the footprint(s) you are contributing
- [ ] An example screenshot image is very helpful 
- [ ] If there are matching symbol or 3D model pull requests, provide link(s) as appropriate
- [ ] Check the output of the Travis automated check scripts - fix any errors as required

PR #495, opened 2018-04-17
More 3M Textool ZIF sockets, generated with script.

ppelleti:
This adds several more [3M Textool](http://multimedia.3m.com/mws/media/494546O/3mtm-dip-sockets-100-2-54-mm-ts0365.pdf) ZIF sockets, using the script I mentioned in my [previous PR](https://github.com/KiCad/kicad-footprints/pull/465).

------------


ppelleti:
Courtyard should be fixed now.

ppelleti:
> All footprints for parts with y pitch = 15.24 have it as 15.25

Looks like you are going by the lower table, while I am going by the upper table.  It's not clear which one should be preferred.

textool-mismatch

> The handle seems a bit long. It should protrude by 12.3 instead of 14.84

OK, thanks, I will fix that.

> We might want to include a bit more info into the footprint names. I would at least include the body size.

Sure.  Can you give an example of the name you'd like to have?

poeschlr:
0.6 inch = 15.24mm. The 15.25 seem to be wrong. (But this alone would not be enough to merit remaking the footprints.)

---

About naming:

We don't really have a rule that fits this part exactly. But i would go with something similar to what we use for tht ic packages. (Rule for ICs: http://kicad-pcb.org/libraries/klc/F3.5/)
Keeping the manufacturer part number makes sense.

I would change the pin count from 02x[num per row] to simply the number of pins.

So in this case maybe:
`[Man]_[MPN]_DIP_Socket-[Pincount]_[X]x[Y]x[Z]_P[Pitch]`

So something like:
`3M_Textool_214-3339-00-0602J_DIP_Socket-14_15.2x33mm_P2.54mm`

Ratfink:
Wouldn't it be nice to try to name these footprints so they match common DIP fpfilters, e.g. `DIP*W7.62mm*`?  For sockets that accept multiple widths, several `W…` fields could be used.  To fit other DIP socket naming, I might suggest putting the manufacturer info at the end.

ppelleti:
@poeschlr and @Ratfink , have you reached a consensus about the naming?

Ratfink:
We haven't had any discussion outside this PR.

My suggestion to match DIP FPFilters, now actually formulated:  
`DIP_Socket-[Pincount]_W[Min width](_W[2.54 mm increments])_W[Max width]_[Man]_[MPN]`

There are some DIP parts with unusual widths, e.g. 10.16 mm, which is why I suggest adding extra widths at 2.54 mm increments, as this would cover many more DIP parts.  This has the drawback of making the names very long and ugly.  More minimally, they could list extra widths at 7.62 mm increments, or even just the min and max.

I didn't include the pitch because we don't in our regular DIP footprints.

ppelleti:
The script that generates these is available on my [zif-poeschlr](https://github.com/ppelleti/kicad-footprint-generator/tree/zif-poeschlr/scripts/Socket) and [zif-ratfink](https://github.com/ppelleti/kicad-footprint-generator/tree/zif-ratfink/scripts/Socket) branches of `kicad-footprint-generator`.

ppelleti:
> Sorry it took so long to get back to you. The naming by @Ratfink is ok to me. (I seem to have missed that he made a suggestion.)

OK, then `zif-ratfink` is the branch you need.  Is there anything more you need from me?

poeschlr:
Would it be possible to create a pull request over on the footpirnt generator repo and either update this pull request with the new footprints or create a new pull request here?

poeschlr:
I finally found time to look at this. Seems good to me now. Thanks

PR #496, opened 2018-04-17
add LGA-16 footprint

ChristianHirsch:
Add LGA-16 footprint based on [MMC5883 datasheet](http://www.memsic.com/userfiles/files/DataSheets/Magnetic-Sensors-Datasheets/MMC5883MA-RevC.pdf).

image


poeschlr:
Looks good. Thanks.

For reference the dimensioned drawing i created while checking this footprint:
screenshot from 2018-04-17 11-54-36

PR #369, opened 2018-02-13
added WSON-8-1EP_5x6mm_P1.27mm_EP3.4x4.0mm fp

cyber-murmel:
Footprint for e.g. 1.8V Flash Memory Chip 

------------

Thanks for creating a pull request to contribute to the KiCad libraries! To speed up integration of your PR, please check the following items:

- [x] Provide a URL to a datasheet for the footprint(s) you are contributing
[Recommended Land Pattern for SST’s, SOIC and WSON Packages](http://ww1.microchip.com/downloads/en/AppNotes/S72030.pdf) page 2
[IS25WP032-064-128](https://www.mouser.com/ds/2/198/IS25WP032-064-128-737458.pdf) page 101
- [x] An example screenshot image is very helpful
screenshot
- [x] If there are matching symbol or 3D model PRs, provide link(s) as appropriate
not yet
- [x] Check the output of the Travis automated check scripts - fix any errors as required

PR #501, opened 2018-04-20
added WSON-8-1EP_5x6mm_P1.27mm_EP3.4x4.0mm fp

cyber-murmel:
Footprint for e.g. 1.8V Flash Memory Chip 

------------

[Recommended Land Pattern for SST’s, SOIC and WSON Packages](http://ww1.microchip.com/downloads/en/AppNotes/S72030.pdf) page 2
[IS25WP032-064-128](https://www.mouser.com/ds/2/198/IS25WP032-064-128-737458.pdf) page 101
screenshot
not yet


cyber-murmel:
Was https://github.com/KiCad/kicad-footprints/pull/369

poeschlr:
Ok now a proper review.

A point about the exposed pad:
We typically have one large copper pad for it. (It can also define mask if there is no special requirement for the mask cutout. Which is the case for this part.)
For the paste layer we use past only pads. These are pads without a pad number, no copper only paste selected in the layer selection list.
Aim for a paste coverage between 50 and 80% (Unless specified in the datasheet.)
You can check any of the other footprints with an exposed pad for inspiration.

---

The courtyard rectangle should have a clearance of 0.25 all around.
You currently have 0.2 in y direction and 0.4 in x direction.

screenshot from 2018-04-20 15-43-59

cyber-murmel:
0.25 inches I guess?

evanshultz:
All dimensions are in mm. :)

cyber-murmel:
Bump.
Sorry that it took me so long.
Footprint should better now.

evanshultz:
Please use more descriptive commit messages in the future. And it's easier for librarians if you do incremental updates instead of nuking everything with each commit.

- Body size is XxY, so it should be `6x5mm`. Fix in footprint name, keywords, 3D model, and description.
- Remove the `;` from the description.
- Include the Microchip doc since it includes all body and footprint info; datasheet not needed then.
- Why is there a paste margin? Unless you found a specific requirement, set this to the default `0`.
- Extend silk lines to entire body width. You can cut back the silk by pin 1 to mark that pin.
- For simplicity, if you don't mind, I'd prefer pads with 0 degrees rotation.
- Your paste pads are OK now, but only because the paste margin is set on the entire footprint (which is wrong). After you fix that, make the paste pads 1.5x1.7mm. This maintains 75% paste coverage.
- The 3D model should match the footprint name here since all differences (presence of thermal pad, thermal pad size) are relevant 3D model parameters. It seems there's also a typo in the 3D model name.

Misca1234:
This push is replaced by a script based version ?
https://github.com/KiCad/kicad-footprints/pull/794

PR #502, opened 2018-04-22
Use NPTH for Adafruit_HUZZAH_ESP8266_breakout_WithMountingHoles.

ppelleti:
I [originally created this footprint](https://github.com/KiCad/Modules.pretty/pull/12) for the [Adafruit HUZZAH ESP8266 Breakout](https://www.adafruit.com/product/2471) with plated mounting holes.  However, this confuses CvPCB, and it thinks the footprint has a different number of pins than the library symbol, so it doesn't show up unless you uncheck the "#" button in CvPCB.

So, this PR changes the type of the mounting holes to NPTH.  This also matches the other `*_WithMountingHoles` footprints in this library, which use NPTH mounting holes.

huzzah-npth

------------


Ratfink:
Thanks for this fix!

PR #506, opened 2018-04-23
Add QFN-28_4x4mm_P0.5mm

Ratfink:
Needed for QFN28 STM32 microcontrollers.  The Travis warning is bogus, due to https://github.com/KiCad/kicad-library-utils/pull/222 not being merged.

Datasheet: http://www.st.com/resource/en/datasheet/stm32f031k6.pdf (figure 41)

screenshot from 2018-04-23 17-03-29

------------


poeschlr:
Everything seems to be ok. (Checked the footprint as well as i can with the limitation that stepup can not work with polygon pads.)

Yea i knew about the fact that i can restart travis from within the ci webside but i did not think about that when i wanted it to rerun.

Anyhow thanks for this fix.

poeschlr:
Ok i updated stepup as i discovered we can now handle polygon pads. Here now the fully dimensioned sketch of the footprint:

It indeed shows that the footprint is done correctly ;) (Lets ignore that the clearance between the corner pads is a bit larger then in the datasheet.)
screenshot from 2018-04-23 23-52-43

Issue #508, opened 2018-04-24
Package_SO.pretty/TSOP-5_1.65x3.05mm_P0.95mm.kicad_mod using pin 6 instead of 5


MarsWarrior:
This 5 pin TSOP package uses pins 1,2,3,4 and 6.

The symbols expect pin 5 instead of 6.

poeschlr:
See PR #509

MarsWarrior:
Ah! I was too late entering my own PR :grinning:

PR #509, opened 2018-04-24
TSOP-5 should have pin numbers 1...5 as in the datasheet.

poeschlr:
The datasheet lists pins with numbers 1...5 https://www.vishay.com/docs/71200/71200.pdf.
The footprint had 1..4+6

From the datasheet
screenshot from 2018-04-24 10-44-10

Reported on the forum:
https://forum.kicad.info/t/sot-23-5-6th-pad-in-v5rc/10443/6

---

Edit:

This footprint is not used as the default footprint for any of our symbols.

This is a fix for #508


Ratfink:
Thanks for fixing this.

PR #510, opened 2018-04-24
Fix ep sizes for cypress footprints (qfn-56)

poeschlr:
As we discovered recently, using the nominal lead size as the size for the EP pad might be a bit small. A good approximation for fulfilling IPC standards is to use the maximum size listed for the lead.

This is the footprint side of https://github.com/KiCad/kicad-symbols/pull/532


Ratfink:
Looks good, thanks.

Will you be fixing the 3D model for QFN-56-1EP_7x7mm_P0.4mm_EP5.6x5.6mm as well?

PR #511, opened 2018-04-24
TEN 20 does not have a package with missing pins

poeschlr:
The TEN 20 converter has no variant with a missing pin.
The pin that is used as common for the dual output version is used as the trim input for the single output version.

For this reason i removed the dual and single version and only left the default option to avoid confusing users.


Ratfink:
According to http://assets.tracopower.com/20180424165119/TEN20/documents/ten20-datasheet.pdf#page=3, the single output TEN 20 converters have no pin 4.  TEN 20WIN converters do use pin 4 as an adjustment for the single output devices. (https://assets.tracopower.com/20180424165119/TEN20WIN/documents/ten20win-datasheet.pdf#page=4).

I'd suggest using better naming somehow, since some single devices use the 6-pin footprint and some use the 5-pin one.  Maybe something along these lines?

* `Converter_DCDC_TRACO_TEN20_THT`
* `Converter_DCDC_TRACO_TEN20-N4_THT`

Ratfink:
Thanks!

PR #512, opened 2018-04-24
Adding script generated VQFP footprints withe measurements by microsemi

poeschlr:
This adds the first patch of footprints needed for https://github.com/KiCad/kicad-symbols/pull/534

Size definition from http://www.microsemi.com/index.php?option=com_docman&task=doc_download&gid=131095 page 50

sfdsfsdf svg


Ratfink:
Overall these look very good.  I'm a little surprised by the pad sizes for VQFP-80_14x14mm_P0.65mm, as they look awfully wide, but your calculations and data look right to me.

One little thing: your script seems to be putting some extra spaces at the beginnings of the descriptions and tags.  Not a big deal, just something you might want to fix to make the script perfect.

poeschlr:
Yes the 80pin version seemed strange to me as well.
The reason is that it is the only one of these that falls into the larger IPC category (pitch > 0.625)

And it has a very wide tolerance range for the lead width.

In detail:
lead width = 0.22 .. 0.38 (tolerance = 0.16)
The side fillet is 0.03 (nominal density, gull wing, pitch > 0.625)
With my choice of manufacturing tolerance = 0.1 and placement tolerance = 0.05 (which are the values suggested in the IPC document) we can calculate the pad width as follows:

W = 0.22 + 2*0.03+ sqrt(0.16**2 + 0.1**2 + 0.05**2) = 0.4752
Rounded to nearest 0.05 results in a width of 0.5 (again a requirement by the ipc standard)

Reducing either the alowed manufacturing tolerance or placement tolerance by a bit would be enough to at least get the pad with to 0.45 (even using 0.099 as manufacturing tolerance already results in this)

---

The spaces come from the fact that i reuse the description generator from the soic scripts.

poeschlr:
I fixed the whitespace problem and i changed the qfp data file such that it results in thinner pads for the 80 pin version.

PR #513, opened 2018-04-24
Script generated PQFP footprints (landing pattern IPC, sizes microsemi)

poeschlr:
This adds another patch of footprints needed for KiCad/kicad-symbols#534

Dimensions from http://www.microsemi.com/index.php?option=com_docman&task=doc_download&gid=131095 page 52

erwerwer svg


poeschlr:
Comparison between the old footprints and the newly generated ones:

PQFP-100 has larger pads now (red: old pads, yellow: new)
screenshot from 2018-04-24 18-55-42

PQFP-160 also has now larger pads (red: old, yellow:new)
screenshot from 2018-04-24 19-41-41

For the PQFP-208 the new footprint is slightly shorter on the outside otherwise they are identical (smaller toe fillet) again old is red, new is yellow but because of the overlay it results as orange
screenshot from 2018-04-24 19-45-54

Ratfink:
Looks good!  I see that the only other parts that use the footprints that now have wider pads are obsolete, so I'm not worried about that.  Also, I notice that the PQFPs have even worse lead width tolerances than the VQFPs, so it's harder to convince myself that fudging it like you did in #512 is a good idea.  So I'd say that this looks ready to merge.

PR #514, opened 2018-04-24
Newly scripted TQFP footprints (landing pattern IPC, dimension microsemi)

poeschlr:
This adds another patch of footprints needed for KiCad/kicad-symbols#534

Dimensions from http://www.microsemi.com/index.php?option=com_docman&task=doc_download&gid=131095 page 50

---

Comparison with old footprints old pads: red, new pads yellow
All new pads have less toe more heel. The 100 pin variant has quite a large difference. The 64 and 144 pin variant are very similar:

64 pin
screenshot from 2018-04-24 20-15-07
100 pin
screenshot from 2018-04-24 20-16-07
144 pin
screenshot from 2018-04-24 20-16-44


poeschlr:
Dimensioned drawings
asdf svg

Ratfink:
Looks great!  Thanks for making a script to produce such beautiful footprints.

PR #517, opened 2018-04-25
Add Amphenol RJMG1BD3B8K1ANR Magjack.

ppelleti:
Footprint for [Amphenol RJMG1BD3B8K1ANR Magjack](https://www.amphenolcanada.com/ProductSearch/Drawings/AC/RJMG1BD3B8K1ANR.PDF).

This footprint is meant to be used with [this symbol](https://github.com/KiCad/kicad-symbols/pull/535).

The KLC script does not like this pin numbering scheme, since there is no pin "1".  But this is the scheme in the manufacturer's datasheet.

rj45-footprint

------------


poeschlr:
Sorry it took us so long. We where quite busy getting the lib ready for v5.

I assume the cross hatched areas in the dimensioned drawings are keepout areas. These should also be represented here. Details about keepout areas see: http://kicad-pcb.org/libraries/klc/F4.5/

---

The footprint looks good otherwise. For reference the dimensioned drawing:
screenshot from 2018-07-28 03-14-44

poeschlr:
I looked at this again and i now think that the marked areas are just the contact points of the connector. I assume they are made from some sort of plastic as they are not really dimensioned. So the footprint should be ok as is. Thanks for your contribution.

PR #521, opened 2018-04-25
Add missing x2son package (kicad stepup generated)

poeschlr:
Another missing footprint as reported in https://github.com/KiCad/kicad-symbols/issues/520

This footprint is needed for the TPS3831 symbol. (Edit: merging this should fix both symbols missing a footprint in Power_Supervisor)
Datasheet: http://www.ti.com/lit/ds/sbvs193d/sbvs193d.pdf

Generated using freecad and kicad stepup. The freecad file is included. I don't think we will need to split them out of the repo as of now. (If we get a lot of such contributions we might consider a footprint-source repo.)


poeschlr:
Footprint is updated

screenshot from 2018-04-27 20-13-31

Ratfink:
Looks great, thanks for fixing!

The one thing I wonder is if the F.Fab RefDes should be reduced in size and centered on the chip body.  But I see that we don't do that for very small parts, so if you think it's better this way I could merge as-is.

poeschlr:
I just noticed a general problem with how we handle special mask and paste pads.
Sadly the project clearance is used if one has a pad with 0 clearance. This means that if we define an exact mask cutout using pads (like we do here) and the user sets a clearance, it is used as a clearance even if that pad has no copper.

I fear we will need to find a small clearance value to guarantee that the project settings do not interfere with these footprints.

poeschlr:
If we assume that kicad 5 will behave the same as my current nightly (on all platforms) then my workaround for the global clearance setting problem should be ok that way.

poeschlr:
I have conformation that 1nm is indeed the minimum. So this should be ok to merge.

evanshultz:
Isn't this footprint already present at https://github.com/KiCad/kicad-footprints/blob/master/Package_SON.pretty/Texas_X2SON-4_1x1mm_P0.65mm.kicad_mod? Using poly pads is nicer, but I don't think this footprint was actually missing. It looks like the relevant symbols at https://github.com/KiCad/kicad-symbols/blob/master/Power_Supervisor.lib were only missing the `Texas_` prefix.

poeschlr:
The fix for the "missing" footprint happens now in the symbol https://github.com/KiCad/kicad-symbols/pull/561

This one will then simply increase the quality of said footprint but it will have to wait till my large workaround PR has been merged as i am not really interested in producing a merge conflict for that one.

poeschlr:
This should now be ready to go.

Ratfink:
The module name in the file is set incorrectly, so Travis is complaining a lot.  Way more than I expected it to for that problem, actually.  Please fix that, then I think this should be ready.

poeschlr:
Hm is there a kicad bug? I saved the footprint in the normal way, it should not have added the lib name. I need to look into this in more detail as @SchrodingersGat had a similar problem recently. (Very strange behaviour)

poeschlr:
I played around a bit. It seems no matter how i save a footprint in the footprint editor with todays nightly it includes the lib name. I asked the devs if this is on purpose or if it is a bug.

Ratfink:
Thanks for all the work on this footprint!

PR #516, opened 2018-04-24
Add STM32 BGA packages

Ratfink:
Generated by the script at https://github.com/pointhi/kicad-footprint-generator/pull/126

Datasheets can be found in https://github.com/KiCad/kicad-footprints/issues/374

screenshot from 2018-04-24 17-35-44

------------

Thanks for creating a pull request to contribute to the KiCad libraries! To speed up integration of your PR, please check the following items:

- [X] Provide a URL to a datasheet for the footprint(s) you are contributing
- [X] An example screenshot image is very helpful 
- [X] If there are matching symbol or 3D model pull requests, provide link(s) as appropriate
- [ ] Check the output of the Travis automated check scripts - fix any errors as required

Issue #374, opened 2018-02-15
Missing footprints for STM32 microcontrollers


Ratfink:
Now that https://github.com/KiCad/kicad-symbols/pull/305 and https://github.com/KiCad/kicad-library-utils/pull/212 have been merged, there are quite a few missing footprints needed for the STM32 microcontroller family.  Below is a list of the needed footprints, with links to datasheets that contain package/footprint drawings.  Naming for WLCSP footprints follows the suggestion [here](https://github.com/KiCad/Housings_CSP.pretty/issues/8#issuecomment-325613650), but without the ball pitch because the packages are already uniquely identified without it.

- [x] Package_BGA:LFBGA-100_10x10mm_Layout10x10_P0.8mm: http://www.st.com/st-web-ui/static/active/en/resource/technical/document/datasheet/CD00161566.pdf
- [x] Package_BGA:LFBGA-144_10x10mm_Layout12x12_P0.8mm: http://www.st.com/st-web-ui/static/active/en/resource/technical/document/datasheet/CD00191185.pdf
- [x] Package_BGA:TFBGA-64_5x5mm_Layout8x8_P0.5mm: http://www.st.com/st-web-ui/static/active/en/resource/technical/document/datasheet/CD00251732.pdf
- [x] Package_BGA:TFBGA-100_8x8mm_Layout10x10_P0.8mm: http://www.st.com/st-web-ui/static/active/en/resource/technical/document/datasheet/DM00166116.pdf
- [x] ~~Package_BGA:TFBGA-216_16x16mm_Layout15x15_P0.8mm~~ Package_BGA:TFBGA-216_13x13mm_Layout15x15_P0.8mm http://www.st.com/st-web-ui/static/active/en/resource/technical/document/datasheet/DM00071990.pdf
- [x] Package_BGA:TFBGA-265_14x14mm_Layout17x17_P0.8mm: http://www.st.com/st-web-ui/static/active/en/resource/technical/document/datasheet/DM00387108.pdf
- [x] Package_BGA:UFBGA-64_5x5mm_Layout8x8_P0.5mm: http://www.st.com/st-web-ui/static/active/en/resource/technical/document/datasheet/DM00039193.pdf
- [x] Package_BGA:UFBGA-100_7x7mm_Layout12x12_P0.5mm: http://www.st.com/st-web-ui/static/active/en/resource/technical/document/datasheet/DM00098745.pdf
[...]

PR #522, opened 2018-04-25
Added STM32 WLCSP footprints

Ratfink:
Right now these use the perhaps overly-precise nominal dimensions of the parts for the fab and silkscreen outlines.  Let me know if you think those should be rounded, and if so I'll change the script to do that.

Generated by the script at https://github.com/pointhi/kicad-footprint-generator/pull/126

Datasheets can be found in https://github.com/KiCad/kicad-footprints/issues/374

No screenshot because holy mackerel it's 40 tiny BGA-like footprints, you would barely be able to make out the pads!  Plus, it's the same script that did #516, so they look very similar to that.

------------


poeschlr:
Connected with my question over at the generator pr:

I found two differently named footprints that seem to be equal.
ST_WLCSP-36_Die440 and ST_WLCSP-36_Die445
Do i miss something?

Ratfink:
I did notice that there were a couple of footprints with identical parameters while I was creating these.  However, that is consistent with ST's naming:

* [ST_WLCSP-36_Die440](http://www.st.com/content/st_com/en/products/microcontrollers/stm32-32-bit-arm-cortex-mcus/stm32-mainstream-mcus/stm32f0-series/stm32f0x1/stm32f051t8.html#samplebuy-scroll)
* [ST_WLCSP-36_Die445](http://www.st.com/content/st_com/en/products/microcontrollers/stm32-32-bit-arm-cortex-mcus/stm32-mainstream-mcus/stm32f0-series/stm32f0x2/stm32f042t6.html#samplebuy-scroll)

poeschlr:
You seem to have calculated the paste ratio by hand. This is quite error prone. It might be a better idea to enter the desired paste diameter and let python calculate the ratio.

---

I checked the datafile of commit https://github.com/pointhi/kicad-footprint-generator/pull/126/commits/cd4ceb721b92644dcc8391f9d70b980f76b5a0a4

The body size for ST_WLCSP-49_Die433 is different from the datasheet you use in the description field (3.06 vs 3.029)

ST_WLCSP-49_Die435 has the wrong pitch (0.4 vs 0.35)

ST_WLCSP-49_Die448 the datasheet does not specify a suggested pad size. You choose the minimum ball diameter. Might be good to check what IPC would suggest.

ST_WLCSP-63_Die427 does not list a suggested pad size either. But here your choice can not be right as you selected 0.22 wheras the minimum ball diameter is 0.24

ST_WLCSP-64_Die441 Size is not quite right (you entered the maximum instead of nominal)

ST_WLCSP-66_Die432 does neither specify a pad size nor a paste opening. You copied the information from the part before it.

ST_WLCSP-100_Die422 you used the minimum body size instead of nominal
same with ST_WLCSP-100_Die452

ST_WLCSP-100_Die461 copy paste error. You forgot to overwrite its size information

ST_WLCSP-168_Die434 does not list pad size information, you copied the one from the previous package

Ratfink:
Good point, I implemented that in the script and updated the data.

---

* Fixed.
* There are two WLCSPs in that datasheet.  The WLCSP-64 has a 0.35 mm pitch, the WLCSP-49 has a 0.4 mm pitch.
* Ball diameter is the same as ST_WLCSP-49_Die447, so I used the same pad diameter.  It looks like IPC would suggest 0.20 mm pads for this ball size (80% of nominal ball diameter), but they also say that 100% of nominal ball diameter is within the acceptable range, so 0.22 mm (88%) looks okay.
* Keeping in the range determined in the previous bullet, I've increased this to 0.24 mm.
* Fixed.
* Same ball size as ST_WLCSP-63_Die427, so I've set that to 0.24 mm as well.
* Fixed and fixed.
* Fixed.
* Changed to 0.22 mm, as determined above.

poeschlr:
Everything looks ok now. Thanks

PR #523, opened 2018-04-25
Add DecaWave_DWM1001 footprint

Ratfink:
[Datasheet](https://www.decawave.com/sites/default/files/dwm1001_datasheet.pdf)

The keep-out zone follows the recommendation on page 16.  There's no silkscreen around the antenna area since that is likely to be hanging off the board edge.  The anchor is centered on the RF shield to avoid putting a pick-and-place nozzle too close to one of its perforations.  I wasn't quite sure what to do with the VDDIO test point on page 18, so let me know if you think the NPTH hole I put there is wrong.

There is a 3D model for this part in the packages3D repository, but it's positioned with the origin at the center of the pads' bounding box.  I'm currently unable to fix that myself, so if somebody else could that would be appreciated.

screenshot from 2018-04-25 19-02-30

------------


Ratfink:
Updated screenshot with F.Fab pin 1 mark (I knew I forgot something!)
screenshot from 2018-04-25 21-26-52

poeschlr:
The landing pattern looks ok. But your fab marking for the shielded area is positioned wrong.

screenshot from 2018-04-26 18-19-04

screenshot from 2018-04-26 18-23-29

---

I am a bit unsure about the shield through holes. Should there be at least a "no blank copper" area?
screenshot from 2018-04-26 18-42-34

PR #524, opened 2018-04-25
Adding footprints for sensor motion

poeschlr:
Footprints for https://github.com/KiCad/kicad-symbols/pull/546


poeschlr:
All needed footprints should now be included.

I used an ipc calculator script (libre office calc file) to arrive at the pad sizes and dimensions (where the datasheet does not define the sizes)

sens svg

poeschlr:
One footprint is modified in kicad with custom pads around the corners to increase the clearance here (without modification 0.075mm clearance, with it 0.125mm)

updated dimensioned drawings
screenshot from 2018-04-28 15-44-52

Ratfink:
The datasheet given for LGA-12_3x5mm_P0.8mm_LayoutBorder1x5y is for an LGA-14 part, so either the datasheet URL is wrong or the footprint is in error.

poeschlr:
The footprint is wrong.

poeschlr:
should be fixed now

Ratfink:
Looks good, thanks!

PR #526, opened 2018-04-26
Add script generated diode footprints for temp sensor (DO-34_SOD68)

poeschlr:
Some more footprints for https://github.com/KiCad/kicad-symbols/issues/520

Needed for KTY83

These are generated using the scripts (i think by @jkriege2)
PR for adding the size specifications: https://github.com/pointhi/kicad-footprint-generator/pull/127

datasheet: https://www.nxp.com/docs/en/data-sheet/KTY83_SER.pdf


poeschlr:
I have bad news. When checking these footprints with kicad stepup i discovered that the courtyard is calculated relative to the silkscreen instead of the body outline.

This means all of our resistor-like footprints have a wrong courtyard clearance. (To make sure i checked a few footprints in the resistor_tht lib. Sadly all of them had a wrong courtyard clearance.)

I will fix the script and make a separate pull request for this fix. (This pr will be updated with a fixed version of this footprint as well.)

Ratfink:
Oh geez, that is bad news.  Thanks for noticing, and for working on the fix!

Here's a PDF for just the SOD68 package: https://assets.nexperia.com/documents/outline-drawing/SOD68.pdf

poeschlr:
I think i have a working fix. (At least for these few new footprints. I will need to check it for all others in a separate PR)

screenshot from 2018-04-27 17-01-47

jkriege2:
you should also make ddrill a bit larger. You use only 0.6mm and the wires are 0.55mm according to DS ... maybe use 0.75mm?

The scripts (yes by me) are quite old and a bit inconsistent here, as you give the DRILL diameter, not the wire diameter.

PS: If I calculated the courtyeard wrong, we might have to regenerate all other diodes too ... could you then do that in another PR?

Thanks,
JAN

poeschlr:
The wire is a maximum of 0.55mm which leaves 0.05 for the drill tolerance. (And this assumes that both the wire and the drill are at the worst tolerance point.)

---

And yes as i wrote above i will look at all other resistor like footprints (not only diodes are wrong)

jkriege2:
Nice,
JAN

jkriege2:
@poeschlr : I added a few of the missing 3D models: https://github.com/KiCad/kicad-packages3D-source/pull/124 and https://github.com/KiCad/kicad-packages3D/pull/303

PR #527, opened 2018-04-26
Add Sod-70 (Thinner but longer TO-92) for temp sensor

poeschlr:
NXP seems to use a special package for KTY81. They call it SOD70. It has a slightly smaller diameter but is also a bit longer than TO-92.

This adds a footprint generated via kicad stepup. (source file is again part of the pull request.)

Edit:
datasheet: https://www.nxp.com/docs/en/data-sheet/KTY81_SER.pdf

and of course this is for https://github.com/KiCad/kicad-symbols/issues/520


jkriege2:
is it intentional that there is a fcstd-file too?

jkriege2:
also I see a variant with 5.08mm pin spacing in the DS ... could you also add that one?

Thanks,
JAN

jkriege2:
Nice work!

Thanks,
JAN

jkriege2:
@poeschlr : I added the missing 3D models: https://github.com/KiCad/kicad-packages3D-source/pull/124 and https://github.com/KiCad/kicad-packages3D/pull/303

PR #528, opened 2018-04-27
Vishay PowerPAK SC70-6L footprints

SchrodingersGat:
- Single version - http://www.vishay.com/docs/70486/70486.pdf
- Dual version - http://www.vishay.com/docs/70487/70487.pdf

e.g. https://www.digikey.com/product-detail/en/vishay-siliconix/SIA453EDJ-T1-GE3/SIA453EDJ-T1-GE3CT-ND/4385553

image

image


evanshultz:
- Did you consider using polygon pads? Or maybe you're only using v4 features since you use KiCad for production?
- Can you add more than just a URL for the description? The part datasheet might be good too since the footprint drawing doesn't include the body size (at least for single; see dual below).
- How did you determine the paste pads for the single one? Just considering the big middle copper pad, there's only 56% paste coverage. That's a bit low, and will only be worse when taking into account the outer pin 1 pads.
- Those paste pads have a pin a number.
- Lastly, would you look at https://github.com/KiCad/kicad-footprints/issues/529? I think your paste pads will need the clearance defined.
- Since the packages are actually different with regards to pin layout, I think they should have unique 3D models.
- http://www.vishay.com/docs/62859/sia931dj.pdf shows the dual body size as 2.05mm sq, but http://www.vishay.com/docs/70487/70487.pdf shows 2.5mm sq. Which is correct?
- I see several dimensional problems with the footprints (dual on left):
image

evanshultz:
Thanks!
- What about extending the silk horizontally across the footprint and removing it around pin 1 or extending it farther near pin 1? That seems to be more common for our SON footprints instead of silk only in each corner?
- Paste: Now at 67% of just the center pad which should be OK since the outer pads have paste too.
- Only a few dimensional fixes left: vertical courtyard on single (is 1.35mm, should be 1.28mm), outer pad width on dual (is 0.35mm, s/b 0.325mm), dual courtyard (horizontal is affected by above), big inner pad offset (is 0.159mm, s/b 0.16mm):
image

evanshultz:
- Single: Pads 3/4 had their width changed to 0.325mm; it should have stayed at 0.35mm.
- Single: Courtyard should be X = +/-1.35mm, Y = +/-1.28mm.
- Dual: Courtyard should be X = +/-1.33mm, Y = +/-1.28mm.

Package_TO_SOT_SMD.pretty/Vishay_PowerPAK_SC70-6L_Single.kicad_mod
```diff
@@ -0,0 +1,43 @@
+(module Package_TO_SOT_SMD:Vishay_PowerPAK_SC70-6L_Single (layer F.Cu) (tedit 5AE7F3BE)
```
poeschlr: Is it possible you used the export footprint feature to save these footprints?

This would explain why your footprint has the lib name included here.

(That is also why travis complains about the value field. It assumes the module name to be correct and does not handle cases that include the library name.)

SchrodingersGat:
What a saga this has turned out to be. I have fixed the offending lines, thanks Rene.

evanshultz:
@SchrodingersGat 
Setting a specific margin on paste and mask pads is no longer needed. Fix in KiCad. Could you fix that? Also, it would be nice to have a description that is more than just a URL. Then this looks good to me.

evanshultz:
@SchrodingersGat 
Can you take care of that last thing and then the footprints and symbols can be merged?

SchrodingersGat:
Alright, hopefully all good to go now.

evanshultz:
Yep. Quite an ordeal. Thanks!

Issue #529, opened 2018-04-27
Problem regarding paste and mask only pads. Project clearance interferes with them.


poeschlr:
I just noticed that our defined workflow for defining paste and mask areas does not really work as soon as the user sets some clearance in the project.
If a pad has it's clearance set to 0, the project setting is applied.

I asked the developers what the minimum clearance value is. We will need to use this as a workaround.
https://lists.launchpad.net/kicad-developers/msg35609.html

---

At least in kicad 5 it seems the smallest value possible is 1e-06 (0.000001)

poeschlr:
The developers confirmed that 1nm (=1e-06 = 0.000001) is the minimum value not changed back to 0.

poeschlr:
We should also include this in the KLC and also in the travis check scripts.

poeschlr:
I choose 1e-06% for the paste ratio as well (results in 1e-08 in the file)
That way it looks uniform in the GUI.

poeschlr:
Detailed explanation https://github.com/KiCad/kicad-footprints/pull/447#issuecomment-385584621

poeschlr:
Fix has been committed to the software side: https://github.com/KiCad/kicad-source-mirror/commit/93c7f65d96130c7165d412945668293f571cf2e9

No need to fix anything on the library side.

PR #530, opened 2018-04-27
Workaraound for global clearance overwritting pad settings if clearance not defined.

poeschlr:
This is a fix for the problem outlined in https://github.com/KiCad/kicad-footprints/issues/529


poeschlr:
Do not merge yet. It seems we also need to set the solderpaste ratio as setting solderpaste clearance alone is not enough. I hope the devs will come back to me about that.

poeschlr:
Ok this should now be also done.

poeschlr:
For more details regarding the reasoning for this workaround have a look at this comment of mine https://github.com/KiCad/kicad-footprints/pull/447#issuecomment-385584621

poeschlr:
Wait with merging, we might get a better solution. See https://github.com/KiCad/kicad-footprints/pull/447#issuecomment-385696311

poeschlr:
Not necessary for kicad version 5 (kicad 4 backport will need such a workaround but that is a topic for another day.)

PR #531, opened 2018-04-27
Rename OMRON → Omron

Ratfink:
The correct capitalization is Omron, not OMRON.  There's no need to shout.

To be merged along with the corresponding symbol PR, https://github.com/KiCad/kicad-symbols/pull/554.


poeschlr:
Thanks

PR #532, opened 2018-04-28
Add missing footprint for Bosch BME680

poeschlr:
Missing footprint for https://github.com/KiCad/kicad-symbols/pull/545

datasheet https://ae-bst.resource.bosch.com/media/_tech/media/datasheets/BST-BME680-DS001-00.pdf#page=44

created with script included in https://github.com/pointhi/kicad-footprint-generator/pull/125

screenshot from 2018-04-28 16-24-37

screenshot from 2018-04-28 16-29-51


Ratfink:
Thanks for making this footprint.  I see just one problem: the recommended pad dimensions are 0.4x0.4mm, but you have 0.4x0.55mm.  Fix that (and double-check the positions—they'll need to be moved slightly as well) and I think this will be ready to merge.

poeschlr:
I used IPC standards to derive these. Bosch does not allow for any misplacement of the part which does not seem right to me. They also do not allow for their own tolerances. (Their suggested footprint is to have the pads on the pcb exactly with the same size as the leads on the chip. I used IPC with 0 heel, toe and even negative side fillet.)

There is enough space for a bit more copper which should make the footprint a bit more resilient.

Ratfink:
IPC-7351B doesn't define "heel", "toe", and "side" for LGA pads, instead using the term "periphery" to refer to all sides of the pad.  For LGAs, they say "Periphery: 0.00", round-off factor the same as always, and "Courtyard excess: 1.00".  However, this part looks more like what they'd call a "DFN with Pullback Leads", which is similar but with 0.25 mm courtyard excess.  So while I agree with your reasoning, I think it would be best to ensure this (and your script) agrees with the current standard.

poeschlr:
What do you mean with "current standard"?

Ratfink:
It looks like the script you generated this with is based on IPC-7351, which is superceded by IPC-7351A, which is in turn superceded by IPC-7351B.  IPC-7351C hasn't been released yet, so IPC-7351B is the current revision of the standard.  (Maybe I should use words like "present" or "latest" instead to avoid confusion with electric current. 🙂)

poeschlr:
well i don't have access to newer standards

Ratfink:
Well, if you think this is better than what Bosch suggests, I guess I can merge it.  Thanks!

poeschlr:
I have more confidence in a slightly out of date standard than in a document that does not list how they derived their footprint. Especially if it does not seem to allow for any tolerances.

You seem to have access to more recent information. Could you give a short update what changed between the two standards (especially interesting would be if the formulas include further therms. And if the fillet margins [or whatever they are called now] changed by a lot.)

Ratfink:
The only access I have is through [a copy](https://www.slideshare.net/NicolaeFlorinAlexandru/ipc-7351-bgenericrequirementsforsurfacemountdesignandlandpatternstandardunlocked) someone uploaded to SlideShare.  Usability is poor: I can only see it as a series of images, so it's impossible to automatically search through the text of the document.  I'm not terribly familiar with either revision, which is why I suggested you took a look yourself before.

poeschlr:
The formulas are still the same. A few values are slightly different 
- J Lead has larger toe fillet (affects recently merged PLCC footprints), 
- square end has heel = 0 instead of negative heel. (affects the resistor, capacitor, ... footprints)
- nearly all round off therms have been reduced. Would maybe allow for better values for the recently merged qfp footprints.

PR #533, opened 2018-04-28
Fix footprint names in Sensor_Current lib (Pitch -> P)

poeschlr:
These footprint where not renamed but their corresponding symbols had their footprint fields updated to the new naming convention.


poeschlr:
Travis error is from the fact that one of the footprint has the same package as a normal qfp but uses much larger pads (they can share the 3d model)

Ratfink:
Looks good, thanks!

PR #538, opened 2018-04-29
Fix courtyard clearance for inductor (chocke) footprints

poeschlr:
Next fix for the courtyard problems. For these the same applies, the old script did calculate the clearance relative to silk instead of body dimensions.

script PR: https://github.com/pointhi/kicad-footprint-generator/pull/127

I did again check all footprints using the footprint browser.

This revealed one error. L_Axial_L29.9mm_D14.0mm_P25.40mm_Horizontal_Vishay_IHA-105 is a bit strange. It is an axial inductor but the pitch is smaller then the length. This results in a totally wrong courtyard. I will look at this tomorrow. (It is 3 in the morning right now.)


jkriege2:
Hi !

I think the IHA-105 was a typo when putting the data into the scripts. The pitch should be 30.48mm or 33.02mm ... Here's the datasheet: https://www.vishay.com/docs/34014/iha.pdf

JAN

poeschlr:
IHA-105 minimum pitch calculation using IPC constrains:
Size B does not include the endcap. The bend can only start outside of the endcap so it needs to be included in the calculation for the minimum possible pitch. The minimum bend radius is the maximum lead diameter.

Pmin = 29.85 (Lmax) + 2* 3.18 (cab) + 2*0.863 (max diameter) = 37.936mm So maybe 38mm would be a good value.


All IHA-xxx footprints had way too small of a pin pitch. One would need to bend the leeds in some kind of "?" form. So i increased them for all. I am too lacy to think about what the nearest imperial pitch would be so i used metric pitch for them.

---

I also just noticed the 3d settings are wrong. There is a scaling factor !=1 in the script.

jkriege2:
About the 3D models:
If you change the pitch ... could you ensure that if old 3D models for these exist, they are removed from the repo?

Also: your PR changes the 1-1-1-scaling to 1/2.54-1/2.54-1/2.54, because I think the scripts still have that old scaling, but when moving over the footprints to the new repo that was fixed by hand. In the scripts you'll need to change s_3d=[1 / 2.54, 1 / 2.54, 1 / 2.54] to s_3d=[1,1,1] in make_Chikes_THT.py.

best,
JAN

poeschlr:
Scaling is fixed. KISYS3DMOD has been added. Now obsolete 3d models are removed in https://github.com/KiCad/kicad-packages3D/pull/304

poeschlr:
Ok why are there still some left with the wrong 3d path?

PR #539, opened 2018-04-29
Add script generated dfn-6-1ep footprint for Si7020-A20

poeschlr:
Another footprint for https://github.com/KiCad/kicad-symbols/issues/520

script PR: https://github.com/pointhi/kicad-footprint-generator/pull/125

needed for https://github.com/KiCad/kicad-symbols/pull/558


Ratfink:
Silicon Labs has a [recommended footprint](https://www.silabs.com/documents/public/data-sheets/Si7020-A20.pdf#page=32) in their datasheet which is also based on IPC-7351 guidelines; wouldn't it be best to use that?

As you point out, this footprint should also work for https://github.com/KiCad/kicad-symbols/pull/493.  However, looking at the datasheets, I see that three different versions of this footprint will be needed for different 3D models: one normal, one with a depression in the center, and one with a protective cover.  Should we prepare for that inevitability now (possibly with different fab layer drawings), or wait until someone makes models for these parts to worry about that?

And since you didn't post one, here's a screenshot:
screenshot from 2018-04-29 13-52-47

poeschlr:
In my mind scripted is always better then the alternatives. I even used the current IPC standard to determine the landing pattern. (I used nominal density flat no leads requirements and flat no leads pull back for the EP.)

Ratfink:
Okay, that seems fair.  What do you think about the second question I asked, though?  I'd be fine with merging this now if there's no immediate plan to make 3D models for these parts.

I also noticed that you set `(solder_paste_margin_ratio 0) (solder_paste_margin 0.000001)`.  The `solder_paste_margin_ratio` setting is probably not supposed to be `0`, so you may want to change that.

poeschlr:
Yes the solder paste margin ratio should not be 0. There seems to be a problem in the renderer.

I will fix that shortly. (I need a distraction from the texas mcu lib anyways.)

Ratfink:
Since the workaround is apparently no longer necessary, please undo it and I'll merge this.

poeschlr:
@Ratfink done

PR #542, opened 2018-05-01
Added a footprint for the TE OZ-SS-103LM1 SPST Power Relay

johnfwhitmore:
Added the footprint for TE OZ-SS-103LM1

I've added the symbol to the symbol library.

http://www.te.com/commerce/DocumentDelivery/DDEController?Action=showdoc&DocId=Data+Sheet%7FOZ_OZT_series_relay_data_sheet_E%7F0411%7Fpdf%7FEnglish%7FENG_DS_OZ_OZT_series_relay_data_sheet_E_0411.pdf%7F1440002-8

screenshot from 2018-05-01 17-01-30


poeschlr:
[I think this drawing is more detailed](http://www.te.com/commerce/DocumentDelivery/DDEController?Action=showdoc&DocId=Customer+Drawing%7F1440002-5%7FH1%7Fpdf%7FEnglish%7FENG_CD_1440002-5_H1.pdf%7F1461251-5) (It is specialized for the OZ-1M series)

Your footprint name is way to specialized. It is possible to create a footprint compatible to all OZ-1M relays. Meaning OZ relay form factor A (NO)

So i would suggest as a name either `Relay_SPST_TE-OZ-1M` or `Relay_SPST_TE-OZ-xx-1xxxMx`

---

The footprint is also slightly wrong in multiple measurements. See annotated dimensioned drawing

screenshot from 2018-05-06 01-24-24

PR #543, opened 2018-05-02
Bourns SDR1806

bdm310:
Added the Bourns SDR1806 series footprint to Inductor_SMD.
https://www.bourns.com/docs/Product-Datasheets/SDR1806.pdf

image

https://github.com/KiCad/kicad-packages3D/pull/305

------------


Shackmeister:
hi @bdm310 

Can you please export your model using Kicad Stepup?
if you could be willing to, it would be great with some copper on where the coil should be :)

poeschlr:
The courtyard clearance should be 0.25mm
The body length is also not quite right. See annotated dimensioned drawing

screenshot from 2018-05-06 03-38-21

bdm310:
Not sure where my 17.78mm came from and I think I did this before finding the 0.25mm spec in the KLC. I did 14.3+0.05 this time around; I'd say it's better to do the courtyard at max tolerance condition.

bdm310:
I must've done it to the silk and not the body. Updated.

Shackmeister:
getting closer :)
please change silkscreen width to 0.12mm instead of 0.15mm
also please move the REF and Value closer to the courtyard :)

bdm310:
I wasn't consistent (one of the arcs was 0.12) so this would've needed an update anyway but it may be helpful to clarify in the KLC that 0.12 should only be deviated from if necessary. The current wording makes it seem relatively loose and I'd call this a "low density" part. It'd also definitely be helpful to include a guideline for where reference designators and values should go.

Shackmeister:
@bdm310 
Thanks for the contribution :)
hope to see more :)

Issue #544, opened 2018-05-02
Too many footprints in several libs


evanshultz:
@poeschlr @jkriege2 @Ratfink @Misca1234 
[KLC says we should only have 250 footprints or symbols in a lib](http://kicad-pcb.org/libraries/klc/G1.2/). Some of our footprint libs are exceeding this number.

I do notice with all current libs that invoking all footprint libraries (in Pcbnew or footprint editor) does take quite a while. I'm not sure if splitting them up would really help this, but when loading only a single library's footprints it should help. (Loading all libs in Eeschema or CvPcb takes an annoyingly long time as well, but this is about footprints.)

The exact qty of footprints changes nearly daily, but as of right now here are some libs that have 250 (or nearly that many) footprints:
- Capacitor_THT: 357 (split by axial, radial, disc, etc.?)
- Connector_JST: 426 (split by part family?)
- Connector_Molex: 582
- Connector_Phoenix_MC: 247 (close enough; split by horiz and vert?)
- Connector_Phoenix_MSTB: 269
- Connector_PinHeader*: 278
- Connector_PinSocket_1.27mm and bigger size: 247 or 279 (1.0mm is only 157)
- Connector_Samtec: 436
- Inductor_THT: 241
- Package_DIP: 233

And here are two things which I noticed when looking at this as well:
- Package_DFN_QFN: 170 (maybe split DFN and QFN since they are distinct package types?)
- SOT is, I believe, only an SMT package. So does it make sense to have Package_TO_SOT_THT (there are no SOT footprints in there)? Should it just be Package_TO_THT? And then would it make sense to split Package_TO_SOT_SMD into Package_SOT_SMD and Package_TO_SMD (resulting in 3 total libs)? Or break those two existing libraries differently into Package_SOT and Package_TO?

poeschlr:
@SchrodingersGat was there any reasoning behind choosing the number 250 as the lib size limit?

---

I am not so sure 250 should be a hard cap but more of a "start to worry if the lib might grow" value.

We truly run into trouble with libs > 1000 footprints as github can not handle that many files well. (Or 500 if you expect every footprint to come with its own 3d model.) 

Splitting connector libs at least should not create problems on the symbol lib side of things. But it will again require updating the 3d model side. (And this must be done at the same time. No exceptions!)

Touching the TO_SOT libs will create a lot of work on the symbol side with very limited gains. (All of that shortly before a major release. So i don't like that idea.)
[...]

PR #546, opened 2018-05-03
Split samtec lib to reduce number of footprints in the lib.

poeschlr:
As mentioned in #544 the only lib where i am sure that we should split it and how we should do it is the samtec lib. This PR takes care of this.

Split away the HLE series connectors into two new libs
split up with respect to mounting technology (THT vs SMD)


poeschlr:
No 3d model update required as we do not have models for these.

poeschlr:
Because this series has > 400 footprints all on its own. (That is why i even split it further into smd and tht) The samtec lib now has 14 left. (We do that quite often that we split away some things where it makes sense and have a lib for everything else. Best example the connector footprint lib.)

---

Fixed the fp-lib-table description

evanshultz:
Alright then.

PR #550, opened 2018-05-03
Split capacitor THT along the lines of round vs rectangular form factor.

poeschlr:
Another #544 inspired improvement.
About 50% of capacitor tht footprints are for caps in rectangular form. the rest is for round footprints so it makes sense to split it along these lines.


evanshultz:
An innocent question: Why not name the libs the same as the footprints? So `Capacitor_Axial_THT`, `Capacitor_Radial_THT`, `Capacitor_Disc_THT`, etc.?

evanshultz:
Yeah, that organization might be nice. Everything is a tradeoff.

I would like to understand the rationale. Sorry if you see this as being pedantic or dragging out getting this merged. A few things:
1. Other libs end with `THT`, like `Package_TO_SOT_THT`. Should `Capacitor_THT_Rectangular` but `Capacitor_Rectangular_THT`?
2. Rectangular footprints have just `Rect` in the name, but the lib is `Rectangular`. Should they match?
3. Using `Round` could capture radial or axial caps, but only from one perspective. From another perspective they're a cylinder. While `Disc` totally describes the body shape. I don't have a suggestion right now, but I wonder if there's a better term than `Round`. Let me think about this.

poeschlr:
A cylinder is round no matter what direction you look at it. Yes the outline on the board is not round but this does not change the fact that the thing mounded on the board is still round. (The lib is named after the shape of the component not the shape of the outline on the board)

Cylinder would be way to restrictive. round would encompass even shapes that are not really cylindrical but also not a cuboid. This would mean that most future contributions will go into the round lib. We can use cylinder but that would mean we will need a third lib sometime in the future when the dipped caps are added.

I deliberately choose to have tht to the front as the first hierarchy is along smd vs tht. only the second split then is by form factor.

I think rectangular is in this case a better more precise name. If you really want we can of course change that. In footprint names it makes sense to use shorthands as the footprint names are long already this is not necessarily the case for library names.

poeschlr:
@evanshultz if we want this in kicad 5 we need to do this today or at the latest tomorrow. By the way if you feel unconformable i suggest to leave this for post kicad 5. (And live with the large lib for the whole kicad 5 cycle.)
So either merge as is or close.

poeschlr:
This will have to wait for after the v5 release

PR #551, opened 2018-05-04
Drv887x footprint

thomasonw:
This submission is in support of the added devices DRV887x  (ref: [https://github.com/KiCad/kicad-symbols/pull/512](https://github.com/KiCad/kicad-symbols/pull/512) )

Footprint is a modification of  HTSOP-8-1EP_3.9x4.9mm_P1.27mm_ThermalVias  to reflect the slightly smaller pad area, and longer / narrower leg pads as specified by TI for these devices. 

Reference documentation:  [http://www.ti.com/lit/ds/symlink/drv8870.pdf](http://www.ti.com/lit/ds/symlink/drv8870.pdf)

image


thomasonw:
OK,  While in the middle of submitting this I updated my Github app, and now am not able to continue to link to this thread - plus edits were lost.   

I am closing this submission and will see if I can recover git-hub on my side before submitting again.

thomasonw:
Was able to recover Github, going to try reopening this request

thomasonw:
OK, to the reviewer:  Seems I have recovered from my Github app issue.

Travis is reporting errors around how the pad was created.  I copied this device from an existing one:  ` HTSOP-8-1EP_3.9x4.9mm_P1.27mm_ThermalVias` and the Travis errors are from how this original footprint was made:
  

- The pad is created out of 4x smaller boxes, but those have their silk and solder-mask disabled.
- The thermal vias are the same, solder mask layer disabled.
- And overall layer was added to create the solder mask separately.

Please advise if this should remain this way, or if I should modify the pad in a different way.

Thank you,

poeschlr:
A better example would be HTSSOP-20-1EP_4.4x6.5mm_P0.65mm_EP3.4x6.5mm_ThermalVias

On the top side make one large copper pad pin number 9 (disable both paste and mask layer on it, size 2.85x4.9mm
Add a mask only pad without pin number size 2.4x3.1mm
Add the 6 vias, tht pads with all copper, pin number 9, no paste no mask.
Add a copper pad on bottom (a bit larger than the area covered by the vias)

Now the tricky part. Make a few paste only pads (no pin number, no copper). Aim for 50 to 80% coverage in area. Try to avoid the areas where the vias are located.

thomasonw:
Great coaching - see how these edits look.  Travis still not happy.  Am wondering, right now the thermal vias are 'tented' on the bottom solder-mask, maybe not the best idea? 

And Travis is still not happy about some of the missing layers...

evanshultz:
The remaining CI errors are bogus because our check script isn't smart enough. Yet. :)

- The datasheet gives an exact stencil opening, so copy that for a single 2.4x3.1mm paste pad and remove the other ones. Usually the guidance above is good, except the paste pads shouldn't overlap as you have, but in this case the document is explicit about what to do.
- Pads 1-8 should have 2.2x0.5mm copper pads.
- Set the mask oversize on the footprint (not per pad) to 0.07mm.
- Remove the paste margin on the top thermal pad.
- Vias should be 0.33mm diameter (with a then 0.63mm annular ring).

Lastly, since this is now a rather specific footprint I'd like to prefix it with `Texas_` so we know it's built to TI's datasheet.

evanshultz:
The symbol using this footprint is actually https://github.com/KiCad/kicad-symbols/pull/512, not the one above.

thomasonw:
@evanshultz :  Have made the edits above, one open question though:  The thermal via's are tented by the bottom solder mask layer.  Is this best approach, or would it be better to leave them open given the solder-paste is being applied over the vias from the top.

evanshultz:
One more fix that was missed above: Remove the paste margin on the top thermal pad.

Also, there are two bottom copper pads.

Yes, add all mask layers to the vias. Thank you.

evanshultz:
Thank you!

evanshultz:
And please use better commit messages in the future. "Edits" doesn't make it clear what or why the commit was pushed.

PR #553, opened 2018-05-04
Infineon BTS7004-1EPP

johnfwhitmore:
The Infineon PROFET BTS7004-1EPP has a proprietary footprint, (I think), PG-TSDSO-14-22 (Thin (Slim) Dual Small Outline 14 pins) 

Datasheet : https://www.infineon.com/dgdl/Infineon-BTS7004-1EPP-DS-v01_00-EN.pdf?fileId=5546d4626102d35a016147550a725555

screenshot from 2018-05-04 18-12-45

I've set the filename of the associated 3D file but not created it. I'm creating this pull request to see what Travis has to say about it.


johnfwhitmore:
screenshot from 2018-05-05 10-30-03

Not 100% sure of this, never seen a footprint with that feature before. Thank you for your help

poeschlr:
The paste pads are normal rectangles. No need to use polygons.

Use a standard smd rectangular pad. No pin number assigned. In the layer selection part remove everything except paste. And select no copper.
This does reduce the possibility for errors and can be easily checked by us librarians. (The footprint can then even be used in kicad 4.)

evanshultz:
@poeschlr 
Is this good to go now? I'd like to look at the symbols that use this footprint.

johnfwhitmore:
Sorry not sure how that pad ended up different. As far as I'm concerned this footprint is good to go but I'm certainly not the expert, quite the opposite. The symbol is in another pull request at https://github.com/KiCad/kicad-symbols/pull/574

poeschlr:
Courtyard clearance in y direction is not quite right:
screenshot from 2018-05-08 12-12-08

poeschlr:
@pushytoxin my workflow is described here: https://forum.kicad.info/t/how-to-check-footprint-correctness/9279

johnfwhitmore:
screenshot from 2018-05-09 17-42-03

johnfwhitmore:
It appears that the courtyard should be 0.3 mm border. But that is an error so should the courtyard be slightly smaller then it should be or slightly larger then it should be?

evanshultz:
- The courtyard on this footprint should be 0.25mm all around. See http://kicad-pcb.org/libraries/klc/F5.3/.
- So the vertical courtyard should be +/-2.7mm. That still needs corrected.
- Your last commit changed the horizontal courtyard placement, but that's also wrong. It should be +/-3.76mm since the courtyard should be on a 0.01mm grid. I round up and I believe that's the expected way to do it (though KLC doesn't clearly say).

johnfwhitmore:
Sorry I've been distracted for a few days. So I've calculated the courtyard to be 2.85+(1.31÷2)+0.25 in the X Direction which is calculated on the footprint dimensions. In the Y direction I've used the mechanical limit of (4.9÷2) + 0.25 
I hope that's correct. Sorry again for the delay.
screenshot from 2018-05-23 22-00-20

johnfwhitmore:
Oh I rounded up in the X Direction as you suggested, to get it on the 0.01

evanshultz:
Yep. We come up with the same dimensions and all looks good to me. Thanks!

johnfwhitmore:
Thank you for merging

PR #555, opened 2018-05-04
Add footprint for E73-2G4M04S-52832 BLE5 module

isundaylee:
Datasheet: http://www.cdebyte.com/en/downpdf.aspx?id=243 (page 5)

Symbol PR: https://github.com/KiCad/kicad-symbols/pull/573

Screenshot: 

screen shot 2018-05-04 at 2 09 13 pm

------------


isundaylee:
@poeschlr I was following the pin number in the data sheet. Not clear why it is like that though.

poeschlr:
The drawing in the datasheet is over constrained. The part can not be both 17.5mm wide while the pads are 1.77mm from the side and the pin pitch is 1.27mm (1.27*11+2*1.77 would be 17.51)

So either we assume that the 1.77mm measurement is rounded (should then be 1.765mm)
Or the with of the part is rounded (and should in reality be 17.51mm)

I would  for the former and move all pads on the bottom row to the left by 0.05mm.

screenshot from 2018-05-06 11-47-49

isundaylee:
@poeschlr Thanks for the review! Changed as requested.

evanshultz:
The pad fix above looks OK to me. Please rename this footprint to just `E73-2G4M04S` since it's compatible with both 52832 and 52810 versions. Then I think it's ready to merge!

isundaylee:
@evanshultz  Thanks! Renamed as requested.

evanshultz:
Thank you!

PR #559, opened 2018-05-04
Amphenol Micro USB Type B 10118193-0001LF

carlosromorgar:
https://cdn.amphenol-icc.com/media/wysiwyg/files/drawing/10118193.pdf

image

------------


evanshultz:
- Please remove the `LF` from the end of the footprint name.
- As you have a F.Fab pin mark a divot/triangle at pin 1 remove the notch in the corner far away.
- Align the F.Fab pin 1 mark lines so they don't overlap the outline.
- Move the F.Fab refdes a bit so it doesn't clash with the pin 1 mark.
- Please add silk lines between pins 6-1 and 5-6 at the "back" of the connector.
- All F.Fab lines should be 0.1mm thick.
- Silk should be offset from fab lines by at least 0.11 so they don't overlap.
- Ensure at least 0.12mm from silk to copper lines. Silk lines are violating all 4 outer pin 6 pads.
- There are a few dimensional issues:
image
- I assume you intend to use this symbol?
image

carlosromorgar:
Hi,
Sorry for my late response. I have done all suggested changes, except dimensions 0.51 and 6.09 mm.

This is the drill
image


This is the distance
image


This is the final footprint
image

The symbol is what you indicate.

best regards

PR #563, opened 2018-05-05
Fixed pin numbering in Connector_Wire and made pin 1 rectangular per KLC

bdm310:
All pads in Connector_Wire were numbered as "1". I fixed that and made pin 1 rectangular on multi pad footprints per the KLC. It looks like these were script generated, I'm not sure where that script is but I'm happy to fix that as well if someone can point me to it.

image

------------


poeschlr:
For reference the revious discussion about that: https://github.com/KiCad/Wire_Pads.pretty/pull/3

poeschlr:
If the pads have different numbers then i would suggest to change the footprint naming such that it aligns with other connector footprints.
So instead of `SolderWirePad_3x_0-8mmDrill` use `SolderWirePad_1x03_P3.175mm_Drill0.8mm`
Meaning `SolderWirePad_[number of rows]x[pads per row]_P[pin pitch]mm_Drill[drill size]mm`

bdm310:
Very much agree. Naming should be consistent now.

poeschlr:
For single pin footprints the pin pitch field does not make sense. So i would remove it for them.

Also could you move the footprint such that pin 1 is at the origin?

poeschlr:
The footprint name in the file header should agree with the file name.

Maybe add some description to the description fields.

bdm310:
These were clearly done by hand and not consistently, I was trying to avoid writing a script to rebuild them completely but maybe I should just bite the bullet. I wasn't trying to make this in to a whole day project...

poeschlr:
The only thing that is really required is the thing about the footprint name. As it might create problems for users if the two places where this info is stored do not agree with each other.

I assume you already know about the scripting lib we have for footprints: https://github.com/pointhi/kicad-footprint-generator

bdm310:
I'm going to skip pad shape changes and just get these to the point where they're usable and as up to KLC as far as I can see.

I've found the 3d model scripts, hadn't made it to the footprint generation ones yet. Thanks for the link.

poeschlr:
Thanks

PR #564, opened 2018-05-05
Added footprint for the E-Switch FS5700 DPDT footswitch

pushytoxin:
Add a footprint for a common stompbox footswitch.

Datasheet link:
https://www.e-switch.com/system/asset/product_line/data_sheet/226/FS5700.pdf

image

image


pushytoxin:
Thanks very much for checking the part out. I corrected the errors that you've pointed out.

Regarding the document link, I certainly would have done it the way you suggested if the KLC included that part.

Issue #562, opened 2018-05-04
FFC connectors are missing the pin number for mounting pads


poeschlr:
It seems the FFC connectors have been forgotten when @Shackmeister assigned pin numbers to the mounting pads

PR #565, opened 2018-05-05
Add pin number to mounting pads for FFC connectors

poeschlr:
This will fix the footprint side of https://github.com/KiCad/kicad-footprints/issues/562

Script PR: https://github.com/pointhi/kicad-footprint-generator/pull/130


Ratfink:
Looks good, thanks for fixing this.

PR #570, opened 2018-05-07
Add CNCTech 1001-011-01101 USB connector footprint

isundaylee:
Drawing: http://cnctech.us/pdfs/1001-011-01101.pdf
Screenshot:

screen shot 2018-05-06 at 8 47 28 pm

Product image:

------------


evanshultz:
- Add `_Horizontal` to the end of the footprint name.
- Rotate the footprint 90 degrees CW so pin 1 is closest to top left.
- After rotating, unless you found a reason not to, please center the footprint. You will probably have to adjust the courtyard to keep it on a 0.01mm grid.
- I would prefer if you can avoid any rotated pads.
- Typically the silk is above the footprint and fab footprint name is below.
- The PCB edge line should be on Dwgs.User layer.
- It would be best to put the fab ref des under the part body that doesn't hang off the PCB.
- For the NPTHs, the pad size and drill size should both be 1.1mm.
- After you do the above there can be one silk line all the way across the part by pins 1-4.
- Your courtyard clearance is 0.245-0.25mm; it should be 0.5m everywhere for connectors.

Dimensioned drawing:
image

isundaylee:
Thanks for the review! List of things to fix: 

- [ ] Add _Horizontal to the end of the footprint name.
- [ ] Rotate the footprint 90 degrees CW so pin 1 is closest to top left.
- [ ] After rotating, unless you found a reason not to, please center the footprint. You will probably have to adjust the courtyard to keep it on a 0.01mm grid.
- [ ] I would prefer if you can avoid any rotated pads.
- [ ] Typically the silk is above the footprint and fab footprint name is below.
- [ ] The PCB edge line should be on Dwgs.User layer.
- [ ] It would be best to put the fab ref des under the part body that doesn't hang off the PCB.
- [ ] For the NPTHs, the pad size and drill size should both be 1.1mm.
- [ ] After you do the above there can be one silk line all the way across the part by pins 1-4.
- [ ] Your courtyard clearance is 0.245-0.25mm; it should be 0.5m everywhere for connectors.

isundaylee:
@evanshultz  Thanks for the review! Fixed the issues in the comments. Some lingering questions: 

- I'm not very sure about where I should place the reference markers on both the silk layer and the fab layer. Especially for the silk layer, it seems like in this case it actually is better to place it on the left instead of on the top, since the left is the only side that would be on the PCB. Does what I have now make sense? 

- I am currently centering the footprint with regards to the fab+copper area. However CI seems to think that I should center it with regards to only the copper pads. What should I do?

- Some of the courtyard has clearance 0.495mm instead of 0.5mm in order to comply with the 0.01m alignment. Is that ok? 

Thanks! Updated screenshot:

screen shot 2018-05-17 at 5 27 41 pm

evanshultz:
- My expectation is that the user will need to move/rotate the silk ref des regardless of what we do. So... what about placing it above the top pad 5 so it is on the board and also in the "correct" rotation? Centered on the top also seems like a silly choice to me.
- Apparently our check script isn't working well for centering. It should be centered with respect to fab and copper. Your footprint is centered properly now (+/-10.9mm). Thank you.
- Yes, the courtyard off by +/-0.005mm is fine.

One more change:
- Correct the 3D model name to match the footprint name change.

Dimensioned drawing:
image

isundaylee:
@evanshultz Fixed and pushed. Thanks for the review! Updated screenshot: 

screen shot 2018-05-18 at 11 46 47 am

evanshultz:
Thank you!

PR #571, opened 2018-05-07
Added new footprint: TO-220-11_Horizontal_TabDown_StaggeredType1_Py5.08mm

jeremysiebers:
Added new footprint: TO-220-11_Horizontal_TabDown_StaggeredType1_Py5.08mm

Added new variant of Added new footprint: TO-220-11_Horizontal_TabDown_StaggeredType1_Py2.54mm --> 5.08mm

lmd18200 footprint

------------


poeschlr:
Sadly the staggered packages still follow the old naming convention. Could you aling yours to the new one?
http://kicad-pcb.org/libraries/klc/F3.5/

So the name should be: `TO-220-11_[body size]mm_P3.4x5.08mm_StaggerOdd_Tab-Down`

(I will look at the other ones and get them to the new clearer convention.)

jeremysiebers:
- [X] So it would then be TO-220-11_[body size]mm_P3.4x5.08mm_StaggerOdd_Lead[lead length]mm_Tab-Down

lmd18200 footprint

jeremysiebers:
Hi Rene,

I'm using the LMD18200 datasheet becuase that is the device I'm using in my design at the moment.
In this datasheet http://www.ti.com/lit/ds/symlink/lmd18200.pdf it is 20.02 +/- 0.25.
But since this is an old device I'll change the width to 20.2mm so that it can accomodate all other devices as well (the drawing itself is 20.2).

lmd18200 footprint

- [X] In your latest screenshot it looks like the courtyard outline is not closed. Could you fix that?

jkriege2:
A comment on the body size: Does this really make sense for such standardized packages? All TO-220 have more or less a size, fully determined by the TO-220-11 ...

Or should we also add these sizes for SOT-23, TO-92, TO-18, ...???

Best,
JAN

poeschlr:
The footprint name should always reflect the footprint measurements. So if you want the footprint specialized for your part it is ok to change it to have a body width of 20.02. 
That would require you to change the fab, silk and courtyard outlines to reflect that.

poeschlr:
@jkriege2 good point. To be honest i would have only included it for consistency with the rest of our lib.

jkriege2:
I think it really does not make too much sense ... I think everybody knows what to expect from TO-220 or TO-247 ... In addition, the very same package will have different sizes, depending on whether it's mounted vertically or horizontally? I would be confused by that size. + The exanple given for the KLC rule also does not have the package size and the size is explicitly an OPTIONAL parameter!

poeschlr:
Ok then in that case, i would agree not to include the body size. (Has the benefit that i will have less work for adapting the other footprint names)

poeschlr:
@jeremysiebers could you remove the dash in TabDown? (That way it would align with how other similar footprints are currently named resulting in less work when we fix up the naming convention in this lib. And yes i need to change that in the pictures used on the KLC site.)

jeremysiebers:
No problem.
- [X] removed the dash in TabDown

poeschlr:
Thanks for this contribution

Issue #569, opened 2018-05-06
TH Test points with non-plated holes?


thomasonw:
I have just noticed the TH Test points mostly use NPTH holes vs. normal plated one.  Example:

TestPoint:  TestPoint_THTPad_1.5x1.5mm_Drill0.7mm

Is this right?  The test points place copper on both sides, but does not connected them.  If indeed it is intended to use NPTH for drilling, why not only have a copper pad on one side?  But more so, I was expected the Through-Hole test points to use plated holes. 

Can someone help me understand this?

Ratfink:
Seems like a bug to me.  It looks like these appeared in https://github.com/KiCad/Connectors_TestPoints.pretty/pull/2, and there's no explanation of why they're NPTH there.

PR #572, opened 2018-05-07
Test point footprints had NPTH pads instead of normal THT (plated) pads

poeschlr:
See https://github.com/KiCad/kicad-footprints/issues/569


Ratfink:
Thanks for fixing this bug!

PR #573, opened 2018-05-07
Added new shunt footprint: WSK2512 for 0.005 to 0.2 Ohms

jeremysiebers:
Added new shunt footprint: WSK2512 for 0.005 to 0.2 Ohms:

http://www.vishay.com/docs/30108/wsk.pdf
http://www.vishay.com/search?query=WSK2512&searchChoice=part

wsk2512

Currently I created this footprint to match the default shunt resistor: 

r_shunt

Or should I add a new shunt resistor mapping specific to the WSK2512?

Secondly, I named this shunt as follows: R_Shunt_2512_6332Metric_0.005-0.2 because the 0.0005 to 0.0049 variant has a slightly different size, is this ok? A THT variant is called R_Axial_Shunt_L[Length]_D[Diameter]_P[Pitch]_[Modifiers]_[Orientation]_[Options].

------------


poeschlr:
As this is a more specialized footprint my suggestion would be to name it `R_Shunt_Vishay_WKS2512_...`

jeremysiebers:
@poeschlr 

- Is the name R_Shunt_Vishay_WKS2512_6332Metric_0.005-0.2Ohm a good name? I'm a bit doubting if it should be .._6350Metric.. because the component is 0.635mm wide?
- So it is not required to make a separate schematic symbol?

evanshultz:
My response to your questions:
- The "6332" part is for 6.35x3.18mm dimensions. The choice of 6.3mm or 6.4mm varies between different manufacturers so you see 6332 and 6432. 3.18mm rounds to 3.2mm. Since [KiCad uses 6332](http://kicad-pcb.org/libraries/klc/F3.1/), let's stick with that here. So you're already good for that part. I don't find that adding `Shunt` helps with the part. Instead of including resistance range in the footprint name, which just looks weird anyway, perhaps including thickness would be better. The only reason to use separate footprints is so the 3D model (reflecting the actual part) is unique, and the difference is the thickness. To that end, http://kicad-pcb.org/libraries/klc/F2.2/ isn't clear if the thickness should be prefixed with `H` or `T`. If using `T`, it would be `R_Vishay_WKS2512_6332Metric_T2.66mm`.
- A new symbol is not required, but to get the benefits of this part the Kelvin connection is required. (It could be done using net ties from pins 1 and 2 but having 4 pins makes it easy and matches the component/footprint design.) It might be nice to make the pin on each side look like a "Y" on its side instead of having the Kelvin pins sticking out weirdly from the resistor body.

Also:
- If you submit a symbol, which I would support, can you add a wiggly US resistor symbol since we're doing that now?
- I think the "main" pins should be numbered 1 and 2 to better match with the non-Kelvin resistor symbols. Then the Kelvin pins are 3 and 4.
- Some of the fluff from the description that the footprint generator script adds can be removed, and I would suggest something like `Resistor SMD 2512 (6332 Metric), 2.6mm thick, Vishay WKS2512 (http://http://www.vishay.com/docs/30108/wsk.pdf)`.

jeremysiebers:
- [X] it would be R_Vishay_WSK2512_6332Metric_T2.66mm
- [X] I think the "main" pins should be numbered 1 and 2 to better match with the non-Kelvin resistor symbols. Then the Kelvin pins are 3 and 4.
- [X] Some of the fluff from the description that the footprint generator script adds can be removed, and I would suggest something like Resistor SMD 2512 (6332 Metric), 2.6mm thick, Vishay WKS2512 (http://http://www.vishay.com/docs/30108/wsk.pdf).

wsk2512

I will create new symbols (EU and US) for this resistor type with Kelvin contacts.

jkriege2:
As discussed over at the symbol-PR (https://github.com/KiCad/kicad-symbols/pull/613#issuecomment-391542398): I think in the end this is a shunt, so we should stick to the symbol we already have ... otherwise we create a mess of symbols (two incompatible generics for the same thing ... or we renumber R_Shunt and thus break existing designs and footprints). 

The style was derived (I think) from the IEC-symbol proposal (see e.g. https://knowledge.autodesk.com/support/autocad-electrical/learn-explore/caas/CloudHelp/cloudhelp/2017/ENU/AutoCAD-Electrical/files/GUID-1D071951-D38E-4B15-8B9E-C30ECBD55C19-htm.html)!

So I would vote for designing pad-numbering to match the existing symbol and not add an additional symbol (US-style would be great though!)

jeremysiebers:
@evanshultz @jkriege2 @Ratfink @poeschlr 
Gentlemen, please decide what we want to have regarding this shunt type so that I can finish this addition.

poeschlr:
Courtyard clearance is not quite right. I will fix that in a separate pull request. (In that i will also add the other two size options.)

PR #574, opened 2018-05-07
Update all staggered to-sot footprint names to new naming convention

poeschlr:
Discussion see https://github.com/KiCad/kicad-footprints/pull/571


evanshultz:
2 questions:
1. `StaggeredEven` and `StaggeredOdd` aren't very descriptive since we have only relative comparison. All pins are staggered either way. It's not any better than `Staggered1` IMO. But, alas, I don't have a better suggestion except to specify both the odd and even pin stagger from the body.
2. Did you update the script?

poeschlr:
I think most of the changes contained here are a better solution to what we now have.
I however agree that StaggeredEven (or odd) could be expressed better, but when i an oliver wrote the original proposal we could not come up with anything better either.

jkriege2:
Did you also update the scripts? ... These are ALL script-generated!

PR #575, opened 2018-05-07
On-Shore Technology USB-B1HSxx receptacle

ppelleti:
[Datasheet](http://www.on-shore.com/wp-content/uploads/2015/09/usb-b1hsxx.pdf).

This footprint is meant to be used with the existing `USB_B` symbol.

usb-b1hsxx

------------


evanshultz:
- How did you determine pad sizes?
- Can you push the silk ref des and fab footprint name text outside the courtyard? I don't find this is a rule, but text is outside (or just barely inside) the courtyard on all USB footprints I see.
- Offset silk from fab lines by at least 0.11mm so it doesn't end up under the part body.
- Please adopt the pin 1 silk mark style as shown below:
image

evanshultz:
Here is the dimensions drawing:
image

ppelleti:
> How did you determine pad sizes?

The *drill* sizes are directly from the "Recommended P. C. Board Layout" in the datasheet.  The *pad* (copper) sizes are just numbers I made up that seemed reasonable; there wasn't really a lot of rationale behind them.  Do you have a better recommendation?

I will fix the other things.

evanshultz:
Yes, I see drill sizes are given and I checked them above. I was wondering about only the pad sizes. I suppose they're OK but I just wanted to know and have it here if you had a reason.

ppelleti:
Usually I just add 1.0mm to the drill size to get the pad size.  In this case, for the signal pads, that would have left too little clearance between them, so I went a bit smaller.  For the shield pads, I added a bit more than 1.0mm, because they are big holes, and 3.5 seemed like a round number.

ppelleti:
The other three issues should be fixed now.

usb-b1hsxx-updated

evanshultz:
Thank you!

PR #578, opened 2018-05-10
CUI PJ-063AH barrel jack

ppelleti:
[Datasheet](https://www.cui.com/product/resource/digikeypdf/pj-063ah.pdf).

Because this barrel jack does not contain an internal switch, it needs [its own schematic symbol](https://github.com/KiCad/kicad-symbols/pull/591).

I have created a version with circular holes, in addition to the version with oval holes as specified in the datasheet.

pj-063ah

pj-063ah_circularholes

------------


ppelleti:
Named the mounting pins `MP`, to coordinate with the change made in the schematic symbol.

evanshultz:
- You might as well include the voltage in the description as you have the current.
- The description says the center pin is 2.1mm, but the datasheet says 2.0mm.
- Please use the official datasheet link of `https://www.cui.com/product/resource/pj-063ah.pdf`.
- Silk should not be under the part body, so offset from fab lines by 0.11mm (now is 0.1mm).

Dimensioned drawing:
image

evanshultz:
All but the OD look good to me. I think 6.5mm doesn't mean anything useful for KiCad users, but 5.5mm (and also 2.1mm or 2.0mm here) does since it's the mating plug's standard size. Can you change it back unless you have some other thoughts?

ppelleti:
OK, I've changed the outer diameter back to 5.5mm.

evanshultz:
Thank you!

PR #581, opened 2018-05-14
Add SOT-563

whitequark:
This package is just the existing SOT-553 one with one more pin. It is required for the [DRL](http://www.ti.com/lit/ml/mpds159c/mpds159c.pdf) variant of [TI SN74LVC1T45](http://www.ti.com/lit/ds/symlink/sn74lvc1t45.pdf). Unfortunately TI does not seem to provide any footprint drawings, only a package drawing.

Footprint:
screenshot_20180514_040738
Footprint aligned with TI's STEP model:
screenshot_20180514_040726
Package drawing:
screenshot_20180514_041354
3D package PR: https://github.com/KiCad/kicad-packages3D/pull/314

------------


evanshultz:
We'll assume SOT563 is an industry standard and we can use a common footprint for TI and other suppliers. Then you may want to look at the following and see if you can develop a common theme:
- https://www.centralsemi.com/PDFS/CASE/SOT-563PD.PDF
- https://www.diodes.com/assets/Package-Files/SOT563.pdf
- https://www.torexsemi.com/file/SOT-563/SOT-563-pkg.pdf
- https://www.monolithicpower.com/pub/media/document/MP5075L_r1.0.pdf
- http://www.bourns.com/docs/Product-Datasheets/CDSOT563-T05C.pdf

A footprint calculator would be another avenue to pursue.

Would you mind taking a look at the above before this gets reviewed?

whitequark:
@evanshultz I took a look. All of these footprints are subtly different from mine, but also all of them match mine in at least one dimension. I'm not really sure what to do since they all deviate in different ways. The Bourns one at least seems clearly wrong (see the pitch) since they seem to have rounded inch to mm incorrectly.

Also, if you want me to change something here, wouldn't it imply that the other SOT-5X3 footprints all need to be changed too?

evanshultz:
They all have 0.5mm pitch, including the Bourns, right? Maybe I'm not seeing what you're talking about.

If the library has a SOT-5x3 footprint, I believe users expect it to work for any part they select in that package. So if we consider the links I sent above, and try to average their variation to produce one universal footprint, it seems like the following would be a a potential path (`min` and `max` are the distance between pads and to the outer extents of the pads):
```
	padx	pady	pitchy	min	max
central	0.6	0.4	1.4	0.8	2
diodes	0.5	0.375	1.7	1.2	2.2
torex	0.45	0.34	1.35	0.9	1.8
mps	0.6	0.25	1.4	0.8	2
bourns	0.51	0.3	1.4	0.89	1.91

KiCad	0.7	0.4	1.5	0.8	2.2
```

Generating footprints according to IPC would be even better.  Probably a generator for them will be developed soon enough.

The other SOT5x3 footprints don't have any datasheets linked so they don't even have a single reference to their parameters. Leaving them alone is OK, but if you wanted to make further changes that's fine too.

So what do you think? You have the same pad X position, so enlarging the pads as above could give us better compatibility with other SOT-563 packages. Or checking out what IPC says. Let me know what you are thinking.

whitequark:
@evanshultz Thanks for the investigation. I've enlarged the pads to 0.7x0.4:
screenshot_20180524_075728
screenshot_20180524_075615
Let me know if anything else needs to be done.

whitequark:
Regarding the Bourns footprint, I mean that the Y pad pitch is specified as 0.51, whereas the pin pitch is specified as 0.50. This seems wrong to me, why would they not match?
screenshot_20180524_080105
screenshot_20180524_080122

evanshultz:
Thanks! Without using IPC or some standard, that's probably good enough. Please change the row-row spacing to 1.5mm (it is still 1.4mm). Then this is ready to merge.

Yeah, probably silly rounding or a typo on the Bourns part. Not the package drawing shows 0.50mm.

whitequark:
Done!

evanshultz:
The pad placement looks good, but the courtyard is +/-0.1mm too small in the X direction. 0.75mm + (0.7mm / 2) + 0.25mm = 1.35mm. If you can fix that I'll merge.

whitequark:
Oops, I was wondering if I enlarged it enough. Fixed.

whitequark:
Thanks!

whitequark:
@evanshultz Looks like this footprint can't be used with 6/6 or even 5/5 rules, this seems wrong. Opinions?

evanshultz:
Can you elaborate on what "6/6" and "5/5" rules are?

whitequark:
@evanshultz 6/6 and 5/5 refer to trace width and trace separation, in mils, respectively. The lower the more expensive the board.

evanshultz:
Gotcha. Thanks. Obvious in hindsight.

From another perspective, SOT-563 is small and perhaps bigger parts should be used if the design rules are a bigger factor in the board design?

Ultimately, if the footprint is designed properly for the part then we should leave it at that. Deviations are possible with unique footprint names, but the default footprint for a part should give the best chance of success in the widest range of uses, considering the vast range of potential users of this library and the absence of information about of their requirements. That's what I think.

Can you detail your thoughts?

whitequark:
> Ultimately, if the footprint is designed properly for the part then we should leave it at that.

I agree.

> Deviations are possible with unique footprint names

Do you think this deviation has place upstream? If so, what would its name be?

jxltom:
The pad to pad distance is only 0.1mm. I think the clearance is too small.

Why don't use pad width as 0.3 as recommended in datasheet?

PR #582, opened 2018-05-14
Add Cypress-specific QFN-56 footprint.

awygle:
For CY7C68013A/14A.

This is a footprint for the Cypress CY7C68013A/14A.

Tagging @evanshultz here because this footprint is a bit weird. See discussion in https://github.com/KiCad/kicad-symbols/pull/289 for more details.

Datasheet: http://www.cypress.com/file/138911/download
Screenshots: 
image
image
image

------------


awygle:
> Please link to the Cypress datasheet and app note in the description.

Done.

> Footprint name on fab layer should be 1x1mm text size.

Done.

> Would you mind unrotating pads and vias? For some of the pads it will make a difference because the X and Y dimensions are flipped, but for vias and the paste pads it serves no purpose.

I did this, but I admit I don't really see the reason for it. Is the expectation that all pads in all footprints be at 0 rotation?

> Doesn't page 62 of the datasheet show that circular mask should be over the top of the vias (it shows "diameter"? You have mask on the via bottoms and a patchwork or lines, polygons, and hollow circles on top mask. I just want to be clear that we both share the same objective and I'm understanding what you've done to meet that objective since this is wacky footprint.

I removed the artwork on the B.Mask layer - that was a mistake. 

The reason for the patchwork of lines, circles, and polygons is because the mask layer is a negative layer - those are the areas where the soldermask will not be applied. So the hollow circles are required to _leave_ mask over the top of the vias, and then all the rest of the mess is to fill in the rest of the square. That's my intention and why I approached it the way I did.

> The vertical mask lines can be a bit longer or mask circles can have a bigger OD to fully bridge the mask layer and cover up the tiny holes in the mask that currently exist around the vias.

Fixed this - no more little holes.

> The top mask should only be the same size as the thermal pad, or 6.22mm square. It's too big.

Done.

>The user can swell (or shrink) mask on the board which could uncover vias if they swell dramatically. Interestingly, lines on the mask layer don't swell according to this board-level setting so we're only in trouble if the user swells the outer pads so much that it covers up vias under the part. That takes a heck of a lot of swell so I'm not sure we have a significant reason to worry about it. (Note that we can control mask oversize on mask-only pads, but since outer pads are the problem I don't think it matters through it might simplify the footprint a bit.)

It would really take some work to end up in this situation and you'd probably have tons of other problems at this point. If I understand right, I could specify a non-zero mask expansion on the outer pads to "lock" them relative to the global setting, but I don't see that as being worth the lost flexibility for people using sane expansion settings.

> Since the package isn't specific, please use a generic 3D model.

There isn't a 3D model for a QFN that's 8x8mm, but I removed Cypress from the name.

whitequark:
> There isn't a 3D model for a QFN that's 8x8mm, but I removed Cypress from the name.

Can you please also add a 3D model?

evanshultz:
Sorry that I forgot about this PR!

Only one change:
- Remove the `Package_DFN_QFN:` from the first line of the file. Some version of KiCad did this for a while and it's wrong. The newer nightlies should fix this, or you can manually edit the file. This will fix more of the Travis violations.

And some follow up discussion points:
- Rotating the pads just increases the file size for no benefit versus changing the X and Y size. Personal annoyance of mine to see rotated pads.
- Right. I get that mask is negative, I just wanted to make sure that I understood what you did and didn't misunderstand anything. We're on the same page. :)
- There is a ring of mask around the outside of the thermal pad, but this is probably fine as it should act like a dam to block solder paste bleeding since solder can't go through the vias. It may make the IC float during reflow, so stencil thickness any any adjustments to the paste openings will have to be used. I see there is ~70% paste coverage which, in the absence of better data, is OK for me.
- I agree that setting a fixed mask swell isn't worth it. Just pointing out a wacky situation that doesn't normally exist in any of our footprints.

I'd be remiss not to mention that Rene wrote a very nice QFN generator script. While this footprint is surely fine, and we did a lot of work to find out what Cypress recommended, using standard IPC values to derive footprints is probably even better. I'll merge this, just FYI for the future.

Will you submit a PR to update the schematic symbol to use this footprint? If not, that's OK and I will.

whitequark:
@evanshultz Done.

evanshultz:
Thanks!!

PR #519, opened 2018-04-25
Add Y14 Relay

asukiaaa:
Add Y14 relay.

http://www.hsinda.com.cn/en/ProductShow.asp?ID=208

screenshot from 2018-04-25 20-23-56

Thanks.

------------

- [x] Provide a URL to a datasheet for the footprint(s) you are contributing
- [x] An example screenshot image is very helpful 
- [x] If there are matching symbol or 3D model pull requests, provide link(s) as appropriate
- [ ] Check the output of the Travis automated check scripts - fix any errors as required

PR #583, opened 2018-05-15
Add footprint of SIMCom900

jneiva08:
datasheet:
http://simcom.ee/documents/SIM900/SIM900_Hardware%20Design_V2.05.pdf

simcom900

This PR is to complete symbol [#519](https://github.com/KiCad/kicad-symbols/pull/519)
------------


evanshultz:
- Assuming there is only this one package for the SIM900 module, it can be called `SIMCom_SIM900`.
- Please add a useful description like `Quad band GSM/GPRS module, 24x24x3mm` in addition to the datasheet link you already have.
- Please add a better description to both keepouts. Area A on page 17 of the datasheet if copper keepout (trace, vias, zone, etc.) and should mention all copper, not just vias. Area B is only a via keepout and it needs some indication of such.
- Line thickness for these keepout areas should be 0.1mm.
- Put the fab layer ref des under the part body.
- SIlk should not be under the body. Offset it from fab lines by 0.11mm.
- For this package, I think a small triangle outside the part will be a better pin 1 silk marker.
- The only dimensional changes is to make the courtyard clearance 0.25mm:
image

evanshultz:
And thanks for taking the initiative to improve the KiCad library!

evanshultz:
When making a major visual change, please post a new screenshot.
image

Also in the future, please use a more salient commit message than "Modified footprint as requested". Someone will see this later and not have any idea what changed, and certainly not why it was changed.

It's looking good! A couple things left to do:
- Typo of "copper" for area A.
- The pin 1 silk is too close to the pad. However, after seeing it again you could eliminate the triangle and remove the vertical silk line by pin 1 for the indication. The removal of silk indicates pin 1 then.

jneiva08:
This is how it looks now after commit.

sim900-footprint

evanshultz:
And you even noticed no pin 1 mark on the fab layer and added something to indicate that. Thanks!

PR #584, opened 2018-05-15
new footprint Würth 7499010121A ethrenet magjack

lorem-ipsum:
[datasheet](http://katalog.we-online.de/pbs/datasheet/7499010121A.pdf)
Matching [symbol PR](https://github.com/KiCad/kicad-symbols/pull/603)
screenshot_20180515_230517
------------


evanshultz:
- Please snuggle the silk lines just 0.11mm offset from the fab lines. You will have to break those lines where they pass too close to a pad.
- Change `-` to `_` in the footprint name and add a trailing `_Horizontal`. See the existing one in https://github.com/KiCad/kicad-footprints/tree/master/Connector_RJ.pretty.
- Please move your keywords info to the description along with the datasheet and add just simple keywords.
- Only the courtyard clearance if off dimensionally. It should be 0.5mm all around:
image

@poeschlr 
On this footprint should the horizontal courtyard be from the body or the outside of the 1.6mm mounting holes?

lorem-ipsum:
I integrated your feedback and fixed the broken courtyard. I think I included the markings on the silk screen, but in this case they were weird.
I did not include the mounting holes in the courtyard as the mounting holes are in line with the package and do not extend further.
screenshot_20180528_133531

evanshultz:
@lorem-ipsum 
- Please use `horizontal` instead of `side entry` in the description.

@poeschlr 
Please see the question above about the courtyard so I can finish reviewing this. Thanks!

PR #585, opened 2018-05-16
add footprint SM-22 for MACOM Transformer

lorem-ipsum:
This adds a footprint for RF transformers produced by MACOM. SM-22 is there name for this housing. I could not find any other manufacturer using this, so I've added their name. [Datasheet](https://cdn.macom.com/datasheets/ETC1-1-13.pdf)
screenshot_20180516_113218

------------


lorem-ipsum:
Hi @Ratfink,

thanks for the review. I did not check at all whether a similar PR already existed. I should do that in the future. I integrated your suggestions. That's how it looks now:
screenschot

Ratfink:
Ordinarily I'd ask for the silkscreen to be a bit closer to the part body than you have it here.  However, given the dimensional tolerances of ±0.2 mm, much closer would risk having silkscreen under the part body.  So, I think this looks good as-is.

Thanks for the good work!

PR #586, opened 2018-05-16
Add chamfer to edge pads fo LGA 14 2.5x3mm

poeschlr:
This allows for a larger pad to pad clearance between the corner
pads. But with the downside that this fp is now incompatible
with kicad 4.

screenshot from 2018-05-16 13-05-00


evanshultz:
Is there a recommended footprint for this package? The link in the footprint (www.st.com/resource/en/datasheet/lsm6ds3.pdf) only has a package drawing. I found www.st.com/resource/en/technical_note/cd00134799.pdf, but again it doesn't have that info.

poeschlr:
This is again one of my script generated footprints using ipc suggestions. It had been added in the past with the only difference being now that i chamfered the corner pads as the script now supports that. (And the clearance between these was quite small)

Pull request: https://github.com/pointhi/kicad-footprint-generator/pull/125

PR #589, opened 2018-05-17
Mini-Circuits Filter Mounts FV1206 and variants

lorem-ipsum:
This is the first time I've placed vias in a footprint, I hope it is correct to use THT connectors for this. 

Datasheets:
- [FV1206](https://ww2.minicircuits.com/case_style/FV1206.pdf)
- [FV1206-4](https://ww2.minicircuits.com/case_style/FV1206-4.pdf)
screenshot_20180517_083419

------------


lorem-ipsum:
Added the remaining variants used by the high pass and band pass filters
[FV1206-1](https://ww2.minicircuits.com/case_style/FV1206-1.pdf)
screenshot_20180517_220826
[FV1206-5](https://ww2.minicircuits.com/case_style/FV1206-5.pdf)
screenshot_20180517_220850
[FV1206-6](https://ww2.minicircuits.com/case_style/FV1206-6.pdf)
screenshot_20180517_220914
[FV1206-7](https://ww2.minicircuits.com/case_style/FV1206-7.pdf)
screenshot_20180517_220939

lorem-ipsum:
I did just realise that the footprints are not only used for filters, but also for other components (example is [this power splitter](https://www.minicircuits.com/WebStore/dashboard.html?model=QCN-3%2B)). Now I am wondering what's the correct package library? Would `Package_DFN_QFN` fit?

evanshultz:
Thanks! These footprints look good.

I'd like to give some general feedback based on the images above and then do a real review once you've been able to tackle those items. Is that OK?
- Put the ref des above the part and value below it.
- Silk lines should be 0.11mm offset from fab lines (yours are farther) and 0.12mm away from copper.
- Make the fab layer ref des as big as possible, up to 1x1mm, so it fills the part body.
- Your keywords should be in the description (keep the URL) and then add generic keywords.
- Via annular rings should be 0.15mm. For 0.3mm drill, use 0.6mm pad. Some are good but not all.
- The pin 1 silk mark can remove silk instead of adding a line. This might be good for FV1206-4, for example. But it's up to you. Play around and see what looks good.
- FV1206 doesn't have a pin 1 silk marker that I can see.
- I don't see a fab layer pin 1 marker on any footprint. These might work with a notched corner or a small triangle inside the fab outline.

poeschlr:
Finished in https://github.com/KiCad/kicad-footprints/pull/882

lorem-ipsum:
Thanks for taking care of this. I am sorry that I did not find the time to finish this properly.

PR #592, opened 2018-05-20
CUI PD-30S connector.

ppelleti:
[Datasheet](https://www.cui.com/product/resource/digikeypdf/pd-30s.pdf).

This is a shielded version of the PD-30 connector which is already in the library.  The PD-30S can use the same schematic symbol as the PD-30.

I've provided the PD-30S in versions with and without circular holes.

pd-30s

pd-30s_circularholes

------------


evanshultz:
- I'm thinking this should go in Connector_DIN.pretty.
- Then the connector name would prefix with `DIN_` so CvPcb can filter footprints nicely. It might be `DIN_3Pin_` or since I think this one falls into the "Connectors for Specific Functions" section of http://kicad-pcb.org/libraries/klc/F3.6/.
- In the description please note that this is DIN connector.
- I feel like there should be a line, maybe on the fab layer, where the circular part of the body sticks out. That part of the component may need to hang off the PCB. Or maybe on Dwgs.User?
- What symbol goes with this? It looks like if there were `_MountingPin` and/or `_Shield` versions of the existing standard DIN symbols it would be fine. I currently see a specific symbol for this footprint but I don't know if it's needed. Can a generic symbol be used?

Dimensioned drawing:
image

@poeschlr 
Can you please confirm the footprint name and also verify that the courtyard should extend from the edge of any NPTHs?

ppelleti:
> It looks like if there were _MountingPin and/or _Shield versions of the existing standard DIN symbols it would be fine.

The only problem I see with that is that the `DIN-3` symbol has a diagram of where the pins are.  The PD-30 is a nonstandard connector with the pins in different positions.  So, using `DIN-3` for it would be somewhat misleading.  (Note that there are already two different versions of `DIN-5` and two versions of `DIN-7`, which differ only in the placement of the pins.)

ppelleti:
> I'm thinking this should go in Connector_DIN.pretty.

It's worth pointing out that none of the footprints in `Connector_DIN` are [DIN Connectors](https://en.wikipedia.org/wiki/DIN_connector) in the colloquial sense of meaning "a round connector."  That footprint library contains rectangular connectors which are specified by some [DIN](https://en.wikipedia.org/wiki/Deutsches_Institut_f%C3%BCr_Normung) standard.

The PD-30 is the opposite, in that it's a circular connector (so "DIN" in the colloquial sense), but is not specified by a DIN standard.

ppelleti:
> In the description please note that this is DIN connector.

I have done this now.

> I feel like there should be a line, maybe on the fab layer, where the circular part of the body sticks out. That part of the component may need to hang off the PCB. Or maybe on Dwgs.User?

I have added the line on the fab layer.

pd-30s-line

As for the other suggestions, in addition to the concerns I already mentioned, I have this concern:

Since this is just a shielded version of [CUI_PD-30](https://github.com/KiCad/kicad-footprints/blob/master/Connector.pretty/CUI_PD-30.kicad_mod), I think that `CUI_PD-30` and `CUI_PD-30S` should be in the same library, use the same naming scheme, and have the same pin numbering scheme.

Therefore, if we pick a new library/naming scheme/pin numbering scheme for `CUI_PD-30S`, we should update `CUI_PD-30` to match.  However, if the current "freeze" prohibits changing the location, name, and/or pin numbering of `CUI_PD-30` (and I'm unclear whether it does or doesn't), then I think we should make `CUI_PD-30S` match the existing `CUI_PD-30`.  (As what I have currently submitted does.)  I think it would be a shame to have two such similar footprints be mismatched.

ppelleti:
> and also verify that the courtyard should extend from the edge of any NPTHs?

Since the NPTHs in this footprint are near the center and far from the edge, I don't really see how they would influence the courtyard.

evanshultz:
You have a number of great points! Especially when read all together. Thanks for thinking about this and putting it down here.

The freeze, as I understand it, is only related to adding or removing libraries so the lib table files don't include libs where aren't there or omit libs. Moving elements between existing libs or adding elements to existing libs should be fine.

Agree that both footprints should be in the same lib and have a similar look. They will need to be separate footprints, if for nothing else than to allow unique 3D models.

I'll move the NPTH question into a separate Issue.

@poeschlr @jkriege2 @Ratfink 
- This part has 3 logical pins like the `DIN-3` symbol, but it doesn't have those pins placed in the same locations as a "normal" 3-pin DIN jack. Should the symbol be shared or should it have a unique symbol?
- Does the `Connector_DIN` library just capture connectors specified by a DIN standard, or also the colloquial "DIN connector" footprints? In other words, which lib should hold CUI PD-30* footprints?

poeschlr:
Are the pins with number 4 used for shielding or only for mounting the part? (If they are for the former then the pin number should be SH otherwise MP.  That way the generic symbols can definitely be used. Not sure if we have DIN connector symbols for shielded versions or for versions including mounting pins)

ppelleti:
Sorry, I've been busy with things other than PCB design lately.

> Are the pins with number 4 used for shielding or only for mounting the part?

I think some/all are used for shielding, but it's a bit confusing to me.  There are five pins that I've given the number 4 to, but one of them is different than the others.  For clarity, I've color-coded them red and green here:

pd-30s-color-coded

Also, it's helpful to view them in the context of the "unshielded" version (PD-30, which is [already in the library](https://github.com/KiCad/kicad-footprints/blob/master/Connector.pretty/CUI_PD-30.kicad_mod)).  In this photo, PD-30 is on the left, and PD-30S is on the right:

p18_3774 pd-30 and pd-30s

So, the "green pin" is present on both the PD-30 and the PD-30S.  It connects to a metal ring that mates with the DIN-style connector.  The "red pins" are only present on the PD-30S, and connect to the metal enclosure (which is what makes the PD-30S different than the PD-30).

I'm not sure exactly what this means, in terms of how the pins should be numbered.

ppelleti:
What would you like me to do next on this?

evanshultz:
I would then that green pin 4 should be called 'SH' since it's the shield of the signals going through the connector.

The red pin 4s are also a shield, but only for the connector housing.

Let us know if you still would like to complete this PR and I'll take a closer look.

ppelleti:
> The red pin 4s are also a shield, but only for the connector housing.

So should all five of the pin 4s be designated `SH`?

> Let us know if you still would like to complete this PR and I'll take a closer look.

I'm happy to complete the PR if you guys can reach a conclusion on the naming issues.

evanshultz:
The associated symbol (I read the conversation above but would argue against a specific symbol and go with a generic symbol) looks like this:
image

So... I think pin 4 there is actually poorly named and should be named 'SH'. However, that boat seems to have already left the dock.

I don't see a suitable symbol present for PD-30S, so there's no way to use this footprint right now. I suppose the most straightforward way is to rename the red pin 4s above to 'SH' and then a new symbol can be added like the one above but with a matching 'SH' pin. Any objections? @poeschlr

ppelleti:
> So... I think pin 4 there is actually poorly named and should be named 'SH'.

When I contributed this symbol and the PD-30 footprint during the version 4 release cycle, the KLC did not allow alphabetic pin numbers.  That seems to be new to the version 5 library.

> I suppose the most straightforward way is to rename the red pin 4s above to 'SH' and then a new symbol can be added like the one above but with a matching 'SH' pin.

So you want to keep the green pin 4 as pin 4, and add the red ones as SH?  Or did I misunderstand?

evanshultz:
Yes, pin names (not numbers) are new. And not documented in KLC AFAIK.

What you wrote above sounds good to me. Thanks!

poeschlr:
Shield and mounting pin numbers are documented in the KLC. See: http://kicad-pcb.org/libraries/klc/S3.7/

evanshultz:
See the datasheet. The red pin 4s above are called "shield" on this connector. The green pin 4 may be a shield for just the cable, but it depends on the cable design. I don't see any of them as mounting points.

ppelleti:
> I don't see a suitable symbol present for PD-30S, so there's no way to use this footprint right now. I suppose the most straightforward way is to rename the red pin 4s above to 'SH' and then a new symbol can be added like the one above but with a matching 'SH' pin.

I've added the new symbol in KiCad/kicad-symbols#1581.

pd-30s-footprint

pd-30s-circular-footprint

myfreescalewebpage:
@evanshultz ping ?

myfreescalewebpage:
I try to continue the review (I have not read all comments above so point me if this was already discussed):

CUI_PD-30S
- [ ] Pins 1 and 2 Hole size X should be 0.7mm
- [ ] Pin 6 Hole size Y should be 0.7mm
- [ ] Position of SH pins not correct, should be at X=5.35mm from pin 1 and pin 2
- [ ] SH pins should be all rectangular, isn't it ?
- [ ] I disagree with the external dimensions of the footprint, 13.4mm instead of 13.5mm for example, and 16mm instead of 16.1mm

Can you make a general review and I will continue to review after ?

Thanks
Joel

ppelleti:
> Pins 1 and 2 Hole size X should be 0.7mm

Pins 1 and 2 are specified to have 0.6x2.5mm holes in the datasheet.  Where does 0.7mm come from?

pd-30s-datasheet-holes

> Pin 6 Hole size Y should be 0.7mm

There is no pin 6.  Pins are numbered 1, 2, 3, 4, and SH.

> I disagree with the external dimensions of the footprint, 13.4mm instead of 13.5mm for example, and 16mm instead of 16.1mm

The leftmost fab line has an X position of -5.5, and the rightmost fab line has an X position of 10.5.  The difference between -5.5 and 10.5 is 16.0.  Can you elaborate on where you're getting 16.1 from?

myfreescalewebpage:
Thanks for the quick answer. All come from the datasheet you have link in the PR and in the footprint: https://www.cuidevices.com/product/resource/pd-30s.pdf

The image you have just shown in your last message do not come from the datasheet, and it is not consistent with the datasheet, where have you retrieve it ?

Datasheet is the wrong one or your image is the wrong one, I don't know which one to follow :)

ppelleti:
This is the datasheet I downloaded from https://www.cui.com/product/resource/pd-30s.pdf on April 29, 2018:

[pd-30s.pdf](https://github.com/KiCad/kicad-footprints/files/5271623/pd-30s.pdf)

The datasheet at that URL now is completely different.

myfreescalewebpage:
I see. The date in your datasheet is 2009. The date in the new datasheet is 2020 so I suggest to update the footprint. According to the table revision in the last datasheet some update have been made on the connector itself probably.

ppelleti:
OK, I have updated the footprints based on the new datasheet.

myfreescalewebpage:
Thanks for this late update, looks great, merging :)

PR #587, opened 2018-05-16
Add some of the qfn packages needed for texas MCU symbols

poeschlr:
This adds some of the footprints needed for fixing https://github.com/KiCad/kicad-footprints/issues/540

Symbol PR is https://github.com/KiCad/kicad-symbols/pull/560

This also fixes parts of https://github.com/KiCad/kicad-symbols/issues/520

PR #593, opened 2018-05-21
Thermal via version of previosly added qfn footprints

poeschlr:
- This adds the thermal via version of some already added qfn footprints
- These are script generated and use custom pads (chamfered rectangles)
  for increased paste coverage while still avoiding paste over vias.
  (At least where texas did not suggest something else)

The original footprints where added in https://github.com/KiCad/kicad-footprints/pull/587

The scripts are found in https://github.com/pointhi/kicad-footprint-generator/pull/125

---

Overview screenshots:

screenshot from 2018-05-21 22-42-42
screenshot from 2018-05-21 22-43-08


evanshultz:
These look great!

A few questions:
1. The pin 1 mark is removing silk from the pin 1 corner? http://www.ocipcdc.org/archive/What_is_New_in_IPC-7351C_03_11_2015.pdf shows a horizontal but not vertical silk line on QFN-like footprint on pages 16-17. We use this style already.
1.  Do the paste pads need to be custom shapes? Can they be square or rectangles?
1. Do you want paste to avoid being over vias?
2. The vias don't have the `*.Mask` layer which I thought we wanted in the official lib.

poeschlr:
> The pin 1 mark is removing silk from the pin 1 corner?

I can change that.

> Do the paste pads need to be custom shapes? Can they be square or rectangles?

Would reduce the paste coverage if you still want clearance to the via. I don't see a reason why not to use custom shapes.

> Do you want paste to avoid being over vias?

Where possible yes. (It reduces solder wicking)

> The vias don't have the *.Mask layer which I thought we wanted in the official lib.

I explained that already in great detail in the other post with this discussion. The vias have no mask set. This means they are bottom tented. At most i would set B.Mask as the top mask should be controlled by the top pad stack. (In the old post i pointed you to an example footprint that is build exactly like this ones)

evanshultz:
2 & 3. Would it be best for official library footprints to avoid paste over vias? I think yes.
4. I think you mean https://github.com/KiCad/kicad-footprints/pull/285.

poeschlr:
Sadly the script is not able to create the texas 24 pin qfn footprint to avoid paste over vias. The screenshot is the result for 55% coverage (i guess it is closer to 50% as the script ignores the chamfers when calculating paste coverage.)
With the normal 65% the paste pads overlap completely. (One option to get more coverage with this is shown in green. We would need to create this footprint by hand.) Also the datsheet suggests a very similar via and paste setup compared to the current footprint.
screenshot from 2018-05-26 03-38-08

For the 64 pin version i managed to avoid the vias with paste.

screenshot from 2018-05-26 03-29-00

evanshultz:
@poeschlr 
Alright then. I'm fine to merge so we can get missing footprints in the library. Is there anything else you want to do with this PR?

poeschlr:
No right now this should be ready. (Anything more complex for the paste layer needs to for after this years shell eco marathon as i really need to finish my electronics for that event.)

evanshultz:
Then thank you!

PR #594, opened 2018-05-22
move Mini-Circuits CDxxx footprints to Package_SO

lorem-ipsum:
they are not only used for transformers, but are a rather generic footprint used by the manufacturer (example is [this](https://www.minicircuits.com/WebStore/dashboard.html?model=ADP-2-1%2B) power splitter. From Wikipedias description of the different packages I assumed them to be part of the SO category.

No changes to the footprints.

------------

### DIFF
--- a/Package_SO.pretty/Mini-Circuits_CD541_H2.08mm.kicad_mod
+++ b/Package_SO.pretty/Mini-Circuits_CD541_H2.08mm.kicad_mod
@@ -1,11 +1,11 @@
-(module Transformer_Mini-Circuits_CD541_H2.08mm (layer F.Cu) (tedit 5A365E23)
+(module Mini-Circuits_CD541_H2.08mm (layer F.Cu) (tedit 5A365E23)
   (descr https://ww2.minicircuits.com/case_style/CD541.pdf)
   (tags "RF Transformer")
   (attr smd)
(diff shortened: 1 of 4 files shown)


evanshultz:
@poeschlr @jkriege2 @Ratfink 
I'm convinced to move them. I suppose `Package_SO` is fine, especially since we removed the `IC` part. They so, after all, have small outlines. Anybody object or find a better library?

Ratfink:
I guess that's a fine place for these to go.  They don't really feel like other packages in that library, but I can't think of a better place for them.  Well, unless we create a whole library for them like we did with Package_DirectFET, but that seems like overkill in this case.

poeschlr:
Moving them is ok for me.
(As long as it does not delete a library or leave a library empty. If the later happens we need to leave some dummy footprint in that library as to not disrupt existing kicad installations that have this lib in their fp-lib-table)

lorem-ipsum:
Is there anything todo or is it just a lack of time right now?

poeschlr:
Before this is merged someone must also update the symbols that reference these footprints. (13 symbols in the transformers library.)

poeschlr:
Ok ignore me. I did miss that the symbol PR is already created. Thanks for the contribution.

PR #598, opened 2018-05-23
SD_TE_2041021: Use non-numeric pin numbers for CD, WP, and SH.

ppelleti:
This is necessary because of [the change being made to the symbol](https://github.com/KiCad/kicad-symbols/pull/592).

sd_te_2041021

------------


ppelleti:
@evanshultz @poeschlr This only makes sense if KiCad/kicad-symbols#592 is also going to be merged.

evanshultz:
Yes. And https://github.com/KiCad/kicad-symbols/pull/644 is also affected. I'll post an update at the PR you linked since it seems like the most relevant conversation.

This may need to be reverted since it's breaking things and wasn't agreed upon.

Issue #595, opened 2018-05-22
Minimum annular ring for "big" pads


evanshultz:
@poeschlr @jkriege2 @Ratfink @Shackmeister 
We specify the minimum annular ring for vias in KLC, but this doesn't apply for big pads. For example, mounting holes or large pins on components.

See https://github.com/KiCad/kicad-footprints/pull/578 and the part datasheet. For many connectors, only drill hole sizes are given and the annular ring isn't. We currently don't have and standard guidance for contributors.

Using IPC-2221 Level A, and thus being conservative for this type of hole, would require a 0.7mm annular ring width. As this user chose 1mm I think it's fine, but I also think KLC should give a recommendation. Would 1mm be a reasonable minimum annular ring to require for this type of hole?

poeschlr:
I took a quick look at IPC-2221 (http://www.sphere.bc.ca/class/downloads/ipc_2221a-pcb%20standards.pdf)

I am a bit confused. Section 9.1.2 (table 9-2) defines the annular rings. The largest one is 0.15mm. (This is the one set in the KLC)
1mm annular ring seems a bit excessive to me. Do you mean 1mm final pad size? If the drill is 0.7mm +2*0.15mm would be 1mm final size. 
1mm annular ring would result in 2.7mm pad size for a 0.7mm drill. So if the 0.7mm drill has a pin pitch smaller than 2.7mm + pad to pad clearance, you can not achieve such a large annular ring. (Annular ring is defined as the smallest copper width anywhere around the pad.)

evanshultz:
Table 9-2 applies to vias, not pads where a component lead sits, and that is 0.15mm. Since we can't control pad size on a per-layer basis, and IPC says "All lands and annular rings shall be maximized wherever feasible" on page 73, I believe that means we need to use the external unsupported size on all layers. That is noted in KLC as you mentioned and is the minimum annular ring for any padstack.

But I'm talking about pads for a component lead which also need to consider Table 9-1. After re-reading I think it may mean diameter so using the equation on page 73, and assuming Level A, we get 2 * 0.15mm + 0.4mm = 0.7mm over the drill diameter for the minimum pad diameter.

1mm over the drill size, as I mentioned above, is a bit generous but not overly so. We could use 0.7mm over the drill for a "big" pad and it should be OK per IPC. So... use 0.7mm oversize for pads like this in lieu of any specific pad size given from the part vendor?

And, of course, users can shrink or change the pads in their library or on the board if they like.

[...]

PR #600, opened 2018-05-24
Added LEM-LTSR-NP footprint

jeremysiebers:
Footprint to be used with LEM-LTSR6-NP, LEM-LTSR15-NP and LEM-LTSR25-NP.
See [https://github.com/KiCad/kicad-symbols/pull/615](url)

[https://www.lem.com/sites/default/files/products_datasheets/ltsr_6-np.pdf](url)
[https://www.lem.com/sites/default/files/products_datasheets/ltsr_15-np.pdf](url)
[https://www.lem.com/sites/default/files/products_datasheets/ltsr_25-np.pdf](url)

ltsr_footprint

------------


jeremysiebers:
Updated footprint regarding comments in: [https://github.com/KiCad/kicad-symbols/pull/615](url)

jeremysiebers:
Renamed footprint to LEM_LTSR-NP see [https://github.com/KiCad/kicad-symbols/pull/615](url)
Updated footprint due to fit-test on PCB test board.
Enlarged holes and set them all to circular shape.
Re-test footprint after new test board is lasered.

image

jeremysiebers:
Updated Footprint contact holes after fit test on test PCB:



Contact numbers 7 to 10 do fit but they are a little to tight, also a bit more copper pad is better for soldering. 
Contact numbers 1 to 6 do not fit so they are also increased

image

image

jeremysiebers:
Holes are now:

image

image

Look in 3D like:

image

jeremysiebers:
@Shackmeister

No, I've downloaded the 3D model from www.lem.com

Misca1234:
Hi

F.Silk line is not allowed to cross the pads, do something like this instead, otherwise it looks good

bild

evanshultz:
- As mentioned above, cut silk around pads. Fill in between pairs of pads which isn't shown above.
- Description can be just `LEM current transducer LTSRx-NP 5V supply voltage series` or similar (and keep URL
- Pin 1 should be in the top left corner.
- While not required, I prefer if you adjust pad dimensions and don't rotate any pads.
- Pins 7-10 are 0.61mm across by sqrt(0.5^2+0.35*2). Drill holes should be 0.2mm bigger according to IPC (which we adopted for KLC), so 0.81mm but yours are 0.7mm. Similarly, pins 1-6 are 0.8mm so the drill should be 1mm. Is there a reason to make those 6 pins bigger?
- Annular rings should be 0.55mm bigger then the drill (see https://github.com/KiCad/kicad-footprints/issues/595), so 1.36mm for pins 7-10. Your 1.4mm is OK.
- For pins 1-6, the pads are a bit oversized but that's OK. You could shrink the small dimension to 1.85mm just to give a bit more space, but 2.8mm for the big dimension seems OK to me.
- How did you come up with the radius of the body near pins 3 and 10?
- Your courtyard is bigger than 0.25, but all other dimensions look good to me:
image

poeschlr:
Finished in https://github.com/KiCad/kicad-footprints/pull/881

PR #604, opened 2018-05-25
Add Texas_S-PWQFN-N20 footprint

andrasfuchs:
Added the Texas_S-PWQFN-N20 to Package_DFN_QFN.pretty. This package is for the Texas Instruments TPS22993. [Symbol pull request is pending](https://github.com/KiCad/kicad-symbols/pull/588).

Datasheet: http://www.ti.com/lit/ds/symlink/tps22993.pdf

image

image

------------


andrasfuchs:
The only Travis error left is that the 3D shape is missing. Could you help me find it (if it exists) or tell me how to generate it?

poeschlr:
You don't need to add a 3d model (step or wrl) file.

But you must setup the footprint such that it can use a 3d model. In other words fill out the 3d settings. (Details see http://kicad-pcb.org/libraries/klc/F9.3/)

evanshultz:
Sorry that I was too busy to pop in earlier and missed this. Way to go!

I do have concerns about the footprint name, but it's bigger than just this footprint so it shouldn't hold anything up here. See the linked issue above.

andrasfuchs:
Thank you guys for your comments!

I tried my best to make the modifications accordingly.

Ratfink:
Looks good!  The one thing I'd like to suggest is that the silkscreen lines should be moved a bit closer to the part body, and preferably aligned to at least a 0.01 mm grid.  They could be made somewhat longer as well if you'd like.  As an example of what I mean, here's a screenshot (current shape in darker blue, my suggestion in lighter blue):
screenshot from 2018-05-25 14-35-07

As for keywords, it looks like we usually have at least one in the form "QFN-[pitch]".  Other than that, I don't think it's really all that important, since footprint associations and filters take care of most of the problem of searching for footprints.

andrasfuchs:
image

Ratfink:
That's better, but please use a grid in mm for the silkscreen lines rather than mils.  Every other dimension in the footprint is metric, so it just makes sense to stick with that (and generally we favor metric dimensions anyway).

andrasfuchs:
image

I hope I understand your suggestion correctly: I moved the silk lines very little just to have more round mm coordinates. I needed to use the 0.05mm grid and moved the lines to be align on that grid. Is this ok now?

Ratfink:
Yes, this looks great.  Thanks for all the work in creating this high-quality footprint!

andrasfuchs:
Thanks for the merge!

I know you guys are all very busy, but as a sidenote I think it would be great if you could modify the footprint generator script(s) just a little to reflect your preferences, like changing the Fab line width to 0.10 and the Silkscreen line width to 0.12.

Also the symbol travis script forced me to use a 100mil grid for the pins (it threw an error if I didn't), so I think something similar would be possible here for the footprints to force mm grid for the Silkscreen.

Thank you!

evanshultz:
@andrasfuchs 
You mean the defaults in KiCad, right? Unfortunately, the devs for the software aren't us and it's not automatic to get those preferences pushed into the tool. It would be nice if the out-of-the-box configuration matched KLC more closely.

andrasfuchs:
@evanshultz 
Well, when I first created the footprint, I had to select a script to run from the "Footprint Generators" dialog. As I can tell, these scripts are in the [KiCad\share\kicad\scripting\plugins](https://github.com/KiCad/kicad-source-mirror/tree/master/pcbnew/python/plugins) folder and they look fairly straightforward Python scripts.
I'm not sure how close the relationship between the script writers and the librarians, but it would be nice if those scripts generated something closer to your expectations ;)

poeschlr:
Oh you mean the footprint wizard scripts. Sadly these produce very bad footprints. 

For v6 we might ask the devs to hand over these scripts to us librarians. First course of action will be to delete the ones that are bad. The QFN generator will be the first to go as it does not create a valid land pattern. (No independent control over the heel and toe fillet sizes. -> not possible to create a valid pads that can be used for anything other than a hand solder variant.)

 Better scripts can be found here: https://github.com/pointhi/kicad-footprint-generator (For qfn there will be new ones available shortly as soon as https://github.com/pointhi/kicad-footprint-generator/pull/125 is merged.)

andrasfuchs:
Thanks @poeschlr for the explanation!

PR #605, opened 2018-05-25
Footprint for ublox SAM-M8Q

jneiva08:
Added footprint for ublox SAM-M8Q
* One is normal and another is for HandSolder

Datasheet:
https://www.u-blox.com/sites/default/files/SAM-M8Q_HardwareIntegrationManual_%28UBX-16018358%29.pdf

ublox_SAM-M8Q

gps_normal

ublox_SAM-M8Q_HandSolder

gps_hand

This PR for the symbol is [#633](https://github.com/KiCad/kicad-symbols/pull/633)

------------


evanshultz:
Very nice footprint. Thanks!
- Silk lines should be 0.12mm thick.
- Pin 1 should be in the top left so rotate your footprint.
- Silk can't be under parts (especially as this has a substrate sitting against the PCB) so I suggest you remove silk in the top left corner (or just the vertical part in the top left corner) to mark pin 1.
- The rounded square can be on F.Fab layer.
- You will need to set the footprints's mask clearance to 1nm to approximate having the mask openings the same as the copper pads (see page 10).
- You are not following the recommended stencil design on page 11.

Dimensioned drawing (I don't see dimension F used but it's probably that I'm blind; the radius of the arc is actually 1mm and I don't know why FreeCAD is showing 1.05mm):
image

jneiva08:
On the last commit have changed:
- Silk lines is 0.12mm,
- Pin 1 is now on the top left,
- Removed Silk under parts,
- Moved inner square to F.Fab layer,
- Added stencil design only in ublox_SAM-M8Q,

Now ublox_SAM-M8Q looks:

footprint1

footprint1_3d

Now ublox_SAM-M8Q_HandSolder looks:

footprint2

One doubt i have is F.CrtYd need to be 0.25 from F.Fab F.SilkS?
Can you explain how can ensure footprints's mask clearance to 1nm?
Will need do stencil design to ublox_SAM-M8Q_HandSolder?

Thanks for your help.

evanshultz:
The courtyard should be 0.25mm from any fab lines or copper. There is no relationship to silk lines. This is kind of like a connector where we use 0.5mm clearance, but 0.25mm is OK unless somebody else objects.

1nm is the smallest number KiCad can handle so making the mask 1nm over the copper will end up with a PCB where they are the same size (as the datasheet requests). The mask (and paste, if you ever need it) clearance can be set for all pads in the Footprint Properties: 
image

On the handsolder one there's no need for a special paste shape so this is fine.

evanshultz:
Looks good. Thank you!

PR #607, opened 2018-05-25
Add TE 282834 terminal blocks

Ratfink:
Generated using the scripts at https://github.com/pointhi/kicad-footprint-generator/pull/131

Based on the datasheet
http://www.te.com/commerce/DocumentDelivery/DDEController?Action=showdoc&DocId=Customer+Drawing%7F282834%7FC1%7Fpdf%7FEnglish%7FENG_CD_282834_C1.pdf

screenshot from 2018-05-25 13-57-25
screenshot from 2018-05-25 13-57-36

------------


evanshultz:
- We have `Connector_TE-Connectivity` so should this be `TerminalBlock_TE-Connectivity`?
- There is no README.md file in this folder.
- http://kicad-pcb.org/libraries/klc/F3.6/ indicates that the `Modifiers` field is required for **Manufacturer Specific Connectors**, but I don't think that's true. It should be optional. It also doesn't mention putting `TerminalBlock` at the beginning of the footprint name; that's not done in the Connector libraries but for some reason it is on all footprints in TerminalBlock libraries I could find.
- Silk is only 0.06mm offset from fab lines.
- As mentioned at https://github.com/KiCad/kicad-footprints/issues/595, a 1mm oversize on the drill holes should be adequate. Would you consider 2.1mm instead of 2.2mm? That also allows mask between pins which should help to avoid bridging during wave solder.
- It's not clear which direction the wires go into and out of this terminal block. Should we use an arrow or text or some indicator so users don't accidentally place this 180 degrees rotated from the desired orientation? I realize this one has the pins in the middle of the body so it will looks fine either way, but not all terminal blocks are this way. (This is a problem for all our terminal blocks, it appears.)
- How did you determine the placement of the horizontal lines at -2.25mm and +2.85mm?

Dimensioned drawing:
image

For completeness and discoverability later, you have also fixed a mix-up between RND and Phoenix library descriptions in sym-lib-table.

As an aside, it's never been really clear to me what demarcates `Connector` and `TerminalBlock` libraries...

Ratfink:
* Good point, I'll change it.
* In other libraries, those are a holdover from the bad old days when we had one repository per library.  A quick directory-vs-README count reveals that about half our footprint libraries don't have README files today.  I don't think they're required anymore.
* They're named similarly to other terminal blocks, but seeing as how the footprint filter for our screw terminal symbols is `TerminalBlock*:*`, maybe that name prefix isn't needed.  @poeschlr What do you think, should we remove that name prefix from our terminal blocks?
* Yep.  I could increase that offset in the script, but that will affect most of our other screw terminal footprints as well.  Which isn't really a problem, I'd just submit another PR once this one is done that re-generates all of the affected libraries.
* Sure thing.
* Yes, that's a problem with all our terminal blocks because they all come from the same script.  If we can agree on a style that clearly shows this, I could implement it (first here, then another PR for the other libraries).
* I measured the base of the trapezoidal section with a caliper.  I could just remove those lines if you prefer.

I fixed that mix-up, but it was fixed separately in https://github.com/KiCad/kicad-footprints/pull/602, which I merged into master.  Then I merged master into this branch to remove the ensuing merge conflict.  Because of that messy history, I'd recommend a squash merge for this branch once it's done.

Yeah, the distinction between connectors and terminal blocks does seem vague.  Maybe they'd be better as a sub-category of Connector.  It's too late to revisit that organization until 6.0, though.

Ratfink:
Actually, with regard to the library name, yes we have a `Connector_TE-Connectivity` library, but we use the shorter prefix `TE` in the names of most of our TE footprints.  So since we can't rename that until 6.0, should I just follow suit, calling the library `TerminalBlock_TE-Connectivity`, and leaving the footprints themselves as `TE`?

evanshultz:
I'm in favor of consistency for now, in absence of other data. (Since we use `Phoenix` instead of `Phoenix Contact` everywhere, for example, I think just `TE` everywhere is the best approach.)

Please adjust the silk line offset and pad diameter and I can merge. If you want to show the "entry" side of the connector my best ideas are a bidirectional arrow or some kind of consistent window showing where the wires can be inserted (some 45 degree entry terminal blocks are obvious but I think we can do something better and consistent for all terminal blocks).

Squash merge it is.

Ratfink:
You said two things in the first paragraph that seem contradictory to me: favor consistency, and use "TE" everywhere.  Since we already have a library with "TE-Connectivity" in the name, it seems inconsistent to use "TE" everywhere in this PR.  To be explicit, how do you think the library and symbols should be named?

I'll make the other script changes shortly once I know the answer to that question.

evanshultz:
I appear to have been thinking about two things with inadequate coordination between my brain and my hands.

I think `TE` is the best solution and would support moving there. Using `Phoenix` in my example above is the evidence for this shorthand already in KiCad.

However, for the time being, let's be consistent with `TE-Connectivity` so we don't break the existing lib. After v5 we can change, if everybody else agrees.

Is that good with you?

poeschlr:
I would suggest we do as little changes to footprint names as possible for the v5 release. So if current footprints use "TE" without "connectivity" then go with that. 
As the Connector_TE-Connectivity lib sets the example for the library name i would suggest to follow that example for the new library.

Ratfink:
Updated screenshots:
screenshot from 2018-06-11 14-55-59
screenshot from 2018-06-11 14-56-21

evanshultz:
Thanks!!

PR #609, opened 2018-05-26
Add ESP-07 footprint

Ratfink:
802.11 b/g/n Wi-Fi Module, similar to ESP-12E.

Datasheet:
http://wiki.ai-thinker.com/_media/esp8266/docs/a007ps01a2_esp-07_product_specification_v1.2.pdf

This is heavily based on the ESP-12E footprint, but with some changes that I feel make it saner: the silkscreen RefDes is moved to the bottom, there is no silkscreen around the antenna because [Espressif recommends making the antenna area hang off the base board](https://www.espressif.com/sites/default/files/documentation/esp8266_hardware_design_guidelines_en.pdf#page=17), and the silkscreen pin-1 mark is moved to the left of the pad.  If these changes are approved, I'll make another PR to make the same changes to the ESP-12E footprint already in the library.

The only script error is a false negative: the anchor is centered inside the module's outline.

Symbol PR: https://github.com/KiCad/kicad-symbols/pull/629

screenshot from 2018-05-26 14-18-49

------------


evanshultz:
- Under the part body should be a place keepout, while a 15mm copper keepout extending to the sides and above the module should be placed. Is that correct? Maybe instead of "Antenna" the text should instead tell the user what to keep out? And why not make the place keepout exactly the size of the body?
- I'd put the fab layer footprint name under the body so it's on the board.

Note that http://wiki.ai-thinker.com/_media/esp8266/esp8266_series_modules_user_manual_v1.1.pdf is useful for this review because it shows the pinout.

Dimensions:
image

Ratfink:
I'm not entirely sure what you mean by a place keepout, but yes, looking more closely it does appear that Espressif recommends a 15 mm copper and component keepout around the antenna.  Since this module has an antenna jack that can be used instead of the on-board antenna, that keepout is optional, so I'll have to either convey that in the text as well or make a keepout-less version of the footprint.  Let me know which you'd prefer.  Given that eventually we expect to have keepouts as an actual feature of KiCad rather than a graphical suggestion, I feel like a second footprint will eventually be needed regardless.

Okay, I'll move the value text to be under the body.  (unless you mean a different sense of "under", in which case, oops)

Ratfink:
Moved value text:
screenshot from 2018-06-05 19-13-17

evanshultz:
Sorry. I'm using Allegro terminology. "Place keepout" means no parts can be located within the keepout boundary. "Copper keepout" is a superset of "Via keepout" and "Route keepout" and "Shape keepout" meaning no copper of any type can be within the keepout boundary (and dynamic shapes are automatically cut out).

So my suggestion when considering the pathetic design rules in KiCad is to make two areas (perhaps hatch them oppositely?) and label the one under the body to indicate no parts can be placed there and the bigger one to indicate no copper should be there. You could also add a line on Dwgs.User to indicate the board edge as we do on some other footprints.

poeschlr:
Isn't the place keepout simply what the courtyard area is?

evanshultz:
Nope. I could dig around in IPC specs and point out the courtyard is for visual inspection, AOI, rework, etc. But instead, just consider a normal vertical XLR connector:
image

The courtyard will go around the body of the part if you look down at the PCB. Three legs will stand the connector several mm off the PCB surface and it can be legal to place other small parts under the XLR body. Under the body is inside the courtyard.

Hopefully one day KiCad has a useful design rule engine that takes this into account and checks for collisions between part bodies (which doesn't have anything to do with the courtyard).

poeschlr:
That would require a 3d model for every part and a full 3d check of them. The 3d model would also need to somehow include clearance settings or you would need some way of analyzing things like temperature influences.

The courtyard area is what we have for ensuring no other part is placed in this area. (DRC Kicad 5 checks for courtyard violations)
For your connector i would define the courtyard to include the ears and let the designer decide if they are ok with putting something in that part of the courtyard. (They then need to ignore a DRC error.)

evanshultz:
Yep, courtyard is all that KiCad has.

Updating the text to explain would be useful, I think. If you can do that I'll merge.

Ratfink:
screenshot from 2018-06-10 14-11-40

evanshultz:
Looks good. You either need a second comma after `components` or else remove the one comma you have. Then it looks good to me.

Ratfink:
I have two commas already for the list "metal, traces, or components".  Adding one after "components" makes the sentence worse, doesn't it?

> No metal, traces, or components, on any PCB layer if using on-board antenna

evanshultz:
Oh. Yes. I mis-read the text.

PR #610, opened 2018-05-26
Adding TE_x-776087-x footprint

thomasonw:
TE Connects, 23 position Ampseal assembly, right angle

This is the addition of an Ampseal connector.  There is no matching symbol, one uses the 1x23 connector.

The URL for the footprint is:
[http://www.te.com/usa-en/product-CAT-AM78-H342.html](http://www.te.com/usa-en/product-CAT-AM78-H342.html)

With datesheet (PDF file) accessible here:
[www.te.com/commerce/DocumentDelivery/DDEController?Action=srchrtrv&DocNm=770669&DocType=Customer+Drawing&DocLang=English&DocFormat=pdf&PartCntxt=1-770669-1](www.te.com/commerce/DocumentDelivery/DDEController?Action=srchrtrv&DocNm=770669&DocType=Customer+Drawing&DocLang=English&DocFormat=pdf&PartCntxt=1-770669-1)

I am not sure if the 'name' is correct, the 1- prefix is used to indicate Gold vs. Tin plating, while the -x postfix is used to indicate the color/keying of the connector.

image


thomasonw:
I have removed the 3D model and put in its place a dummy .wrl link

kleini:
Why do you open a new merge request. Same part is already submitted in https://github.com/KiCad/kicad-footprints/pull/287

PR #614, opened 2018-05-29
Added LQFP-48_7x7mm_1EP3.6x3.6mm

SchrodingersGat:
See http://www.analog.com/media/en/technical-documentation/data-sheets/LTC7810.pdf

Use the same 3D model as the footprint without exposed pad.


SchrodingersGat:
Simply adds an exposed pad (of the specified dimension) to the existing LQFP-48 model

image

poeschlr:
The paste layer for the exposed pad should be split. Aim for about 65% paste coverage. (Look at any other footprint with an exposed pad for inspiration.)

The pad is a bit on the small side. (No allowance at all for any tolerances. And yes i know the datasheet suggest it that way.) I would use the maximum "lead" size here instead of nominal. There is enough clearance to the normal pads and a larger pad can not really hurt.
(Using maximum size is a good approximation for getting 0 fillet around the pad without needing to to a lot of calculations. If you really want you could do the full calculation. It will result in a slightly larger pad than with maximum lead tolerances as placement and manufacturing tolerances will also be included.)

---

Sadly only my qfn generator includes exposed pad creation. (Otherwise you could have simply added it there) I really need to add this to the QFP generator as well. https://github.com/pointhi/kicad-footprint-generator/pull/125 (But you could trick it into generating the exposed pad for you and copy that part from the generated footprint into yours.)

SchrodingersGat:
The EP has been split now, and the pad size matches the maximum dimension on the datasheet.

SchrodingersGat:
image

poeschlr:
Footprint name should be:  `LQFP-48-1EP_7x7mm_P0.5mm_EP3.7x3.7mm`

poeschlr:
Looks good now. Thanks

PR #615, opened 2018-05-29
Footprint for SIM800C

jneiva08:
Added footprint for SIMCOM SIM800C

Datasheet: http://simcom.ee/documents/SIM800C/SIM800C_Hardware_Design_V1.05.pdf

Footprint

sim800c_footprint

This PR for the symbol is [#634](https://github.com/KiCad/kicad-symbols/pull/634)

------------


Misca1234:
Should not the "SIMCom_SIM800C" be outside of the f.CrtYrd line, for example at the bottom?
The %R REF** is not placed at origo (0.0, 0.8)
You made the large F.CrtYrd as a rectangle, but according to the data sheet it is jagged but I guess it is ok.

jneiva08:
Made the modifications sugested.

footprint looks now:

footprint

evanshultz:
The value can be inside the fab outline if the part is big enough. The fab layer should help with component installation and location, and putting it under the part works nicely for this purpose. But outside is OK too.

Similarly, the %R doesn't need to be at the origin because it has the same purpose. (0,0) is common if there's no reason to put it elsewhere, but the original placement of both was fine too.

@Misca1234 
Did you do any dimensional checks?

Misca1234:
>@Misca1234
> Did you do any dimensional checks?
Not more than a visual check in the link " Files changed 1 ",
I have not downloaded the foot print and checked in depth, but I do it during today

Misca1234:
I checked the dimensions and all looks good,

The distance between the F.Silks and F.Fab is 0.11, The KLC http://kicad-pcb.org/libraries/klc/#_footprint_guidelines, chapter F5.1 
does not mention this distance anymore but earlier it was 0.12
But is 0.11 ok then all is good

poeschlr:
Silk to fab offset is not critical. The only things important are that it is fully outside the part (So fab outline plus half of silk width plus part size tolerance would be the absolute minimum.)
There is no hard maximum. However it should be reasonably close to the part as otherwise it will be hard for the user to see where to place the component. (during manual placement) and machines might also struggle in the quality control step if the outline is too far away.

PR #616, opened 2018-05-29
added CPG-236/CP-236 footprint for Artix-7 FPGAs

herostrat:
Mechanical properties:
UG475, p. 264: https://www.xilinx.com/support/documentation/user_guides/ug475_7Series_Pkg_Pinout.pdf

grafik

additional pad and mask specification:
UG1099, p. 43: https://www.xilinx.com/support/documentation/user_guides/ug1099-bga-device-design-rules.pdf

grafik

Resulting in my proposed footprint:
grafik

I think it is an atomic footprint which should be connected to the XC7A35T-CPG236 Symbol.
Is this possible and wanted? Can anybody point me to where I can do this?

Thanks a lot for reviewing.
Please let me know if there is anything wrong.

------------


herostrat:
updated footprint:

grafik

Is a 3d model really required?

evanshultz:
- For this manufacturer-specific part we need to mention the manufacturer. Also, do you know the difference between CP236 and CPG236 (maybe it's the material of the package since it doesn't seem to have a dimensional effect)? If we resolve that, then the footprint name will be something like `Xilinx_CP236`. Again, that's assuming this footprint is specific to only Xilinx which I think is reasonable since it has odd missing balls and we don't have a [resolution for BGA naming in that case](https://github.com/KiCad/kicad-footprints/pull/442).
- Yes, a 3D model is required. If/when one is added the 3D viewer will show the model and just work without needing to touch the footprint.

Other items:
- Please add a description along with the datasheet. See other BGA footprints.
- For the silk pin 1 mark there are two modern styles I see in the library you can copy: Bevel the corner under pin 1 (err, A1) or add a short vertical and horizontal line outside the existing silk square. Either way is OK.
- Set the mask margin for the whole footprint instead of on each pad.
- Also, the mask margin is for each side so you will only need 0.15mm and not 0.3mm.
- Our BGA courtyard clearance is 0.5mm. You have 0.25mm.

image

herostrat:
Hey, thanks for reviewing.
I am really sorry for the stupid mistakes, I am just learning KiCad and its conventions. I will be more careful in the future.

 The naming schema of Xilinx is really strange. There are a few instances in the Artix Series alone, where they name CSG/CS, CPG/CP etc., see [UG475, p. 95](https://www.xilinx.com/support/documentation/user_guides/ug475_7Series_Pkg_Pinout.pdf)

What I could find is, that e.g. in [cs484](https://www.xilinx.com/support/documentation/package_specs/cs484.pdf) spec they write:
grafik
So it seems the difference is just the solder ball material.

Another thing to notice: the CPG-238 is the same (footprint-wise) as the CPG/CP-236.

Mouser sells only the CPG variant of the XC7A35T, so only the CPG variant seems relevant.

They call the package "238-LFBGA, CSPBGA" (so its a low profile, chip scale bga) and lists the suppliers device package name as "236-CSBGA (10x10)", which I could not find any reference to from Xilinx. Searching for those names results in only Xilinx FPGAs, so I think it is safe to assume its Xilinx specific.

So my proposal for the name would be: "Xilinx_CPG236", although it has 238 balls.
The reason is that the Xilins symbols are also called CPG236 and not CP in the KiCad library.

The ball-names are the same for 236 and 238, so if someone wants to add XC7A12T or XC7A25T (both of them I could not find to buy) in the CPG238 package, he also can use the CPG236 footprint.

I will work on the other issues asap and try to make a 3D model happen.
Thanks for your patience

evanshultz:
There were no stupid mistakes here! We're all working together to improve the KiCad library. :)

I don't find any existing Xilinx BGA footprints we there may not be any precedent (good or bad). Since leaving off the `G` means a leaded part, I agree with your footprint name. No reason to even try to capture a footprint that is non-RoHS these days (and as you found it's apparently unavailable).

CPx238 supports a different number of I/O, but the differences don't appear to be related to the footprint or 3D model from what I can see. Having a unique package would be best, I think, so that it's easier for the user to find the package called out for their component, but we can cross that bridge if a CPx238 symbol is ever added.

Thanks for making the other changes I noted above!

herostrat:
So, I think I finished this footprint according to the rules

Note:
The mask clearance is now global.
But because the size of the pad is 0,27mm in diameter, the mask clearance has to be 0,015mm in order to generate a combined size of 0,3mm.

herostrat:
grafik

herostrat:
One point I found just now in UG475 p.332:
grafik

There the soldermask opening is 0.375mm and not the 0.3mm as shown in UG1099.

I am not sure which one should be used here...
Has someone an idea?

herostrat:
Hey Tom, 

no I did not produce a board with this footprint just yet, and as it looks like I won't for the future.

That's because it does not seem to be possible to produce a board for reasonable prices (or at all really).
Most PCB manufacturers in Germany at least cannot produce vias (normal or even filled and plated for usage in pad) with a 0,01mm hole and 0,25mm plating diameter. Another problem was the 0.07 traces needed for fanout (0,1 was the lowest I could find).

Therefore I will use the same FPGA with a 1,00mm pitch package.

It was a stupid mistake on my part not to research more into the matter before I started the layout.
I thought it is (more) expensive, but nor that hard/impossible.

So if you do not plan to make 10k boards and therefore can accept the cost of laser-vias, I wouldn't recommend 0.5mm pitch.

TomKeddie:
@herostrat I have a friend doing 0.4mm BGA ball pitch using http://quickturnpcb.co.kr - they're still a few hundred dollars though.  I might switch to the 0.8mm pitch, the 0.4mm parts are temptingly cheaper.

evanshultz:
Just getting back to this. Sorry.

UG475 is newer so I'd be tempted to accept it. It's also specific for this family of parts (7 series) instead of being generalized, which I would normally be inclined to favor. I'd like to go with the UG475 design (now on page 334 of the current rev at https://www.xilinx.com/support/documentation/user_guides/ug475_7Series_Pkg_Pinout.pdf).

That being said, if you're not interested in finishing up this up because you're not using the footprint, you can close this PR. If you are interested, I can merge this but I'd prefer to see this migrated to a script-generated footprint.

herostrat:
nbd, I agree to generate this via the footprint generators.
I try to come back to this in the near future.

If you have no problem with it I would like to keep this PR open (so that I do not forget it) and I will close everything as soon as I can merge the generated ones

herostrat:
Closing it as the script generated version is in https://github.com/KiCad/kicad-footprints/pull/1557.
Thanks for the help :)

PR #620, opened 2018-05-30
Add/update LEM footprints

evanshultz:
For https://github.com/KiCad/kicad-symbols/pull/615 I reviewed some LEM footprints. Seeing how awful they were I went ahead and spent time on all non-Allegro and non-AKM footprints in this library.

LEM_HO-NP
- Updated descriptions
- Wrong pin placement and number of pins! (It seems to be a copy of the item below)
- Some dimensions guessed since no 3D model is available.
- Update datasheet to https://www.lem.com/sites/default/files/products_datasheets/ho-np_0100__1100_series.pdf

LEM_HO-NP_SP33
- Replace underscore within footprint name with hyphen
- Updated descriptions
- Should it be named `LEM_HO-NP-SP33` (hyphen before "SP33") since the underscore usually moves to a new section of the footprint name?
- Did not capture the thinner part of the body near pins 8-13
- Update datasheet to https://www.lem.com/sites/default/files/products_datasheets/ho-np_sp33-1000_series.pdf

LEM_HTFS
- Near-total redesign to match datasheet dimensions and add fab lines
- Fix silk line width to 0.12mm and fab line width to 0.1mm
- Avoid silk under part body
- Add datasheet of https://www.lem.com/sites/default/files/products_datasheets/htfs_200_800-p_sp2.pdf

HX*_HallEffectCurrentTransducer
- Remove `_HallEffectCurrentTransducer` from footprint name
- Add `LEM_` to footprint name
- Correct dimensions per datasheet
- All silk lines to 0.12mm
- Add/separate footprints where pin sizes differ and by MPN
- Note: HX 50-P/SP2 kept rather large pads from existing footprint
- Updated datasheets

LEM_LA25-P
- Rename footprint to accomodate all LA variants
- Moved pins to correct Y position
- Moved fab value to inside part body
- Avoid silk under part body
- Make pin 1 silk mark equal X and Y value from fab lines
- Update datasheet to https://www.lem.com/sites/default/files/products_datasheets/la_25-p.pdf

LEM_LV25-P
- Set pad drill size to match datasheet for pins 4-6
- No drill size is given for pin 1 and 3, but made them the same oversize from the pin diameter as pins 4-6
- Moved fab value to inside part body
- Make pin 1 silk mark equal X and Y value from fab lines
- Minor correct to bottom courtyard line location
- Update datasheet to https://www.lem.com/sites/default/files/products_datasheets/lv_25-p.pdf

Dimensions not provided were taken from manufacturer's 3D models. Some dimensions were not specified and may fall outside the nominal range but were kept on the assumption that these dimensions were originally taken from measuring an actual component.

I realize there is a lot here, so please feel free to check things out in detail if you want or just quickly scanning through it to get the fixes merged faster.

Images (old on left and new on right; only a smattering of the HX* footprints shown):
image
image
image
image
image

------------


evanshultz:
Apparently the Notepad++ macro feature doesn't work worth a damn. Or I'm using it wrong. But surely that's not the issue...

poeschlr:
The LEM-HO-NP looks a bit strange. Differently numbered pads overlap each other. If they are intended to be always connected it might be better to give them the same name. If the part can be used in cases where they are not connected in that way, then the pads should be designed such that there is a clearance between them.

Also the LEM-HO-NP and SP33 variant where designed for the same basic part (the sp33 is just the 3v3 version whereas the one without suffix is the 5V version. I am not sure if we need both footprints as they might be pin compatible. I am too tired to check this now but i will look into this sometime next week. If not remind me again as i might forget about it.)

The one you added is the one for larger current ranges. Meaning the original footprint could be named LEM-HO_8-NP and the new one would be for example LEM-HO_40-NP (both are compatible with different current ranges. [8,15,25] and [40,60,120,150] The name given is for the smallest current range of the respective pin compatible parts.)

---

Note i only looked at the screenshot for now. I did not fully check this PR.

Misca1234:
LEM_HO-NP:
People could argue about the shape of F.Fab or even a millimeter here and there,
but data sheet and drills are ok so I would say good to go

LEM_HO-NP_SP33
Data sheet, drills and lines holes are ok so I would say good to go

LEM_HX02-P
https://www.lem.com/sites/default/files/products_datasheets/hx%202_6-p_e%20v5.pdf
The "height" is in foot print 15.51 but in datasheet is 15.4
I do not think it matter but the drill holes are larger on secondary side than primary

LEM_HX04-P
https://www.lem.com/sites/default/files/products_datasheets/hx%202_6-p_e%20v5.pdf
The "height" is in foot print 15.51 but in datasheet is 15.4

LEM_HX05-P-SP2
https://www.lem.com/sites/default/files/products_datasheets/hx%203_50-p_sp2_e%20v07.pdf
The "height" is in foot print 15.51 but in datasheet is 15.4

LEM_HX05-NP
https://www.lem.com/sites/default/files/products_datasheets/hx%205_15-np_e%20v10.pdf
The "height" is in foot print 15.51 but in datasheet is 15.4

LEM_HX15-NP
https://www.lem.com/sites/default/files/products_datasheets/hx%205_15-np_e%20v10.pdf
The "height" is in foot print 15.51 but in datasheet is 15.4

LEM_HX15-P-SP2
https://www.lem.com/sites/default/files/products_datasheets/hx%203_50-p_sp2_e%20v07.pdf
The "height" is in foot print 15.51 but in datasheet is 15.4

LEM_HX20-P-SP2
https://www.lem.com/sites/default/files/products_datasheets/hx%203_50-p_sp2_e%20v07.pdf
The "height" is in foot print 15.51 but in datasheet is 15.4

LEM_HX50-P-SP2
https://www.lem.com/sites/default/files/products_datasheets/hx%203_50-p_sp2_e%20v07.pdf
The "height" is in foot print 15.51 but in datasheet is 15.4
You made pin 5 and 6 as big holes, are you sure of this ?
According to the data sheet its indicate that they are pins after all ?
bild



LEM_LA-P
https://www.lem.com/sites/default/files/products_datasheets/la_25-p.pdf
The bottom F.Fab line is placed at 7.8 mm (14.35 - 6.55)
But if the drawing is made in "normal" fashion, it would mean that the lower right image on page 3
would mean that the lower F.Fab line should be placed at 6.55 but that contradicts the 
"Recomended PCB layout" image up to the right (if "Recomended PCB layout" is seen from 
top copper side)
I am inclined to think that the "Recomended PCB layout" is seen fro mthe bottom copper side.

LEM_LV25-P
https://www.lem.com/sites/default/files/products_datasheets/lv_25-p.pdf
All good

LEM_HTFS
Missing data sheet
Should it not be rotated 180 degree? Is not pin 1 supposed to be in the upper right "corner"?
Should not the NPTH holes be made for 1 mm pins ?
The F.Fab line to the right of the pin 1, should it not be at distance 4.5 ?

bild



Beside the 0.11 mm difference in hegith for some of them and the question about LEM_LA-P, all looks good

Shackmeister:
to me it looks like the current FP for LA_25-P is wrong.
found the official step file if you guys wanna use it for checking
[90.13.19.000.0_20090916.zip](https://github.com/KiCad/kicad-footprints/files/2131124/90.13.19.000.0_20090916.zip)

evanshultz:
Thanks for looking!

@poeschlr 
LEM_HO-NP
This is how the datasheet shows it, with unique pin numbers and overlapping pins. I think this is correct; if you look at page 11 of the datasheet you will see the connection matrix and if we want the user to be able to follow that for their particular design we will need to retain unique pin numbers.

*SP33
This is a totally different part, wrong in our existing library, and the impetus for this entire PR.

From the datasheet, it looks like the lower-current one is 5V only and the higher-current one is 3.3V only. So the naming is mutually exclusive and my footprints names are OK. Is there something I missed?

@Misca1234 
HX*
I don't see the width/height dimension mistake. 3.89mm + 11.51mm = 15.4mm. `LEM_HX50-P-SP2` with this dimension:
image

LEM_HX50-P-SP2
The datasheets shows 1.2x6.3mm pin size on page 3 of the datasheet, and I retained the 1.5x8mm slot hole from the existing footprint (I can only assume it works as I do not have an actual part in my hands to check).

If the datasheet is showing three 1.2x1.3mm rectangular pins, then with a 0.2mm oversize we could use 1.97mm (maybe just round to 2mm) holes with 2.55mm annular ring. That would be a potential, unproven option.

A picture I found online would seem to indicate 3 individual holes might work too, but again I don't have evidence the existing hole doesn't work:
image

LEM_LA-P
The drawings should be 90 rotations from each other, and it looks like I'm wrong. The dimension from the upper horizontal fab line to the row of pins should be 6.55mm instead of the 7.8mm I have. After reading https://github.com/KiCad/kicad-packages3D/pull/272 I'm not sure what to think. I'm wrong, right?

Also, should this footprint be named `LEM_LA25-P` since there's only one current option? I tried to make it generic but maybe that's too much?

LEM_HTFS
I see a datasheet in my footprint and listed above. Please check again.

I think you're right about rotation, according to http://kicad-pcb.org/libraries/klc/F4.2/. It's got more than 2 pins and pin 1 isn't in a quadrant. I've rotated it 180 degrees. Good catch!

The NPTHs in the corners are 1.2mm, which matches IPC Level B's recommended 0.2mm oversize for a 1mm hole.

The 4.5mm dimension you mentioned is correct for me (before rotation):
image

evanshultz:
Branch conflicts fixed.

One other question: Should we have a new library called `Sensor_Voltage` since `LEM_LV25-P` is voltage transducer and not for current measurement? Not in this PR, but as I was looking at this it seems like the correct way to go.

Misca1234:
Schackmeister found an official step file of LEM_LA25-P

bild

which shows where the "6.55" should be, the data sheet is is not rotating the objects in a "normal" way

Misca1234:
I can not find that 15.51 any more
LEM_HTFS looks fine
Only thing that remains is the elusive LEM_LA25-P

evanshultz:
FYI, once this gets merged I will be submitting symbols for all of them that were added or changed.

poeschlr:
Regarding Lem HO and your remark about the lower current ones being 3v3 only
One can by the 5V LEM HO 8-NP-0000 from digikey.
https://www.digikey.com/product-detail/en/lem-usa-inc/HO-8-NP-0000/398-1141-ND/4331974

Here the official lem product page:
https://www.lem.com/en/ho-8np0000

and the datasheet:
https://www.lem.com/sites/default/files/products_datasheets/ho-np-0000_series.pdf

From the datasheet link (and from digikey) it would on first glance seem one can use the 4 digit code at the end for distinguishing them. But this code is used for requesting specially configurations (Different reference voltage, programmable, response time and settings for over current detection)

---

Regarding pad overlapping of the high current version:
Well on page 11 they describe how you can use them for lower currents without loosing precision. They suggest to increase the number of primary turns in that case. (Meaning the 4 measurement leads are not used in parallel but in series.)
This usecase is not possible with the current footprint so we might want to add one footprint per such variant. (So the default all parallel, the half and the quarter current version.)

evanshultz:
High-current version with 4x3mm ovals holes and both versions re-named as you suggested. Updated the datasheet of the low-current one.

Since you guys all like adding more footprints during a PR, and y'all made me look at the website again, I also added the -NSM version (https://www.lem.com/sites/default/files/products_datasheets/ho-nsm-0000_series.pdf; only comes in the low-current version; the CI violation is invalid because it doesn't account for pads and fab lines when determining footprint center).
image

Anything else?

Misca1234:
I guess we are all good to go ?

evanshultz:
I forgot to delete the old footprints. They're gone now. I also updated the HTFS footprint since I found there's a more generic name for it.

@Misca1234 
Yes, I believe so, unless the HX50 footprint needs pad changes. That's the only outstanding thing I can see above. I submitted the matching symbols at https://github.com/KiCad/kicad-symbols/pull/700 and the only errors there are missing footprints.

poeschlr:
Looks good to me as well. Thanks for the great effort.

PR #621, opened 2018-05-31
Add Assmann A-2004-2-4-LPS-N-R horizontal RJ45 jack

ianrrees:
Add Assmann A-2004-2-4-LPS-N-R, which is similar to a commonly available generic transformerless horizontal RJ45 jack.

screen shot 2018-06-01 at 7 43 17 am

------------

- [na] If there are matching symbol or 3D model pull requests, provide link(s) as appropriate


poeschlr:
I don't think generic footprints for such connectors are a good idea. When we last looked into it most of these had slight differences. (Slightly different pin sizes, Slightly different pin arrangements, ...)

Such a generic footprints just gives beginners a wrong sense of security. More experienced users could use such a footprint as they know what they need to check. (But these users most likely prefer specialized footprints as they result in more reliable boards.)

So my suggestion would be: Rename the footprint to represent the part after which you modeled it.

---

The alternative is a very complicated naming convention that encodes all the information necessary for a user to decide if this footprint is right for their part. Things like pin pitch, drill size, shield pad positioning, shield drill, Mounting pin position and size, ....

It is just easier to make specialized footprints here.

evanshultz:
First, thanks for contributing to KiCad! Without the community the library never grows and improves.

Second, please always provide a datasheet for any submission. While at first post didn't have one, by picking a specific part we now have http://www.assmann-wsw.com/fileadmin/datasheets/ASS_7323_CO.pdf as a datasheet.

Lastly, please append `_Horizontal` to the footprint name.

ianrrees:
I'm sorry for not adhering to convention - have updated the footprint to match a specific Western manufacturer's datasheet.

evanshultz:
- Add hyphens in the part name, like `A-2004-2-4-LPS-N-R`.
- I'm not sure about this PN anyway. Looking at page 6 of http://www.assmann-wsw.com/fileadmin/catalogue/05_Modular_rev4-0.pdf, I see `A-2004-x-LPS` seems valid, but surely the `-N-R` can be removed (probably RoHS compliance or something). What I can't reconcile is the `-2-4` in the datasheet above with the link I just posted only showing a single character there. Are you able to sort this out?
- Only the upper courtyard is correct, an the others need adjustment.
- Your connector it a bit too long. Should be 18.6mm and yours in 18.75mm. Shrink it at the top.
- Offset the silk lines from the fab lines by 0.11mm.
- Snuggle the silk lines about their line width (0.12mm) from copper. Right now they're far away.
- Add a 3D model even if you didn't make one. When (if?) it's added later it will just work without touching the footprint.
- The pads for you pin 9s are excessively large. Change them from 3mm diameter to 2.6mm please.
- The symbol to be used in below. To use this, your pin 9s will have to be changed to pin number SH. (Yes, a letter for the pin number.)
image

Dimensioned drawing:
image

ianrrees:
With regard to the name - I was under the impression from Travis that we need to have name match filename, and filename isn't to have hyphens.  Is it better to have hyphens in the filename (which also goes against the convention in the RJ library), or to have a mismatched file name from part name?

Makes no difference to me which hyphenated parts we keep - if it's poked in to digikey as I did it, then it finds the right part, but either way my intent is for the footprint to work with the original aliexpress link which isn't an Assmann part and I doubt has a standardised datasheet...

I'll try to implement the other adjustments soon.

evanshultz:
The file name and footprint name can contain hyphens. It may be that no other footprints in the RJ lib happen to use hyphens, but they are legal characters in all OSes that run KiCad and in KiCad footprints. The footprint name and file name do need to match, but that shouldn't be a problem. I do want to remove any parts of the footprint name which are for RoHS compliance or component packaging or other attribute which does not affect the footprint or 3D model. I wrote down the info I could find but there's still something I can't quite reconcile.

Thanks for making the updates!

ianrrees:
Have updated the footprint, except I'm not clear on what you're wanting with the 3D model.  The UI seems to want a specific model - since there isn't one for this part, I feel like we're better off not picking a wrong one.  Please feel free to modify the footprint to fix the 3D model, I'll be able to see what you've changed, and it seems easier than the back-and-forth.

poeschlr:
You just need to fill out the 3d settings to allow for easy adding of a model.
http://kicad-pcb.org/libraries/klc/F9.3/

ianrrees:
@poeschlr - I appreciate the link to the relevant section of the KLC, and have added the path as described there.  Apologies for missing the mention of the model being required.

Two suggestions to someone more familiar with the KiCad culture and state of development:

  * It's not obvious to someone in my position - who wants to add a footprint to the library - that there needs to be a link to the 3D model, because the only mention in the KLC is in the bottom of the 3D model section there.  If I'm not planning on making a 3D model, why would I read that?

  * As I said in my comment above, the UI isn't particularly helpful for doing this, and to expand a little further: this isn't the sort of thing that should need to be done manually.  Both the producer (footprint editor) and consumer (pcbnew) could construct that default path - so one of them should "just" do it.

Ratfink:
F9.3 *is* in the footprint section.  I do think that it would make sense to put the requirement that all footprints have a 3D model link at the top of that rule.  I have suggested this in https://github.com/KiCad/kicad-website/issues/288.

I agree with the second point, but we're not the KiCad developers, so we're not in a position to change that.

evanshultz:
- Fab line on the right side isn't vertical.
- Courtyard should be on a 0.01mm grid so round up or down. The left one is off grid.
- After searching around on Assmann's site I found http://assmann.us/index.php?main_page=index&cPath=589_601_603 which links to http://assmann.us/specs/A-2004-2-x-LPS-N.pdf. That's not exactly the same connector, but it shows the `-4` is the pin finish which we wildcard. And you can also find http://assmann.us/specs/A-2004-2-4-LPS-N-R.pdf (why are there multiple datasheets for the same part!) which seems to indicate the `-N-R` is required for this exact footprint. I've asked Assmann is they can clarify anything to stay tuned.

All else looks good!

ianrrees:
Sorry about the delay in doing this!

PR #622, opened 2018-05-31
Fix to match datasheet

evanshultz:
When reviewing https://github.com/KiCad/kicad-symbols/pull/637 I discovered that the footprint didn't match the datasheet. It was quite off in some places, in fact. Many minor updates so the footprint actually matches the datasheet and best meets the current version of KLC. Note the two paste pads added give ~72% paste coverage.

I left the naming alone but brought it up at https://github.com/KiCad/kicad-footprints/issues/436 since this is a footprint with missing pins and we haven't really resolved how to name them. That Issue has some ideas about this particular one but I've left it alone for now.

Datasheet: http://www.analog.com/media/en/technical-documentation/data-sheets/3630fd.pdf (I don't see a drawing of the footprint with missing pins online)

image

------------


evanshultz:
image

You will notice I centered the inside edges of the rows of pads in the 3.20-3.45mm range which means the outer edges are 5.103mm apart. 3um seems rather in the noise and it's obviously an non-manufacturable tolerance. But if you want me to meet the singular dimension of 5.1mm and then the inner edge still falls in the range of dimensions I can update that.

poeschlr:
looks good. thanks

PR #379, opened 2018-02-16
Wuerth 7499151120 dual RJ45 jack

evanshultz:
As mentioned at https://github.com/KiCad/kicad-symbols/issues/301.

In addition to adding this footprint, the other two RJ45 jacks have no manufacturer information I can see. Should I remove them? Can anyone identify the vendor and MPN?

Notes:
- I chose pin names that follow the datasheet, using 1-14 and appending D or U for DOWN or UP.
- The housing pins are named SH, in accordance with out latest plan (I believe).
- Pad sizes chosen according to IPC Level B
- The F.Fab footprint name extends past the footprint outline, but will shrink once it's packaged to match the shorter symbol name (which doesn't have "_Horizontal" in it).

Datasheet: http://katalog.we-online.de/pbs/datasheet/7499151120.pdf

Image:
image

------------

Thanks for creating a pull request to contribute to the KiCad libraries! To speed up integration of your PR, please check the following items:

- [ ] Provide a URL to a datasheet for the footprint(s) you are contributing
- [ ] An example screenshot image is very helpful 
- [ ] If there are matching symbol or 3D model PRs, provide link(s) as appropriate
- [ ] Check the output of the Travis automated check scripts - fix any errors as required

PR #623, opened 2018-05-31
Add (reanimate) FCI 54602-x08

evanshultz:
@thomasonw mentioned at https://github.com/KiCad/kicad-footprints/pull/379#issuecomment-393545589 that some generic footprint was for a specific component. Indeed, it appears `RJ45_8` was rather similar to the mentioned part (which could use a wildcard for plating type), but the footprint was a bit old and had some non-orthogonal lines. I fixed it up and have submitted it here.

Datasheet: https://cdn.amphenol-icc.com/media/wysiwyg/files/drawing/c-bmj-0102.pdf

image

------------


evanshultz:
image

PR #624, opened 2018-06-01
Footprint for Molex 1041681620

jneiva08:
Added footprint for Molex 1041681620

Datasheet: https://www.molex.com/pdm_docs/sd/1041681620_sd.pdf

Footprint:

footprint

This PR for the symbol is

------------


Misca1234:
I suggest you do the following changes

Move the tecxt microSD_microSIM_Molex_1041681620, below, outside the F.CrtYrd line
Centre the R% to (0,0)

I guess it is a metal casing a and in the data sheet the text "NO PATTERN AREA all socket space"
indicates that you should make this area to a keep out zone, 
if you not already know how to, check foot print ESP-WROOM-02 and you can see how a keep out zone is marked.

Is there a reason for renaming G1-G7 (GND) to HS (if renamed, why not just GND) ?

jneiva08:
Sorry for taking long time to do.
The footprint looks like this:

footprint

evanshultz:
Man, I have certainly seen better-dimensioned drawings before...

Review notes:
- Location of S3 is wrong (S3 through T7 are OK).
- The X placement of the THT SH pins is wrong (too far outside).
- The back of the connector (G4/G5 side) could use a bit more detail on the F.Fab layer.
- I believe the card sticking out (bottom in the pic above) is usually put on the fab layer.
- The keepout area is the size of the body, not inset.
- Place silk above and below the SW pin.
- Rear courtyard clearance is a bit too much.
- Front courtyard doesn't need to account for the card sticking out (no card during assembly and it can/should be removed for rework).
- From the "Connectors for Specific Functions" section of http://kicad-pcb.org/libraries/klc/F3.6/, I think the correct footprint name is `SD-SIM_microSD-microSIM_Molex_1041681620` (notice how the underscore separates section, but here is separating within the `[Function]` section). (The last example there doesn't start with `SD_` either.)
- You can use a custom pad shape to make exactly the pad called out in the datasheet. Try this (it's the lower right one):
image

Dimensioned drawing:
image

evanshultz:
I ought to include the footprint since the above doesn't show all about the custom pad shape, right? It is attached here.
[microSD_microSIM_Molex_1041681620.kicad_mod.txt](https://github.com/KiCad/kicad-footprints/files/2280409/microSD_microSIM_Molex_1041681620.kicad_mod.txt)

evanshultz:
Thanks!

I mean that above and below the SW pin there is no silk and there should be:
image

I'll check the rest shortly.

jneiva08:
the footprint looks like this now:
sd

jneiva08:
@evanshultz is missing something to be able to merge?

evanshultz:
Yes, the pin naming. I've been pushing on this at https://github.com/KiCad/kicad-symbols/pull/592 but without an answer. It seems Rene is rather busy lately. Deciding how to name the pins on these types of connectors will unlock several PRs, including this one.

jneiva08:
@evanshultz have changed pin names in conform the [symbol PR](https://github.com/KiCad/kicad-symbols/pull/644)

gmsotavio:
Most Molex symbols and footprints uses the name convention: Molex_XXXXXX-XXXX rather than Molex_XXXXXXXXXX. My recommendation is to change the name of your footprint from SD-SIM_microSD-microSIM_Molex_1041681620 to SD-SIM_microSD-microSIM_Molex_104168-1620.

jneiva08:
Now the Molex part-number is identical tConnector_Molex

PR #625, opened 2018-06-02
Use custom pad shape for SMA_Jack_SMD_Vertical

gauravjuvekar:


Ratfink:
[KLC]: http://kicad-pcb.org/libraries/klc/ "KiCad Library Convention"
[G1.1]: http://kicad-pcb.org/libraries/klc/G1.1/ "Only standard characters are used for naming libraries and components"
[G1.2]: http://kicad-pcb.org/libraries/klc/G1.2/ "Each library is limited to 250 items"
[G1.3]: http://kicad-pcb.org/libraries/klc/G1.3/ "Libraries are organized by functionality"
[G1.4]: http://kicad-pcb.org/libraries/klc/G1.4/ "English language should be used throughout libraries"
[G1.5]: http://kicad-pcb.org/libraries/klc/G1.5/ "Plural naming is to be avoided"
[G1.6]: http://kicad-pcb.org/libraries/klc/G1.6/ "Capitalization conventions"
[G1.7]: http://kicad-pcb.org/libraries/klc/G1.7/ "Library files use Unix style line endings"
[G1.8]: http://kicad-pcb.org/libraries/klc/G1.8/ "Components can only contain features supported in latest stable release"
[G1.9]: http://kicad-pcb.org/libraries/klc/G1.9/ "Dimensional units"
[G2.1]: http://kicad-pcb.org/libraries/klc/G2.1/ "Defining generic and atomic parts"
[S1.1]: http://kicad-pcb.org/libraries/klc/S1.1/ "Symbol libraries should be categorized by function"
[S2.1]: http://kicad-pcb.org/libraries/klc/S2.1/ "General symbol naming guidelines"
[S2.2]: http://kicad-pcb.org/libraries/klc/S2.2/ "Non-functional variations in part number should be replaced with wildcard"
[S2.3]: http://kicad-pcb.org/libraries/klc/S2.3/ "Where atomic parts are available in multiple footprint options, a separate symbol must be drawn for each footprint"
[S3.1]: http://kicad-pcb.org/libraries/klc/S3.1/ "Origin is centered on the middle of the symbol"
[S3.2]: http://kicad-pcb.org/libraries/klc/S3.2/ "Text fields should use a common text size of 50mils"
[S3.3]: http://kicad-pcb.org/libraries/klc/S3.3/ "Symbol outline and fill requirements"
[S3.4]: http://kicad-pcb.org/libraries/klc/S3.4/ "Symbols with complex functionality may incorporate simple functional diagrams"
[S3.5]: http://kicad-pcb.org/libraries/klc/S3.5/ "Pin connection points must be placed outside of the symbol"
[S3.6]: http://kicad-pcb.org/libraries/klc/S3.6/ "Pin name position offset"
[S3.7]: http://kicad-pcb.org/libraries/klc/S3.7/ "Pin numbering for exposed pads"
[S3.8]: http://kicad-pcb.org/libraries/klc/S3.8/ "Multi unit symbols"
[S4.1]: http://kicad-pcb.org/libraries/klc/S4.1/ "General pin requirements"
[S4.2]: http://kicad-pcb.org/libraries/klc/S4.2/ "Pins should be grouped by function"
[S4.3]: http://kicad-pcb.org/libraries/klc/S4.3/ "Rules for pin stacking"
[S4.4]: http://kicad-pcb.org/libraries/klc/S4.4/ "Pin electrical type"
[S4.5]: http://kicad-pcb.org/libraries/klc/S4.5/ "Pins not connected on the footprint may be omitted from the symbol"
[S4.6]: http://kicad-pcb.org/libraries/klc/S4.6/ "Hidden pins"
[S4.7]: http://kicad-pcb.org/libraries/klc/S4.7/ "Active low pins should be designated using a bar above the symbol name"
[S5.1]: http://kicad-pcb.org/libraries/klc/S5.1/ "Symbols with a default footprint link to a valid footprint file"
[S5.2]: http://kicad-pcb.org/libraries/klc/S5.2/ "Footprint filters should match all appropriate footprints"
[S6.1]: http://kicad-pcb.org/libraries/klc/S6.1/ "Component Reference Designator (RefDes)"
[S6.2]: http://kicad-pcb.org/libraries/klc/S6.2/ "Component fields must be filled appropriately"
[S6.3]: http://kicad-pcb.org/libraries/klc/S6.3/ "Component metadata is added to symbol (and all aliases)"
[S7.1]: http://kicad-pcb.org/libraries/klc/S7.1/ "Power flag symbols"
[S7.2]: http://kicad-pcb.org/libraries/klc/S7.2/ "Graphical symbols"
[F1.1]: http://kicad-pcb.org/libraries/klc/F1.1/ "Footprint libraries are categorized by function"
[F1.2]: http://kicad-pcb.org/libraries/klc/F1.2/ "Connector footprint libraries"
[F2.1]: http://kicad-pcb.org/libraries/klc/F2.1/ "General footprint naming conventions"
[F2.2]: http://kicad-pcb.org/libraries/klc/F2.2/ "Footprint naming field prefixes"
[F2.3]: http://kicad-pcb.org/libraries/klc/F2.3/ "Manufacturer specific version of generic footprints"
[F2.4]: http://kicad-pcb.org/libraries/klc/F2.4/ "Footprint naming for non-standard pin numbering"
[F2.5]: http://kicad-pcb.org/libraries/klc/F2.5/ "Footprint naming conventions for specific components"
[F3.1]: http://kicad-pcb.org/libraries/klc/F3.1/ "SMD chip package naming conventions"
[F3.2]: http://kicad-pcb.org/libraries/klc/F3.2/ "Resistor naming conventions"
[F3.3]: http://kicad-pcb.org/libraries/klc/F3.3/ "Capacitor naming conventions"
[F3.4]: http://kicad-pcb.org/libraries/klc/F3.4/ "SMD IC package naming conventions"
[F3.5]: http://kicad-pcb.org/libraries/klc/F3.5/ "THT IC package naming conventions"
[F3.6]: http://kicad-pcb.org/libraries/klc/F3.6/ "Connector naming conventions"
[F3.7]: http://kicad-pcb.org/libraries/klc/F3.7/ "Fuse naming conventions"
[F4.1]: http://kicad-pcb.org/libraries/klc/F4.1/ "Datasheet recommendations take priority"
[F4.2]: http://kicad-pcb.org/libraries/klc/F4.2/ "Pin 1 should be located at the top left"
[F4.3]: http://kicad-pcb.org/libraries/klc/F4.3/ "Connected copper elements have the same pad number"
[F4.4]: http://kicad-pcb.org/libraries/klc/F4.4/ "Thermal pads"
[F4.5]: http://kicad-pcb.org/libraries/klc/F4.5/ "Specifying footprint keepout areas"
[F4.6]: http://kicad-pcb.org/libraries/klc/F4.6/ "Local clearance and settings should be set to zero"
[F5.1]: http://kicad-pcb.org/libraries/klc/F5.1/ "Silkscreen layer requirements"
[F5.2]: http://kicad-pcb.org/libraries/klc/F5.2/ "Fabrication layer requirements"
[F5.3]: http://kicad-pcb.org/libraries/klc/F5.3/ "Courtyard layer requirements"
[F6.1]: http://kicad-pcb.org/libraries/klc/F6.1/ "Footprint placement type must be set to surface mount"
[F6.2]: http://kicad-pcb.org/libraries/klc/F6.2/ "Footprint anchor should be placed in the middle of the component body"
[F6.3]: http://kicad-pcb.org/libraries/klc/F6.3/ "Pad requirements for SMD footprints"
[F7.1]: http://kicad-pcb.org/libraries/klc/F7.1/ "Footprint placement type must be set to Through Hole"
[F7.2]: http://kicad-pcb.org/libraries/klc/F7.2/ "Footprint anchor should placed at the location of Pin-1"
[F7.3]: http://kicad-pcb.org/libraries/klc/F7.3/ "Pin 1 should be rectangular, and other pads circular or oval"
[F7.4]: http://kicad-pcb.org/libraries/klc/F7.4/ "Pad requirements for THT footprints"
[F7.5]: http://kicad-pcb.org/libraries/klc/F7.5/ "Minimum annular ring width"
[F7.6]: http://kicad-pcb.org/libraries/klc/F7.6/ "Minimum hole diameter"
[F8.1]: http://kicad-pcb.org/libraries/klc/F8.1/ "Virtual components"
[F9.1]: http://kicad-pcb.org/libraries/klc/F9.1/ "Footprint meta-data is filled in as appropriate"
[F9.2]: http://kicad-pcb.org/libraries/klc/F9.2/ "Footprint properties are as default values unless otherwise required in datasheet"
[F9.3]: http://kicad-pcb.org/libraries/klc/F9.3/ "Footprint 3D model requirements"
[M1.1]: http://kicad-pcb.org/libraries/klc/M1.1/ "Models should be created by contributor"
[M1.2]: http://kicad-pcb.org/libraries/klc/M1.2/ "Manufacturer models should not be duplicated"
[M1.3]: http://kicad-pcb.org/libraries/klc/M1.3/ "Source files for 3D models should be supplied"
[M2.1]: http://kicad-pcb.org/libraries/klc/M2.1/ "Permitted 3D model file types"
[M2.2]: http://kicad-pcb.org/libraries/klc/M2.2/ "Model alignment and scaling"

Thank you for volunteering to fix up this part in the KiCad library.  I see this is your first contribution as well, so allow me to welcome you to the community!  As you may already know, we follow a set of guidelines called the KiCad Library Convention, which can be found [here][KLC].  This document is currently a bit outdated in some sections and in need of work, so a few things I say might disagree with it.

In the future, please provide a screenshot and datasheet link when submitting substantive changes.  This makes it easier for us librarians to review contributions.

This footprint is named as a generic footprint, but it seems that different manufacturers suggest different footprints for their surface mount SMA connectors.  Also, the footprint has no datasheet link provided, so it's hard to know if it's correct or not.  Here are a few datasheets I've found:

* https://www.amphenolrf.com/media/downloads/4007/132134-10.PDF (seems to be a fairly close match to this part)
* https://belfuse.com/resources/Johnson/drawings/dr-142-0711-202.pdf (another manufacturer suggesting the same footprint)
* https://www.mouser.com/datasheet/2/238/CONREVSMA001-SMD_D_131108-1-1175053.pdf (similar shape but different dimensions)
* https://www.molex.com/pdm_docs/sd/732511352_sd.pdf (similar part, but wildly different footprint)

So, I would suggest renaming this footprint to have a specific manufacturer part number.  Following [F3.6], I suggest the name `SMA_Amphenol_132134-10_Vertical` (unless you're using a different part, in which case, feel free to use its manufacturer and part number instead!).  I would also like to see the datasheet link for this part added to the end of the "Document link" text in the Footprint Properties dialog.  And once that's done, please tweak dimensions as needed to make the footprint really match that datasheet.  @poeschlr @evanshultz @jkriege2 Do you agree that this would be the best course of action for this footprint?

Other changes I'd like to see while you're fixing it up:

* Please move the reference designator on the F.Fab layer to the center of the connector, and reduce its size so it doesn't intersect any graphical elements ([F5.2]).
* The value text and silkscreen reference designator could both be moved a bit closer to the connector body.  It wouldn't be a problem for them to overlap the courtyard outline.

gauravjuvekar:
Okay, I'll try to change to the new KLC. However, I just realized that thermal relief is not supported yet for custom pad shapes. What's the librarians take on this? Should we actually include footprints with custom pads yet?

gauravjuvekar:
@poeschlr @Ratfink 
Okay. Regarding naming convention, http://kicad-pcb.org/libraries/klc/F3.6/ , where how should gender be specified? Is it the subseries? Also, are there any conventions (Male/Female vs Jack/Plug/Receptacle)?

evanshultz:
Yep. Should be a specific footprint. I have not seen SMA on-board connectors to have gender, so it should be good as suggested above.

Thermals are a concern in general. Wave solder will only touch the bottom of the PCB, meaning top and inner layers with a solid connection will likely suffer in regards to barrel fill (even with pre-heating). Not that it still won't (or can't) be acceptable, but it will be less fill than with thermal spokes on inner and top layers. But since this is an SMT connector that will use reflow instead of wave solder (as mentioned above), it's probably OK as long as a proper heating profile is used.

gauravjuvekar:
@evanshultz 
THT SMA on board connectors do have a gender
https://www.amphenolrf.com/132133.html
https://www.amphenolrf.com/132134.html

Ratfink:
None of our other manufacturer-specific coaxial connector footprints have a gender specified in the footprint name, so I don't think this one should as well.  They also don't have whether the connector is surface mount or through-hole specified explicitly (ignore U.FL_Hirose_U.FL-R-SMT-1_Vertical, the "SMT" in the name is actually in the part number).  The part number has both gender and mounting covered anyway.  So please use names like the one I suggested above, and if we decide to change the naming format later we can do so for all the coaxial connector footprints at once.

I agree not to worry about the thermal relief.  The old footprint worked pretty horribly with thermal relief anyway:
screenshot from 2018-06-04 13-47-58

Ratfink:
Please, in the future, don't keep adding more footprints to the same pull request.  Doing so doesn't get the new parts merged any faster, and only delays merging the original part until they're *all* perfect.

I'll try to give all these a close look later today.

Ratfink:
#### General
* Please remove the 2.04 mm circle from all these footprints *except* 132134-10, where it serves the purpose of easily checking the size of the circular cutout from the pads 2.  For that footprint, it would be better placed on the Dwgs.User layer.
* The F.Fab outline should be simple, so please remove all the pin outlines, as they make it more complicated and we don't normally have those.
* Please center the F.Fab reference designators once you've done that.  Many could probably be made larger as well.
* On most of these, the silkscreen lines could extend closer to pads (KLC currently says to keep a 0.20 mm clearance, but we plan to change that to the same as the line width, 0.12 mm).

#### SMA_Amphenol_901-144_Vertical
* Body outline should be 6.35×6.35 mm, but is currently 6.34×6.34 mm

@poeschlr @evanshultz @jkriege2 Should the pad diameters be larger on the plated-through holes?  They're not too small according to any of our rules, but a 0.25 mm annular ring around a 1.7 mm hole seems small to me.

evanshultz:
Yes, the pads are too small. I brought this up at https://github.com/KiCad/kicad-footprints/issues/595 and pointed out that when a lead is to be soldered into a plated drill hole, IPC Level B (typical) tells us the pad diameter should be 0.7mm bigger than the drill diameter.

Going even bigger isn't an issue generally, but it will suck away a bit more heat and since this is a metal connector it will already absorb a lot of the wave's heat so going bigger may not be advisable. On the contrary, during rework a bigger pad allows the PCB to soak up more heat before delamination. But since even connector must be soldered and hopefully only a tiny percentage should be removed, I say 0.7mm is good (with 1mm still being OK).

gauravjuvekar:
@evanshultz Okay, lets hold off this PR till its completely decided in #595

evanshultz:
I think we can go with Level B (typo on my part above) which means 0.55mm diameter annular ring over the drill diameter. So if you want to update this we may be good.

gauravjuvekar:
@evanshultz 
Pushed the changes, though they aren't significant (0.05 increase in pad diameter if I understood Level B correctly)

I've also set the thermal connection to solid for SMD footprint.
Please merge if its correct now (screenshots not updated since there is no significant visible change.)

Ratfink:
Looks good!  Thanks for contributing all these footprints to the KiCad library.

PR #626, opened 2018-06-02
Qfn 48 1 ep 6x6mm p0.4mm ep4.30x4.30mm

Misca1234:
This is two footprints used by ESP32
https://www.espressif.com/sites/default/files/documentation/esp32_datasheet_en.pdf

Symbols and 3D models comes later

QFN-48-1EP_6x6mm_P0.4mm_EP4.30x4.30mm.kicad_mod
Very similar to QFN-48-1EP_6x6mm_P0.4mm_EP4.66x4.66mm.kicad_mod

QFN-48-1EP_5x5mm_P0.35mm_EP3.70x3.70mm.kicad_mod

bild

bild

bild

bild

bild

bild

bild

bild

bild

bild

------------


poeschlr:
As i noticed that my exposed pad calculation is probably wrong i will need to do some work with them. Also my script has not yet been merged (Is awaiting review for a months now.)

But yes i think it would be best if we use scripts as much as possible. (To be fair i would not suggest to force anybody to use my scripts.)

---

A note about this contribution: Where does the 4.66x4.66 EP size come from? The drawing on page 38 gives 4.3 +/- 0.05 for the slug. Even with my old idea of using tolerances i would at most expect 4.4x4.4.

As i learned today it seems more likely that the better results will come with using the nominal slug size. Meaning i would more likely use 4.3x4.3 (unless i looked at the wrong drawing. It is 2 in the morning for me after all.)

Ratfink:
There's a pretty significant error in the 5x5mm footprint: two sides are supposed to have 14 pins, and the other two are supposed to have 10.  Also the pads are the exact size of the terminals on the chip, leaving no heel, toe, or side fillets whatsoever (unless they're off-center, which I didn't check.  Either way, the pad size is surely too small).

Misca1234:
I gave the foot print scripts several "tries", but I did not understand how to use them so I skipped them and hand made them.
I have earlier used for for connectors but the qfn was to much threshold.

> A note about this contribution: Where does the 4.66x4.66 EP size come from?

It is form my initial comment where I point out that the new QFN is very similar to the existing foot print QFN-48-1EP_6x6mm_P0.4mm_EP4.66x4.66mm.kicad_mod 
Indicating that this QFN in this thread might be unneeded.

> There's a pretty significant error in the 5x5mm footprint: two sides are supposed to have 14 pins

Yes, a copy past error, I fix it

Misca1234:
Something like this

bild

bild

poeschlr:
As the 5x5 was an interesting testcase i added these two to the script if you want to use them. (Only the silk generation needed slight modifications)
The changes to the generator are contained in this commit: https://github.com/pointhi/kicad-footprint-generator/pull/125/commits/6722aacba47ebac8b524622b9ada096a5537174d

I will open a PR for the footprints. You can then decide if you want the manually created one or the scripted footprints. (@Misca1234 i hope you are not mad at me for doing this so late in the process of this PR.)

Misca1234:
No, absolute not, much better with script, footprints just takes times and is no fun, especially with therminal vias, can not ever get them right.

I am moving over to script for footprint at the moment, done a keep out class which will pushed son along with a "standard box" script.

I close this push

PR #633, opened 2018-06-06
Te x 776163 x

thomasonw:
TE Connects, 35 position Ampseal assembly, right angle.

This is a companion footprint to pull #610 

The URL for the footprint is:
http://www.te.com/usa-en/product-1-776163-1.html?te_bu=ict&te_type=disp&te_campaign=oct_glo_oct_feed&elqCampaignId=13728#guest-click

With datasheet (PDF file) accessible here:
www.te.com/commerce/DocumentDelivery/DDEController?Action=showdoc&DocId=Customer+Drawing%7F776163%7FJ8%7Fpdf%7FEnglish%7FENG_CD_776163_J8.pdf%7F1-776163-1

As with #610, I am not sure if the 'name' is correct, the 1- prefix is used to indicate Gold vs. Tin plating, while the -x postfix is used to indicate the color/keying of the connector.

image

And also as with #610, I am unsure if I should add the existing 3D .STEP file as-is, or if more work is needed.

I am hopeful to get these reviewed with corrections before adding the other connectors in the family.


Misca1234:
Hi

The F.Silk line should not be placed on top of the F.fab line.
F.Silk should in principle be just outside the contours of the device, 
If you place the F.Silk with a distance off 0.11/0.12 mm to the outer most F.Fab, it should be enough.
This also means that you don't have to place F.Silk "along" the inner lines of the F.Fab

PR #635, opened 2018-06-06
Add Infineon_PG-HDSOP-10-1

evanshultz:
https://www.infineon.com/cms/en/product/packages/PG-HDSOP/PG-HDSOP-10-1/

Seems similar to the PG-HSOF footprints which are in this lib. Plus, it's conceptually a pair of DPAK devices end-to-end and DPAK is in the lib.

image

------------


evanshultz:
Dimensioned drawing (I fixed the top and bottom courtyard after this):
image

Also fixed 3D model directory and added a few more keywords.

poeschlr:
Why is the courtyard clearance 0.5mm in y direction and 0.25 in x?
Everything else seems good to merge.

evanshultz:
A brain fart was the root cause, but see above the drawing where I caught this error and already addressed it. This should be ready to merge. Thanks for checking it out!

poeschlr:
Ok i did not notice that remark.

PR #636, opened 2018-06-07
Add wuerth electronics barrel jack connector footprint

poeschlr:
Datasheet: http://katalog.we-online.de/em/datasheet/6941xx301002.pdf

screenshot from 2018-06-07 13-02-32

screenshot from 2018-06-07 13-14-39

I checked the footprint also against the 3d model provided by the manufacturer: katalog.we-online.de/em/download/STEP-694106301002-rev1.stp
screenshot from 2018-06-07 13-56-05


poeschlr:
I noticed that pin 1 is slightly larger then the other two. (2.8 vs 2.4mm in the long direction)
To get the same clearance between the pin and the "drill" for all pins i increased the size of pad 1.

Updated screenshots:
screenshot from 2018-06-07 14-00-42

screenshot from 2018-06-07 14-00-34

Misca1234:
All looks good to me, any objections to merge ?

poeschlr:
@Misca1234 seems like no objections have been made.

Ratfink:
Yep, this looks great!  Since @Misca1234 seems too shy, I'll merge it myself to be sure it's in for the stable release.

Issue #545, opened 2018-05-03
Murata footprints incorrectly named "muRata"


Ratfink:
We have a variety of footprints for parts made by [Murata](https://www.murata.com/).  Some include the company's name in the footprint name.  Of these, we have some that capitalize it as `muRata`, and others that capitalize it as `Murata`.  We should reconcile these before the 5.0 freeze begins.

In the company's logo, it's capitalized `muRata`.  However, in ordinary text, they seem to be quite consistent with `Murata`, so I suggest we use the latter.

poeschlr:
Acourding to the "about murata" text on the official website the correct writing is "Murata" (muRata is only the logo)

https://www.murata.com/en-us/about?intcid5=com_xxx_xxx_cmn_nv_xxx

Ratfink:
The fix for this will involve symbol, footprint, and 3D model changes.  I know for sure it would conflict with https://github.com/KiCad/kicad-packages3D/pull/294, so we should wait for that to be merged before working on this.

Misca1234:
The person who undertakes this task should in parallel fix the scripts in 
https://github.com/easyw/kicad-3d-models-in-freecad/pull/161
otherwise that person will get some unneeded workload to guess afterwards which footprint was associated with which 3D model.

Not to mentioned if murata footprints is moved from Converter_DCDC.lib to Regulator_Switching.lib as it should with symbol CRE1S0505S3C

Shackmeister:
The 3D model fixes has now been fixed

poeschlr:
Has this been done? If not we might want to keep this in mind for new contributions. Existing footprints should really only be renamed if there is a danger of confusion (While we are in the stable release cycle)

JacobEFO:
I suppose this issue still needs to be resolved.

I'll be making a PR in a jiffy (both here and on 3d-packages).

antoniovazquezblanco:
Thank you @JacobEFO !

antoniovazquezblanco:
> Existing footprints should really only be renamed if there is a danger of confusion (While we are in the stable release cycle)

AFAIK there should be no problems if a footprint gets renamed during a stable release although this should not be that necessary.

antoniovazquezblanco:
@JacobEFO Is there any symbol that will need footprint or fpfilter updating?

Thanks!

JacobEFO:
@antoniovazquezblanco  Can you elaborate this? I don't understand what you mean with "footprint or fpfilter updating".

[...]

PR #637, opened 2018-06-07
Add Murata NCS1 DC/DC converter module

johnbeard:
This footprint fits:

* NCS1S1203SC - 4.5-18V in, 3.3V out
* NCS1S1205SC - 4.5-18V in, 5V out
* NCS1S1212SC - 4.5-18V in, 12V out
* NCS1S2403SC - 9-36V in, 3.3V out
* NCS1S2405SC - 9-36V in, 5V out
* NCS1S2412SC - 9-36V in, 12V out

Datasheet: https://power.murata.com/data/power/ncl/kdc_ncs1.pdf

Footprint screenshot

Notes: 

* The Fab layer is the maximum material condition of the body (i.e. the rectangle outside of which the body cannot be according the data sheet). 
* Silk is slightly outside that so it would be visible even at the limits of the body size and positions allowed. 
* Courtyard is +0.25mm on the MMC.

Symbol PR: https://github.com/KiCad/kicad-symbols/pull/656

------------


poeschlr:
There are some problems with this footprints dimensions. Example: pin pitch in the footprint is 2.45 instead of 2.54mm. More problems are pointed out in this annotated screenshot of the dimensioned drawing derived from your footprint. ([tutorial for creating such a drawing](https://forum.kicad.info/t/how-to-check-footprint-correctness/9279))

screenshot from 2018-06-16 01-47-12

johnbeard:
The pin pitch is clearly wrong - I must have typo'd 2.45! Thanks for the catch.

For the other dimensions, is it standard to just average the upper and lower bounds in this case to get "nominal"? And then use this on the F_Fab?

If you take the nominal size of the case and use that on the F_Fab layer, then outset by a little to get the F_SilkS outline, the variation in the case position (up to 0.7mm long-wise and 0.65mm cross-wise) could result in invisible silk once assembled.

Ratfink:
The name should say `Murata`, not `muRata`: see https://github.com/KiCad/kicad-footprints/issues/545.

johnbeard:
@ratfink OK, I'll change that too. 

@poeschlr Can you advise on where the silkscreen box and courtyard should go? Outset from nominal, or where it currently is, outset from the maximum variation for the case position?

johnbeard:
Also for IPC B pad sizes: should that now be 1.075 + 0.1 + 0.5mm = 1.675? Or use the minimum hole from the datasheet -> 1.6?

poeschlr:
If you use IPC you need to start at the lead sizes not at the suggested hole.
So it would be maximum lead diameter + 0.2 = 0.65 + 0.2 = 0.85 (rounded to nearest 0.05)
screenshot from 2018-06-21 13-00-25

from that i would say the suggested hole diameter is quite large. The question would now be why. (Does muarta know something they do not tell us?)

johnbeard:
I imagine it's to accommodate the pin position variance?

From the datasheet: "All pins on a 2.54 (0.100) pitch and within 0.25 (0.010) of true position."

poeschlr:
Regarding your previous question: we draw the fab outline after the nominal measurements for all our parts. (And therefore also the silk and courtyard.)

In your case using everything as max tolerance might still result in parts being outside that area. (if the side to pin is at minimum for example.)

---

regarding drill: well with the tolerance for the pin position we would arrive at 1mm drill size. (or maybe a bit less if we use root mean square error addition.)

Misca1234:
This push has been overtaken
https://github.com/KiCad/kicad-footprints/pull/1044

PR #639, opened 2018-06-07
Changed padsize as the old pad size did not work whith our

hamtam1:
We had a lot of tombstones with the current pad size. The main problem is the gap between the pad which was more than 0.8mm.

https://www.vishay.com/docs/28745/soldpads.pdf


evanshultz:
Not sure why this was closed, but I was already writing this...

At https://github.com/KiCad/kicad-footprints/pull/469 all SMD chip padstacks were updated according to IPC-7351 section 3.1 (this is the original IPC-7351, not IPC-7351B). Chip dimensions were taken from IPC-SM-782 section 8.1. I did the math just now by hand on a 0603 and I come up with the same dimensions as the current footprint except I calculate pad X locations should be +/-0.805mm to get G=0.79mm and Z=2.43mm (they are located at +/-0.8mm now).

So all that being said, I believe the existing 0603 footprint is correct. If you have issues with tombstoning, I can suggest the following:
- Review your resistors to check if they conform to the dimensions in IPC-SM-782 section 8.1
- Submit footprints that are Vishay-specific or specify pad and lead spacing parameters according to KLC
- Adjust your manufacturing process to avoid tombstoning using these IPC footprints

All information for our generator script can be found at https://github.com/pointhi/kicad-footprint-generator/tree/master/scripts/SMD_chip_package_rlc-etc.

Also, the existing 0603 pads are (just barely) less than 0.8mm. They are 0.82mm wide pads placed 1.6mm apart which gives 0.78mm gap between pads.

poeschlr:
The document he linked is older than the ipc standard that was used for generating the footprints.

hamtam1:
My first commit was wrong and I was hoping you did not see it yet. All I can say is that our manufacturer had a lot of tombstones. The current FP is within the specification but the most manufacturers recommend to have a smaller gap e.g.:

https://www.vishay.com/docs/28745/soldpads.pdf
https://katalog.we-online.com/pbs/datasheet/885012106003.pdf

We have good experience with the gap of 0.8mm.

evanshultz:
The current 0603 footprint gap is 0.78mm. Your change goes to exactly 0.8mm. 0.02mm is inconsequential and in the noise for PCB fabrication tolerances. I do not understand why you're proposing such a tiny and insignificant change.

hamtam1:
Your consideration is compelling ;-)
My fault. The footprint which we hat a problem with hat a pad size of 0.67mmx1mm. This was the size of the footprint at April.5th. The current footprint is probably OK.
Sorry I will check my requests more precisely.

poeschlr:
I will still update the footprints to use IPC-7351B as it is newer than the standard originally used.

The differences between the IPC-7351 and IPC-7351B are as follows:
For square end < 0603 (nominal density):
 - Toe old = 0.1, new 0.2
 - Heel old = -0.05, new 0.0
 - Side unchanged at 0.0
 - Courtyard unchanged at 0.15
 - rounding base is now 0.02 for all measurements. (Was 0.05)

For square end >= 0603 (nominal density):
 - Toe unchanged at 0.35
 - Heel old = -0.05, new 0.0
 - Side unchanged at 0.0
 - Courtyard unchanged at 0.25
 - rounding base is now 0.05 for all measurements. (Was 0.02 for heel and toe direction and 0.1 for side direction)

PR #640, opened 2018-06-07
changed padsize as we had problems with Tombstoning

hamtam1:
The gap must not be grater than 0.8mm.

Signed-off-by: Martin Thomas <mtgithub@hamtam.de>


evanshultz:
See https://github.com/KiCad/kicad-footprints/pull/639#issuecomment-395523169.

A little more info might work. You initially submitted bigger pads in the PR above, but now your pads are smaller! You changed from a gap much less than 0.8mm to exactly 0.8mm gap while the existing footprint is 0.78mm. This doesn't make sense to me.

PR #642, opened 2018-06-07
Update two terminal smd chip packages to IPC-7351B

poeschlr:
Some time ago @evanshultz or @Ratfink found a copy of IPC-7351B as a series of images on the net. (I can't quite remember who it was.) I now updated the two terminal chip packages to use the fillet definitions of this new standard.

The script update PR is found at: https://github.com/pointhi/kicad-footprint-generator/pull/133

---

The differences between the IPC-7351 and IPC-7351B are as follows:
For square end < 0603 (nominal density):
 - Toe old = 0.1, new 0.2
 - Heel old = -0.05, new 0.0
 - Side unchanged at 0.0
 - Courtyard unchanged at 0.15
 - rounding base is now 0.02 for all measurements. (Was 0.05)

For square end >= 0603 (nominal density):
 - Toe unchanged at 0.35
 - Heel old = -0.05, new 0.0
 - Side unchanged at 0.0
 - Courtyard unchanged at 0.25
 - rounding base is now 0.05 for all measurements. (Was 0.02 for heel and toe direction and 0.1 for side direction)

for tantal and castelated on the round base did change. (The new standard universally uses 0.05 as the round base for them. The old standard had a mix of 0.02 and 0.1)

---

As an example here the difference for the 0603 footprints:
screenshot from 2018-06-07 22-03-06

---

This should also help with the issue reported in https://github.com/KiCad/kicad-footprints/pull/639 as the G dimension is now even smaller than it was. (The contributor complained that it was too large in the past. But there was already one fix between the footprints he used and the ones that are in the repo right now.)


evanshultz:
A couple spot checks seem correct. Are you OK to just merge this and if anything happens the fix the script and re-generate?

poeschlr:
I trust the script to be honest. The only question mark is the [new ipc definition](https://github.com/poeschlr/kicad-footprint-generator/blob/255c1d42da748e9078bf7b8d93489f6ce2049df7/scripts/SMD_chip_package_rlc-etc/ipc7351B_smd_two_terminal_chip.yaml)

evanshultz:
Right. So if a spot check looks good are you fine if I merge without checking every single one?

poeschlr:
Yes of course. That after all is the good thing about scripts. If you check a few footprints you can be sure that all of them will look ok. (Especially as the same script already generated the previous footprints. All problems that have been found with them where related to the ipc definition file.)

PR #645, opened 2018-06-08
Pulse PA200xNL footprints

evanshultz:
https://productfinder.pulseeng.com/products/datasheets/P663.pdf

I don't have all parts in hand so I don't have any way to check dimensions that aren't given; therefore I used a big box on the fab layer instead of a more descriptive outline.

Even though it's for RoHS compliance, I left the `NL` at the end since Pulse uses it everywhere. It seems like they make it a core part of the MPN.

Because these parts are in an obvious transformer repo, and for most parts we don't include the type of component at the front of the footprint name, I didn't here. KLC doesn't have footprint naming guidelines for transformers so maybe using `Transformer_` is legacy and should be cleaned up?

image

I dimensioned one of each physically-unique parts.

PA2001NL:
image

PA2002NL:
image

PA2005NL (courtyard fixed after this):
image

------------


evanshultz:
Note that there is an apparent mistake on page 3 of the datasheet. Pins 2 and 5 are clearly supposed to be centered 2.54mm north of pads 3-4. This means the top of a 1.78mm pad is 0.255mm above the center of the part (as shown above).
image

The dimension given of 0.64mm is not correct unless the pads are not 1.78mm tall. They would need to be 2.55mm tall to get 0.64mm above the center of the part body, and if the middle row of pads is not the same height as the top row, what height is the bottom row? If the middle and bottom row were both 2.55mm they'd overlap so that can't be right as each row is a different winding and there needs to be isolation between them. The top and bottom rows could be 1.78mm with the middle row being 2.55mm but that's not shown at all and all "pins" of the part appear to be the same so having unequal pads doesn't make a lot of sense to me. (1.78mm top row and 2.55mm middle row still gives the stated 1.4mm creepage/clearance but it's very unclear above.)

Anyway, I asked Pulse and hopefully they can clear this up. I do believe the footprint above is correct.

Misca1234:
Have Pulse come back with an answer ?

evanshultz:
Yes, that is what I wrote above.

Pulse confirmed their dimension is wrong and promised an updated one, but nothing yet. They said they believe my analysis is correct and the pads are all equal sizes and so this PR would be correct.

If you want to merge we have high confidence this is correct and I will modify the footprint if something changes. Otherwise, you can wait.

Misca1234:
All looks good but one minor questions, any reason why there is no pin 1 ref in F.Fab layer ?

evanshultz:
> All looks good but one minor questions, any reason why there is no pin 1 ref in F.Fab layer ?

Yes. Because I'm daft. Sadly. Fixed now.

I also improved the pin 1 silk mark on some of them that could use a bit of work. Inspired by the latest IPC 7351 docs, I used primarily silk removed to indicate pin 1:
image

evanshultz:
I got a response from Pulse about the wrong footprint: They simply removed the errant dimension on the left side and added "6X " to the pad height on the right. So I was correct and this PR can be merged.

PR #646, opened 2018-06-08
add QFN-64 footprint with 4.7x4.7mm exposed pad

HubertD:
as used in Microchip SAMC2x: http://ww1.microchip.com/downloads/en/DeviceDoc/DS60001479B.pdf
qfn64-ep47


poeschlr:
The pad placement looks a bit odd to me. From what i see the toe fillet is way too large whereas the heel fillet is too small. (In other words the pads should be moved inwards.)
Is it possible you used the footprint wizard found in pcb_new? (That generator is rubbish as it does not allow proper definition of the pads.)

If you want you can add the dimensions to the ipc dfn_qfn generator found in this pull request to the footprint generator repo. (Simply fork the repo, checkout the last commit listed in that pull request, do your changes and poke me such that i can pull in your addition as well. The script is found in script/Packages/Packages_DFN_QFN/ipc_dfn_qfn_generator.py The definition should be added to the dfn_qfn_definition.yaml file.)

You can of course also do it by hand.

HubertD:
Thanks Rene for your feedback.
I updated the PR with a 4.8x4.8mm exposed pad.

However, I have trouble following your other advice.
Trying to run ipc_dfn_qfn_generator.py from your commit fe7e505cb leaves me with "ImportError: No module named KicadModTree". 
Also, looking at dnf_qfn_definitions.yaml (sic), I don't have the feeling that I would get the myriad of options right :-)

Anyway, I did not generate my footprint in the first place, but copied the existing QFN-64-1EP_9x9mm_P0.5mm_EP7.25x7.25mm footprint, just shrinked the exposed pad and removed some paste pads.
So, if the pad design looks wrong to you, then probably there are also problems in the existing qfn-64 footprints?

poeschlr:
@HubertD wrote:
>Trying to run ipc_dfn_qfn_generator.py from your commit fe7e505cb leaves me with "ImportError: No module named KicadModTree".

That is strange. The path to KicadModTree is added by the script. Did you download the full repo or just the script directory? Did you execute the python script from within the scripts directory or did you execute it from somewhere else?

---

>Also, looking at dnf_qfn_definitions.yaml (sic), I don't have the feeling that I would get the myriad of options right :-)

Most of the things you would need to enter are directly from the datasheet. (only exception are the parameters controlling the paste layer layout.) So i have  a bit more confidence on that part. (But i agree that it might seem a bit much to comprehend at first.)

---

>So, if the pad design looks wrong to you, then probably there are also problems in the existing qfn-64 footprints?

Yes this is correct. I will look into that. (Sadly both qfn-64-1EP footprints look a bit odd. Both to not have a datasheet link present. Some of the symbols who use these should also have a footprint with slightly different exposed pad size then the current footprints.)

poeschlr:
The pad layout of QFN-64-1EP_9x9mm_P0.5mm_EP7.35x7.35mm looks a lot better than the one you copied. So if you base yours on that one it should be ok.

poeschlr:
Ok by starting to fix all qfn-64-1ep assignments i think i by accident generated this footprint already. It will be named `QFN-64-1EP_9x9mm_P0.5mm_EP4.85x4.85mm` i will also include a version with thermal vias which will be named `QFN-64-1EP_9x9mm_P0.5mm_EP4.85x4.85mm_ThermalVias`.

poeschlr:
I think this would be replaced by one of the footprints added in https://github.com/KiCad/kicad-footprints/pull/649

HubertD:
okay, I only need the footprint for my symbols; 
if they're from a common script, that's even much better from my point of view.

as for the ipc_dfn_qfn_generator.py problem, i try to investigate further and come back to you,  @poeschlr 

will close this request and reference QFN-64-1EP_9x9mm_P0.5mm_EP4.85x4.85mm from my SAMC2x parts.

PR #648, opened 2018-06-09
add WLCSP-56_3.444x3.170mm_P0.4mm

HubertD:
Absolutely not sure about this one. I have no experience in creating BGA, let alone WLCSP packages.

Anyhow, I want to add the whole SAMC2[01] family of microcontrollers to kicad-symbols, and therefore I need a WLCSP-56 package.
Maybe someone could look over this?

The chip I'm targeting is the the Microchip SAMC20J18A-U;
Datasheet: http://ww1.microchip.com/downloads/en/DeviceDoc/DS60001479B.pdf 
The WLCSP-56 package is roughly described on page 1075.

Since there is no footprint recommendation in the datasheet, I got some inspiration from a Microchip AppNote:
http://ww1.microchip.com/downloads/en/AppNotes/00002091A.pdf

wlcsp56

Not sure even about the naming of the footprint, I would gladly take some advice..?


poeschlr:
The courtyard clearance should be 1mm for bga like footprints. (To allow rework)

Font thickness for the reference on the fab layer is a bit high. (should be between 10 and 15% of font size)
Line width for fab layer should be 0.1mm and for silk layer it should be 0.15mm

The fab outline should exactly represent the nominal body size (no rounding)

Rotate footprint such that A1 is on the top left corner. (meaning size x and size y will be exchanged in the footprint name)

For reference the dimensioned drawing of the footprint:
screenshot from 2018-06-16 00-34-41

HubertD:
thanks for your feedback @poeschlr!
a little bit better now?

HubertD:
would have loved to use @Ratfink 's generator script instead of my manual drawn version.

unfortunately, the columns in this datasheet: http://ww1.microchip.com/downloads/en/DeviceDoc/DS60001479B.pdf (page 1075) are numbered from right to left (G1 is top left, A1 is top right) and I don't think there is currently a way to configure that for the bga.py script?

Anyways, I hope they didn't mix up top and bottom view in the datasheet?!

Ratfink:
In the top view in the datasheet, A1 is shown in the top right corner.  However, as Rene mentioned above, this footprint should be rotated so that A1 is in the top left corner, which in addition to being the standard orientation for BGA footprints, makes the ball numbering match the standard numbering my script uses.

HubertD:
you are right, @Ratfink - maybe I should get some more sleep.
will update this PR later today.

HubertD:
wlcsp-56

@poeschlr this should be good now?

used @poeschlr 's script to generate;
used parameters:
---
WLCSP-56_3.444x3.170mm_Layout8x7_P0.4mm:
  description: "WLCSP-56, 7x8 raster, 3.170x3.444mm package, pitch 0.4mm; see section 48.2.4 of http://ww1.microchip.com/downloads/en/DeviceDoc/DS60001479B.pdf"
  pkg_width: 3.170
  pkg_height: 3.444
  pitch: 0.4
  pad_diameter: 0.225
  paste_diameter: 0.25
  layout_x: 7
  layout_y: 8
---

poeschlr:
Ping me next week. I am not at home right now.

poeschlr:
You rotated the footprint correctly but did not change the name to fit the new rotation. 
The name should be:  `WLCSP-56_3.170x3.444mm_Layout7x8_P0.4mm`
Otherwise the footprint looks good. Could you then also make a pull request on the footprint generator repo that adds your parameters to the yaml file and link it here?

poeschlr:
Looks good, thanks

PR #630, opened 2018-06-04
QFN-64 footprint for LAN951x

jneiva08:
Created by the script the footprint QFN-64 1EP 9x9mm P0.5mm EP7.3x7.3mm

Datasheet: http://ww1.microchip.com/downloads/en/DeviceDoc/00002304A.pdf

qfn64

This PR for the symbol is [#651](https://github.com/KiCad/kicad-symbols/pull/651)

------------

Thanks for creating a pull request to contribute to the KiCad libraries! To speed up integration of your PR, please check the following items:

- [x] Provide a URL to a datasheet for the footprint(s) you are contributing
- [x] An example screenshot image is very helpful 
- [ ] If there are matching symbol or 3D model pull requests, provide link(s) as appropriate
- [ ] Check the output of the Travis automated check scripts - fix any errors as required

PR #649, opened 2018-06-10
Add new qfn-1ep footprints

poeschlr:
As part of checking https://github.com/KiCad/kicad-footprints/pull/646 i noticed that a lot of symbols that had one of the two qfn-64-1ep footprints assigned would need a footprint with very different ep sizes.
This pull request adds them.
The footprints are script generated. (added by latest commit to https://github.com/pointhi/kicad-footprint-generator/pull/125)

The symbol side of things is done in https://github.com/KiCad/kicad-symbols/pull/661

Datasheet links:
- http://ww1.microchip.com/downloads/en/DeviceDoc/60001477A.pdf (page 1083)
- http://ww1.microchip.com/downloads/en/DeviceDoc/atmel-8285-8-bit-avr-microcontroller-atmega165a_pa_325a_pa_3250a_pa_645a_p_6450a_p_datasheet.pdf (page 827)
- http://www.ti.com/lit/ds/symlink/tusb8041.pdf
- http://ww1.microchip.com/downloads/en/DeviceDoc/doc7593.pdf (page 432)
- http://ww1.microchip.com/downloads/en/DeviceDoc/Atmel-2549-8-bit-AVR-Microcontroller-ATmega640-1280-1281-2560-2561_datasheet.pdf (page 415)

All footprints come as a thermal via version and a normal option.

Overview screenshot of all 6 added general versions (only thermal via variant shown)
screenshot from 2018-06-11 19-28-05


poeschlr:
As decided in https://github.com/KiCad/kicad-footprints/issues/658 the footprints now use rounded rectangle pads.

evanshultz:
Good catch! Images above look good and a quick check of a few roundrect ratios gives the correct radius values. Merging on assumption of correctness otherwise.

jneiva08:
Sorry for bothering,  Probably no not make sense the footprint on PR [630](https://github.com/KiCad/kicad-footprints/pull/630), for QFN-64 1EP 9x9mm P0.5mm EP7.3x7.3mm.
is possible to add to your script?

PR #476, opened 2018-04-07
Add BatteryHolder Keystone 1048 into Battery.pretty

ghost:
Add BatteryHolder Keystone 1048 into Battery.pretty
------------

Thanks for creating a pull request to contribute to the KiCad libraries! To speed up integration of your PR, please check the following items:

- [ ] Provide a URL to a datasheet for the footprint(s) you are contributing
- [ ] An example screenshot image is very helpful 
- [ ] If there are matching symbol or 3D model pull requests, provide link(s) as appropriate
- [ ] Check the output of the Travis automated check scripts - fix any errors as required

PR #650, opened 2018-06-10
Added BatteryHolder_Keystone_1048_2x18650

Misca1234:
Replacement of push
https://github.com/KiCad/kicad-footprints/pull/476#issuecomment-396054078

Batter holder for 2x 18650 cells
http://www.keyelco.com/product-pdf.cfm?p=920

bild

bild

bild

bild

------------


poeschlr:
I remember from the original some question about pin numbers regarding a fitting symbol. Do we have a symbol that fits this pin numbering scheme? If not it might be a good idea to add one.

Misca1234:
Yes, I started a discussion about it in 
https://github.com/KiCad/kicad-symbols/pull/663

with an initial idea, that idea seems to be impossible to make due to several of the foot prints,

So I will make a symbol for each foot print instead, I have not removed the thread
https://github.com/KiCad/kicad-symbols/pull/663

so we could have a discussion of how a battery hold symbol should look like, should it already have a power flag etc.

Misca1234:
working on a script that generates this foot print, closing it

PR #651, opened 2018-06-11
SO-4 footprints - silk too close to pads

johnbeard:
* SO-4_4.4x2.3mm_P1.27mm
* SO-4_4.4x3.6mm_P2.54mm
* SO-4_7.6x3.6mm_P2.54mm
* SO-5_4.4x3.6mm_P1.27mm

In the SO-4 footprints, the silkscreen was within 0.05 mm of the
pads. KLC guidelines are 0.2mm or more.

This means that the silk might be removed by the board manufacturer, and
if this happens, it will remove the entire pin-1 marker line from the
physical PCB.

This moves the silk in these footprints to be 0.2mm or more from pads.

Example before (local clearance only for illustration)

before

And after:

after

------------


poeschlr:
Looks good, thanks

PR #652, opened 2018-06-11
Add Panasonic GaN FET and Driver footprints

evanshultz:
- https://industrial.panasonic.com/content/data/SC/ds/ds7/c0/PKG_HQFN016-A-0404XZL_EN.pdf
- https://industrial.panasonic.com/content/data/SC/ds/ds7/c0/PKG_HSON008-A-0808XXI_EN.pdf

It's a bit confusing to me how the info is layed out, but it makes enough sense to follow after starting at it a bit. Note that the 2nd link above has the file named "HSON", but it looks like a DFN and is called "DFN" at https://industrial.panasonic.com/ww/products/semiconductors/powerics/ganpower/gan-power-devices so I went with that.

image

------------


evanshultz:
The weird `Val**` text and broken script checker stopped me, but I had to leave for vacation (without a computer) before I could figure it out. It turns out that if the `reference` field (ref des) is changed to `user`, then text of `Ref**` is created. Similarly, if the `value` field is `user` then you get the `Val**` text. This is probably intentional since those two fields are created for every new footprint and KiCad is generating them again if they're missing.

I fixed the other footprint text fields as mentioned above, added silk to both, and updated the paste pads on the HSON (I used 0.9mm square paste pads as 1mm square runs over the 6.94mm copper pad) in the last commit (not sure why it seems to have overwritten the previous files instead of incrementally update them):
image

For the HQFN the "metal mask" (which is the stencil opening) is specifically show as 2.05mm square; we do usually split them up but this is a manufacturer-specific footprint built according to their recommendation. So is it OK?

poeschlr:
I would still split the paste layer. There is no reason why not to do it. (Unless the datasheet gives one i would just assume the guy drawing it up was lacy or did not know better.) You can easily split it and keep the same overall area.

evanshultz:
Alright. They had exactly 50% paste coverage before, so I've done that using 4 small paste squares.

poeschlr:
Looks good, thanks

PR #653, opened 2018-06-11
Use rounded rectangle pads for two terminal chip parts

poeschlr:
The newer IPC standards suggest to use rounded rectangle pads instead of
normal rectangle pads.

One source for this claim is: http://www.ocipcdc.org/archive/What_is_New_in_IPC-7351C_03_11_2015.pdf (page 9)

---

The only changes to the files should be rect -> roundrect and the addition of the roundrect ratio.
The script changes are in commit: https://github.com/pointhi/kicad-footprint-generator/pull/125/commits/9a835417fae6881c4e8a79119ddcaec395705047 (part of my pull request https://github.com/pointhi/kicad-footprint-generator/pull/125)


poeschlr:
This does now also fix https://github.com/KiCad/kicad-footprints/issues/655

PR #662, opened 2018-06-16
Added new HSOIC-8 for the LMR14030

johnfwhitmore:
I've looked at the Package_SO:SOIC-8-1EP_3.9x4.9mm_P1.27mm_EP2.35x2.35mm but it's thermal pad is the wrong size so I've created this symbol. I might be looking in the wrong library for the package I require, in which case this footprint is not required, which will be great.

I'm going to push and check Travis output

datasheet - http://www.ti.com/lit/ds/symlink/lmr14030.pdf

screenshot from 2018-06-16 15-41-36


johnfwhitmore:
I hope that's more or less it but I'm sure my courtyard, Fab and Silk Screen aren't perfect.

johnfwhitmore:
screenshot from 2018-06-16 16-54-53

poeschlr:
The silk outline should not go around the pins but only the body. So in the case of SOIC one line on top and bottom side of the footprint (the top side on the left elongated as a pin 1 marker.

The fab outline should be the exact dimension of the body. With a chamfer near pin 1.

For inspiration look at other SO footprints.

---

The datasheet also suggests a much larger copper pad than you used. The current pad would be mask layer only and a larger pad added with the dimensions given in the datasheet.

have a look at for example `HTSSOP-16-1EP_4.4x5mm_P0.65mm_EP3.4x5mm_Mask3x3mm_ThermalVias` for inspiration with that

johnfwhitmore:
I guess you're referring to the shaded area labelled "Solder mask over copper"

poeschlr:
yes the shaded area is what i meant. (add a pad with the size of the shaded area. That pad does not have mask or paste assigned. From the current large pad remove the copper layer and the pin number. Meaning it will be a mask only pad.)
The footprint name should mention the copper pad size for the ep size and also the mask size.

johnfwhitmore:
I think that's a bit better?

screenshot from 2018-06-16 18-16-56

johnfwhitmore:
Oh sorry forgot to mark pin 1!

johnfwhitmore:
screenshot from 2018-06-16 19-36-34

johnfwhitmore:
that's the paste layer split into four.
Think that's all your comments addressed?

johnfwhitmore:
Can I ask about the status of this pull request? Just like to clean up my branches if possible.

johnfwhitmore:
Ping... Sorry to be back to you, and no panic, but just wondering about status of this PR.
Thank You.

poeschlr:
Sorry for the long wait. We where quire busy with the release.

The courtyard clearance should be 0.25mm all around (in y direction measured with respect to the body dimensions = Fab outline, in x direction measured with respect to the outer pad edges)

screenshot from 2018-07-28 21-20-27

Misca1234:
@johnfwhitmore 
Hi, could you address the issues Rene pointed out?

johnfwhitmore:
Sorry this one slipped through the cracks! I'll get on it

poeschlr:
We now have a script that can create such footprints quite easily. So i would really prefer this one to be entered into it. Would you be willing to look into this? Or if not would you be mad at me if i take care of this?

If you are interested the script can be found here: https://github.com/pointhi/kicad-footprint-generator/tree/master/scripts/Packages/Package_Gullwing__QFP_SOIC_SO

PS: sorry it took us so long.

johnfwhitmore:
I'd not be mad at all if you took care of it. You are probably better qualified to know how to correctly run the script and get things correct.

Misca1234:
It is very easy, you enter the basic parameters for the foot print as parameters in a data file and the run the master script via python, and viola the foot print is done.

We prefer the script because then we only need to match the data in the data sheet against the parameters in the parameter file and then decide for a merge or not, the reason is that we trust that the script is correct.

If the contribution is made by hand we must check every little line and pad, and layers for the foot print manually before a merge.

I tend to stay away from reviewing hand made foot print with thermal pads => less people that takes on review => longer time before merge.
But with the script made ones, it is dead easy to review even for me.

PR #667, opened 2018-06-17
Add UDFN-6_1.6x1.6mm_P0.5mm package

Y0ba:
Package Information: http://www.onsemi.jp/pub/Collateral/517AU.PDF

Example: 
udfn6

------------


Y0ba:
Seems like there is another package with same name: https://www.onsemi.com/pub/Collateral/517AP.PDF

Added case name into footprint name.

poeschlr:
The courtyard clearance should be 0.25mm all around
screenshot from 2018-07-28 21-31-38

I am a bit unsure about the pin numbering of this footprint.
Do you have a datasheet of a part that uses this footprint? That way we could look at how the pin numbering is used in a component.

I am also not really sure about the footprint naming. If we add the onsemi specific package name then we should add a prefix onsemi

The second datasheet you link is for a completely different footprint that would have the name `UDFN-6-1EP_1.6x1.6mm_P0.5mm_EP0.61x1.26mm` (So that one does not really influence the naming of the footprint committed here)

Y0ba:
> Do you have a datasheet of a part that uses this footprint?

http://aosmd.com/pdfs/datasheet/AON1611.pdf

I used pin numbering from this datasheet, so that I could use symbol Q_NMOS_DGS with this footprint. I can make numbering 1/2/3/4 and add a separate symbol for AON1611.

> I am also not really sure about the footprint naming.

It's not specific to onsemi.

poeschlr:
I fear we will take a bit to find a good footprint naming convention for this. (future proof)
And also a pin numbering scheme. (The manufacturers using such a package really are of no help here)

PR #670, opened 2018-06-18
Use rounded rectangle pads for scripted LGA footpritns

poeschlr:
As decided in #658 these use rounded rectangle pads


evanshultz:
Only one question about `Bosch_LGA-8_3x3mm_P0.8mm_ClockwisePinNumbering.kicad_mod`: The pads went from 0.4x0.55mm to 0.5x0.55mm. But looking at the datasheet they should be 0.4x0.4mm. What's going on?

evanshultz:
I do get that. And it makes sense what you say.

My concern is not adhering to the manufacturer's recommendation, however. Perhaps there is a reason to make the footprint exactly as specified. If a user has a problem and finds out the KiCad footprint doesn't match the footprint specified by the vendor, isn't that our problem? If this was a generic footprint it would be harder to support this argument but with a vendor-specific footprint it's concerning to me that KiCad would not follow the vendor's suggestion. Certainly the vendor may suggest a poor footprint, but do we know that is the case or only that this footprint doesn't naively follow IPC tolerancing?

poeschlr:
The corrected pad calculation equations result in a footprint matching the suggestion by bosh much closer.

evanshultz:
LGA-14_3x2.5mm_P0.5mm_LayoutBorder3x4y:
- I'm not getting your value for the X location of pin 1. ( 3mm wide package / 2 ) - 0.1mm pad offset - 0.45mm (middle of pad width range) gives me 1.175mm. You have 1.1625mm. Similarly, I find pad 6 to have a Y location of 0.925mm and you have 0.9125mm. Why are we not getting the same value?
- Should only pin 1 have a custom shape? All other pins look to be rectangular in the datasheet.

LGA-14_3x5mm_P0.8mm_LayoutBorder1x6y:
- Here pin 1 shown with a notch in the datasheet but a normal rectangular pad in the footprint. This is in contrast to the above footprint.

LGA-16_3x3mm_P0.5mm_LayoutBorder3x5y:
- Probably similar to the first bullet above, I calculate an X location for pin 1 of 1.225mm. You have 1.2125mm.

evanshultz:
@poeschlr 
Can you look at the above? Are these correct or is there any adjustments to be made?

evanshultz:
@poeschlr ?

poeschlr:
This is a bit on the backburner right now. (The LGA generator needs a bit of work. I will reopen a new PR once i have it up and running again)

PR #671, opened 2018-06-18
Use rounded rectangle for scripted PLCC

poeschlr:
As decided in #658 these use rounded rectangle pads


evanshultz:
Can you please post screenshots? It's helpful not just for reference in this PR but they make it easier to track what happened in the future.

poeschlr:
Here you go. One of them as a representation of all of them:
screenshot from 2018-06-19 18-50-15

evanshultz:
PLCC has pin 1 at the top middle, not top left. That will also affect the pin 1 mark on top silk and fab layers.

poeschlr:
Well spotted. I missed that when i originally wrote the generator scripts. (Will fix that after i dealt with the terminal blocks)

poeschlr:
Pin 1 is now at the correct position
screenshot from 2018-06-20 00-45-52

evanshultz:
Nice! Great job.

For reference, package drawings are at http://www.microsemi.com/index.php?option=com_docman&task=doc_download&gid=131095. And starting at page 96 of https://www.microsemi.com/document-portal/doc_view/130699-ds2136-40mx-and-42mx-fpga-families-datasheet one can see the pinout goes CCW around the part.

Issue #672, opened 2018-06-19
MPT Phoenix Terminal block generates assert on 3d viewer


FabianInostroza:
When viewing the following footprints 
TerminalBlock_Phoenix_MPT-0,5-2-2.54_1x02_P2.54mm_Horizontal.kicad_mod
TerminalBlock_Phoenix_MPT-0,5-3-2.54_1x03_P2.54mm_Horizontal.kicad_mod

in the 3d viewer there are some asserts that fail.
The reason is that these footprints contain NPTH with zero size

poeschlr:
The script for these is build kind of wrong. (uses size 0 do indicate NPTH instead of using the drill size for that.)

poeschlr:
I discovered another problem with the script. It adds pin numbers to NPTH drills.

FabianInostroza:
Seems you already fixed this, thanks @poeschlr

PR #673, opened 2018-06-19
Scripted terminal blocks used size 0 for NPTH pads

poeschlr:
- NPTH pads need to have the pad size set equal to the drill size.
- Removed pin number from npth pads
- Includes other inprovments of the footprint-generator-script made
since the last export (Example corrected rounding of angle fields)

This is a fix for https://github.com/KiCad/kicad-footprints/issues/672

Changes to the script: https://github.com/pointhi/kicad-footprint-generator/pull/135


evanshultz:
I don't see any NPTHs with a pin number changed. Searching in the commit there are no hits for "np_". But the rounding of silk arcs and adjustments on courtyard seem OK.

Also, many of them have a comma in the tags. Would you want to address that simple item now?

TerminalBlock_Phoenix.pretty/TerminalBlock_Phoenix_MPT-0,5-3-2.54_1x03_P2.54mm_Horizontal.kicad_mod
```diff
@@ -38,15 +38,15 @@
...
   (fp_line (start -2 -3.6) (end -2 3.6) (layer F.CrtYd) (width 0.05))
-  (fp_line (start -2 3.6) (end 7.1 3.6) (layer F.CrtYd) (width 0.05))
-  (fp_line (start 7.1 3.6) (end 7.1 -3.6) (layer F.CrtYd) (width 0.05))
-  (fp_line (start 7.1 -3.6) (end -2 -3.6) (layer F.CrtYd) (width 0.05))
+  (fp_line (start -2 3.6) (end 7.08 3.6) (layer F.CrtYd) (width 0.05))
+  (fp_line (start 7.08 3.6) (end 7.08 -3.6) (layer F.CrtYd) (width 0.05))
+  (fp_line (start 7.08 -3.6) (end -2 -3.6) (layer F.CrtYd) (width 0.05))
   (pad 1 thru_hole rect (at 0 0) (size 2.2 2.2) (drill 1.1) (layers *.Cu *.Mask))
-  (pad 1 np_thru_hole rect (at 0 2.54) (size 0 0) (drill 1.1) (layers *.Cu *.Mask))
+  (pad "" np_thru_hole circle (at 0 2.54) (size 1.1 1.1) (drill 1.1) (layers *.Cu *.Mask))
```
poeschlr: @evanshultz here is an example for the changes regarding npth pads

poeschlr:
We decided with other phoenix parts that the comma is ok as it is after all part of the real part number used. (The idea behind not allowing comma in the tag field is that we wanted to avoid comma separated lists. This is not the case here)

evanshultz:
OK. Thanks. I will merge with that, then.

While it doesn't affect the footprint, do you want to do any rounding in the description and/or tags as can be seen below?
```
(descr "Terminal Block Phoenix MPT-0,5-7-2.54, 7 pins, pitch 2.54mm, size 18.240000000000002x6.2mm^2, drill diamater 1.1mm, pad diameter 2.2mm, see http://www.mouser.com/ds/2/324/ItemDetail_1725672-916605.pdf, script-generated using https://github.com/pointhi/kicad-footprint-generator/scripts/TerminalBlock_Phoenix")
(tags "THT Terminal Block Phoenix MPT-0.5-7-2.54 pitch 2.54mm size 18.240000000000002x6.2mm^2 drill 1.1mm pad 2.2mm")	+  (tags "THT Terminal Block Phoenix MPT-0,5-7-2.54 pitch 2.54mm size 18.240000000000002x6.2mm^2 drill 1.1mm pad 2.2mm")
```

poeschlr:
Yea this should be rounded. I will look into that.

poeschlr:
Rounding of the documentation fields is introduced. Sadly python has no good format string for "show at most two digits after comma". The options are: show exactly two digits after comma or show at most 3 significant digits. (I chose the later. But that means that if there are 2 digits before the comma, only one digit after the comma is shown.)

evanshultz:
How sad and surprising; I hadn't come across this use case either.

We're getting into the possibly-useless minutiae here, and it's likely not worth doing, but what about `"{:.2f}".format(x).rstrip("0").rstrip(".")`? It is awfully lame and ugly, but for a one-by-one solution it works on all the following numbers for me:
```
3.1415926
3.1
3
3.0000
3.00001
0.00025419
```

It would only need to be applied to a few specific numbers, or the formatting could be a pre-process to the formatting/printing.

But if you just want me to merge as it is, I'll happily just do that.

poeschlr:
I think this is overkill for the description field. If you are unhappy that some things are rounded to one digit after the comma then i will use .2f (will create a bit more noise but well i am not prepared to completely rewrite the generators for this.)

evanshultz:
Nah. Just asking since I was as sad as you about this. Seems like something other folks would want to do and Python would support, but alas...

Thanks for the fixes! Will merge now!

PR #675, opened 2018-06-20
Fix PLCC 32 (original had wrong pin layout)

poeschlr:
The original plcc 32 had 8 pins on every side. All datasheets of parts we have in the lib show a layout of 7 on the top/bottom side and 9 on left/right.

Datasheet: http://ww1.microchip.com/downloads/en/DeviceDoc/doc0015.pdf
Symbol PR: https://github.com/KiCad/kicad-symbols/pull/677

The original:
screenshot from 2018-06-20 02-44-39

The new one:
screenshot from 2018-06-20 02-43-45


evanshultz:
So our PLCC32 footprint didn't match any existing package? Hard to believe, but possible. And there's no datasheet on the footprint to check. I looked at Mouser and DigiKey, plus a quick Google search, and all I found was this unequal package so I guess so.

However, removing the sockets may not be right. See https://www.digikey.com/products/en/connectors-interconnects/sockets-for-ics-transistors/409?FV=ffe00199&quantity=0&ColumnSort=0&page=1&k=plcc+32&pageSize=25&pkeyword=plcc+32 and you will see there are various layouts of sockets available, included one with 8 leads on each side. That leads me to believe there are packages available even though I couldn't quickly find them. So what about leaving the sockets alone since there are (or may be) compatible square PLCC32 sockets.

poeschlr:
I checked quite a few of these products. As far as i can see all of them have the 7/9 pin layout. Could you point me to the one with the 8/8 layout? 

---

More details:
The 3M datasheets have a picture of the 32 pin version which is clearly 7/9, the milli max has a footprint layout that is 7/9, CNC Tech does not explicitly state a layout but the side lengths point strongly to the fact that it also has the 7/9 layout. 

I did not look at the datasheets of the rest of them. But i looked at the pictures from digikey. All of these who have 32 pins in the picture have 7/9 layout.

The same is true for a google image search for plcc-32.

---

I would assume if there would be a 8/8 layout version out there it would be possible to find either a picture or a datasheet for it. I really searched for quite a while when i designed the footprint and again today.

If you are unconformable removing them i could move them to legacy instead with a note that we can not find a single component fitting them.

evanshultz:
I couldn't find any packages with the 8x8x8x8 layout as mentioned, only https://www.digikey.com/product-detail/en/preci-dip/540-88-032-24-008/540-88-032-24-008-ND/4109891 which is a square PLCC32 SMT socket:
image
image

I also find the 7x9x7x9 to be the clearly most common variant, but did see this single square one.

poeschlr:
I could not find the source for your screenshot.  Could you provide a link?

I checked both datasheet links on your linked product page and the 3d model. All 3 tell me it is 7/9 layout.

[The pdf datasheet](http://www.precidip.com/AppHost/9699,1/Scripts/Modules/Catalog/Default.aspx?c=12&i=1819&p=375&pdf=1&dsku=540-88-XXX-24-008) lists quite different sizes to your screenshot. In particular it has the same side lengths as the 7/9 versions i listed above. (The picture in there is for the 52 pin version, as is the picture you posted.)

The acrobat reader version is even clearer. There is a suggested footprint drawn for the 32pin variant. Again 7/9 layout. (In the list where all versions are listed, the 32 pin variant has a * next to it. The note then tells the reader that it is rectangular.)

The 3d model linked from the product is also 7/9 layout.

---

And the current 4x8 version is not square but has the side lengths that i would expect from the 7/9 version.

evanshultz:
It is possible we are getting different datasheets because we're in different locations. I've attached the one I downloaded from DigiKey which shows square sides:
[Catalog.pdf](https://github.com/KiCad/kicad-footprints/files/2124213/Catalog.pdf)

It could also be outdated info, and I don't see a date on the datasheet I found so I went to their website. Here is the product page showing square ones: http://www.precidip.com/en/Products/PGA-BGA-PLCC-Sockets/pview/540-88-XXX-17-400.html

http://www.precidip.com/en/Products/PGA-BGA-PLCC-Sockets/pview/540-88-XXX-24-008.html is a slightly different family which shows the rectangular PLCC32 socket.

I find two JEDEC standards referenced but I can't find those standards to review them.

evanshultz:
OK, I got it now. I was going by the A1/A2 numbers but indeed you are right and the socket's casing size does not reflect the PLCC package size even here where the socket is square. My apologies.

Issue #678, opened 2018-06-20
I think my advice for EP sizes was wrong


poeschlr:
As you guys might know, i advised using either IPC standards or at least the maximum tolerance for the ep size.
Most datasheets that give a suggested footprint select the nominal size as the size for the pad on the pcb.
I asked over at the forum for advice: https://forum.kicad.info/t/size-of-exposed-pads/11209

A user there found this [TI document](http://www.ti.com/lit/an/slua271a/slua271a.pdf) that explicitly states that the ep size should be the same size as the "slug" (Which i would interpret as "use the nominal size") Sadly that document does not give a reasoning behind this statement. (The user suggested that it could have to do with improving self alignment.)

---

So i guess my interpretation of the IPC standard in this regard was wrong. (Meaning some of the later added footprints with exposed pads might not be ideal. I doubt they will be so wrong as to merit immediate response but i think in future we can go back to using nominal sizes for the EP.)

poeschlr:
Yes i am convinced that IPC-7351 only applies to the lead pads.
The only thing mentioned about exposed pads is that for qfn the clearance between it and the other pads should be at least 0.2mm and that to achieve that it is advisable to reduce the heel fillet.

poeschlr:
[Microchip seems to suggest EP sizes all over the place.](http://ww1.microchip.com/downloads/en/PackagingSpec/00000049BQ.pdf) (Sometimes the maximum slug size, but also the minimum and also values in between including nominal.) So i would guess it is not that critical.

poeschlr:
Yes i would say so. (There will of course be exceptions if there is a good reason given.)

evanshultz:
Just one general question: I have seen bridging between metal elements on a package and traces running under the package. Rivets on snap caps, exposed slugs, formed metal "wings", etc. Is using the nominal value the best choice since we only have an unattributed TI doc saying to do it and other docs giving different guidance?

I got a response from TI and they said basically that this doc is intended to be general and is only pointing out to maintain clearance to other pads. The datasheet supersedes this doc.

I'm fine with using the nominal value, but I just wish we had a more authoritative source of info on this...

poeschlr:
Using nominal size is only the advice for exposed pads for parts similar to qfn.

[...]

PR #679, opened 2018-06-21
Correct my mistake regarding EP size for qfn-64

poeschlr:
This partly fixes my recent mistakes regarding EP size. See https://github.com/KiCad/kicad-footprints/issues/678
It holds the fixed footprints for https://github.com/KiCad/kicad-symbols/pull/661

---

The 5.6 and 5.55 square ep pad versions both had nominal size 5.4 but different tolerances. (meaning they now are the same.)

---

A screenshot showing the footprints (ep size increase from left to right and top to bottom)
screenshot from 2018-06-21 04-11-51


poeschlr:
For these using chamfered pads to avoid the vias is not really necessary as the suggested radius for rounded rectangle is already enough.
Updated screenshot of the thermal via versions
screenshot from 2018-06-26 00-45-35

PR #680, opened 2018-06-21
Fix my mistake with ep sizes with regards to texas mcu footprints

poeschlr:
This fixes the ep size as described here https://github.com/KiCad/kicad-footprints/issues/678 for the texas mcu footprints originally added here https://github.com/KiCad/kicad-footprints/pull/587 (the thermal vias version was added in a later pull request it seems)

This also changes the pad shapes to round rect as described here https://github.com/KiCad/kicad-footprints/issues/658

screenshot from 2018-06-21 05-01-05


poeschlr:
Last update now uses round rect where possible instead of chamfered pads.
I also changed the N24 version to place paste over vias (but increased the coverage to combat the resulting paste wicking.) This was necessary as the clearance between the past pads was way too small even with only 50% coverage.

screenshot from 2018-06-26 01-14-52

PR #681, opened 2018-06-21
Fix ep size mistake for dfn-6

poeschlr:
Fixes the dfn footprint with regards to the problem described in https://github.com/KiCad/kicad-footprints/issues/678. The original footprint has been added in https://github.com/KiCad/kicad-footprints/pull/539.
The symbol side of things is done in https://github.com/KiCad/kicad-symbols/pull/679.

screenshot from 2018-06-21 04-56-30


evanshultz:
@poeschlr 
Should the old footprint be removed?

For the record, here is the datasheet for the part driving the initial footprint creation: https://www.silabs.com/documents/public/data-sheets/Si7020-A20.pdf.

PR #688, opened 2018-06-23
Fix footprints of RECOM DC/DC converters.

rschaeuble:
Fix footprints of RECOM DC/DC converters, where the 'bulge' was toward the back while it should have been to the front.

The footprints of two of the RECOM DC/DC converters had the buldge of the body toward the wrong side.

Wrong:
recom-wrong

Correct:
recom-correct

This affected the R-78B-2.0 (see https://www.recom-power.com/pdf/Innoline/R-78Bxx-2.0.pdf, page 5) and the R-78HB-0.5 (see https://www.recom-power.com/pdf/Innoline/R-78HBxx-0.5_L.pdf, page 5).

The changes were quite simple: moving everything except the pins downwards for 4.5mm.

The 3D models also need fixing, for this I created another pull request: https://github.com/KiCad/kicad-packages3D/pull/327

------------


Shackmeister:
Thanks for the fix! :)

Shackmeister:
Hi @rschaeuble these footprints were actually generated via scripts like the 3D models. Could you please make a PR for the script with your changes? This way the bug won't come back if somebody re run the entire series from the script https://github.com/pointhi/kicad-footprint-generator/blob/master/scripts/Recom_DCDC/Recom_SIP.py

rschaeuble:
Sure thing.

In the worst case it might take me until the weekend, but I'll get it done.

rschaeuble:
I have created a pull request for the generator script: https://github.com/pointhi/kicad-footprint-generator/pull/141

While fixing the generator, I found a few other small errors in the footprints (wrong package widths, slightly wrong offset of pin 1 etc.). Once the generator has been accepted by its author, I will create new merge request for the footprints files in this repo. I will then also make the corresponding fixes to the 3D models.

pointhi:
Please be careful about the recom regulators. They have both versions to my knowledge (bulge in front/back)

For example: https://www.recom-power.com/pdf/Innoline/R-78Exx-1.0.pdf

The question is now: does the Letter (B, HB, E,...) specify where the bulge is? I don't know.

rschaeuble:
It's not only the bulge There's also the issue that some datasheets specify the horizontal offset of pin 1, while others specify pin 3's offset, so pin 1 has to be calculated.

I've double and tripple checked all values when I made the changes in my pull request to the generator. I opened all footprints in KiCad, measured all relevant distances and compared them to the datasheets.

evanshultz:
@rschaeuble 
We typically merge footprints before the script, so that the output is checked and verified to be OK. Once footprints are merged then the script is merged. This has worked well for us so if you don't object would you please do footprints first?

rschaeuble:
@evanshultz: no objections at all. Here's the PR: https://github.com/KiCad/kicad-footprints/pull/721

PR #689, opened 2018-06-24
2 terminal smd chip: Pad equations where outdated

poeschlr:
- New industry standards have updated not only the fillet sizes but also
the way how tolerances are respected.
- For devices that directly give the lead to lead spacing no complicated
calculations are necessary. (was wrong previously)

---

Sorry for again getting it wrong. [I noticed my mistake when i made this post on the forum](https://forum.kicad.info/t/v5-heads-up-devs-dont-explain-here-the-upcoming-changes/11123/27)
(The new equations make a lot more sense to be honest.)

---

This results in slightly increased pad to pad clearances for a lot of devices. (All tantals and most parts larger than 1210. For smaller parts there is no change as rounding hides the mistake.)

Some of the current tantals had extremely small pad to pad clearances. So they are not usable without this update.

---

The change to the script is done in this commit: https://github.com/pointhi/kicad-footprint-generator/pull/125/commits/c419ebb0d5573f119eb134bcd1a6ec4151a38bda


Misca1234:
214 file changes  .. man
I guess we just take your word that all are good and merge  :) ?

poeschlr:
Look at what changed in the script :) (That is why scripts are good. You can see the changes made and can infer what that change will do. In this case you can even check the equations against the industry standard.)

Misca1234:
That is fine with me, new process comes with new tools

jcyrax:
For information 0201 footprint was also influenced by this. It had 0.12mm clearance in previous (wrong) version, now 0.18mm.

Regarding changes, I can confirm that they look correct to me. I compared script formulas to IPC-7351B document and also calculated by hand the same clearance S for 0201 footprint.

PR #453, opened 2018-03-26
Add TI 20-pin S-PVQFN 2.4mm EP variant

xl0:
Datasheet: http://www.ti.com/lit/ds/symlink/cc1101.pdf page 103.
pvqfn
QFN-20-1EP_4x4mm_P0.5mm_EP2.5x2.5mm 3D model matches reasonably well.

PR #691, opened 2018-06-24
Replace PR #453: add Texas_S-PVQFN-N20_EP2.4x2.4mm

poeschlr:
Replacement for stale PR https://github.com/KiCad/kicad-footprints/pull/453
- This footprint is script generated

Script can be found in: https://github.com/pointhi/kicad-footprint-generator/pull/125
Datasheet: http://www.ti.com/lit/ds/symlink/cc1101.pdf

screenshot from 2018-06-25 01-46-04


Misca1234:
I guess you notice that travis complain about the 3D model name for Texas_S-PVQFN-N20_EP2.4x2.4mm_ThermalVias ?

poeschlr:
Yes this is a known problem of travis. It does not make sense to use a different 3d model for the thermal via version as it is intended for the same physical part. So we typically point its 3d model path to the same 3d model as the one without thermal vias.

Misca1234:
ok

PR #693, opened 2018-06-25
Add 121-ball 0.8mm pitch BGA footprint.

awygle:
Based on Lattice caBGA package, found at http://www.latticesemi.com/view_document?document_id=213.

Intended for the Lattice ice40HX FPGA in BG121 package - keep an eye out for the symbol PR.

Screenshot:

image

------------


whitequark:
@evanshultz can you please take a look?

herostrat:
Hey guys,

is this a lattice only package or does anyone else use it?

One remark, looking at: www.latticesemi.com/-/media/LatticeSemi/Documents/ApplicationNotes/PT/PCBLayoutRecommendationsforBGAPackages.ashx?document_id=671 one can see that the solder mask opening and pad dimensions are not the same.
Unfortunately i cannot find data there about the 0.8mm pitch 121 caBGA package.
The picture on page 2 shows what I mean:
grafik

Do you have a pcb recommendation for this specific footprint?

awygle:
I based this footprint on Table 1 from that document. Specifically, the 121-ball caBGA on page 60 of the PackageDiagrams document has e = 0.8, b = 0.4, as does the 381-ball caBGA on page 99. The 256-ball caBGA has b = 0.45. So I used the PCB Solder Land Diameter from the NSMD column of the 2nd row of the 0.8mm Ball Pitch section of that table, which includes the 381-ball caBGA.

herostrat:
Ok , than I do not get the dimensions for the solder pad and solder mask opening.

I understand the physical dimensions of the balls (pitch 0,8mm and diameter 0,4mm as you said)
I also understand the arguments about similar footprints and using their definition as seen in the table 1 from TN1074:
grafik

Because the package only uses SMD for the device pads we need a NSMD pad on this footprint.
Which means that the solder land diameter should be 0.35mm as you designed.

The only thing that bugs me is that there is no definition for the solder mask opening.
The document shows that there is a gap between pad and mask.
But I cannot see how big that has to be.
This is rather important for soldering this packages correctly in my experience.

The Artix package I pulled (https://github.com/KiCad/kicad-footprints/pull/616) has an example of what I mean.


For the name, I could only find lattice products which mach the dimensions of this footprint.
So I think a more fitting name would imho be something like this
Lattice_caBGA-121_9x9mm

awygle:
> The only thing that bugs me is that there is no definition for the solder mask opening.

You're right - the document does not specify the size of the mask opening for NSMD, or the size of the pad for SMD. I agree that they should, but they don't. Given that, I left it as 0 as per the KLC, allowing the user to choose the value which matches their PCB process requirements. If you'd prefer I set it to a specific value, just let me know what that is.

> For the name, I could only find lattice products which mach the dimensions of this footprint.

I didn't run into any of these either, however, there's really nothing Lattice-specific about a fully-populated 11x11 BGA with 0.8mm pitch. If it were partially unpopulated (as many of the ECP5 packages, for example), then I'd agree with a vendor-specific name, but this package seems quite generic to me. The fact that nobody uses it now doesn't mean they won't in the future. If you tell me to add Lattice to the name again, I'll do it, but I wanted to express my disagreement once.

herostrat:
Please have in mind that I am no maintainer and do not decide on anything.

I'm just trying to help out a bit in order to speed up the merging of PRs.
I think asking questions about things that don't seem obvious, writing stuff down/summarizing the thought process behind some design choices ad correcting errors could help out a bit.

I don't have a problem with the name of your footprint, just making sure we thought about it :)

So we'll have to wait for an official response

awygle:
Absolutely :) I appreciate your insight and feedback!

awygle:
The corresponding symbol PR is https://github.com/KiCad/kicad-symbols/pull/777 and it now passes all Travis tests except the lack of this footprint.

awygle:
Just a reminder that this PR exists - is someone available to review?

Shackmeister:
sure thing :)

awygle:
> the Silkscreen lines should be 0.12mm instead of 0.1mm

Will fix.

> the name says "Pad_0.5mm" which I believe should be Pad_0.35mm"

Will fix.

> I can see there are 2 dominant styles for pin 1 indicators. Perhaps one should be picked as the main one

Sure, just let me know what to do.

> according to this lookup table http://www.topline.tv/SMD_vrs_NSMD.html a soldermask opening of 20 mil (0,508mm) is recommended. If you add a footprint solder mask clearance of 0,075mm I believe it should be fine

As I mentioned above, I went back and forth on this - so you want the explicit value to be set, rather than 0 as the KLC specifies?

awygle:
The above three issues have been corrected.

Shackmeister:
thanks for the contribution :)

PR #695, opened 2018-06-25
added TFBGA144 package

nerdyscout:
unfortunately I do not have a datasheet which is publicly avialable.

grafik


Ratfink:
Without a drawing of the package or footprint, there's no way for us to verify that this is correct (or even what parts use it).  If you really can't provide any such reference, I think we'll have no choice but to close this PR.

nerdyscout:
yes, I understand. Unfortunately I have not any alternate part using the same footprint.

PR #697, opened 2018-06-25
Create SOT-227 footprint

evanshultz:
I couldn't find any reference footprint online. I did look at several datasheets and they were all quite close.

I used https://www.vishay.com/docs/95423/sot227g2.pdf as the primary datasheet.

Here are others I looked at too:
- http://www.galco.com/techdoc/vish/ufb200fa20_ds.pdf
- https://www.microsemi.com/document-portal/doc_view/125317-mounting-considerations-for-the-sot-227-package
- http://www.icbanq.com/icbank_data/semi_package/sot227_dim.pdf
- https://www.ixys.com/Documents/AppNotes/IXAN0073.pdf

Drill holes are 4.2mm to accomodate M4 screws (nominally 4mm but see 3.85mm in the last link above). And the annular ring is 8mm to account for the max tolerance of the specified screw head in the same document.

IPC 7351B section 3.1.2 states that land pattern tolerances are to be done with max size leads. Since none of the lead styles in 7351B seem appropriate, I added the Level B fab allowance from Table 9-1 of IPC 2221A of 0.25mm to the max pin size to get the pad dimensions. If there are other suggestions about how to design the drill and pads for this footprint please share.

Pin numbers were taken from https://www.vishay.com/docs/95793/vs-fc420sa10.pdf.

Other dimensions not given were taken from the attached Vishay 3D model. I believe this was only the corner bevel of 2mm (measured 1.97mm in the model) and the distance between the centers of the pins of 3.7mm.

I chose this library because, even though the package does not have leads that go through the board, the footprint does have through holes and screws (which do go through the board) are required to use this package.

image

image

------------


evanshultz:
image

evanshultz:
Hmm. Right, need bottom copper around the screw head. A custom pad shape can't do that either (AFAIK), so I guess I do need two pads for each pin number.

Would it be best to break up the pads by top and bottom pads instead of up by SMT and THT?

poeschlr:
The THT pad already creates copper where the screw head will meat the pcb (or do i misunderstand the way this is mounted? Is the screenshot out of date?)
Is this package even intended to be mounted directly on a PCB? (We might want to add some information about that fact to the documentation field of the footprint.)

Is it intended that this will be soldered or will it simply connect via the screws? 
If you want the top pads to be soldered then i would increase them towards the outside as i would guess such a large package will be hand-soldered.

For the drill size i would suggest to use mechanical standards instead of electrical ones. (There will be a screw going through them.)
For M4 the typical drill size would be 4.3mm (Tolerance H12 -> fine), 4.5mm (Tolerance H13 -> medium) or 4.8mm (tolerance H14 -> coarse). These dimensions come from DIN EN 20273 or ISO 273-1979 (http://www.metricmcc.com/catalog/ch10/10-1040.pdf)

evanshultz:
Yes, it is intended to be mounted on a PCB and I've only seen it mounted using the 4 screw lugs (never soldered). I think the use of this package is well-understood in the field where it is used.

4.5mm THTs is fine with me. With the holes enlarged is this OK to merge?

image

poeschlr:
Yes looks ok to me. Thanks

PR #701, opened 2018-06-26
Texas VQFN RHL

jneiva08:
Datasheet: http://www.ti.com/lit/ds/symlink/bq51050b.pdf (pag.40)

Screenshot:
texas

Later will add symbol

------------


poeschlr:
Is this the same footprint as https://github.com/KiCad/kicad-footprints/pull/454?

jneiva08:
it seems to be the same

jneiva08:
forgot to add 3d
3d

poeschlr:
As this one includes the Thermal vias it should have a suffix _ThermalVias

jneiva08:
Renamed footprint to have "_ThermalVias" on the name

poeschlr:
Paste pads (the ones without a pin number) should not have copper enabled.

I would suggest to change the pad types to rounded rectangle with 25% round radius (maximum radius 0.25mm.)

It might also be a good idea to change pad 21 into a polygon pad (not necessary for merging this.)
And there is a small mistake with the layout of pad 1 (the blue dimension should be 0.4 instead of 0.405)

screenshot from 2018-07-28 01-51-50

jneiva08:
* Changed the pads to be rounded to be near 0.05,
* Fix small mistake to be 0.4mm instead 0.405,
* Removed F.CU from F.Paste pads.

vqfn

@poeschlr 
Where can find maual for make polygon pad's?

poeschlr:
Looks good, thanks (sorry for the delay in merging, i kind of forgot about that one)

PR #707, opened 2018-06-28
Fix some further mistakes for two terminal smd

poeschlr:
- Fuse 2920: terminal length was entered as terminal spacing (Resulted
in too little pad to pad spacing)
- Resitor 1218: body width max was 1mm too large
- 2816: Increased tolerance range for terminal length to fit more
devices.
- Tantal 3528: wrong tolerance range for terminal length
- Tantal 3216 and 6032-28 had wrong tolerance ranges but their effect is
hidden by rounding (meaning no change in the footprint)


evanshultz:
All looks good to me. Can you link to the script update and I'll merge.

poeschlr:
https://github.com/pointhi/kicad-footprint-generator/pull/125/commits/49c01ce9ae50e04f458aa19e780542fefa267e77

PR #708, opened 2018-06-28
created GB042-34S-H10

nerdyscout:
Hi,

this is the socket of an PCB to PCB connector, i will commit the corresponding plug soon.

grafik

datasheet: http://www.lsmtron.com/pdf/Connector&Antenna_catalog.PDF page8

I wasn't completly sure in which connector package i should push. For me the part does somehow look like an FPC/FFC, but according the datasheet it isn't. Do you want to have it somewhere else?

best regards,
ns


poeschlr:
From the screenshot i can already tell you that it should be rotated 90 degree clockwise such that pin 1 is on the top right corner.

poeschlr:
I now had a bit more time to take a closer look.
The pad width is wrong (should be 0.23 not 0.17)

The outline on the fab layer should represent the exact body outline of the part.
The silk outline should be fully outside the part body (meaning it will result outside the correctly drawn fab outline. Typically we use an offset of somewhere between 0.1 and 0.15mm between the body dimension and silk artwork.)
The courtyard rectangle should then be offset from the body outline (=fab outline) or pads (whatever results in the larger area. Meaning if the pads are outside the body then the pads are used in the respective direction)

More details see dimensioned drawing
conn

nerdyscout:
updated version:

image

Issue #711, opened 2018-06-28
0805 rounded rectangle pad change


stambaughw:
The recent change to rounded rectangular pads for the SMT caps and resistors made the gap between the pads on the 0805 footprints less than the 0603 footprint.  I check a bunch of datasheets and other resources and in none of them can I find where the recommended gap between pads of the 0805 footprint is less than the 0603 footprint.  The attached image shows an 0603, 0805, and 1206 resistor (left) and capacitor (right) footprints for comparison purposes.  If this is correct, would someone please attach the document that was used to generate these footprints so I can understand why this occurs.
![resistor-capacitor-footprints](https://user-images.githubusercontent.com/5891750/42044289-9aa83c00-7ac6-11e8-811a-cec373d29c07.png)

poeschlr:
I looked into it a bit more. It is even easier to understand than what i wrote above.
For both 0805 and 0603 the terminal to terminal distance is listed in IPC-SM-782. (As written in my previous response i checked these against 20 random resistors and capacitors available from farnell.)

Smin is 0.55 fro 0805 and 0.7 for 0603. If we ignore PCB manufacturing tolerances the pad to pad clearance (Gmin) calculates directly from Smin. (In IPC-7351B the heel fillet is given as 0 meaning Gmin will be the same as Smin.)

If we respect manufacturing tolerances (0.1 fabrication tolerance and 0.05 placement as suggested in IPC-7351B) we get the exact result as the current footprints have.

stambaughw:
On 6/28/2018 12:26 PM, Rene Pöschl wrote:
> I already answered to this over on the mailing list
> <https://lists.launchpad.net/kicad-lib-committers/msg00549.html>.
I guess I should subscribe to the library committers mailing list before
I post a message to it ;)

> TlDr: The reason behind this is the tolerance range of the 0805 is much
> larger then the one for the 0603.
> 
> For reference my answer on the mailing list
> 
>     This is because 0805 has larger tolerance ranges when compared to 0603.
> 
>     For 0603 the tolerances are body length 0.2mm, "lead" length 0.25mm
>     For 0805 they are body length 0.3mm, "lead" length 0.5mm
> 
>     I verified the tolerance ranges by checking 20 random resistors on
>     farnell. For both 0603 the components fall into the tolerance ranges
>     given by this ipc document. (I even found some 0805 parts that would
>     require increasing the tolerance ranges even further.)
> 
[...]

PR #712, opened 2018-06-29
Better fitting dimensions for 0805

poeschlr:
Fix for #711 (reported by @stambaughw
Originally the dimensions for 0805 where taken from IPC-SM-782. These
measurements seem to no longer fit the range of parts available right
now. An arbitrary selection of 15 different parts was used to determina
a better fitting footprint.
The research can be found under https://docs.google.com/spreadsheets/d/1BsfQQcO9C6DZCsRaXUlFlo91Tg2WpOkGARC1WS5S8t0/edit?usp=sharing

---

Script change: https://github.com/pointhi/kicad-footprint-generator/pull/125/commits/31e88bac09c6a017f1e8a3aab38ce4a71b6dca80


poeschlr:
I can not see different pad placements in the old footprints. They all have the pads on +/-0.85 with size 1.15x1.4 (The handsolder and castelated variants are of course different from this.)

---

As i mentioned in the original issue, i will do some research later on. If i find that indeed caps and resistors (for example) are different i will correct for that. (I do not expect a large enough difference to merit immediate action.)

PR #716, opened 2018-07-01
Convert some pads of jumpers to polygons instead of combination of normal pads

pointhi:
In KiCad 4 there were no polygon pads which resulted in some overlapping pad constructs. This is no longer required, as well as the old system causes some nasty problems like dupliation of pin numbers and unconnected warning. This PR solves the issue for the jumpers by converting some of the pads to polygon pads:

kicad_polygon_jumper

------------


Ratfink:
Looks great, thanks for fixing up these footprints!

evanshultz:
Sorry to bring this back. Is https://lists.launchpad.net/kicad-developers/msg36445.html resolved with these new footprints? Has anyone checked? (I'm busy now but will try to check next week if there is no response.)

matthijskooijman:
@pointhi, wouldn't it be possible to use a graphic rectangle on the copper layer? A quick test (adding one to the open solder jumper to connect two pads) shows this does not give any DRC errors (tested on 5.0.2 and a recent nightly).

PR #717, opened 2018-07-02
UQFN-10 for USB3740

jneiva08:
The footprint were created by [script](https://github.com/pointhi/kicad-footprint-generator/tree/master/scripts/Housings_DFN_QFN) with custom yml. But needed to retouch bay hand because the horizontal pad was bigger.

UQFN-10_1.3x1.8mm_P0.4mm
ufn_p0 4

UQFN-10_1.6x2.1mm_P0.5mm
ufn_p0 5

This footprint is for PR [#716](https://github.com/KiCad/kicad-symbols/pull/716)

------------


poeschlr:
I would suggest to change the pad type to rounded rectangle with 25% round radius ratio. (New IPC suggestion. I am currently working on getting this lib to that standard)

The chamfer size for the pin 1 mark is defined as 25% of the shortest side with a maximum of 1mm. So it would be better to reduce its size here. (Just discovered that the legacy script has an absolute chamfer size of 1mm. Fixed that in https://github.com/pointhi/kicad-footprint-generator/pull/149/commits/21f64005ea06f343d70c5b91c748b2b796a5f1c7)

Otherwise the footprint matches the datasheet
screenshot from 2018-07-28 09-08-42

poeschlr:
Ok you asked me if my script would support this. Well i looked into it again and found it was quite easy to implement. So i did that and made a pull request with the script generated version of these: https://github.com/KiCad/kicad-footprints/pull/790

Issue #540, opened 2018-04-30
Missing footprints for MCU Texas 


poeschlr:
Footprints missing for https://github.com/KiCad/kicad-symbols/pull/560

 - [ ] 'MSP430F1101AIDGV': http://www.ti.com/lit/ds/symlink/msp430f1101a.pdf
Package_SO:TVSOP-20_4.4x5mm_P0.4mm
 - [ ] 'MSP430G2444IDA38': http://www.ti.com/lit/ds/symlink/msp430g2444.pdf
Package_SO:TSSOP-38_6.1x12.5mm_P0.65mm
 - [x] 'MSP430F1101AIRGE': http://www.ti.com/lit/ds/symlink/msp430f1101a.pdf
Package_DFN_QFN:VQFN-24-1EP_4x4mm_P0.5mm_EP2.6x2.6mm
 - [x] 'MSP430G2001IRSA16': http://www.ti.com/lit/ds/symlink/msp430g2001.pdf
Package_DFN_QFN:Texas_S-PVQFN-N16_EP2.7x2.7mm
 - [x] 'MSP430FR5720IRGE': http://www.ti.com/lit/ds/symlink/msp430fr5720.pdf
Package_DFN_QFN:Texas_S-PVQFN-N24_EP2.25x2.25mm
 - [x] 'MSP430F1122IRHB': http://www.ti.com/lit/ds/symlink/msp430f1122.pdf
Package_DFN_QFN:Texas_S-PVQFN-N32_EP3.55x3.55mm
 - [x] 'MSP430F5232IRGZ': http://www.ti.com/lit/ds/symlink/msp430f5232.pdf
Package_DFN_QFN:Texas_S-PVQFN-N48_EP5.2x5.2mm
 - [x] 'MSP430F5217IRGC': http://www.ti.com/lit/ds/symlink/msp430f5217.pdf
Package_DFN_QFN:Texas_S-PVQFN-N64_EP4.4x4.4mm
 - [ ] 'MSP430F2232IYFF': http://www.ti.com/lit/ds/symlink/msp430f2232.pdf
Package_BGA:Texas_RXBGA-49_3.33x3.488mm_Layout7x7_P0.4mm
 - [ ] 'MSP430F5217IYFF': http://www.ti.com/lit/ds/symlink/msp430f5217.pdf
Package_BGA:Texas_RXBGA-64_3.415x3.535mm_Layout8x8_P0.4mm
[...]

PR #718, opened 2018-07-03
Add missing MSP430 footprints

evanshultz:
Identified and symbols updated for these at https://github.com/KiCad/kicad-symbols/pull/710.

TSSOP:
http://www.ti.com/lit/ds/symlink/msp430g2744.pdf

DSBGAs:
http://www.ti.com/lit/ds/slas718g/slas718g.pdf
http://www.ti.com/lit/ds/symlink/msp430f2234.pdf
http://www.ti.com/lit/an/snva009ag/snva009ag.pdf

MicroStar Junior:
http://www.ti.com/lit/ml/mpbg674/mpbg674.pdf
http://www.ti.com/lit/wp/ssyz015b/ssyz015b.pdf

image

------------


evanshultz:
Merging this will resolve https://github.com/KiCad/kicad-footprints/issues/540.

Ratfink:
The courtyard clearance on the BGA footprints is wrong (should be 1 mm).  Also, the F.Fab outline bevel at the A1 corner is the wrong size.  It may be easier to create these using my [BGA footprint generator script](https://github.com/pointhi/kicad-footprint-generator/tree/master/scripts/Package_BGA).

Also, from page 10 of the MicroStar reference guide, the mask openings should be 0.30 mm diameter, but you have 0.35 mm.

evanshultz:
Good catches. All fixed. Thanks!

Ratfink:
* Your fix accidentally left the courtyard for Texas_DSBGA-64_3.415x3.535mm_Layout8x8_P0.4mm open (top and bottom lines aren't long enough on the right side).
* In Texas_DSBGA-49_3.33x3.488mm_Layout7x7_P0.4mm, the top and bottom courtyard lines were rounded down instead of up (should be Y=±2.75, you have ±2.74).
* Corner bevel size is still too small for the DSBGA footprints.

evanshultz:
- My, that's sad. Fixed.
- This is actually fine. The Fab lines are 0.996mm away, and at https://github.com/KiCad/kicad-footprints/pull/292#issuecomment-360926500 Rene accepted rounding down. But I pushed them out.
- Yeah, I wasn't too precise on that. Increased bevel to specified amount with a bit or rounding. KLC should probably say `1mm or 25% of the smallest package side` since not all BGAs are square.

I just finished these manually but I'll look at the script. Probably should be using it instead. Hopefully it can handle missing pads.

I hope this is it. Sorry for the mistakes last commit. Let me know if you see anything else.

Ratfink:
Looks good now, thanks a lot for adding these footprints!  I'll take a look at the corresponding symbol PR next.

evanshultz:
@Ratfink 
I added the following the bga.yml:
```
Texas_DSBGA-49_3.33x3.488mm_Layout7x7_P0.4mm:
  description: "Texas Instruments, DSBGA, 3.33x3.488x0.625mm, 49 ball 7x7 area grid, NSMD pad definition (http://www.ti.com/lit/ds/symlink/msp430f2234.pdf, http://www.ti.com/lit/an/snva009ag/snva009ag.pdf)"
  pkg_width: 3.33
  pkg_height: 3.488
  pitch: 0.4
  pad_diameter: 0.225
  layout_x: 7
  layout_y: 7
Texas_DSBGA-64_3.415x3.535mm_Layout8x8_P0.4mm:
  description: "Texas Instruments, DSBGA, 3.415x3.535x0.625mm, 64 ball 8x8 area grid, NSMD pad definition (http://www.ti.com/lit/ds/slas718g/slas718g.pdf, http://www.ti.com/lit/an/snva009ag/snva009ag.pdf)"
  pkg_width: 3.415
  pkg_height: 3.535
  pitch: 0.4
  pad_diameter: 0.225
  layout_x: 8
  layout_y: 8
Texas_MicroStar_Junior_BGA-113_7.0x7.0mm_Layout12x12_P0.5mm:
  description: "Texas Instruments, BGA Microstar Junior, 7x7mm, 113 ball 12x12 grid, NSMD pad definition (http://www.ti.com/lit/ml/mpbg674/mpbg674.pdf, http://www.ti.com/lit/wp/ssyz015b/ssyz015b.pdf)"
  pkg_width: 7
  pkg_height: 7
  pitch: 0.5
  pad_diameter: 0.25
  layout_x: 12
  layout_y: 12
  row_skips: [[], [], [[4, 11]], [3, 10], [3, 10], [3, 6, 7, 10], [3, 6, 7, 10], [3, 10], [3, 10], [[3, 11]], [], []]
```

I find a couple issues/questions (haven't looked at the code, just comparing with my footprints):
1. No support for global or per-pad mask/paste clearance.
1. It appears the code places the floor(layoutX) and floor(layoutY) row/col on the centerlines. It works OK for my 7x7 BGA because there is an odd number of rows/cols, but the 8x8 and 12x12 layouts don't work because those should have the centerlines between a row/col. Does that make sense?
1. The ref des silk and fab value text is within the courtyard. Isn't it common (even if not specified in KLC) to place that outside the courtyard? Is there a guideline on this that I don't recall?

Here is the 12x12 BGA showing these 3 items (compare with screenshot above):
image

evanshultz:
See a couple updates above.

Also:
4. Why is the silk offset hard-coded to 0.125mm? Could you use \scripts\tools\global_config_files\config_KLCv3.0.yaml to import all configuration options instead of using magic numbers? (We should probably have those config files higher in the script hierarchy so it's obvious and shareable by all scripts.)
5. Lines 32-43 could be a combined using the .format() method in a few ways.

Ratfink:
1. That's largely a limitation of the kicad-footprint-generator script, at least at the time I wrote the BGA script.  I think Rene's big unmerged PR (https://github.com/pointhi/kicad-footprint-generator/pull/125) improves the situation, so once that's merged I could revisit that part of the script.
2. I believe that's due to differences between Python 2 and 3.  It looks like changing lines 74 and 83 to use floating-point division would make Python 2 behave the same as 3 here.
3. I never knew of any requirement to place those fields outside of the courtyard.  It seems unnecessary to do so, given that the courtyard is only for placement and is never displayed outside of Pcbnew.  Rene must agree since he approved the script's output (#516, #522).
4. Oh yes, I could use that configuration file.  Making that higher would make it more discoverable for folks writing new scripts.
5. Sure enough, that would be an improvement.

PR #720, opened 2018-07-03
added part: Wells TSOP Socket 48pin

nerdyscout:
Hi,

this is an 48pin TSOP Socket from Wells:

http://www.farnell.com/cad/316865.pdf?_ga=2.37208032.177107060.1530611323-249019997.1498114824

updated picture
grafik

------------


poeschlr:
The annular ring of the pads is quite small (0.15mm). I would suggest to use elongated pads (oval) to make soldering easier while still having enough clearance to neighboring pads.

The courtyard clearance should be 0.25mm
Silk should be outside of the part body.
Line width should be 0.1mm for F.Fab and 0.12 mm for F.Silk.

Drill size for the pads should be 0.7mm is 0.5 right now.
The outermost pad rows have the wrong x position.

screenshot from 2018-08-27 09-12-50

nerdyscout:
updated screenshot
image

i hope i got all suggestions resolved.
drill holes are now even bigger than pins to fit easier

nerdyscout:
does the marking for pin1 on F.Silk has to be within the courtyard? with 0.25mm it is outside now, seems pretty tight to me.

poeschlr:
The marking does not need to be within the courtyard but it can definitely be nearer to the outline compared to now. You could also use a different type of marking. (Example leave out half of one line near the corner that is nearest to pin 1)

nerdyscout:
ups. something strange happend...

Misca1234:
Just a correction for my comment above, it should be 
"Normally you should provide a url NOT to a re seller, "

Thx

PR #647, opened 2018-06-09
Esp32 wroom

Misca1234:
These are the foot prints used by Espressif's RF modules

ESP-WROOM-02 (existed before but modified)
https://www.espressif.com/sites/default/files/documentation/0c-esp-wroom-02_datasheet_en.pdf

ESP-WROOM-32
https://www.espressif.com/sites/default/files/documentation/esp32-wroom-32_datasheet_en.pdf

ESP-WROOM-32U
https://www.espressif.com/sites/default/files/documentation/esp32-wroom-32d_esp32-wroom-32u_datasheet_en.pdf

ESP-WROOM-32D
https://www.espressif.com/sites/default/files/documentation/esp32-wroom-32d_esp32-wroom-32u_datasheet_en.pdf

The keep out area is described in
https://www.espressif.com/sites/default/files/documentation/esp-wroom-02_pcb_design_and_module_placement_guide_0.pdf

(I can not figure out why travis complain about 
  Footprint anchor is not located at center of footprint
     - Footprint center calculated as (0.0,0.775)mm
for ESP-WROOM-32 and ESP-WROOM-32U, but I guess it is OK)

ESP-WROOM-02
bild

bild

bild

ESP-WROOM-32
bild

bild

bild

ESP-WROOM-32U
bild

bild

bild

ESP-WROOM-32D
bild

bild

bild

------------

Thanks for creating a pull request to contribute to the KiCad libraries! To speed up integration of your PR, please check the following items:

- [ ] Provide a URL to a datasheet for the footprint(s) you are contributing
- [ ] An example screenshot image is very helpful 
- [ ] If there are matching symbol or 3D model pull requests, provide link(s) as appropriate
- [ ] Check the output of the Travis automated check scripts - fix any errors as required

PR #722, opened 2018-07-03
Add Footprint for ESP32-WROOM-32 RF module

HubertD:
Datasheet: https://www.espressif.com/sites/default/files/documentation/esp32-wroom-32_datasheet_en.pdf
esp32-wroom-32

------------


HubertD:
i think the travis KLC F6.2 warning should be ignored, since it refers to the center of the pads, not the center of the module?

herostrat:
Hey Hubert,

have a look at: https://github.com/KiCad/kicad-footprints/pull/647
It also tries to push the footprint.
You two could probably join forces, e.g. by reviewing it.

Otherwise: the keepout of your footprint is too small, the reference specifies a 5mm wide block around the antenna.

HubertD:
hrm, i did search for PRs before starting this footprint, but did not find any.
I suspect I was searching for symbols first, and as I found none, concluded there would be no Footprint-PRs either.
Anyways, closing this in favour for #647

PR #724, opened 2018-07-04
Adding Texas_R-PWSON-N12_EP0.4x2mm

ki5libs:
Adding Texas_R-PWSON-N12_EP0.4x2mm
[Datasheet](http://www.ti.com/lit/ds/symlink/tpd6f003.pdf)
wson-12-1ep_1 35x2 5mm_p0 4mm_ep0 4x2mm
[3D package PR](https://github.com/KiCad/kicad-packages3D/pull/333)
------------


poeschlr:
For other specialized TI packages we use the TI naming convention. So this one should be named `Texas_R-PWSON-N12_EP0.4x2mm` to fit that scheme as well.

The courtyard clearance in x direction is too small. (should be 0.25 is 0.125)
screenshot from 2018-07-28 01-10-39

poeschlr:
Another thing: could you convert the pads to rounded rectangle pads with 25% round radius ratio? (would fit newer industry standards and kicad finally supports this)

ki5libs:
It's changed. Is it right to make EP as rounded rectangle too?

poeschlr:
Yes as far as i can tell, all pads can be rounded rectangle with the rule 25% radius with max radius 0.25mm (even paste only pads can be set that way.)

ki5libs:
All pads are rounded for a long time, please review and merge if everything else is OK...

poeschlr:
I missed the fact that the paste layer is included in main pad. please split it in two with coverage between 50 and 80% (you could use the suggested sizes given in the TI datasheet or calculate them your self)

ki5libs:
@poeschlr I split the earth pad according to datasheet, but not sure that did it right way. Could you check it, please?

poeschlr:
There should still be one large copper pad. That pad should have pad number 13, top copper and mask enabled but paste disabled.

On top of that add your paste pads. These should have no pad number, no copper, mask disabled, paste enabled.
Look at any other footprint with an exposed pad for inspiration. (One example would be `DFN-6-1EP_3x3mm_P1mm_EP1.5x2.4mm`)

poeschlr:
Could you convert the paste pads to rounded rectangle as well? Then this would be finished. (25% round rect radius with maximum of 0.25mm)

ki5libs:
Done

poeschlr:
Thanks

PR #728, opened 2018-07-04
Add footprint for TE MQS 936287-2

stephanhu:
datasheet link: http://www.te.com/commerce/DocumentDelivery/DDEController?Action=showdoc&DocId=Customer+Drawing%7F936287%7FA2%7Fpdf%7FEnglish%7FENG_CD_936287_A2.pdf

3D model: http://www.te.com/commerce/DocumentDelivery/DDEController?Action=showdoc&DocId=Customer+View+Model%7FCVM_936287-2%7FA%7F3d_stp.zip%7FEnglish%7FENG_CVM_CVM_936287-2_A_c-936287-2-a.3d_stp.zip%7F936287-2

screenshot_te_mqs

------------


poeschlr:
The silk outline is missing.
The outline on the fab layer should be closed.
Add a reference to the correct 3d model (prepare the footprint such that it works with a 3d model. You are not required to provide a 3d model file.)

I did not yet check the dimensions of the footprint.

Misca1234:
@stephanhu 
please see comments above from Rene

stephanhu:
@poeschlr  I've added the silkscreen and corrected the f.fab layer as well as provided a path to the 3d model in the footprint

poeschlr:
Pin 1 markers on silk and fab are missing.
The silk layer outline should be outside the fab outline (or more precisely: outside the area covered by the part)

The 3d settings must be: scale 1, rotation and offset 0 (all 3 axis)
The file ending for the 3d path must be wrl (step export automatically changes the file ending)
The file name of the 3d model must match the footprint name.

---

The outline on the fab layer must match the nominal part dimensions. These are however not really given in the datsheet you linked so we need to trust that the 3d model is correct and drawn after nominal dimensions.
Right now your footprint does not match the dimensions i measure from the 3d model. If you can provide a document that shows your fab outline is correct then this should be ok. Otherwise make sure you match the dimensions given in my screenshot.

---

The courtyard clearance must be 0.5mm (rounded to nearest 0.01mm grid) all around. (It can be a simple rectangle like you drew. But the clearance must be corrected.)

---

It might be a good idea to change the pads from circular to oval. (slightly reduce the x dimension. slightly increase the y dimension) This would allow for more clearance between the pads (Decreasing the annular ring in x direction down to 0.27mm would result in 0.8mm clearance which should be enough for getting a trace through.)

---

For reference the dimensioned drawing. (Of both the footprint and the 3d model) Bright pink means wrong value. violet means ok/reference.

te svg

PR #732, opened 2018-07-05
Added ST_HLGA-10... for LPS25HB

Ratfink:
ST_HLGA-10_2.5x2.5mm_P0.6mm_LayoutBorder3x2y, for the LPS25HB pressure sensor.  Fixes yet another part of https://github.com/KiCad/kicad-symbols/issues/520.

Datasheet: https://www.st.com/resource/en/datasheet/lps25hb.pdf#page=46

screenshot from 2018-07-05 19-35-42

Symbol PR: https://github.com/KiCad/kicad-symbols/pull/735

Created using @poeschlr's Package_LGA script in https://github.com/pointhi/kicad-footprint-generator/pull/125, using this config (please add this to the PR when you're back, Rene!):

```yaml
ST_HLGA-10_2.5x2.5mm_P0.6mm:
  device_type: 'HLGA'
  manufacturer: 'ST'
  #part_number: 'mpn'
  size_source: 'https://www.st.com/resource/en/datasheet/lps25hb.pdf#page=46'
  body_size_x_min: 2.4
  body_size_x: 2.5
  body_size_x_max: 2.6
  body_size_y_min: 2.4
  body_size_y: 2.5
  body_size_y_max: 2.6
  lead_width_min: 0.25
  lead_width_max: 0.35
  lead_to_edge_min: 0.08
  lead_to_edge_max: 0.13
  lead_len_min: 0.40
  lead_len_max: 0.50
  pitch: 0.6
  num_pins_x: 3
  num_pins_y: 2
  #chamfer_edge_pins: 0.25
  #pin_count_grid:
  #pad_length_addition: 0.5
  #suffix: '_ClockwisePinNumbering'
```

------------


evanshultz:
- Should the name be `ST_HLGAL...1 since the datasheet shows the `L`?
- There is a double space in the description between the first two words (`ST  HLGA`).
- The X location of pin 1 is -0.9125, but the datasheet looks like it should be -0.92mm.

Ratfink:
* `ST_HLGA-10L…`, actually.  Good catch.  Doesn't look like the script supports that though.
* Script bug.
* I don't know what's going on there.  I think I entered all the information correctly based on the datasheet.

Maybe this should wait until Rene gets back (i.e. after the release), so we can be sure there's no problem with the script with regards to pad placement.

poeschlr:
The pads in the script are calculated using ipc. The datasheet might use a different way to get their pad sizes. So if there is a difference in the result between the script generated footprint and the datasheet the decision is up to you if you want to use the scripted one.

evanshultz:
I'm talking about pad placement (location), not pad size.

poeschlr:
The location and pad size are linked. (Or more precisely the heel and toe size will result in a pad size plus pad position.)

evanshultz:
OK, I see how that could happen. And the pads on this footprint cover up the pins so it should work. However, on all the other LGA parts the part's pin locations were exactly at the pad locations and it make me think twice since only this one was different. So... merge it?

poeschlr:
The drawing by st is made completely different to what other manufacturers use. (Or more precisely: It is made in a way that i did not expect.)
Meaning the current version of the generator might use wrong tolerance calculations. I don't have the time right now to check if @Ratfink made the conversion correctly. 
For now we can merge it but maybe keep in mind that we might update the generator later on such that it can better deal with this dimensioning scheme.

---

For now i added the dimensions listed above to the generator in commit https://github.com/pointhi/kicad-footprint-generator/pull/125/commits/50b95fe60bf6a9f150d650d7aa8ac7d33bd70ede

PR #735, opened 2018-07-06
Fix mounting hole size in Arduino Uno footprints

### LINKED ISSUE #590: Module: Arduino mounting holes too small?
Some of the Arduino footprints in the Module library have mounting holes. These holes use a 1.78mm drill size, which seems rather small. For example: https://github.com/KiCad/kicad-footprints/blob/bb3185e1b563b7b3890cd754baaa7d6a8ac32410/Module.pretty/Arduino_UNO_R3_WithMountingHoles.kicad_mod#L96-L99

 Looking at an actual Arduino Uno r3, these have drill holes of around ⌀3mm. Previous (third-party) footprints I've been using also use around 3mm, the [huzza footprint](https://github.com/KiCad/kicad-footprints/blob/master/Module.pretty/Adafruit_HUZZAH_ESP8266_breakout_WithMountingHoles.kicad_mod) has 2.9mm.

@jkriege2, you originally submitted these footprints. Did you intentionally use such small mounting holes?

Ratfink:
Issue #590 pointed out that it was wrong before.  I looked at the actual Eagle source (from https://store.arduino.cc/arduino-uno-rev3), and found that the correct diameter is 3.2 mm.  This fixes the hole sizes.

Fixes #590.

------------


Shackmeister:
Thanks! :)

PR #738, opened 2018-07-09
Add Kingbright AAA3528ESGCT footprint

evanshultz:
Datasheet: http://www.kingbrightusa.com/images/catalog/SPEC/AAA3528ESGCT.pdf

Image:
image

I'd like help with the footprint name. https://www.digikey.com/products/en/optoelectronics/led-indication-discrete/105?k=aaa3528 lists a bunch of Kingbright MPNs that all start with `AAA3528` so that would seem an obvious choice since all of those parts could fit the footprint. The suffix varies by the LED colors. However, https://www.digikey.com/products/en/optoelectronics/leds-circuit-board-indicators-arrays-light-bars-bar-graphs/106?k=aaa3528 shows another pile of Kingbright MPNs that all start with the same 7 characters but are physically incompatible. Those also vary by LED colors. Should it stay like this, with the footprints all individualized, since the 3D model would have different LED colors anyway?

Note: there is not currently a symbol that fits this footprint, but I'll do that once the footprint is merged.

------------


herostrat:
Hey evanschultz,

researching into this a bit here is the official catalog:
http://www.kingbrightusa.com/kcpartno.asp?sltSearch=PN&txtPartNo=aaa352

They state in the product features that this a "Industry standard PLCC-4 package"

So I would, similar to other PLCC LEDs name it:
LED_Kingbright-PLCC-4_3.2mmx2.8mm or something like that (it is probably the same as the LED_Cree-PLCC4* one)

Speaking of other PLCC types, could it be that the naming is off there?

grafik

Shouldn't they be all have the "-" between PLCC and the number?

evanshultz:
@herostrat 
Thanks for looking and commenting!

That's the thing: it's not an industry standard! There are 4-pin PLCC packages in many shapes and sizes. So just a LED_PLCC-4 footprint wouldn't be descriptive enough.

Since the 3D model and footprint are matched 1:1 if the 3D model were to show LED colors we would need a unique footprint (to get the unique 3D model) for each MPN. Or if the actual component even with the same LED colors varied between vendors, a different 3D model would be required. So that's why I initially didn't generalize this footprint at all and went with the full MPN. Is that the best way?

(Note that we need not be so detailed with 3D models, but I'm thinking ahead and "out loud".)

Typically we do have a hyphen between the package type and pin count, but in the case of a manufacturer-specific footprint we would use their nomenclature. http://www.cree.com/led-components/media/documents/ds-CLV1A-FKB.pdf, for example, doesn't have a hyphen in "PLCC4". https://www.mouser.com/datasheet/2/678/V02-0479EN_DS_HSMx-A4xx-xxxxx_2018-04-09-908900.pdf does have a hyphen. https://cdn-shop.adafruit.com/product-files/3484/3484_Datasheet.pdf doesn't contain the text "PLCC" at all. Neither does https://cdn-shop.adafruit.com/datasheets/WS2812.pdf. So putting in a hyphen even if it's missing isn't a problem to me, but I'm not aware of this topic being brought up before. And you've definitely pointed out an issue where the manufacturer shows a hyphen and our footprint name doesn't have it.

poeschlr:
All other footprints in that lib have the prefix LED. Please add that to this one as well.

In what form are other parts that start with the same 7 number prefix incompatible with this one? If it is some part dimension we can simply include that dimension in the naming.
I am however prepared to merge with the full part number as well.

The footprint it self is ok. For reference the dimensioned drawing:
screenshot from 2018-07-25 09-35-29

evanshultz:
@poeschlr 
Thanks for looking!

Updated the footprint name.

I made a footprint for this part:
image

But the other parts I mentioned above look like this:
image

Totally different physically, yet they start with the same 7 characters. After those 7 characters, the second link has `AVU`. So that would work to separate the footprints physically, but I think we still would want individual footprints since the LED colors differ, right?

evanshultz:
@poeschlr 
Is this good to merge now? The footprint name is updated and using the full MPN allows for an accurate 3D model with LED colors so I'm OK with that as well since you are.

poeschlr:
Ok for me.

PR #742, opened 2018-07-11
Add TSOPII-54 Package

fvollmer:
The TSOP-II-54 Package is used by various SDRAM chips. For example the MT48LC8M8A2TG is available from the memory library, but the TSOPII-54 Package is missing.

It was kind of hard to find the best landing pattern. I hope I got all the dimensions right. :)

Datasheet: https://www.micron.com/~/media/documents/products/data-sheet/dram/64mb_x4x8x16_sdram.pdf, http://www.cypress.com/file/42846/download

Screenshot:
tssopii-54

There are currently no other related pull requests.

------------


poeschlr:
How did you arrive at that strange naming? (The -II part) The datasheet uses normal TSOP naming.
After the pin count should come a underscore followed by the body dimensions (not sure what the stuff is you have there)

The body size in y direction is not quite right (should be 22.22 not 22.365)
screenshot from 2018-07-24 14-50-19

fvollmer:
@poeschlr Thank you for your review!

I fixed:
- the y dimension on F.Fab to 22.22mm
- the body dimensions in the filename and the footprint value to `22.2x10.16mm` (not sure why I got this wrong)

The new name is now: `TSOP-II-54-22.2x10.16mm_P0.8mm`. The `-II` naming is also used in
- the datasheet. They use the name at the first page:
tsopii
- the jedec has the II in their specification for sdram modules (https://www.jedec.org/sites/default/files/docs/4_20_03R13.pdf).
- other SDRAM ICs e.g. http://www.issi.com/WW/pdf/42-45R-S_86400D-16320D-32160D.pdf
- the existing `TSOP-II-32_21.0x10.2mm_P1.27mm.kicad_mod` footprint
- the official TSOP - TYPE II specification: https://www.jedec.org/system/files/docs/DGuide4-15b.pdf

poeschlr:
You forgot to adjust the courtyard after fixing the body size:
screenshot from 2018-07-25 09-19-09

fvollmer:
Good catch!

I updated the footprint:
- Moved the silkscreen along the y Axis.  It should now have a clearance of at least 0.2mm to the nominal body dimensions and the pads, placed on a 0.1 grid
- Moved the silkscreen along the y Axis for a 0.25mm clearance to the footprint _including_ the silkscreen on a 0.01 grid

Would you mind to clarify the courtyard rules of the KLC?
- According to [F5.3](http://kicad-pcb.org/libraries/klc/F5.3/):
  > The component courtyard is defined as the smallest rectangular area that provides a minimum electrical and mechanical clearance around the combined component body and land pattern boundaries
  
  Does this include the silkscreen?

tssopii-54

poeschlr:
No silk is not part of the courtyard considerations. In fact only copper features and the nominal body of the part are. (As the fab layer represents the nominal body we can "simplify" by stating fab outline and pads.)

Silk does not really need much clearance to the body. (only to pads) It should just be fully outside the nominal body. (So we typically move it out from the fab line by 0.1 to 0.15mm. This is not really critical and can be done by eye.)

fvollmer:
Thank you for the clarification! I modified the courtyard and silkscreen according to your recommandations.

tsop-ii-54

poeschlr:
For the future: it is easier for us reviewers if you make new commits for every requested change. That way we can be sure what has changed since our last review. (We will then merge it via the squash merge feature of github that then converts your many commits into one single commit.)

---

The footprint looks good, thanks

fvollmer:
Thank you for the review process!

I wasn't aware of the squash merge feature. I'll do seperate commits in the future.

PR #743, opened 2018-07-11
Add 1206 reverse mount LED

fauxpark:
I have a handful of [these](https://www.digikey.com.au/product-detail/en/dialight/5976301607F/350-2310-1-ND/2224782) lying around from a previous project and saw the existing reverse mount 1206 footprint was removed. This footprint isn't scripted but it is just a slightly modified 1206 footprint. Perhaps it is possible to make the script spit it out.

The NPTH may need to be elongated slightly to accomodate for more "square" lenses such as in [this LED](https://www.digikey.com.au/product-detail/en/kingbright/APTR3216SURCK/754-1173-2-ND/1747572), or it could be a separate, Dialight-specific footprint.

fp
3d
fp2

------------


evanshultz:
Whoa!?!?!? The footprint was added at https://github.com/KiCad/kicad-footprints/pull/129, and removed with https://github.com/KiCad/kicad-footprints/pull/469 (I assume accidentally when other chip-size footprints were updated). Is that what you were seeing?

@poeschlr @Shackmeister 
Is that the right sequence of events and removing the reverse-mount LED footprint was an accident?

Do you want to add reverse mount LEDs to the script?

@fauxpark 
You just added the 1.8mm hole to the existing footprint, right? 1.8mm is fine for the LED bulb/lens being round and 1.5-1.6mm. However, it won't accomodate the Kingbright LED and the footprint name isn't super clear. If this is to be a generic footprint for many vendors I think we should do something like `LED_1206_3216Metric_ReverseMount_Hole1.8mm` (round hole) or `LED_1206_3216Metric_ReverseMount_Hole2.1x2.3mm` (square hole). This will also allow a more accurate 3D model to fit the footprint.

fauxpark:
Yeah, I'm guessing it was removed because it now didn't match the scripted footprints. I don't see any other reverse mount LED packages apart from some oddball sizes that aren't scripted anyway.

I also pulled back the inner sides of the pads slightly to accommodate the hole. An oval hole would fit both the circular and square lenses, but I can make another footprint with a square hole if need be.

poeschlr:
It was removed by accident. 
I did not notice that there is a manually created footprint right in the middle of my scripted ones. 

When i changed the pad size calculation i needed to delete the handsolder versions as they include the pad size in their name.
To make it easy for me i deleted everything between the smallest and largest standard chip resistor sizes and replaced it with the newly created ones. (I should have deleted only the handsolder versions. Sorry about that.)

fauxpark:
Speaking of which, I just noticed the _HandSolder diodes and LEDs look exactly the same as the normal footprints. SMD caps are not. Is that intentional?

poeschlr:
The Handsolder version has slightly longer pads. (The toe side is elongated.)

More detailed: right now the Zmax dimension (outside to outside between both pads) is increased by 0.35mm compared to the normal footprint. This means each pad is longer by 0.35/2 = 0.175mm.

Originally i determined this value by taking my own footprints that i regularly handsolder and looked how much larger their pads are when compared to the ipc suggestion. If somebody thinks this size increase is too small i can of course change that value. (I would not increase it by too much.)

fauxpark:
> The Handsolder version has slightly longer pads.

That's not what I'm seeing:

```diff
--- /Users/fauxpark/Git/kicad-footprints/Diode_SMD.pretty/D_0603_1608Metric.kicad_mod	Thu Jun 28 21:05:39 2018
+++ /Users/fauxpark/Git/kicad-footprints/Diode_SMD.pretty/D_0603_1608Metric_Pad0.88x0.95mm_HandSolder.kicad_mod	Thu Jun 28 21:05:39 2018
@@ -1,11 +1,11 @@
-(module D_0603_1608Metric (layer F.Cu) (tedit 5B301BBE)
+(module D_0603_1608Metric_Pad0.88x0.95mm_HandSolder (layer F.Cu) (tedit 5B301BBE)
   (descr "Diode SMD 0603 (1608 Metric), square (rectangular) end terminal, IPC_7351 nominal, (Body size source: http://www.tortai-tech.com/upload/download/2011102023233369053.pdf), generated with kicad-footprint-generator")
-  (tags diode)
+  (tags "diode handsolder")
   (attr smd)
   (fp_text reference REF** (at 0 -1.43) (layer F.SilkS)
     (effects (font (size 1 1) (thickness 0.15)))
   )
-  (fp_text value D_0603_1608Metric (at 0 1.43) (layer F.Fab)
+  (fp_text value D_0603_1608Metric_Pad0.88x0.95mm_HandSolder (at 0 1.43) (layer F.Fab)
     (effects (font (size 1 1) (thickness 0.15)))
   )
   (fp_line (start 0.8 -0.4) (end -0.5 -0.4) (layer F.Fab) (width 0.1))

```

poeschlr:
The LED_handsolder definition is missing the pad size increase. will fix it shortly.

evanshultz:
@fauxpark 
Back to the reverse mount LEDs...

I proposed a name change above to be more clear and extensible. Is that OK with you and can you rename these footprints?

fauxpark:
@evanshultz 
Done. I also decided to make the hole slightly wider to fit both lens shapes:

3d

The handful of manufacturer STEP files from Digikey that I tested against seem to fit very well.

poeschlr:
I will not script this.

evanshultz:
@fauxpark 
I think this should be fine. The inner pad clearance goes from 1.55mm to 2.15mm, but that's not a solderable surface on the LED. A ~0.5mm fillet can be formed on each end which should be OK. Thanks!

PR #745, opened 2018-07-14
Adding Transformer_Ethernet_HALO_TG111-MSC13

ki5libs:
Adding Transformer_Ethernet_HALO_TG111-MSC13
[Datasheet](https://www.haloelectronics.com/pdf/discrete-genesus.pdf)
transformer_ethernet_halo_tg111-msc13

[3D package PR](https://github.com/KiCad/kicad-packages3D/pull/341)

------------


Misca1234:
The F.Fab line at the short ends are at 7,43, which give the size of the trafo to 14,86
according to 
https://www.haloelectronics.com/pdf/discrete-genesus.pdf
it should be 14.22

evanshultz:
@ki5libs @Misca1234 
The newest IPC 7351C update is heading towards removing silk for pin 1 instead of adding double silk. This has been used in the library many places already. Since adding parts near the pin 1 silk of this transformer will eat away the silk and lose the pin 1 mark, silk removal is a good way to go in the future.

Misca1234:
So, as little silk as possible and no pin 1 marker if possible.
Otherwise little silk in the corners where pin 1 silk corner marker should look different from the other corners.

poeschlr:
I would not go with the full standard here.
The silk 1 marker as shown in this example was never part of any standard but kind of a kicad specialty.
In this case a line along pin 1 would be what the old standard suggests and what we typically use in other similar parts. (It might even be that the new standard suggests the same. IPC-7351c is not yet released as far as i am aware. So it could still change)

ki5libs:
So summarizing the discussion, silk removal as first pin mark is more preferable then double silk anyway. I got it and will do it this way in the future.

PR #753, opened 2018-07-17
QFN-28-1EP_4x5mm had the wrong ep size.

poeschlr:
The symbols who used this single footprint did require a different EP
size.

Symbol PRs (Split to reduce danger of manual merging):
 - https://github.com/KiCad/kicad-symbols/pull/764
 - https://github.com/KiCad/kicad-symbols/pull/763

---

For LM5175RHF:
Datasheet: http://www.ti.com/lit/ds/symlink/lm5175.pdf

Screenshot of the thermal via version:
screenshot from 2018-07-17 12-05-34

Script params:
``` yaml
QFN-28-1EP_4x5mm_P0.5mm_EP2.55x3.55mm:
  device_type: 'QFN'
  size_source: 'http://www.ti.com/lit/ds/symlink/lm5175.pdf (Page 37)'
  body_size_x:
    minimum: 3.9
    maximum: 4.1
  body_size_y:
    minimum: 4.9
    maximum: 5.1
  lead_width:
    minimum: 0.18
    maximum: 0.3
  lead_len:
    minimum: 0.3
    maximum: 0.5

  EP_size_x:
    nominal: 2.55
    tolerance: 0.1
  EP_size_y:
    nominal: 3.55
    tolerance: 0.1
  EP_num_paste_pads: [2, 3]

  thermal_vias:
    count: [3, 4]
    drill: 0.2
    paste_via_clearance: 0.1
    paste_avoid_via: False

  pitch: 0.5
  num_pins_x: 6
  num_pins_y: 8
```

---

For LTC3555:
Datasheet http://www.analog.com/media/en/technical-documentation/data-sheets/3555fe.pdf

Screenshot:
screenshot from 2018-07-17 12-05-12

Script params:
``` yaml
QFN-28-1EP_4x5mm_P0.5mm_EP2.65x3.65mm:
  device_type: 'QFN'
  size_source: 'http://www.analog.com/media/en/technical-documentation/data-sheets/3555fe.pdf (Page 32)'
  body_size_x:
    nominal: 4
    tolerance: 0.1
  body_size_y:
    nominal: 5
    tolerance: 0.1
  lead_width:
    nominal: 0.25
    tolerance: 0.05
  lead_len:
    nominal: 0.4
    tolerance: 0.1

  EP_size_x:
    nominal: 2.65
    tolerance: 0.1
  EP_size_y:
    nominal: 3.65
    tolerance: 0.1
  EP_num_paste_pads: [2, 3]
  heel_reduction: 0.05 #for relatively large EP pads (increase clearance)

  thermal_vias:
    count: [3, 4]
    drill: 0.2
    paste_via_clearance: 0.1
    paste_avoid_via: False

  pitch: 0.5
  num_pins_x: 6
  num_pins_y: 8
```


evanshultz:
I again am uncomfortable not using the footprint recommended by the vendor. I know you disagree, and I'm with you that IPC-derived footprints should be best, but I think that goes in the absence of any other information. If the vendor gives a good footprint and KiCad deviates from that footprint, and there are issues, I think KiCad is then at fault. The issue is knowing how good the vendor's footprint is. That being said, both of these footprints look good to me.

What is the method of using heel reduction? What criteria are used to determine this? Are you targeting 0.2mm clearance from thermal pad to outer pads? If so, wouldn't 0.025mm heel reduction work for the Analog footprint?

poeschlr:
The vendor has no suggested footprints in the datasheet. The appnote you originally found was completely unrelated! (or can you point to a document that disagrees with the information i have?)

poeschlr:
Sorry for my first response. I thought this was the PR for the microchip stuff.

---

I do not think there will be an issue with the footprints as IPC is the industry standard for a reason. (It is also not as critical as some people make it out to be. If somebody is really concerned about their yield they need to make a personalized footprint anyways. As manufacturing tolerances also play a role in the pad size calculations.)

The pad sizes are a bit different either because TI used different manufacturer tolerance settings (We use the default as suggested by IPC. Maybe TI assumes this is added by the pcb manufacturer. That workflow is also an option according to IPC. Or they use the older standards to arrive at the dimensions.)

The ep sizes agree with the manufacturer suggestion. The only thing a bit different is the number of thermal vias. (higher in my footprints as my script is not able to create the pattern suggested by TI.)

evanshultz:
What does VQFN mean? Apparently just the package height, from https://en.wikipedia.org/wiki/Quad_Flat_No-leads_package and https://e2e.ti.com/support/logic/f/151/t/267283?VQFN-and-QFN-package-for-SN74LVC08ARGYR. So for our general footprint, I think just `QFN` is fine. While I didn't check, to match the 3D model it could be `VQFN` but there may be many thicknesses of VQFN packages and adding all of them adds little value to me.

Yes, the significant difference in TI pad/paste size on the outer pads was one of my bigger concerns. Maybe because the TI datasheet shows "D" pads instead of square ones they use smaller pads? Or maybe the stencil height (TI is using 0.125mm thick here but I don't know the criteria used by IPC)? I'm not sure why they are so divergent but it wasn't a small difference.

I'm not too concerned about using this footprint for the Analog part. Having a slightly longer toe shouldn't be a problem and it's otherwise quite close.

Please also see my comments about providing pad size in the footprint names at https://github.com/KiCad/kicad-footprints/pull/752.

poeschlr:
Regarding pad sizes. I would not really change them other then maybe increasing the heel reduction for the TI footprint. (Would give a bit more clearance between the outer pads and the center pad.)

The paste stencil should not have anything to do with the pad size. (It would only influence the paste clearance set by the user. And the paste coverage used for the exposed pad. I think our goal of 50% to 80%  should fit stencil thicknesses of about 0.1mm. At least that was the value given in the appnotes i read. I guess such a large range would fit a large range of stencil thicknesses.)

evanshultz:
The TI footprint has 0.225mm clearance now, which should be OK according to IPC (and so it should match well with the rest of the footprint construction methodology). So I'm fine with leaving it alone.

If you want to rename to VQFN, go ahead. If not, let me know and I'll merge.

poeschlr:
Changed to VQFN

PR #755, opened 2018-07-17
Button_Switch_SMD: Add KSC7 button from CK switches

antoniovazquezblanco:
https://www.ckswitches.com/media/1973/ksc7.pdf

imagen

------------


antoniovazquezblanco:
BONUS: Added KSC6

https://www.ckswitches.com/media/1972/ksc6.pdf

imagen

evanshultz:
- I think we can ditch the `LFx` or use `xxx` since RoHS compliance is non functional.
- Silk can be offset by just 0.11mm but this is OK.
- If we use the dimensions they give as nominal (not min or max), pads should be 2.8mm wide on +/-2.9mm X location.
- KSC6 button circle should be 1.85mm radius.
- Courtyard offset should just be 0.25mm unless I misunderstand this requirement for switches. So yours are too big.

poeschlr:
I think these would get the name prefix:

`SW_push_1P1T_NO`

See: https://github.com/KiCad/kicad-symbols/issues/580#issuecomment-450632899

antoniovazquezblanco:
Performed requested changes!

evanshultz:
1. Should we have all lowercase in the footprint name? Not `Push`?
1. Is the trailing `xxx` required? Would that be important in the 3D model?
1. Pads should have X location of +/-2.9mm, right? Then courtyard X location can be +/-4.55m.
1. KSC6: Should the circle be 1.45mm in diameter, half of 2.9mm?

antoniovazquezblanco:
Done!

evanshultz:
Looks good. Thanks!

antoniovazquezblanco:
Thank you!

PR #756, opened 2018-07-17
Add Inolux IN-PI554FCH footprint

diegoherranz:
Started from LED_WS2812B_PLCC4_5.0x5.0mm_P3.2mm
but mark indicates pin 1 in this case whereas it
indicates pin 3 on the WS2812B.

New symbols at https://github.com/KiCad/kicad-symbols/pull/766

in-pi554fch_footprint

------------


diegoherranz:
Thanks, @Misca1234, for your comments. You're right.

All this originates from the fact that this symbol was derived from WS2812B and that LED has a mark on pin 3 (who knows why).

I forgot to move the diagonal line on F.Fab to pin 1. Now fixed.
I guess there was a text field saying "1" on Silkscreen to reiterate that that is pin 1 and not the one with the diagonal marking. Since this Inolux LED has a "sane" marking (that is, pin 1 marked), it is not necessary to reiterate which pin is pin 1.

Thanks!

Shackmeister:
I recall something about the marker on the actually component being on pin 3. If you chwck the history of the footprint you can probably fint the discussion

diegoherranz:
I've tightened up the silkscreen. Something like this?
inolux_updated

Regarding WS2812B, I think it is all correct regarding Pin 1 vs Pin 3 because of the marker on the device being on pin 3! See the [datasheet](http://www.world-semi.com/DownLoadFile/108).
We could potentially do some cleaning/improving on a few SMD LEDs by the look of it.

Thanks!

Misca1234:
@Shackmeister 

I am not so sure about that, for this one, checkout the data sheet
http://www.inolux-corp.com/datasheet/SMDLED/Addressable%20LED/IN-PI554FCH.pdf

bild

Misca1234:
@diegoherranz 

The F.Fab pin marker should be a cut corner, like this

bild

diegoherranz:
I have cut the corner as requested.
inolux_update2

Regarding pin 1 vs pin 3:
- Inolux LED (this PR): it has the marker on pin 1, and that is what's reflected here.
- WorldSemi WS2812B: it has the marker on pin 3 and that is what LED_WS2812B_PLCC4_5.0x5.0mm_P3.2mm reflects.

So I think it all looks OK in terms of pin 1 / pin 3 stuff. I guess some "style" improvements could be made on the WS2812B footprint (silkscreen, F.Fab...) but that's a separate story.

Thanks.

diegoherranz:
Thanks!

PR #761, opened 2018-07-18
Button_Switch_THT: Add CK OS102011MS2QN1 footprint

antoniovazquezblanco:
https://www.ckswitches.com/media/1428/os.pdf

imagen

------------


poeschlr:
@antoniovazquezblanco what is the state of this?

poeschlr:
Regarding fp name i think it would be something like:

`SW_Slide_1P2T_CK_OS102011MS2QN1`

I am however not so sure if we should have all these options for the type (rocker and slider would look the same on the symbol side). 
They would specify the footprint closer but might either require us to have more symbols or have multiple filters per symbol.

See: https://github.com/KiCad/kicad-symbols/issues/580

antoniovazquezblanco:
> According to IPC-2221A, the switch pads can be 1.35mm in diameter and the mounting holes can be 2.05mm. They don't have to be exactly this size, but since there's only 0.15mm between the pads I think we should go smaller to avoid shorting.

I did not get this one. What should I do? 🤔 

The rest is done!

evanshultz:
@antoniovazquezblanco 
I was saying that the distance between pads is only 0.15mm, which is very small. I'm suggesting we go with smaller pads than IPC would recommend to give more space and avoid accidental shorts. However, when I calculate now I get 0.25mm (2.1mm - 0.75mm - 1/1mm).

Here is the footprint now:
image

Looks good to me. Ready to merge?

antoniovazquezblanco:
Yes! Thank you very much!

PR #762, opened 2018-07-19
Update RGB LED (1 footprint) + add 2 new RGB LED footprints with larger, more spaced, easier-to-solder pads

ElectricRCAircraftGuy:
1. This is a minor update to the keywords and name of the RGB LED: from "LED_D5.0mm-4" to "LED_D5.0mm-4_RGB".
2. I added 2 new footprints with larger, easier-to solder pads:
LED_D5.0mm-4_RGB_Staggered_Pins and LED_D5.0mm-4_RGB_Wide_Pins

Original one with _RGB added to end of name:

image

New footprint 1:
image

New footprint 2:
image

------------


poeschlr:
Please sign the license agreement

ElectricRCAircraftGuy:
Done. I have signed the CLA.

ElectricRCAircraftGuy:
I'm guessing you want me to fix the Travic CI failed build stuff before you accept this pull request?

ElectricRCAircraftGuy:
What does this mean exactly?

ElectricRCAircraftGuy:
Also, why did Travis CI check my first 2 footprints but not the 3rd one?

poeschlr:
The official KiCad library has a set of rules that must be followed (within reason) for your contribution to be accepted. The rules are documented here: http://kicad-pcb.org/libraries/klc
Travis does check a lot of these rules but the scripts can not check everything so the final decision will always be made by one of the volunteers maintaining the library. (Most likely you will get some feedback about your contribution at some point in the future. Right now i am more worried about getting the lib ready for the version 5 release.)
So in short yes the travis errors should be fixed.

---

Regarding travis error about second ref des:
All footprint in the official kicad library have two reference designators. The main one is on the silk layer but we add a copy to the F.Fab layer. (The copy is added by adding a normal text field with content "%R" to the F.Fab layer. More details see: http://kicad-pcb.org/libraries/klc/F5.2/)

Look at other similar footprints in the lib to get an idea of what we expect.

---

I restarted the test script, maybe it will now check your third footprint as well.

ElectricRCAircraftGuy:
I just learned how to use `check_kicad_mod.py`, and I've spend a substantial amount of time (many hours) getting familiar with the KLC (http://kicad-pcb.org/libraries/klc/). I'll be making the fixes this week hopefully so you can accept this pull request.

ElectricRCAircraftGuy:
Okay, I just updated all 3 footprints to conform to KLC. The pull request is done and ready to be accepted! Here is the result of running `check_kicad_mod.py` on all 3:

> Checking footprint 'LED_D5.0mm-4_RGB':
>   Violating F7.5
>     Annular ring must be at least 0.15mm
>     Pad 1 x-dimension annular ring (0.085mm) is below minimum (0.15mm)
>     Pad 2 x-dimension annular ring (0.085mm) is below minimum (0.15mm)
>     Pad 3 x-dimension annular ring (0.085mm) is below minimum (0.15mm)
>     Pad 4 x-dimension annular ring (0.085mm) is below minimum (0.15mm)
> Checking footprint 'LED_D5.0mm-4_RGB_Staggered_Pins':
>   Violating F7.3
>     Pad 1 should be denoted by rectangular pad
>     Pad 1 should be rectangular
>      - Ignore for non-polarized devices
> Checking footprint 'LED_D5.0mm-4_RGB_Wide_Pins' - No errors

Explanations for the violations still present are as follows: 
* F7.5: ignore. This is no different from the current footprint in the repo, and this exception needs to be made due to the extremely tight pin spacing of these pins on 4-pin (RGB) LEDs.
* F7.3: ignore. I made pin 1 a rounded rectangle, which still denotes is clearly from the other 3, which are circular. The rounded rectangular is required to give proper clearance between all pins. Using a rectangular pad here would make the corner of the pad reduce clearance too much.

jkriege2:
Is there a reason why you added the `_RGB` to the footprint name? We don't have that for most of the other LEDs (only a few are available in special colors, mostly IR and clear ones so you can also use them for IR-Photodiodes/-transistors/-...). I think for the RGBs we don't need that ... anyways, if we want a coloring, we should take the color of the package, not of the LED, as only the package-color is descriptive of the 3D model/footprint (they don't show an ON-state!), so iff we want that ist should be `_Clear`or `_White`, not `_RGB`.
This way the footprint is also usable with other LEDs that might happen to have 4-pins, but are not RGB.

@poeschlr What do you think?

cu
JAN

jkriege2:
Also: In one footprint pin 1 is an oval-rect, while in the others it is a simpe rect ... it should all be a simple rect, because that's in line with all the other footprints!

poeschlr:
I would from now on allow the use of rounded rectangles for marking pin 1. (I plan to update all scripted parts in the near future.)
This can be done now as i do not plan to make any further version 4 backports of the lib.

If we add _RGB to these footprints, how many would also need renaming. I am not really prepared to allow massive renamings until version 5.1 has been released. (After that we can branch away a version 6 branch and do such heavy work on that branch. The version 5 branch would then only get bugfixes from that point on.)

ElectricRCAircraftGuy:
> If we add _RGB to these footprints, how many would also need renaming.

What do you mean? What other renaming would be required? "RGB" just means "4 pin" here.

poeschlr:
If there are other rgb footprints already in the lib then all of them would need to be renamed. We can not simply add a few footprints with a different naming scheme.

If there are more then just a few of them then i would suggest we keep the current naming scheme as i would then classify this change as a major reorganization which is not allowed until we start the development for version 6. (In that case this contribution would need to use the same naming scheme as the existing footprints.)

ElectricRCAircraftGuy:
I see. Makes sense. I'll carefully check for other rgb leds in the library and get back to you.

ElectricRCAircraftGuy:
I looked through every LED (THT and SMD) in the Kicad libraries. This means I visually inspected every single footprint, as well as their names. Not one requires renaming except the one I am trying to rename here (my pull request renames 1 single footprint, and adds 2 new footprints). Prior to my pull request we had:

**LED_THT.pretty:**
* LED_D5.0mm-4
* LED_BL-FL7680RGB

**LED_SMD.pretty**
* LED_RGB_1210
* LED_RGB_5050-6
* LED_RGB_Cree-PLCC-6_6x5mm_P2.1mm
* LED_RGB_Getian_GT-P6PRGB4303
* LED_RGB_PLCC-6

I visually inspected every one, as I said, and these are all the RGB LEDs in existence in the library. Also note that all of them *except* the "LED_D5.0mm-4" one already have "RGB" in their names, so this change would actually make this LED more in-line with the ones already in the library.

ElectricRCAircraftGuy:
Here's the datasheet for the LED_BL-FL7680RGB one: https://cdn-shop.adafruit.com/datasheets/BL-FL7680RGB.pdf. It's name follows its actual part name, so I think its name is correct as-is.

ElectricRCAircraftGuy:
As for the LED_THT types. They are named according to "LED", then diameter, then # of pins. Adding "_RGB" to the end makes the most sense here so as to not mess up their sort order, so I do not want to add RGB after "LED", as that would distort the current form. I think how I've done it is the most logical and cohesive for the LED_THT library.

poeschlr:
Looks good, thanks.

ElectricRCAircraftGuy:
Awesome! I'm glad it was accepted! This is my first library contribution to KiCad. I look forward to many more. I'm excited about what KiCad is doing and want to see it succeed.

Issue #764, opened 2018-07-20
TO-220 footprint pitch is wrong?


evanshultz:
@poeschlr 
At https://github.com/KiCad/kicad-footprints/pull/574 you updated the TO-220 footprint names with staggered pins to the new convention. I have concerns if the naming was done right. Or perhaps I'm confused.

Let's take `TO-220-4_P5.08x2.54mm_StaggerEven_Lead5.84mm_TabDown`:
![image](https://user-images.githubusercontent.com/1936989/42983459-b31459b4-8b9b-11e8-9c4a-74186606e8a3.png)

`P5.08x2.54mm` is, I believe, the distance from pins 1-3 horizontally and then from pins 1-2 vertically. The first dimension is correct, but the second should then be 3.8mm instead of 2.54mm.

The `Lead5.84mm` is the distance vertically from the bottom of the body to the farthest lead, I think. That is 2.05mm + 3.8mm which is nearly 5.84mm.

Many of the footprints deviate in only one dimension and a few match all 3 of the assertions above, so I think this is correct. I read the conversation linked from the PR above but it's not clear to me exactly what is being proposed so I want to check. http://kicad-pcb.org/libraries/klc/F3.5/ contains some information, but it doesn't say the `P<x>x<y>` section is required with staggered leads and `Lead<distance>mm` isn't shown at all. And the fact that the footprint naming is inconsistent made me want to be sure.

I'm going to add a footprint, and the script was never updated for this new convention, so if someone can confirm the above (updating KLC would be nice!) I can address this.

poeschlr:
The lead distance is always from body to pin 1. (No matter if it is the farthest or nearest. See note in the picture of the KLC rule you linked.)

It is possible that the pitch measurement is indeed wrong. (I calculated the measurements by hand when i renamed the parts.)

---

It seems the simplifications made by oliver form my original naming convention did mess that up a bit. (My original intent is documented here: https://github.com/KiCad/kicad-library/issues/1447#issuecomment-323785988)

evanshultz:
Thanks for the info! So KLC could use an update and/or clarification?

Fixed footprints and a missing one needed for https://github.com/KiCad/kicad-symbols/pull/765 submitted at https://github.com/KiCad/kicad-footprints/pull/765.

jkriege2:
[...]

PR #765, opened 2018-07-20
Update TO-xxx stagger footprints

evanshultz:
It seems https://github.com/KiCad/kicad-footprints/pull/574 had a few mistakes as I noted at https://github.com/KiCad/kicad-footprints/issues/764.

This PR should correct all footprints. I checked this several times and checked all footprints, but I'm afraid I've gotten bleary-eyed or I didn't catch something in the nearly-indecipherable script that makes them wrong.

The impetus for this was to add the TO-220F-4 footprint missing that I noted at https://github.com/KiCad/kicad-symbols/pull/765. See page 10 at https://www.fairchildsemi.com/datasheets/KA/KA5M0265R.pdf. I made both stagger configurations just for the heck of it.
image

------------


evanshultz:
Script update at https://github.com/pointhi/kicad-footprint-generator/pull/151.

poeschlr:
The information this is based on was wrong. For some reason i used the pin furthest from the package as the reference for the lead lenght field instead of pin 1 as i originally wrote in the KLC. It might be better to document how it is done right now in the KLC instead of renaming all footprints. (This would allow us to include the lib changes in future bug fix releases of kicad.)
I also suspect that i had my reasons for changing my mind. (My guess is that there are some footprints where lead length does not make sense if it is calculated between pad 1 and nearest body outline)

evanshultz:
In some cases pin 1 could be under the body, and without a sign (plus/minus) on the `Lead` field there's no way to know this. Take http://www.ti.com/lit/ds/symlink/lm675.pdf but imagine the stagger pattern is opposite (swap near and far pin stagger). How to handle that? Would this require using the farthest lead from the body, then?

It's fine for me to do the work if we can devise a solid methodology. :)

poeschlr:
Ah that would explain why i used the farthest pin as reference.
I would in that case really change the KLC to fit our current naming convention.

poeschlr:
I update the KLC in https://github.com/KiCad/kicad-website/pull/367. This clarifies that the lead length is measured between the body and the pad furthest away.

Could you align the naming to this? (meaning revert the changes to the lead length)
Also note that i left the definition of x pitch unchanged to get the least disruptive change. (So the pitch is therefore defined as the distance between pads of the same row)

poeschlr:
Is CLA broken? Or did you change your mail address that is connected to github?

evanshultz:
No change in email address. Not sure what's going on.

PR #787, opened 2018-07-27
Move InvenSense QFN footprints to Sensor_Motion

poeschlr:
- These are highly specialized footprints for motion sensors
- Include top copper keepout as suggested in their app note
- Second version with removed solder mask to ensure even more clearance

The original footprint adding these can be found here: https://github.com/KiCad/Housings_DFN_QFN.pretty/pull/23

Changes done compared to the original footprints:
- changed pad shapes to round rect (new ipc suggestion)
- set mask pad to max EP size
- removed pin number from mask only pad
- corrected 3d model name (points to correct generic qfn footprint now)
- added keepout area

Datasheets used to create footprints:
 - https://www.invensense.com/wp-content/uploads/2015/02/MPU-9150-Datasheet.pdf (4x4mm)
 - https://store.invensense.com/datasheets/invensense/MPU9250REV1.0.pdf (3x3mm)
 - https://www.invensense.com/wp-content/uploads/2015/02/InvenSense-MEMS-Handling.pdf (appnote describing keepout on page 1)

Symbol PR: https://github.com/KiCad/kicad-symbols/pull/788

---

screenshot from 2018-07-27 03-15-01


SchrodingersGat:
This all seems reasonable, the KLC errors are acceptable. Thanks @poeschlr

PR #789, opened 2018-07-27
Added Palconn UTC16-G connector

Duckle29:
Added the USB2.0 only USB type C connector from palconn

------------

~~[Datasheet](http://www.palpilot.com/wp-content/uploads/2017/05/UTC027-GKN-OR-Rev-A.pdf)~~
[Correct datasheet](http://www.palpilot.com/wp-content/uploads/2016/10/C-UTC016-02.pdf)

Screenshots: 
image
image

As you can see, both from the datasheet and from the screenshots, the connector is a bit special, in that it combines the power-pins on single pads, and only has 8 pins appart from the data-pins.
image

I decided to make the footprint with individual pads for the A1/B12, A4/B9, B4/A9, and B1/A12 pairs. This way the footprint will work with the symbol already in kicads libraries. 

However that does mean that there's two pads stacked on top of each-other. I chose to do 2 pads on top of eachother, instead of two smaller pads next to each other, to allow better routing in PCBnew


Duckle29:
I have also used this in a design just now. It doesn't cause any DRC errors :)

Misca1234:
Which symbol do you intent to match this one with, if it is "USB_C_Plug_USB2.0"
Notice that the FP filter for that symbol is "USB*C*Plug*"

Duckle29:
I plan to pair this up with USB_C_Receptacle_USB2.0, which has the filter `USB*C*Receptacle` so that should work right?

Duckle29:
@poeschlr I ordered some test PCBs, I can verify if this footprint works in a few weeks.

poeschlr:
I finally came around to check this.
The curtyard clearance does not quite agree with the KLC (must be 0.5mm all around for connectors) and the smd pads are too long.

Screenshot from 2019-06-02 12-05-03

Duckle29:
Hmm I don't see how the pads are too long. from what I can see on the datasheet recommended footprint, the bottom of the pads should be 5.79 + 0.48 mm up from the edge, and the top should be 5.79 + 1.64mm up from the edge, giving me the 1.16mm

As for the other changes, I'll get on that when I'm back at my desktop :)

poeschlr:
The pink measurement of 1.88 is given as 1.64 in the datasheet.

Duckle29:
@poeschlr Gotcha thanks. I must've moved those up to make them easier to hand solder for a test, and forgotten to move them back down. The requested changes have been fixed. I also noticed the connector wasn't centered so I fixed that too :)

poeschlr:
Only y directions changed (except courtyard) meaning this is all i needed to double check. Dimensions look good now. Thanks (drawing for reference)

Screenshot from 2019-06-03 20-15-17

stephendpmurphy:
Well this is on me... But I just ordered boards with this part on it and I can't seem to source them anywhere? I've looked at Mouser and Digikey and neither have this part. Where was this part originally being sourced from?

greyltc:
The datasheet link is wrong?

PR #792, opened 2018-07-28
Pulse pa4320 2

Misca1234:
This is a re push of, probably abandon, foot print
Pulse_PA4320
http://productfinder.pulseeng.com/products/datasheets/P787.pdf

original fp push
https://github.com/KiCad/kicad-footprints/pull/398

bild

bild

bild

bild

------------


poeschlr:
Why does an inductor include a polarization mark?

Misca1234:
Is not pin 1 supposed to be marked out according to the KLC in F.Silk layer ?

poeschlr:
Only for polarized parts. (Should be added to the KLC. It was in KLC 2.x but is not really clear in KLC 3.x)

Misca1234:
The corners looks the same now

poeschlr:
I would make the corner markers on silk a bit longer to make aligning the part easier. (The component has rounded corners. Your silk mark should be long enough to reach the straight parts of it.)

To be honest i do not quite get why you only have marks on the corner instead of a full outline as used for all other parts in the library.

Misca1234:
We discussed it the other day, the tendency was to use as little F.Silk as possible, but I can changed them to go all the way around

Misca1234:
Now so

bild

poeschlr:
Thanks, looks good.

---

The reason behind reducing silk is to reduce cost in series production. There you have camera systems that check component placements. These systems only need a very small mark on the corners as reference. In small scale (or even prototype) production settings you do not have the luxury to use such a modern system. In such a setting having a full outline helps a lot. (Especially for hand populating the pcb)

For this reason i would not go with the (future) ipc suggestions as long as there is not clear downside of using our current system.

PR #767, opened 2018-07-21
Esp32 wroom 2

Misca1234:
(replace push for https://github.com/KiCad/kicad-footprints/pull/647)

These are the foot prints used by Espressif's RF modules

ESP-WROOM-02 (existed before but modified)
https://www.espressif.com/sites/default/files/documentation/0c-esp-wroom-02_datasheet_en.pdf

ESP-WROOM-32
https://www.espressif.com/sites/default/files/documentation/esp32-wroom-32_datasheet_en.pdf

ESP-WROOM-32U
https://www.espressif.com/sites/default/files/documentation/esp32-wroom-32d_esp32-wroom-32u_datasheet_en.pdf

ESP-WROOM-32D
https://www.espressif.com/sites/default/files/documentation/esp32-wroom-32d_esp32-wroom-32u_datasheet_en.pdf

The keep out area is described in
https://www.espressif.com/sites/default/files/documentation/esp-wroom-02_pcb_design_and_module_placement_guide_0.pdf

bild

bild

bild

------------

Thanks for creating a pull request to contribute to the KiCad libraries! To speed up integration of your PR, please check the following items:

- [ ] Provide a URL to a datasheet for the footprint(s) you are contributing
- [ ] An example screenshot image is very helpful 
- [ ] If there are matching symbol or 3D model pull requests, provide link(s) as appropriate
- [ ] Check the output of the Travis automated check scripts - fix any errors as required

PR #795, opened 2018-07-29
Remove ESP32-WROOM-32D

diegoherranz:
Follow up of PR #767.
It was eventually agreed that both ESP32-WROOM-32 and ESP32-WROOM-32D footprints are identical so there is no point in maintaining both. Also, this way, ESP32-WROOM-32D can be an alias of ESP32-WROOM-32 in the schematic symbols (work in progress in https://github.com/KiCad/kicad-symbols/pull/724).

The only difference between these two devices is the bottom exposed pad size, but still the footprints are the same:
https://www.espressif.com/sites/default/files/documentation/esp32-wroom-32_datasheet_en.pdf (Figure 8)
and
https://www.espressif.com/sites/default/files/documentation/esp32-wroom-32d_esp32-wroom-32u_datasheet_en.pdf (Figure 8)

I guess the only reason not to do this would be the 3D model, in case we showed the different exposed pad in the model, but the 3D model PR for this (https://github.com/KiCad/kicad-packages3D/pull/337) doesn't show the exposed pads so the same 3D model can be used for both too.

Thanks.

------------


poeschlr:
Thanks

diegoherranz:
Thanks. I'll do the same on the kicad-packages3D repo so we don't mantain duplicated models.

PR #797, opened 2018-07-29
Adding L_Bourns_SRN6045TA

ki5libs:
Adding L_Bourns_SRN6045TA
[Datasheet](https://www.bourns.com/docs/product-datasheets/srn6045ta.pdf)
srn6045
[3D package PR](https://github.com/KiCad/kicad-packages3D/pull/350)

------------


Misca1234:
What do you say about putting the pads on distance 2.1 (2.075) mm?

And change the F.Silk o something like this 
(symmetric corners, my image is a quick fix in paint, expet for the corner at pin 1)

bild

ki5libs:
I've fixed the distance between pads. It was a mistake. I've also changed F.Silk and F.Fab layers. I'm not sure that completely understood your idea so changed the screenshoot. Do you think it's OK now?

poeschlr:
Why does a footprint for an inductor include a polarization mark? (As far as i can tell the part is even completely symmetric)

Misca1234:
Yes, this one is, but is not pin 1 supposed to be marked out according to the KLC in F.Silk layer ?

ki5libs:
I've checked another inductor SMD footprints and most of them don't have any polarization mark, but several have. I couldn't find direct answer in the KLC, but it tells that 1 pin should be marked. There are no reasons to do this from the logical point of view... But do you think that way I did it last time has any reasons to be not merged? Please make some decision and let me know.

poeschlr:
It  was better worded in klc 2.x. There we wrote that pin 1 only needs to be marked where it makes sense (meaning where the direction in which the component is placed matters. For inductors it does not make any sense to mark pin 1)

ki5libs:
OK, I made it symmetrical again

poeschlr:
Looks good. Thanks for your contribution

For reference the dimensioned drawing of the footprint
screenshot from 2018-08-01 20-08-39

PR #798, opened 2018-07-29
Adding R_Shunt_Vishay_WSKW0612

ki5libs:
Adding R_Shunt_Vishay_WSKW0612
[Datasheet](https://www.vishay.com/docs/30332/wskw0612.pdf)
wskw0612
[3D package PR](https://github.com/KiCad/kicad-packages3D/pull/351)

------------


Misca1234:
Hi

Should not pad 1 and 2 be 2.67 (2.668 mm)  wide ?

ki5libs:
According to the datasheet:
pad 1 and 2 width (a) = 1.01 mm
pad 1 and 2 length (b - d - c) = 3.43 - 0.381 - 0.762 = 2.287 mm
Is it OK, isn't it?

ki5libs:
Fixed the courtyard. Checked the symbol - you are right
What do think I should better do: adding new symbol with connected pairs of pins 1-4 and 2-3 or change pins numbering on this footprint? Datasheet describes pin enumeration as I1, I2, E1, E2 and 1-2 locate on opposite sides so the first variant seems preferable to me...

poeschlr:
I would prefer if we would make all such footprints fit the generic symbol. (otherwise we will have a lot of symbols and users will have a hard time finding out which symbol fits what footprints)

ki5libs:
So is the solution just reenumerate pins clockwise?

poeschlr:
Yes that would be perfect.

ki5libs:
Done. Screenshot also updated

poeschlr:
Thanks

PR #801, opened 2018-07-31
Added Wuerth HCF inductor footprints

SchrodingersGat:
Added footprints for Wuerth HCF series of inductors

http://katalog.we-online.de/en/pbs/WE-HCF#vs_ct:2

image

image


poeschlr:
For reference, the suggested land patterns are in the datasheets for specific parts. (examples below)
- 2013: http://katalog.we-online.de/pbs/datasheet/7443630070.pdf
- 2815: http://katalog.we-online.de/pbs/datasheet/74436410150.pdf
- 2818: http://katalog.we-online.de/pbs/datasheet/7443640330.pdf

Courtyard clearance should be 0.25 all around the parts (current values see screenshot below)

The datasheets are very badly dimensioned. (Missing dimensions and missing tolerance specifications) So i really can not check the footprint in the detail i would like.
I however have found some things that seem strange even with the dimensions given in the datasheet. (see annotations in the drawings)


2815: you increased the toe fillets it seems. was this on purpose? (And the side fillets of the two bottom pins) For body size you used maximum condition in y direction but something else (27.4mm) in x direction (maximum would be 28mm)


2818: again increased pad sizes. Strange body position. Strange body size.

SchrodingersGat:
I have now reduced courtyard to 0.25mm around each footprint. The toe sizes are still marginally different from the datasheet(s) but by ~30µm simply because it's easier to draw with the KiCad grid (and round numbers are nicer).. The pads are slightly larger than spec but should be absolutely fine for loading. I have test aligned them with the 3D models provided by Wuerth...

evanshultz:
@SchrodingersGat Ping!

poeschlr:
@SchrodingersGat are you still working on this?

PR #804, opened 2018-08-03
added ST_VL53L0X and ST_VL53L1X

5N44P:
Footprint for both VL53L0X and VL53L1X ST's Time of flight sensors.
Both sensors are pin compatible, and could use the same footprint. I'm not really sure on how to specify it so i left that part off the name. Should i make two separate identical footprints?

VL53L0X datasheet: https://www.st.com/resource/en/datasheet/vl53l0x.pdf
VL53L1X datasheet: https://www.st.com/resource/en/datasheet/vl53l1x.pdf

image

No symbol or 3D model so far. I'll make the symbol PR whenever this is approved, so that i can link it

EDIT: 
Current versions:
ST_VL53L
image
ST_VL53L_HandSoldering
image
 

------------


5N44P:
First contribution here, anybody can help with TravisCI errors?

``` 
  Violating G1.7
    Library files must use Unix-style line endings (LF)
    Incorrect line ending
     - Library files must use Unix-style line endings (LF)
```

The file is generated directly from kicad, so i really can't understand what could be going wrong with the line ending. I'm on windows, but seems at least strange that i can't contribute if i'm a windows user.

```
  Violating F9.3
    3D model settings
    No 3D model provided
```
ST has not released any 3D model. I could make one but not now, and i don't know what would be the final name if i contribute it. What should i do?

poeschlr:
You do not need to provide the 3d model it self. But you need to prepare the footprint for it. (So fill out the 3d settings as suggested by travis)

Regarding file endings. This should be taken care of by a proper git client. We have it setup such that such a client would automatically convert the line endings to unix stile. What client do you use? What is your git version?

5N44P:
> You do not need to provide the 3d model it self. But you need to prepare the footprint for it. (So fill out the 3d settings as suggested by travis)

Fixed, thanks! I thought that adding also the step file link would be better but travis gave an error on that. Now it passes.

> Regarding file endings. This should be taken care of by a proper git client. We have it setup such that such a client would automatically convert the line endings to unix stile. What client do you use? What is your git version?

Actually i'm using github's drag and drop commit system. I'm not very confident with CLI git interface.

5N44P:
@poeschlr can i push also the `_HandSoldering` version on this branch and on this PR? Or is it better to do a separate one?

5N44P:
@poeschlr The last commit was done with git CLI on linux and is giving the same problem. I guess that is the file, not the client (Git 2.16.3)

5N44P:
Now the _HandSoldering footprint is failing travis test because it uses the same 3D model as the other footprint. 
To me is ok for merging, waiting the response of the mantainers

Misca1234:
> Now the _HandSoldering footprint is failing travis test because it uses the same 3D model as the other footprint.

That is ok, it is a known "problem" and it will be ignore.

5N44P:
Now also the _HandSoldering version footprint is committed. No relevant errors on travis, only the ones above. Waiting for mantainers to review 
image

5N44P:
@Misca1234 @poeschlr 
Hi guys, i'm waiting for a review for this PR. When you have time can you look into it? 

I've tested the _HandSoldering version and works OK (reflow soldering with stencil).

Misca1234:
We are only a handful "reviewers" and it is done on volunteer basis so you will never know when there is time ^^.

It usually works like this, 
If a (new) push does not have any comments and someone want to review it, it will be reviewed.
It is also come down to personal interest, I prioritize 3D models, over symbols and the foot prints at last, others have different interest.

But let me check into this one

PS
If you are interested in helping out with review others pushes and become a "guy with merge rights" please contact Rene and have a chat with him
DS

Misca1234:
This implies that you should make two different foot prints, one for VL53L0X and one for VL53L1X.

5N44P:
Thank you, will fix ASAP!

Misca1234:
@5N44P 
Hi, how are the foot prints going ?

5N44P:
I hope to get it fixed by next week

5N44P:
@Misca1234 It took me a bit more than a week but we're finally there.
I've fixed the dimensions, dropped the HandSoldering version (it was too arbitrary and overall didn't work that well), removed pad 1 notch, fixed silkscreen width and finally we are there i think. 
Now it should be ok for merging, let me know

poeschlr:
We use IPC marking style. In this case use the bottom only pad marking style that we already use for most qfn footprints. https://ocipcdc.org/archive/What_is_New_in_IPC-7351C_03_11_2015.pdf#pate=16

`Bosch_LGA-14_3x2.5mm_P0.5mm` could be a source of inspiration from the lga lib.

---

If the pads to not protrude from the package at all then the bga style marking would also be acceptable.

poeschlr:
The courtyard clearance is not quite right. 

screenshot from 2019-02-10 09-38-42

5N44P:
Something went wrong with travis i think. It started checking a lot of footprints other than the ones in this pr

poeschlr:
Travis fails similarly to what i found in https://github.com/KiCad/kicad-footprints/issues/1384

Please keep this open i need it as a way to analyze the problem.

5N44P:
@poeschlr the trick appears to be! 
+ make an empty commit 
+ `git reset --hard "HEAD@{1}"` (this also works for git resetting to another head, in #1183 i resetted to the head 5 before a rebase)
+ force push 
+ make an empty commit again, to restart travis 
+ push 

I don't know which one of these operations resolves the issue, but together they worked for this and the other

5N44P:
@poeschlr can you please change the label to pending review? Thanks!

poeschlr:
travis somehow behaves as if the diff range is given with .. instead of ... which is why it shows much more files as different. This is either a bug by the git version used or a dot is somehow lost somewhere between the echo command and git diff.

poeschlr:
I somehow forgot to check this. looks good now, thanks

5N44P:
@poeschlr why is this merged? The footprints were merged by #1621, this was left open for debugging the problems with travis

poeschlr:
Damn today is not my day it seems. Luckily the footprints were identical so in essence i added an empty commit to the master branch (not good but not really a problem either)

PR #805, opened 2018-08-03
Added USB_Mini-B_AdamTech_MUSB-B5-S-VT-TSMT-1_SMD_Vertical

5N44P:
image

DS: http://www.adam-tech.com/upload/MUSB-B5-S-VT-TSMT-1.pdf
------------


5N44P:
See https://github.com/KiCad/kicad-footprints/pull/804#issuecomment-410227192 for the remaining Travis errors

5N44P:
To me this is ready for merging. All travis errors and warnings have been cleared and the footprint is already tested on a pcb. 
Waiting for the mantainers response

5N44P:
Tonight i'll fix the fab and silk lines. 
The pin order is standard for all usb connectors, but i'll double check also that. 
The measures on the ds aren't missing, they are only a bit too implicit. 


Btw i've tested the footprint in production, minus the data pins. The power pins are correct, so i'd assume that also the others are. Mechanically fits perfectly.

5N44P:
@poeschlr I've fixed the silkscreen and the fab layer. 

see http://pinoutguide.com/PortableDevices/mini-usb-cable_pinout.shtml for pinouts.

Outlines can be obtained adding the metal thickness (0.4mm, showed in the upper right corner for the tab) to the inner measurement (showed in the lower left corner). Measure with the caliper on the connector confirms the dimensions.

poeschlr:
One courtyard direction is not quite right (see drawing)
Add pin 1 markers on silk and fab layers (remove the silk line next to pin 1 for marking that pin. Add a chamfer on the fab outline edge near pin 1)
Rotate the footprint counter clockwise by 90 degree such that pin 1 is nearest to the top left corner.

screenshot from 2018-08-09 11-55-25

poeschlr:
You are right, as the pins protrude the body, the silk line along pin 1 is the marker.

5N44P:
I'm looking into KLC. Pin 1 marking on Fab layer for connectors requires an arrow (F.5.2.2.c) 
Is it better to make the arrow or the chamfer?

poeschlr:
The pictures in the KLC are just examples of different posibilities. IPC suggests what i described above. Where sensible we should go with the standard. (So chamfer on fab plus line alongside pin 1 on silk. If you do not include the area covered by the pins in the fab outline then we might use something similar to what is used for molex picoblade connectors on the fab layer.)

5N44P:
Yeah that's also what's happening with the only other mini usb. I'd propose do that to keep consistency across the usb lib and adhere to the KLC

image

5N44P:
Current proposed footprint:

image

poeschlr:
Looks good. Thanks

PR #807, opened 2018-08-04
Add QingPu WQP-PJ398SM aka Thonkiconn

robkam:
QingPu 3.5 mm vertical audio jacks. Commonly used for Eurorack synth modules. Popularly referred to as Thonkiconn (https://www.thonk.co.uk/shop/thonkiconn/). 

For symbol Connector:AudioJack2_Ground_Switch

No data sheet, only a web page at http://www.qingpu-electronics.com/en/products/WQP-PJ398SM-362.html

thonkiconn mono footprint

Pad holes are round not slotted for more economical manufacture, Hole diameter is the maximum lead diameter (on the diagonal) plus 0.2mm.


poeschlr:
As this one has very elongated leads it might be a good idea to add a second version with oval drills. (Better connection, more pad to pad clearance possible, ...)

In any case for the upper two pins at least i would suggest to use elongated pads. Something like this should do the trick (used the drill sizes you chose): 
screenshot from 2018-08-09 18-17-44
Even 1mm clearance could be reasonably optained:
screenshot from 2018-08-09 18-33-07

---

The courtyard clearance should be 0.5mm all around for connector footprints (relative to pads or body, whatever results in the larger courtyard area)

The body dimensions (fab outline) are not quite right. The center of the line should represent the nominal body size. (the body should be 9mm wide, 10.5mm high. The bottom edge should be 6mm from the center of the plug)

screenshot from 2018-08-09 18-39-14

robkam:
capture

poeschlr:
Looks good, thanks

robkam:
Thanks :-)

xxv:
The symbol that seems to match this, `Connector:AudioJack2_Ground_Switch` uses the pin labels `S`, `T`, and `TN` instead of numbering them. Should this footprint be updated to match the symbol or am I misunderstanding how it's intended to be used?

robkam:
S, T and TN would be better. All the current KiCad audio jack symbols are S, T, etc. and footprints 1, 2, etc..  When reading the netlist in Pcbnew this gives errors like:
Error: Component “J1” pad “S” not found in footprint 
Error: Component “J1” pad “T” not found in footprint 
Error: Component “J1” pad “TN” not found in footprint

xxv:
Cool. I created a [PR](https://github.com/KiCad/kicad-footprints/pull/930) to fix them.

PR #810, opened 2018-08-05
Scripted qfn-56-1ep used for cypress mcu (replaces #400)

poeschlr:
Originally submitted in #400.

datasheet used for getting the sizes are:
- http://www.ti.com/lit/an/scea032/scea032.pdf (From 2003 so i did not use the recommended footprint from there as it is now outdated.)
- http://www.cypress.com/file/138911/download (Has the same sized part but with slightly smaller tolerances on nearly every measurement. I used the TI measurements as this was used in the original PR. I can easily change it to use the cypress measurements)

---

Symbol that uses this has been merged in https://github.com/KiCad/kicad-symbols/pull/289

A very strange footprint with top tented vias does also exist. But that one is not connected to the symbol. (I researched this a lot. It seems there is no real consensus what the optimal solution is for avoiding solder wicking. It seems right now the solution most used is to simply increase solder paste coverage to make up for the fact that some of it is lost inside the vias. In addition to using very small vias. Bottom tenting what we normally use is also here to help out even more.) 

This footprint is independent of the decision which footprint should be connected to that cypress part.

---

Generator parameters:
``` yaml
QFN-56-1EP_8x8mm_P0.5mm_EP4.8x5.5mm:
  device_type: 'QFN'
  size_source: 'http://www.ti.com/lit/an/scea032/scea032.pdf (page 4)'
  body_size_x:
    nominal: 8
    tolerance: 0.15
  body_size_y:
    nominal: 8
    tolerance: 0.15
  lead_width:
    nominal: 0.23
    tolerance: [0.05, 0.07]
  lead_len_min: 0.3
  lead_len_max: 0.5

  EP_size_x_min: 4.35
  EP_size_x_max: 4.65
  EP_size_y_min: 5.05
  EP_size_y_max: 5.35
  EP_num_paste_pads: [4, 4]

  thermal_vias:
    count: [5, 5]
    drill: 0.2
    paste_via_clearance: 0.1
    EP_paste_coverage: 0.6

  pitch: 0.5
  num_pins_x: 14
  num_pins_y: 14
```

screenshot from 2018-08-05 15-33-31


Misca1234:
Nice, gotta love script, but I got one question, on page 11, PCB design guidelines
they want the EP size to be minimum 4.8x5.5 but I guess that is not a big issue ?

poeschlr:
Yea and if you look at the cypress document they suggest something different. (Or the cypress footprint is made wrong as it has ep size of 6.22x6.22) So a generic footprint can not really follow one particular datasheet too closely (If we would want a truly generic footprint we would need to get every datasheet that is out there and find some compromise between their given measurements. I think you will agree that this is way too much work.)

Also look at the publishing date of that datasheet. Back then the manufacturing processes where quite different. IPC-7351 did not even exist yet. (Was published in 2005. The predecessor was from the the year 199x if i remember correctly.)

---

In general we normally use the nominal size of the "lead" for determining our exposed pad sizes. At the most i would use the maximum lead size. (microchip seems to sometimes use that. I would guess to increase copper area for better cooling.) I can increase copper further but in that case i would suggest to switch from non solder mask defined to solder mask defined. (meaning make the mask cutout equal to the maximum lead size with a separate mask pad.) 

Our decision was motivated by an TI document mentioning without giving a reason that this would lead to the best results. The observation that a vast majority of datasheets use the nominal size. And the suggestion by the community over at the forum. They gave as a possible reason that self alignment might be better when using nominal sizes. (See old discussion: https://github.com/KiCad/kicad-footprints/issues/678)

Misca1234:
then let us go for that

Issue #802, opened 2018-08-02
Connector DIN pin numbers do not fit any of our symbols


poeschlr:
The current din number connectors use uppercase letters in their pin numbering. All our symbols use lower case letters.

The script generating these is found here: https://github.com/pointhi/kicad-footprint-generator/tree/master/scripts/Connectors_IEC_DIN

PR #811, opened 2018-08-05
Fix pin numbers of DIN connectors to fit symbols

poeschlr:
fixes issue #802

script update: https://github.com/pointhi/kicad-footprint-generator/pull/157


evanshultz:
There are a few errors with 3D model library name, but those apparently aren't scripted since your script is correct here.

poeschlr:
Yea i would guess that the footprint names will discussion anyways. Right now there is a lot of information missing from them but i wanted to do the least amount of changes while still fixing the reported problem. (I and a few devs got a mail from an infinion guy who wants to sponsor a bit to kicad. But they want some specific features in return. One of these is more DIN connector footprints/symbols in the library.)

PR #815, opened 2018-08-07
Adding Samtec PCIE-038-02-X-D-RA (PCIe 1x)

suzizecat:
Adding the PCIe 1x connector footprint from Samtec

[Footprint specification here](http://suddendocs.samtec.com/prints/pcie-xxx-02-x-d-ra-footprint.pdf)

image

Samtec provides 3D model but I don't think that they are correctly licensed to fit in KiCAD
------------


poeschlr:
There is no courtyard offset. (Yes samtec gives a minimum area but we still have a definition that is larger as our minimum area.) 
The silk outline should be outside of the body (as fab represents the body it should be outside the fab line)
The fab outline is not drawn correctly. (offset in x and y direction. Does not include every part of the body)

screenshot from 2018-08-09 11-29-25
screenshot from 2018-08-09 11-26-42

Misca1234:
@suzizecat 
Will you address the points above ?

suzizecat:
I will try if i find some time to do so... Sorry about that

suzizecat:
I *think* that I fixed the issues, please do not hesitate to let me know

poeschlr:
Move the silk reference designator to the top and the value to the bottom. (outside of the outline)
There is no need to decrease the font size of the value field.
Silk should not overlap any pads. (Not even non plated ones) Keep a clearance to pads. (0.2mm as this is a very large part)

Add a pin 1 marker on the fab layer.

The hole size and pad size must be the same for non plated through holes (NPTH)

Where did you get the part size in y direction from? (You used 19.5mm where does this value come from. The 3d model i posted above seems to suggest 18.85)
The body has not been fixed.

samtec svg

poeschlr:
Oh and i think the footprint name is not quite right. There is no 38 pin version in the datasheets. I would assume it should be named `PCIE-036-02-X-D-RA` (36 pins per row)


---

And for reference a better dimensioned drawing generated from the samtec 3d model:

screenshot from 2019-02-09 16-45-53

suzizecat:
Damn, I'm rusty on footprint making. Thanks to point out the problematic dimensions in your reply (hadn't understood that on the previous reply)

I found the 19.5 (19.6 actually, fixed) in the datasheet link at the beginning of the post : 
http://suddendocs.samtec.com/prints/pcie-xxx-02-x-d-ra-footprint.pdf
Editing the name and then pushing.

Thanks for your time :heart:

poeschlr:
I think the 19.6 is for the "keepout" area not necessarely for the part outline. Sadly the datasheet is a bit bad for this part to be honest.

poeschlr:
You corrected the distances in y direction wrong. (Introduced a new error and overshot the one you tried to correct.)

I am still convinced that we should go with the 3d model size for the body in y direction.

screenshot from 2019-02-12 19-15-20

suzizecat:
I fixed the overshoot. 
However, I don' see how could I reduce the size of the component, since 18.85mm will fall just on the last pin row (If I don't screw the other measurements...)

poeschlr:
This is how the part is. Look at the 3d model. The plastic support just covers the pins enough to give strengths. (I don't quite get where your problem is. The fab outline can be on top of pads, silk does not need to go all around. Simply leave it out where it would interfere with pads. Courtyard will likely be calculated against the pads in this case.)

suzizecat:
I would have used Fab to define the "overall" outline, running outside of the pads.
However I get your point (and the fact that what I mention above is simply the courtyard job).
For the Silkscreen, do I have to draw the horizontal line which will go over the pads (removing silk over pad) or can I just leave it out ?

I will update the FP this evening.

suzizecat:
I fixed the size with respect to the 3D model dimensions. I also fixed the "pin 1 indicator" on the fab layer which has a 0.12mm width.
The silkscreen is 0.5mm away from the fab outline (splitted near pads) and the CrtYrd is 0.5mm away from the limit of the pads (center + whole size/2 + 0.5)

So, it should be fine now.

poeschlr:
The distance between mounting holes and front edge is not quite right
Should be 7.55mm

Screenshot from 2019-03-24 17-49-34

suzizecat:
Fixed, that snap to lines end in nigthlies is great ! :heart:

poeschlr:
The snap to line ends is in all version 5.1.x releases. It is not suggested to use nightly builds for contributions to the official library as it can mean the contribution is incompatible with the current stable version.

I will take a look at this later. Maybe today.

poeschlr:
As feared your use of nightly builds has negatively impacted this file for us. simply saving it without any change in stable will change nearly every line in the file and make future maintenance a nightmare.

So please save the file using a stable version.

PR #818, opened 2018-08-12
Add _RoundHole to name (now _CircularHoles)

robkam:
regular footprint Jack_3.5mm_QingPu_WQP-PJ398SM_Vertical.kicad_mod with oval holes to follow


poeschlr:
We use the suffix CircularHoles in other such footprints

poeschlr:
The 3d model should still point at the non circular hole variant (yes travis will complain but it is just too much work to teach travis that it is ok that way)

poeschlr:
thanks

Ratfink:
@poeschlr It isn't too much work; in fact there's an open PR that does it: https://github.com/KiCad/kicad-library-utils/pull/247

PR #820, opened 2018-08-12
create mono oval holes QingPu_WQP-PJ398SM

robkam:
Jack_3.5mm_QingPu_WQP-PJ398SM_Vertical.pretty except with with oval pads and holes.

http://www.qingpu-electronics.com/en/products/WQP-PJ398SM-362.html (can't find a pdf)

capture


poeschlr:
Is it on purpose that you opened two pull requests for this?

robkam:
Oops, no, the second one can be deleted.

poeschlr:
The drill size for pads 2 and 3 is a bit too short. If you want the same clearance as you have at pin 1 then the drill of pad 2 should be about 2mm long and the one for pad 3 should be 1.93mm long.

This will result in a clearance of 0.065mm at the nearest points.

robkam:
I don't follow how 2 mm and 1.93 mm are derived?
pin 2 measured at it's widest/diagonally + 0.2 mm = 1.82
pin 3  measured at it's widest/diagonally + 0.2 mm = 1.78

Also I've now noticed in the plan view pin 1 is 0.8 mm but I've gone by the PCB layout which gives 0.6 mm?

robkam:
btw pin 1 measured at it's widest/diagonally + 0.2 mm = 1.63

poeschlr:
Another option would be to interpret the given pcb layout as the minimum needed size. In that case all oval holes would be: width the same as the given width, length would be given length plus two times half the with (so given length plus width)
This would result in:
- Pin 1: 0.5x2
- Pin 2: 0.6x2.1
- Pin 3: 0.6x1.8

robkam:
There is a [Thonkiconn Eagle v6 Library Part – thonkiconn.lbr](https://www.thonk.co.uk/shop/3-5mm-jacks/) available. I've converted it to KiCad and attached here, (it's an earlier model but same dimensions and pins). All pads are round, 1.65 mm diameter and with 1.1 mm hole.

[WQP-PJ301M-12_JACK.zip](https://github.com/KiCad/kicad-footprints/files/2290765/WQP-PJ301M-12_JACK.zip)

robkam:
I'd be happy to abandon the oval hole versions and revise the other two (mono and stereo) to 1.1 mm diameter holes and 1.65 mm width pads?

poeschlr:
The eagle lib just uses round holes as eagle does not really support oval drills.

robkam:
I've no special requirement for the oval hole version.

poeschlr:
You might not have them. But i would guess that the soldering quality would be better if one uses oval holes. This might be interesting for people concerned with the longevity of their product.

robkam:
I'm going to have to do some oval holes for other parts. Is there some rule of thumb to estimate for the clearance from the pin center to corner to oval milled hole edge?

robkam:
extra 0.2 mm from the corner/on the diagonal =sqrt((pin length + 0.02) + (pin width + 0.02))

robkam:
To get a clearance of 0.2 mm at each corner add 0.28 mm to the length and 0.28 mm to the width. Would this work?

poeschlr:
I am not sure what the question here is.

robkam:
Sorry it's not clear. I'd like to work out a way to have the corner of the pin at 0.2 mm from the edge/corner of the oval hole?

I'd worked out the above incorrectly anyhow. 

Extending lines from the pin to the edge of the hole. a for the length to edge, b for the length from width to edge. Have a = b and c is 0.2 mm. What length do a or b need to be, (remembering to use double these lengths so as to also include the gap at the far corner).
   
capture

poeschlr:
Your drawing is not quite right as your line c is not normal on the tangent of the arc. To get it tangent make a line from the center of the arc through the corner of the lead till it reaches the arc outline.

If you do this you will notice that the largest clearance you can have at the arc is the same as you have between the sides of the lead to the side of the hole. This is the case when the straight sides of the hole have the same length as the leads. (So then the hole size would be hole_width = lead_width + 2 *clearance and hole_length = lead length + 2 * hole_width)
screenshot from 2018-08-27 00-19-28


Any other arrangement will result in a smaller clearance to the corner compared to the side. You seem to assume that the line c is at 45 degree. This arrangement would result in negative clearance with the current way you made the footprint.
screenshot from 2018-08-27 00-19-18

My suggestion is to go with either the maximum clearance (so the second paragraph in my response) or use a program like the freecad sketcher to arrive at something you want.

robkam:
So with [measuring the leads](https://github.com/KiCad/kicad-footprints/pull/828#issuecomment-413913186), adding 0.2 mm for clearance and using [hole length = lead length + hole width ](https://github.com/KiCad/kicad-footprints/pull/820#issuecomment-413192881) :
pin (w x l): robkam / poeschlr / data sheet's rectangular holes
1 (0.28 x 0.82): 0.68 x 1.9  / 0.5 x 2 / 0.6 x 1.3
2 (0.28 x 1.02): 0.68 x 2.1 / 0.6 x 2.1 / 0.6 x 1.5
3 (0.17 x 1.03): 0.57 x 2  / 0.6 x 1.8 / 0.5 x 1.5

poeschlr:
your dimensions seem to err on the save side so i would be ok with them.

robkam:
capture

robkam:
The pads look a touch too long.

poeschlr:
Sorry i totally forgot about your pull request. (I started a new job recently and needed to finish some old stuff before that.) Did you close it because you lost patience with us? (or for some other reason?)

robkam:
I thought I'd made a mess of the PRs and that you couldn't face going through them. I was starting to attempt to resubmit them as a single PR, (I don't yet know how to do that for a bunch of new files).

poeschlr:
The reason really is simply a lack of time on my side.

If you want them in one pull request then simply create a new branch from the current master, add the files, commit them, and don't forget to link to the original pull requests such that we can remember what is going on.

If you need more help just ask. (You can even send me a PM over at the forum. I am easer to reach over there right now.)

robkam:
Okay done three new footprints in #1073. 
Also #1069 to revise mono with circular holes QingPu_WQP-PJ398SM 

Thanks, I'll PM you on the forum.

PR #821, opened 2018-08-12
Top tented version of qfn-56 footprint for cypress

poeschlr:
Cypress suggests the use of top tented vias in the exposed pad.
Hand modified the current generic footprint such that it uses this
tenting as an example of how it can be achieved.

Followup for https://github.com/KiCad/kicad-footprints/pull/810
This one showcases how i would make a top tented footprint. (If i am motivated i will add this to the script.)
It is an alternative to the version added in https://github.com/KiCad/kicad-footprints/pull/582 (Using custom pads instead of relying on overlapping graphical lines.)

screenshot from 2018-08-12 13-43-12


evanshultz:
Could you change the link to the Cypress datasheet which specifically shows tented vias?

poeschlr:
Description updated

PR #822, opened 2018-08-12
create stereo oval holes QingPu_WQP-PJ366ST

robkam:
QingPU stereo 3.5 mm vertical jack WQP-PJ366ST suits symbol AudioJack3_Ground

http://www.qingpu-electronics.com/en/products/WQP-PJ366ST-364.html (no pdf)

jack_3 5mm_qingpu_wqp-pj366st_vertical


poeschlr:
Another new pull request for the same part?

robkam:
stereo version, the body is .5 mm wider

robkam:
There should be four PRs but I can only see 3. I've lost one somewhere.

poeschlr:
Ah ok. In future if you submit similar things you can put them into the same pull request. Or if you make different pull requests at least make it clear what the small difference is as it is hard to see otherwise.

robkam:
I've no special requirement for the oval hole version.

robkam:
stereo oval holes

PR #823, opened 2018-08-12
create stereo circular holes  QingPu WQP-PJ366ST

robkam:
QingPU stereo 3.5 mm vertical jack WQP-PJ366ST with more economical round drill holes, suits symbol AudioJack3_Ground - this is the 4th of 4 types of Qingpu 3.5 mm vertical audio jacks.

http://www.qingpu-electronics.com/en/products/WQP-PJ366ST-364.html (no pdf)

jack_3 5mm_qingpu_wqp-pj366st_vertical_roundholes


robkam:
stereo circular holes

PR #824, opened 2018-08-13
Add vertical RJ-45 connector On-Shore PJ012-8P8CX

cpresser:
Add vertical RJ-45 connector On-Shore PJ012-8P8CX

Screenshot
[Datasheet](http://www.on-shore.com/wp-content/uploads/2015/09/PJ012-8P8CX.pdf)
There is no 3D-Model.


poeschlr:
The land pattern including fab and courtyard looks good. But the silk screen should be outside the body dimension (meaning outside the fab outline as it represents the body dimensions.)

For reference the dimensioned drawing:
screenshot from 2018-08-15 17-26-11

A pin 1 marker might also be a good idea on silk and fab layers.

cpresser:
Hi,
thanks for the hints.
I did take a look at other RJ connectors to see how their pin1 indicator looked like and did reproduce it.
The silkscreen-outline was moved as requested.

How did you create the annotated drawing? I could use that for upcoming footprints I want to contribute.

Misca1234:
@cpresser 
All looks good but a minor issue is left
The F.Fab line width is 0.12, according to KLC
"Outline uses line width between {0.10mm and 0.15mm} (recommended 0.10mm)"
If you change the F.Fab width to 0.1 I am happy to merge

poeschlr:
Thanks.

PR #826, opened 2018-08-14
added DSBGA-28_3.0x1.9mm_Layout4x7_P0.4mm for bq51050b

cyber-murmel:
This is one of the footprints for the bq5105xB High-Efficiency Qi v1.2-Compliant Wireless Power Receiver and Battery Charger.

------------

  - http://www.ti.com/lit/ds/symlink/bq51050b.pdf
  - 
  - https://github.com/KiCad/kicad-symbols/pull/822


poeschlr:
The dot as pin 1 marking is outdated. For BGAs a chamfer seems to be standard like is done with `BGA-324_15.0x15.0mm_Layout18x18_P0.8mm_Ball0.5mm_Pad0.4mm_NSMD`.

The chamfer on the fab layer is a bit large. It should be at most 25% of the shorter body side length.

The courtyard clearance also looks a bit small. IPC suggests 1mm clearance all around to allow for rework. (to get the heat gun in there without desoldering nearby components. Yes some of our bga footprints to not follow this rule. The BGA lib in general is very inconsistent right now.)

Misca1234:
@cyber-murmel 
Do you want to make this script based instead?
I can show you how to use them

Misca1234:
Yes, they are easy to use and makes the reviewers life easier, we just check the data file and never measure in the foot print it self, because the understanding is that the scripts always do right.

Do the following.

1.
Fork the following repository and check it as usually.

For this foot print, BGA, there is a specific script for them in
...../kicad-footprint-generator/scripts/Package_BGA

2.
Open the bga.yml file in a text editor and add your foot print and the dimensions, see the other examples in the file.

3.
Execute python script and the yml file like this
python bga.py bga.yml

And then all .kicad_mod foot prints are generated, including yours

4. 
Replace the generated .kicad_mod file in this push with the content that was generated and re push 

5. 
Push the updated bga.yml to 

As you can see there are a lot of scripts there for various purpose.

There are scripts for 3D models as well, are you interested in them too ?

Misca1234:
@cyber-murmel 
How is it going, did you sort out the script approach ?

cyber-murmel:
I think i generated this FP with the wizard and then modified it so that he silk screen doesn't overlap a pad

cyber-murmel:
Jup, looks good now :-)

cyber-murmel:
Ah, sorry, here you go https://github.com/pointhi/kicad-footprint-generator/pull/248

Misca1234:
Nice, just of curiosity, you set the mask_margin to 0.05, did you take it from 
http://www.ti.com/lit/an/snva009ag/snva009ag.pdf ?

Misca1234:
Thx

PR #828, opened 2018-08-15
create mono circular holes QingPu_WQP-PJ398SM 

robkam:

capture


poeschlr:
Where is your evidence that the pads where too large. From the dimensioned drawing that you provide i would gather that 1.1mm is too small.

robkam:
1.65 x 1.1 are the dimensions in an Eagle library from a UK supplier of these parts, and probably widely used without problems afaik. 

I was mistakenly using the [PCB layout/hole dimension given at QingPU](http://www.qingpu-electronics.com/en/products/WQP-PJ398SM-362.html) for the pin dimensions.

Alternatively I'll change the hole diameters to the lengths given on the QingPU PCB layout?
Pin 1: 1.3 mm
Pin 2: 1.5 mm
Pin 3: 1.5 mm

1.1 mm does look too small.

poeschlr:
I would guess there is a reason why the manufacturer suggests their sizes. Maybe their parts have large tolerance ranges that would create problems in automatic assembly if you use too small drill sizes. (If the tolerances are in how the leads are bent then they might not matter that much for hand soldering which could explain why the UK supplier has smaller drill sizes. Especially if they target hobbyists and small patch production.)

So i would really suggest to make the drill sizes large enough to include the area specified by the datasheet. (The sizes you listed in your last post seem reasonable) It would be nice if the manufacturer would supply better drawings but right now our hands are a bit tied.

robkam:
capture

poeschlr:
Please keep the suffix _CircularHoles. The datasheet suggests ovals so it should really be visible that this is a modified footprint compared to what the manufacturer suggests.

robkam:
BTW I used digital calipers to roughly measure the pins (there's no taper) at the base of the body and a PCB with round holes already available to suit these jacks. 
pin 1 - 0.82 x 0.28 mm and PCB 1.31 mm
pin 2 - 1.02 x 0.28 mm and PCB 1.58 mm
pin 3 - 1.03 x 0.17 mm and PCB 1.55 mm

robkam:
This PR now has the correct holes sizes as in the QingPU data sheet, PCB layout:
Pin 1: 1.3 mm
Pin 2: 1.5 mm
Pin 3: 1.5 mm

Once it's merged I'll do the other variations.

robkam:
capture

evanshultz:
@robkam 
The title of this PR has been changed 4 times. Perhaps it's just a pet peeve of mine, but it breaks the email chain for subscribers and makes it harder to track all the comments in a PR in my inbox. There are times when a rename is useful, but it's not adding any value here. Please resist the urge to rename.

robkam:
Oops apologies, I'll leave the name now as is. However please note I'm submitting four very similar footprints here: Mono circular and oval holes, and stereo circular and oval holes.

evanshultz:
Thanks for leaving the title alone.

I know you're submitting 4 PRs. Each librarian gets an email each time you commit to or comment on every single PR. IMO It would have been easier to review and manage one PR for all 4 footprints.

PR #794, opened 2018-07-29
Scripted WSON 8pin 6x5mm with 3.4x4mm EP (replaces #501)

poeschlr:
used dfn generator with pull back ipc settings
This replaces #501 but uses ipc to arrive at the footprint (compared to the footprint supplied there mine has the pads moved a bit outwards. -> more toe less heel)

Datasheet: http://ww1.microchip.com/downloads/en/AppNotes/S72030.pdf

parameters:
``` yaml
WSON-8-1EP_5x6mm_P1.27mm_EP3.4x4.0mm:
  device_type: 'WSON'
  library: Package_SON
  size_source: 'http://ww1.microchip.com/downloads/en/AppNotes/S72030.pdf'
  ipc_class: 'qfn_pull_back' #'qfn' | 'qfn_pull_back'
  body_size_x:
    nominal: 6
    tolerance: 0
  body_size_y:
    nominal: 5
    tolerance: 0
  lead_width_min: 0.35
  lead_width_max: 0.48
  lead_len_min: 0.5
  lead_len_max: 0.75

  EP_size_x: 3.4
  EP_size_y: 4.0
  EP_num_paste_pads: [2, 2]

  pitch: 1.27
  num_pins_x: 0
  num_pins_y: 4
```

screenshot from 2018-07-29 11-30-50

Edit: Updated parameters with corrected values for lead length

PR #832, opened 2018-08-17
Added two DFN-8 variants

SchrodingersGat:
2x DFN-8 variants

* http://rohmfs.rohm.com/en/products/databook/datasheet/discrete/transistor/mosfet/qh8ma3tcr-e.pdf
* https://www.vishay.com/docs/74483/si5504bdc.pdf

image


poeschlr:
DFN footprints can be generated via a script. The current version of the script can be found in my pull request https://github.com/pointhi/kicad-footprint-generator/pull/149 so if you base your addition on my branch then i can cherry pick the size definitions from your branch.

The script it self is found in the scripts/Packages/Package_DFN_QFN
The script is run with python3 ipc_dfn_qfn_generator.py size_definition/dfn.yaml

If you use the script show the parameters used in the pull request description. (You can use this pull request of mine as an inspiration: https://github.com/KiCad/kicad-footprints/pull/794)

---

If you do not want to use the script then at least change the pads to rounded rectangle with 25% corner size (maximum 0.25mm radius)

evanshultz:
@SchrodingersGat 
I see two more symbol PRs from you, and they are missing footprints. Considering your capabilities, would you make script-generated footprints as the first step to merging your symbols?

poeschlr:
I looked at these more carefully. They both do not really seem to be DFN packages. (The leads protrude from the package.)
---

So the ROHM one could be named `ROHM_TSMT8` (or optionally: `ROHM_TSMT-8` or `ROHM_TSMT-8_2.5x3.1mm_P0.65`)

The vishay part could be named `Vishay_ChipFET_1206-8` (or `Vishay_1206-8_ChipFET` optional including dimensions.)

I am not really sure they fit into the DFN library. They might fit into the TO-SOT lib. Or we could make a Package_Other lib that holds such strange parts that do not really fit a category. If vishay has a lot of ChipFET packages then we can go a similar route then we did with the DirectFET parts and make a Package_ChipFET library.

---

The suggested footprint for the vishay part is very different to what you committed here. (It has two pads connected with each other and a very large copper area for them.)

I will add the functionality to generate such footprints with the dfn generator later today. (Such that it can generate parts where the leads protrude from the package -> there will be a way to enter outside lead to outside lead dimensions.)

evanshultz:
@poeschlr 
As DFN means "dual flat no-lead" I think these still apply. I've not seen one with leads sticking out either, but the basic naming may still fit (at least until we had a large number of them and it was worthwhile to split them off, as you said).

Did you update the script yet?

poeschlr:
I had no time yet for the script. The opto device stuff took up all of my time.

evanshultz:
@poeschlr 
Will you have time to do any more work here or should I review these footprints while requesting roundrect pads on your behalf?

evanshultz:
Is the ROHM package unique? Maybe it should be `ROHM_TSMT8`? Should it be vendor-specific as I don't see any other companies using it.

ChipFET seems to be a Vishay package name but used by other vendors. See https://www.onsemi.com/pub/Collateral/AND8061-D.PDF. I only see a `1206-8` version at https://www.vishay.com/mosfets/, but perhaps we should be flexible and go with something like `ChipFET_1206-8`?

evanshultz:
@poeschlr @SchrodingersGat Ping!

evanshultz:
Ping again!

poeschlr:
I think adding these special parts to the script is a bit too much work right now. (I have higher priorities for scripting. Maybe first get most of the normal packages transferred to the script.)

My comments above about the vishay footprint would need answering for this to go further. (The point about the fact that that suggested footprint requests a large copper area connecting two pins.)

Regarding the fitting into the dfn lib: If we are that loose then why do we have a wson lib? These are more like dfn then these parts are. Additionally i would not put too many new things into the qfn/dfn lib if we can help it. That one will grow out of control otherwise.
So my suggestion would really be a Package_other lib for odd packages like these that do not really fit anywhere.
Or we can add a Package_FET library to hold specialized packages for FETs. (Similar way we handle opto devices and sensors)

evanshultz:
@SchrodingersGat 
Will you review Rene's comments and make updates so this can get merged?

antoniovazquezblanco:
This has been untouched for too long. I will mark as abandoned in case someone wants to take over it.

Thanks!

PR #833, opened 2018-08-18
JST connectors: Use roundrect for pin 1 + small fixes

poeschlr:
- Pad 1 is now a rounded rectangle for JST connectors (except ZE
vertical which uses a chamfered pad to reach the required clearance)
- Angle field in arcs is does not unnecessarily include zeros after the
decimal point
- Use circular instead of oval pads where the x and y size are equal

Script updates are in pull request https://github.com/pointhi/kicad-footprint-generator/pull/161 (Last 4 commits.)


evanshultz:
Quick look seems fine. I'll start merging this series of updates. Thanks!

PR #837, opened 2018-08-18
Connector Molex THT: Pad 1 rounded rectangle + fixes

poeschlr:
- Pad 1 is now a rounded rectangle instead of rectangle
- Molex nano fit now uses a chamfered pad for <= 3 pins per row to
inctrease clearance to the NPTH
- Some minor fixes to silk screens (fixes to mirror operator took affect
after reexporting)

Script: https://github.com/pointhi/kicad-footprint-generator/pull/161


evanshultz:
The following THT families don't appear to be included (maybe they're not scripted?):
Picoblade
Picoflex
Sabre
SPOX

Merge?

poeschlr:
They are scripted and i thought i included them. Will take a look when i come home.

poeschlr:
I checked it more closely. I am not sure why some of them where not included here. I had them in my local directory.

The sabre series currently has overlapping rectangle pads for all of its pins. So i did not touch that. (It has two pins per pin number to increase the current rating.)

The microfit-3.0 stuff got merged after i branched away my stuff so i forgot them. (I merged these changes now so i will include them shortly.)

poeschlr:
Everything should now be included

evanshultz:
Travis errors are with rectangular pads, which are fine, and some annular rings that are too small because of the pad size. I'm OK with this. Please confirm this is what you wanted and I'll merge.

poeschlr:
I think the annular size calculation simply can not handle polygon pads very well. It only looks at the reference pad which is a lot smaller than the real annular ring.

PR #842, opened 2018-08-18
add VSSOP-10 package with a 0.5mm pitch

penoud:
I based design according ADS1115 datasheet. I checked the part with the python script but get several errors. But I did not understand why. I got error violating 9.3. I did also the 3D. 
the 3D is in another commit.
------------

- [X ] Provide a URL to a datasheet for the footprint(s) you are contributing
http://www.ti.com/lit/ds/symlink/ads1115.pdf
- [ X] An example screenshot image is very helpful
image
 - [] If there are matching symbol or 3D model pull requests, provide link(s) as appropriate
I will create a pullrequest with the 3d package in the future
- [ X] Check the output of the Travis automated check scripts - fix any errors as required
Still get errors for point F9.3


poeschlr:
The 3d settings you used are wrong. We require scaling equal to 1, rotation and offset equal to 0.
Also the directory 3d path is wrong as the 3d library is required to have the same name as the footprint library.

---

The outline on the fab layer should represent the exact body dimensions.
The silk outline should be fully outside the body. (Take a look at other similar parts as an inspiration)
The silk line width should be 0.12mm
Place the value field such that it is not in the way of other stuff.
Reduce the font size of the second reference (the one on the fab layer) such that at least 3 characters find space within the body dimensions.

For reference the dimensioned drawing of the footprint:
screenshot from 2018-08-19 02-38-29

penoud:
I updated component according comment. But I have question regarding the 3d package. I found one on the 3dcentral web site in STEP. Could I use it or I should absolutly redesign the 3d wiht Freecad 0.17.

poeschlr:
Regarding 3d model: Submitting a 3d model is not a requirement for this being included in the library. the only thing that needs to be correctly done is setting up the footprint such that it can accept a 3d model without modification.

We can not accept 3d models from 3d model sharing sites as their license is not compatible with the license used in our repository. (For most such sites uploading them to github would already be a violation of their terms. So you would violate their license as soon as you open a pull request.)

---

You did not fully fix the 3d model settings.

Something went wrong with your fixes regarding body size and courtyard clearance. They are unchanged.

penoud:
I update the component according last comment. I made the 3D model with Freecad. I will open a PR for 3D models.

penoud:
Arg something went wrong with the orientation. I will correct it.

penoud:
Corrected the 3D package orientation.

PR #843, opened 2018-08-19
Rounded rectangle pad smd single row + mounting pad connectors

poeschlr:
This converts all footprints generated with the "SMD single row plus
mounting pad" generator script over to using rounded rectangle pads.

script changes: https://github.com/pointhi/kicad-footprint-generator/pull/162


evanshultz:
Some JST ZE are not happy:
image

Molex Pico EZmate also has a slightly wrong offset on the angled silk lines but it's not a big issue:
image

poeschlr:
The JST ZE should not be part of this pull request. What you show is how it always looked like.
The Silk offset of the pico EZ did not change compared to the past. (I just could not be bothered to calculate the correct offset correctly. Too much work for too little payoff.)

poeschlr:
I looked at it again. There are JST ZE in this pull request but the one you show in the screenshot is not one of them. I will split this PR into three in the hope that github can deal with that better. (Right now i can not even check what has changed as github refuses to show it because "the list is too long")

PR #846, opened 2018-08-19
Rounded rect Molex smd single row + mounting pad conn

poeschlr:
This converts all Molex footprints generated with the "SMD single row plus
mounting pad" generator script over to using rounded rectangle pads.

Partly replaces #843 (That pull request was too large for github. Split up by manufacturer)


evanshultz:
Centering errors are bogus. Some of them have pin 1 in the lower left corner instead of upper left, but that's not worse than they were before.

PR #862, opened 2018-08-24
Add Kyocera 5638 SD card holders

evanshultz:
https://global.kyocera.com/prdct/electro/product/pdf/5638.pdf

image
image

Notes:
- I kept the origin at the center of the SD card as that is the centerline of dimensioning in the drawings. This makes dimensioning easier and since our scripts only consider pad locations the centering is fairly meaningless of these footprints.
- There are several layers used to indicate card placement in this library. I chose F.Fab which was used in the majority of footprints.

------------


evanshultz:
These footprints should go with the following symbol, but it hasn't gotten pin names updated to SH:
image

evanshultz:
Dimensioned drawings of the top two in the image above:
image
image

The bottom two are just mirrored and shifted versions of the top two so the dimensioning isn't duplicated.

Misca1234:
Brutal, you even got the x position difference between the HS-pad and the other to be 0.05 correct.
 an pin 10 smaller on SD_Kyocera_145638109211859+

Strange that SD_Kyocera_145638109511859+ do not have a keep out area as the others

But there is on position that I am uncertain of
If I try to calculate the x-position for the upper right pad of SD_Kyocera_145638109511859+
and starts from the upper left npth hole I get the x pos for the pad to 
-12.1 + 25.6 + 0.9 = 14.4 but the pad is at 14.1 ?

bild


Otherwise everything looks ok

Misca1234:
hmm, the arrow seems not go to the npth hole center

bild

Would you agree to that, if so, all measurements are ok

evanshultz:
Thanks for reviewing this!

The "109" parts do not need a keepout because the card is flipped and the metal contacts are not close to the PCB.

Agreed the arrow is not to the NPTH pad center. You've got me questioning the pad placement now, but I think I'm just getting paranoid (and the drawing isn't super clear) so if you don't find any issues go ahead and merge.

PR #863, opened 2018-08-24
Footprint for Taoglas GPS antenna

jneiva08:
added footprint for 2 GPS antennas from Taoglas

Taoglas CGGBP.18.4.A.02
datasheet: http://www.taoglas.com/images/product_images/original_images/CGGBP.18.4.A.02.pdf
symbol:
gps_18_2

gps_18

Taoglas CGGBP.25.4.A.02
datasheet: http://www.taoglas.com/images/product_images/original_images/CGGBP.25.4.A.02.pdf
symbol:
gps_25_2

gps_25

------------


poeschlr:
Is there a reason for choosing 0.5mm courtyard clearance?

jneiva08:
No reason at all, should be 0,25?

poeschlr:
yes unless there is a special reason, kicad footprints should use 0.25mm courtyard clearance

poeschlr:
The y position of the pad relative to the outline is not quite right for the smaller of the two parts:

screenshot from 2018-08-27 09-12-50

jneiva08:
I did check the math it appears to be fine.
Next week i will try to verify on frrecad mod (drawing dimensioning) in my work pc, in my old laptop it don't works well.

poeschlr:
Did you intent to push a change that goes with your comment?

Misca1234:
Should there not be a ground plane as well, like this ?

bild

jneiva08:
I have small experience is patch/ceramic antennas.
The few testes i did with iridium antenna, i did have better better if the ground plane/area was bigger. Normally the antenna recommend the best area.
In the following image image can see the top layer. Only call attention for arround the solder pin.
top
If you want can post two pcb's using this two gps antennas.

Shackmeister:
I believe the ground plane shouldn't be part of the footprint.

Misca1234:
@Shackmeister 
According to page 13 they recommend a ground plane, on both sides of the PCB, I guess the images on bottom of page shows top and bottom side

Misca1234:
@jneiva08 
Shackmeister is correct, neither the ground plane or the suggested pin 2 should be included in the foot print, it is up to the user to define the ground plane later.
So remove the copper around the pin

jneiva08:
@Misca1234 
The ground place and pin 2 was never added on the footprint.
I'm thinking changing pin 1 to be only B.Cu with B.Paste and B.Mask. i don't know how should mark the limit of clearance, probably with two circles in F.SilkS and B.SilkS.
What is your opinion?

Misca1234:
To follow Shackmeisters recommendation it would mean not to even include the copper plane, just the "pin"

chschlue:
@jneiva08 reopen if you're still on this.

PR #831, opened 2018-08-17
Added variant of DFN-8 3x3mm

SchrodingersGat:
* Datasheet - http://www.ixysic.com/home/pdfs.nsf/www/IX4426-27-28.pdf/$file/IX4426-27-28.pdf
* Used round-rect pads

image

PR #865, opened 2018-08-25
Add scripted DFN-8 3x3mm 1.7x2.05mm EP for IX4426-27-28

poeschlr:
This replaces #831

The script stuff is included in my larger scripting pull request https://github.com/pointhi/kicad-footprint-generator/pull/149

Script parameters:
``` yaml
DFN-8-1EP_3x3mm_P0.65mm_EP1.7x2.05mm:
  device_type: 'DFN'
  #manufacturer: 'man'
  #part_number: 'mpn'
  size_source: 'http://www.ixysic.com/home/pdfs.nsf/www/IX4426-27-28.pdf/$file/IX4426-27-28.pdf'
  # ipc_class: 'qfn' # 'qfn_pull_back'
  #ipc_density: 'least' #overwrite global value for this device.
  body_size_x:
    nominal: 3
    tolerance: 0
  body_size_y:
    nominal: 3
    tolerance: 0
  lead_width:
    nominal: 0.3
    tolerance: 0.05
  lead_len:
    nominal: 0.4
    tolerance: 0.05

  EP_size_x: 1.7
  EP_size_y: 2.05
  # EP_paste_coverage: 0.65
  EP_num_paste_pads: [2, 2]
  # heel_reduction: 0.1 #for relatively large EP pads (increase clearance)

  pitch: 0.65
  num_pins_x: 0
  num_pins_y: 4
  #chamfer_edge_pins: 0.25
  #pin_count_grid:
  #pad_length_addition: 0.5
  #suffix: '_PullBack'
  #include_pad_size: 'fp_name_only' # 'both' | 'none' (default)
```

screenshot from 2018-08-25 10-37-42


evanshultz:
1. No thermal vias version?
2. Since the package shows a 1.65x2mm nominal pad, even though their reference footprint is 1.7x2.05mm, should the EP on the footprint match the EP on the package as decided previously?

poeschlr:
@SchrodingersGat did not include a thermal via version that is why i did not add one.

The thermal pad is created in accordance with the recommendation of the datasheet. (In this case the datasheet suggests the use of the maximum size.) Neither nominal size nor the recommended size fit with what oliver used.

evanshultz:
But if we're using a standard IPC generator, and not what the vendor suggests, shouldn't we at least be consistent? If your nice IPC generator is using the nominal package EP dimensions elsewhere, why not use that here as well?

poeschlr:
I used the suggested EP size for all footprints where there was a suggestion in the datasheet. (Look for example at the microchip stuff. There are a lot of them who do not use nominal but either min or max EP size depending on what was suggested.) I mostly started it that way to avoid changing too many symbols. And now i kind of set a precedent that the datasheet suggestion is ok as long as it is within the tolerance of the EP.
To be honest i do not have strong opinions for new additions. Existing footprints however should if possible not change their name till we start with v6 development.

evanshultz:
OK. I didn't see this philosophy written down anywhere so thanks for stating it.

poeschlr:
Yea the klc is a bit lacking right now as i was focused on other stuff.

PR #872, opened 2018-08-26
add lvk12 4 contacts resistor

penoud:
add lvk12 4 contacts resistor. Datasheet reference link bellow.
------------

https://www.ohmite.com/assets/docs/res_lvk.pdf
image
Yes PR #382 in the package-3d repository.


poeschlr:
There should not be silk below the part as it would increase the risk of tombstoning. (There is no need for marking pin 1 as these parts are symmetrical.)

The courtyard clearance should be 0.25 (+/- 0.005)mm all around.
The outline in y direction is not quite right.
The pad spacing in x is not correct.

Add the manufacturer name as prefix to the footprint name.

screenshot from 2018-08-26 23-28-40

penoud:
New screenshot:
image

poeschlr:
Another suggestion: rotate the reference on the silk layer by 90 degree. This could allow it to use a larger font size while still fitting inside the outline.

penoud:
Change according last comment. New screenshot:
image

poeschlr:
I meant the reference that is shown in gray (fab layer) not the greenish one (silk layer)

penoud:
Done. 
I got it now. I was not sure but your ref to silk confuse me ;-)

poeschlr:
thanks

PR #873, opened 2018-08-26
add VSSOP-10 0.5mm pin pitch

penoud:
add VSSOP-10_3x3mm_P0.5mm pin pitch. Reference datasheet in following survey. 
------------

http://www.ti.com/lit/ds/symlink/ads1115.pdf
image
https://github.com/KiCad/kicad-packages3D/pull/392


penoud:
Should I move the footprint or the 3d models. The footprints created with the script generator for Kicad place the 3d models in the folder named Package_SSOP.3dshapes. But the previous VSSOP-8 in the footprint were placed in the library Package_SO.pretty.

Misca1234:
Do you have the link to the script that made the foot print ?

penoud:
I didn't find the script. I've done a new one. Here the link to the PR
https://github.com/pointhi/kicad-footprint-generator/pull/209
Updated screenshoot:
image

Misca1234:
> I reduced manually the font size from 0.75 to 0.65 for REF** after script generation.

Don't do that, we trust the script 100%, so just accept what ever the script output.

Of course there is an exception to this, and that if you find something that is wrong, 
then you should bring up the issue and the script can be corrected, try to avoid to "hand fix" them afterwards.

The reason is that if we find an error in the script later on re generate all the foot prints, and this foot print is not affected by the error, the script should produce a foot print equal to the one have been merged.

penoud:
Ok make sense. 
I revert to original font size:
image

poeschlr:
Why is the second reference on the bottom fab layer instead of the top one?

If you used the ipc scripts could you post the relevant section of the size definition yaml file here?

poeschlr:
Looks good, thanks

PR #491, opened 2018-04-15
Add KLS5-CR1220-01 Battery Holder

dsa-t:
Datasheet: http://www.klsele.com/admin/product_upload/20151124100234KLS5-CR1220-01.pdf
batholder

------------

Thanks for creating a pull request to contribute to the KiCad libraries! To speed up integration of your PR, please check the following items:

- [x] Provide a URL to a datasheet for the footprint(s) you are contributing
- [x] An example screenshot image is very helpful 
- [ ] If there are matching symbol or 3D model pull requests, provide link(s) as appropriate
- [x] Check the output of the Travis automated check scripts - fix any errors as required

PR #876, opened 2018-08-27
Finish #491: Add KLS5-CR1220-01 Battery Holder

poeschlr:
This replaces #491 

Datasheet: http://www.klsele.com/admin/product_upload/20151124100234KLS5-CR1220-01.pdf

screenshot of the fixed footprint:
screenshot from 2018-08-27 11-15-54

The area covered by the spring is not really included in the drawing. So i measured it directly from the dimensioned drawing. (Using pdf scaling and a caliber on the monitor.) I added a lot of tolerance to what i determined and drew it that way. (The vertical line is elongated by 2mm and connected to the normal octagon outline.)


dsa-t:
Note that the drawing does not match the actual part. The center of the circle should be on pin 2. So there is more work to be done. I can provide photos and part scans later. For now you can verify that on google images if you don't have the part.

poeschlr:
If we can not trust the official datasheet then we might not want this part included in the official library.

Misca1234:
Should there not be holes for the "mechanical plastic pins" as well ?

PR #879, opened 2018-08-27
Add M.2 connector footprint - socket

blazra:
Hi,

adding a footprint for the M.2 socket. All the M.2 connectors I can find have the same footprint, except for the midplane type. The shape of the plastic differ across manufacturers, but only in details.

TE Connectivity: http://www.te.com/commerce/DocumentDelivery/DDEController?Action=showdoc&DocId=Customer+Drawing%7F2199230%7FB2%7Fpdf%7FEnglish%7FENG_CD_2199230_B2.pdf%7F1-2199230-6
Amphenol: https://www.amphenolcanada.com/ProductSearch/drawings/AC/MDTXXXXXX001.pdf
JAE: https://www.jae.com/z-en/pdf_download_exec.cfm?param=SJ113503.pdf

m 2_screenshot


herostrat:
Hey, seems fine for me from the size

grafik

Do you plan to make a symbol too?

blazra:
About the symbol: I am not sure, because there are at least 4 pinout versions - one for each of the key/notch positions (A, B, E, M). Also, you can use the connector as a generic edge connector, which is actually, how I am using it. So, maybe I could make the generic one? Is it worth it?

blazra:
- :thinking: Just realized that maybe making a generic symbol for all of the key options would be confusing. 

- There are also combinations of keys. :open_mouth:  I have no idea how that should be represented by the symbols.

- On another note, should I add the word "socket" to the footprint name? To differentiate it from potential footprint of the card edge itself.

blazra:
So, I have made a symbol for the M key version. :muscle:
https://github.com/KiCad/kicad-symbols/pull/869

herostrat:
Hey, I do not know whats the best way to deal with such special connectors is.
For my personal use, I like dedicated (named) connectors in order to be sure that the pin-symbol connection is save and sound.

I guess the generic version should be sufficient for normal use cases, because a designer should know the pinout specification of the product himself.
The reason I personally would make a dedicated generic symbol for this connector is that 

 - there is a specific reason one would choose this connector type
 - the numbers of pins is really special and imho not covered by the standart con library
 - one can be a bit more flexible for geometry (e.g. make a scheme of the outline for the shape, so that that the user can better visualize it)

I made one one symbol for SODIMM sockets, where I just used the pinnumbers as names (because there are also a huge number of different use-cases)

I am not sure what the official position is and what is generally accepted as solution, as I am also only a contributor myself

poeschlr:
We can not provide a specialized connector symbol for every type of connector out there (We would need 2000 symbols just for the current Molex, JST and Phoenix connectors plus another 1000 for the rest. That would be a bit much to maintain.)

So we normally use generic symbols for connectors.
We have symbols for dual row connectors with an odd number of pins but the highest pin count is 69. We can however easily add a few more of these (The cutoff was chosen somewhat arbitrarily.)

---

How does the pin-out change with the notch position? I would assume that this would mean that some pads would then not exist. This means one would need one footprint per notch position. If we decide to include these footprints then we would need specialized connector symbols. I would however put them in a separate symbol library (As of yet i have not really an idea for how to name it.)

blazra:
- Dual row 75 pin generic connector symbol would be nice - that would actually cover my use case. Also, it would be correct physical representation of the connector.
- Here are the pinouts depending on the notch position: 
abemkey_pinout
- So the name should be something like: "M.2-NGFF_2Rows-75Pins_P0.5mm_Socket" ?

herostrat:
Hey blazra, sorry for the late reply!

 - I modified the script that generates the symbols to include 75 pin generic connector
See: https://github.com/KiCad/kicad-library-utils/pull/272

 - The addition NGFF could be unnecessary, but I guess we have to ask the maintainers (quick shout-out to @poeschlr ).

You should sign the license agreement (see lincense/cla in the checks box) if you agree and want to merge something afaik.
After we get confirmation on the name I think this PR mergeable imho.

myfreescalewebpage:
Hello @blazra 

I agree with @poeschlr but this connector is very specific and used only for M.2. A specific symbol (https://github.com/KiCad/kicad-symbols/pull/869) for each keying is a good choice to help people designing such kind of card.

I'm currently working on MiniPCIe symbol (https://github.com/KiCad/kicad-symbols/pull/1196) and footprints (https://github.com/KiCad/kicad-footprints/pull/1124) and we had some exchanges with @evanshultz. The solution chosen is that the symbol has no default footprint, so it's possible to use any connector from any manufacturer. A generic footprint is defined (that's my PR). Also I have designed several footprint to easily place screws on the PCB where to mount the M.2 card. Have a look to my PRs, should be good if we found some consistent solutions for both buses :)

Cheers
Joel

PS : I'm currently reviewing your M.2 key M symbol.

evanshultz:
Let's stick with the footprint first and then consider the symbols (we usually don't merge symbols without footprints, but these are generic).

I reviewed footprints in the datasheets you linked above and found just one difference: Vertical from the 1.1mm hole to the top of the upper pad row is 6mm on JAE and 6.05mm on others. Are the parts that fit this footprint all built to the same standard and so should be identical? And therefore, one 3D model should suffice (if it's ever added)?

NGFF was the proposed name, with M.2 being the "official" name, I understand, so let's dispense with NGFF in the footprint name. I think it's much more clear to just notate the keying and get rid of the rows and pins. Would `M.2_M` and `M.2_B`, etc. work? It seems sparse but I think it captures all the info. Are all the key letters mutually-exclusive so this works?

Can you beef up the description by adding the keying type, rows and pins, etc.?

blazra:
Is this PR still relevant when there is the https://github.com/KiCad/kicad-footprints/pull/1404 ?

One 3D model wouldn't suffice, because the connector exists in a few different heights.

I just found out that AVX/Kyocera makes M.2 connectors too. They have the nicest drawings and have all connector versions in one document. Also, they nicely indicate the key positions in footprint drawings. https://global.kyocera.com/prdct/electro/product/pdf/6411.pdf

Yes, the keys are mutually exclusive.

I like the simple naming, but the midplane (H1.8) type has a different footprint. See the Kyocera doc above or TE's [ENG_SS_114-115006_B_114-115006_REV_B.pdf](https://www.te.com/commerce/DocumentDelivery/DDEController?Action=showdoc&DocId=Specification+Or+Standard%7F114-115006%7FB%7Fpdf%7FEnglish%7FENG_SS_114-115006_B_114-115006_REV_B.pdf%7F2199230-3). So maybe, it would be nice to indicate that somehow. Maybe list the heights or make something like aliases for different heights.

myfreescalewebpage:
@blazra if you think @1404 cover your PR, you have the choice, you have made your contribution before.
I think a generic footprint is good (that follow the official spec), then some specific from manufacturers forp people who want to submit them.

blazra:
I think #1404 covers this PR.

Personally, I would add the #1404 footprints as generic ones - not just as a part from TE...

myfreescalewebpage:
Ok, thanks for the update @blazra so we will will focus on #1404.
@evanshultz PR closed (duplicated) suggested by the original author.
Thanks,
Joel

PR #883, opened 2018-08-28
Finish #584: Würth 7499010121A ethrenet magjack

poeschlr:
This finishes the remaining points of #584 
Symbol PR: https://github.com/KiCad/kicad-symbols/pull/603

Datasheet http://katalog.we-online.de/pbs/datasheet/7499010121A.pdf

screenshot from 2018-08-28 02-12-45


Shackmeister:
Hey @poeschlr 
Can you please adress the following;

- Courtyard not on grid (as travis says)

- Chamfered corner on the fab line isn't really pointing to pin 1, a triangle would probably fit better

- I think we usually only use a line to indicate pin 1 on silk, but I might be wrong

- a few silk lines are hitting the solder mask with the 0,2mm clearance

poeschlr:
Fixed:

screenshot from 2018-09-06 23-08-15

Shackmeister:
@poeschlr 
is it on purpose you set the pad clearance to 0,2mm? or was it only intended for checking? also should Travis accept this?

and if we are being extremely pedantic, this is still in the 0,2mm zone :p
rj45_wurth

poeschlr:
Are these pegs connected to the shield? Or are they only for adding mechanical strength?

cp-aquila:
They definitely are requited for mechanical strength.
And they are connected to the metal housing as well. So I would say yes, they are part of the shield.
That would mean the symbol needs another pin :/

PR #703, opened 2018-06-27
added Footprint for TE Miniature PCB Relay PCH_A

daviid5:
added Footprint for TE Miniature PCB Relay PCH_A

http://www.te.com/commerce/DocumentDelivery/DDEController?Action=showdoc&DocId=Data+Sheet%7FPCH_series_relay_data_sheet_E%7F1215%7Fpdf%7FEnglish%7FENG_DS_PCH_series_relay_data_sheet_E_1215.pdf

Regarding Pullrequest kicad-symbols #704 https://github.com/KiCad/kicad-symbols/pull/704

te_pch_a_footprint

------------

Thanks for creating a pull request to contribute to the KiCad libraries! To speed up integration of your PR, please check the following items:

- [x] Provide a URL to a datasheet for the footprint(s) you are contributing
- [x] An example screenshot image is very helpful 
- [ ] If there are matching symbol or 3D model pull requests, provide link(s) as appropriate
- [ ] Check the output of the Travis automated check scripts - fix any errors as required

PR #884, opened 2018-08-28
Finish #703: TE Miniature PCB Relay PCH_A

poeschlr:
This finishes the remaining points of #703 
Symbol PR: KiCad/kicad-symbols#704

Datasheet: http://www.te.com/commerce/DocumentDelivery/DDEController?Action=showdoc&DocId=Data+Sheet%7FPCH_series_relay_data_sheet_E%7F1215%7Fpdf%7FEnglish%7FENG_DS_PCH_series_relay_data_sheet_E_1215.pdf

screenshot from 2018-08-28 02-53-24

screenshot from 2018-08-28 03-13-37


Shackmeister:
Hi @poeschlr 

Can you please fix the Travis outputs?

poeschlr:
Fixed

myfreescalewebpage:
Hello,
@antoniovazquezblanco I think this PR is ready to be merged.
Cheers,
Joel

myfreescalewebpage:
@antoniovazquezblanco  up ?

antoniovazquezblanco:
It's been a while since the last time I reviewed a footprint. I will find some time to get my skills back in shape this holidays. I will assing this to myself as soon as I start.

evanshultz:
Since the dimensioned footprint is above reviewing is easy, assuming you trust Rene. :)

@poeschlr 
The link in the footprint is to the product page. Can you instead use the (annoyingly long) link you posted above in the PR body?

Also, the description is missing the "1" in the relay type and could perhaps be organized differently. I suggest `Miniature PCB Relay, PCH Series, 1 Form A (NO), SPST` and then the link.

The holes are a bit big. For the coil, 0.45mm square gives 0.64mm across so 0.74mm holes should be fine by IPC 2221 Level B. (0.8mm would be OK.) For the contact pins, they're 0.8mm so 1mm drill is good. Then the pads should then be 0.5mm bigger in diameter than the holes. In all cases this is bigger.

Lastly, the coil voltage is taken as a single entry in the PN and can be a single digit ("5"), so the double "x" may not be the best choice. If we're cutting off everything after "M", then I suggest only "xx" and not "xxx" in the footprint name.

poeschlr:
Regarding holes: These are the suggestions from the datasheet. And if we did not use the datasheet then we would use KLC which suggests diameter increase from maximum lead dimensions by 0.2mm. (We don't really know the maximum diameter as there is no info about tolerances in the datasheet. Neither for the lead sizes nor for the lead positions.)

If i read the datasheet correctly then the coil voltage codes are always two chars long (there is an example given for 5V where the code is 05. One of the examples is:  `PCH-105L2M,000`

evanshultz:
OK. You're right about the MPN (I missed the "05") and the footprint is fine then. I also made a typo above which made my comment confusing.

You would mind changing the URL and description as I mentioned above? I think it's good then.

poeschlr:
Fixed the description as requested.

evanshultz:
Thanks!

PR #886, opened 2018-08-28
SOP-16_7.5x10.4mm_P1.27mm

hildogjr:
I used the SOP-16_4... to create this "wide" footprint version used by the IC https://docs.broadcom.com/docs/AV02-0717EN.
I used a 3D already present in the first footprint just re-scaling in one dimension.


SchrodingersGat:
Thanks for the contribution @hildogjr

PR #888, opened 2018-08-28
Vishay WSR2/3 footprint

jneiva08:
Datasheet: https://www.vishay.com/docs/30101/wsr.pdf

Footprint:
r

r2

------------


poeschlr:
Did you link the correct datasheet? At first glance the suggested footprint in it looks very different to what you submitted here.

jneiva08:
fixed datasheet

poeschlr:
@jneiva08 sorry for the late answer.

I would assume your small extension to the pads is meant as the connection point for the sensing connection when using this as a shunt. If this is the case then using a separate pad for it might be best.
This would provide a point to attach a second trace. (It would still need the same pin number assigned as the pads are touching.)

jneiva08:
@poeschlr sorry for bothering.
When added the footprint added that extra pad like datasheet recommended.
It will be wise to remove or add extra pad like you say, but without F.Mask?

poeschlr:
Why not both?
Add an option with extra pad and give it a suffix _KelvinConnection (The second pad must sadly have the same pad number as the first one.) And an option without it allowing users to use this part like a normal resistor.

jneiva08:
@poeschlr  have made the alterations. need add the new screenshots and measurements?

poeschlr:
That would help a lot.

jneiva08:
Screenshot for R_Shunt_Vishay_WSR2_WSR3:
wsr

Screenshot for R_Shunt_Vishay_WSR2_WSR3_KelvinConnection
wsr_2

Sorry i cant put the measurements because the freecad is giving segmentation fault :(

poeschlr:
Looks good, thanks

jneiva08:
thanks

PR #890, opened 2018-08-29
Add inner courtyard to RF shields

### LINKED ISSUE #77: Some RF shielding footprints miss an inner courtyard.
Reported by @pointhi in the pull request that added these components here. https://github.com/KiCad/kicad-footprints/pull/75

poeschlr:
Fixes #77


Shackmeister:
Hi @poeschlr 
should this be merged as is, or should the KLC erros be fixed before merge?

evanshultz:
@poeschlr 
Proxy ping

evanshultz:
Sure. Take your time. I'll check into this and see what response I can get in the meantime.

evanshultz:
I talked to Laird. Reflow soldering is the common way for all SMT shields. Sometimes due to the need for visual inspection, AOI, or repair a two-piece or "openable" shield is used instead of a one-piece, but still reflow soldering is chosen. So we should put paste on all pads.

Here are a few things I noticed without a super thorough check of all footprints:
- 97-2002: The inside top left courtyard lines don't have the same endpoint.
- 97-200*: Is the wall thickness 0.13mm? If so, 97-2002 pad centers should be on 12.76mm square but they're currently 12.635mm. 97-2003 is also off.
- If we know wall thickness, should the fab layer have one thick rectangle, two rectangles showing both edges, something else? I only see a possible wall thickness dimension on the 97-200* ones.
- Should the "location holes" be marked, perhaps on the fab layer?
- These all have a square pad in each corner that should not be required if two rectangle or a single poly pad is used.
- Fab and silk lines are 0.15m thick.

evanshultz:
And none of them have a 3D model.

evanshultz:
@poeschlr 
Do you have time to work on this PR yet?

poeschlr:
I have not yet found time in my schedule to fix this PR.

I can however answer a few of your points:
>These all have a square pad in each corner that should not be required if two rectangle or a single poly pad is used.

I assume this is to get more connection points for thermals or traces.

> If we know wall thickness, should the fab layer have one thick rectangle, two rectangles showing both edges, something else? I only see a possible wall thickness dimension on the 97-200* ones.

My guess would be that a lot of them have a width of less then the default line width. If they have more than twice that then a second rectangle could make sense. But to be honest i assume the outside line would be enough to communicate what is going on.

> Should the "location holes" be marked, perhaps on the fab layer?

This will definitely need to be done in a separate pull request. (Would require going through all datasheets one more time and adding the details to the fab layer. A bit more than i am prepared to invest into this.)

poeschlr:
Fixed as suggested.

Regarding pad centers: I checked with a freecad sketch against the dimensions of the datasheet, looks like the footprint is correct:
screenshot from 2018-10-09 16-08-45

evanshultz:
On the basis of "some fix is better than nothing" I'll merge. I'm not sure anything else is important to do so I'll leave it as is.

PR #891, opened 2018-08-29
KLC fixes for OptoDevice lib

### LINKED ISSUE #124: Footprint naming and placement of second reference designator not ideal in OptoDevice lib
Examples for badly placed second references:
screenshot from 2017-11-28 10-11-39
screenshot from 2017-11-28 10-11-04
screenshot from 2017-11-28 10-10-47

Naming convention is a bit harder. We need to find a system for part prefixes that would group similar devices together

poeschlr:
fixes #124

- Move auto generated fab references to better places
- If necessary fixup silk and fab layers
- For some parts the courtyard layer needed fixing as well


evanshultz:
I propose to look at the footprints visually and not dimensioning every one. Is that OK with you?

Do you want to do more? Travis found line widths, 3D model file extension, etc. that's still wrong. This can just be partial fixes if you want to merge as-is.

poeschlr:
Yes a visual check should be enough. I did not check any dimensions of any of the parts (I only fixed courtyards where they where so obviosly wrong that i could see it from the footprint browser)

poeschlr:
- Fixed 3d file ending
- Annular ring size for LaserDiode_TO3.3-D3.3-3 and LaserDiode_TO38ICut-3
- Fixed linewidth for fab outlines (search and replace for all footprints)

poeschlr:
I want to avoid unnecessary renames such that we can supply bugfix updates of the library together with kicad bugfix updates. (Wayne was quite clear about the fact that we can not supply newer versions of the library if it includes changes that unnecessarily break old designs. I hope we can sell the changes we already made and just promise that we will be a lot stricter in the future.)

So the only reason i would rename a part (from now on) is if the current name is misleading. (For everything else i would document needed changes somehow for when we start with v6 development)
The alternatives are either users do not get bugfix updates of the lib or we need to maintain two branches. (The later is too much work i think. I would really like it if users can get better quality libs by default.)

Similarly with rotating (or moving) of footprints. They could break setups of users who have personal 3d models (Or if we have a 3d model then it would break old boards who have not updated the footprints on the board)

---

the thing in ADNS-9800 is not a keepout zone. It is a edge cut pocket as this part looks through the pcb.

I will go through all the parts one more time and fix what i can see. (including changing smd pats to rounded rectangle)
Right now it seems every part i look at has some problems with the courtyard area. This one lib really is one of the worst libs in the library.

evanshultz:
Here are still some items, but I think this is a huge improvement and we go ahead now. Please confirm I should make an issue for the items below and merge this PR.
- Still many pads that could be round rect (like Maxim OLGA)
- Inconsistent pin 1 silk marks
- ADNS-9800 has no vendor name prefix (PixArt, it seems?)
- AGILENT_HFBR-152x and HFBR-252x look the same (and don't need Agilent in all caps)
- Everlight (and others) don't have pin 1 at top left

Thanks!

poeschlr:
As i said above:
Renaming footprints should be avoided unless the current name could mean that users confuse the footprint with another part (So the points about ADNS and the two HFBR parts can not really be fixed with that restriction)

Fixing rotations can also break existing designs so also not advised.

---

Which leaves the pin1 markers and the one smd part which could need rounded rectangle pads. I will look into these two points.

poeschlr:
There is now another round of fixes done.

evanshultz:
And as I said above, I think we can merge as-is and put my bullet list as an issue. Here are a couple new things I noticed after reviewing the last commit:
- Agilent footprints have an apparent break in the silk:
image
- Vishay CNY70 pin 1 isn't roundrect

Otherwise it looks like great improvements and I'll round up the outstanding items in an issue and merge what you're already done.

evanshultz:
One more thing: Travis mentioned the Luna footprint doesn't have the standard 3D model params and I don't see a model at https://github.com/KiCad/kicad-packages3D/tree/master/OptoDevice.3dshapes so it should be safe to correct now.

poeschlr:
I fixed some more things.

Regarding issues: I already started one for reminding us that this lib needs a better naming convention in general. If you find something else then we might open one issue per topic.

PR #893, opened 2018-08-29
Fixup button MEC 5GTH9 (fixes #796)

### LINKED ISSUE #796: SW_MEC_5GTH9 incorrect pad spacing
SW_MEC_5GTH9 pad spacing should be  400 x 300mil  (10.16 x 7.62mm) as per the spec sheet, current footprint in the Buttons library is 450 x 300mil.

I should have paid more attention when the pads didn't fit the 3D model pins (which are correct), but it was my first attempt at using KiCad and I scaled the model! Switches don't fit the holes in my PCB now, a good lesson!
Local fix was easy enough, I'm starting to like the tool-set. I'm not up to pushing a library fix as yet though.

Edit: Added data sheet
[MEC_5G.pdf](https://github.com/KiCad/kicad-footprints/files/2239588/MEC_5G.pdf)

poeschlr:
- Fix pad placement (both x and y pitch was wrong)
- Change pad numbers such that they fit our symbols
- courtyard clearance
- Fix body outline (was too small) on silk and fab layers

The datasheet can be found here: https://cdn.sos.sk/productdata/80/f6/aabf7be6/5gth9358222.pdf
This fixes #796

screenshot from 2018-08-29 11-47-42

screenshot from 2018-08-29 11-42-18


Shackmeister:
Looks good, Thanks Rene

Slightly off topic, I was visiting MEC a few weeks ago, (their factory is located 5 km from my work) and they seemed very interested in getting their footprints and maybe also their 3D models in to the kicad libraries

poeschlr:
If they can supply them under our license then they are welcome to do so. If they do not want to purplish them under our license then they could make their own library under their own license and request including theirs under the third party library list.

PR #896, opened 2018-08-30
Add footprint for Mini Circuits FG873

poeschlr:
Datasheet: https://ww2.minicircuits.com/case_style/FG873.pdf

This is the footprint for https://github.com/KiCad/kicad-packages3D/pull/365

I am not sure the DFN_QFN lib is the right one for this part. I think we might want to add some library that holds specialized packages that just don't fit anywhere else.

screenshot from 2018-08-30 13-32-22

screenshot from 2018-08-30 14-14-34


poeschlr:
Why should the area be increased that way. IPC now allows closely fitting courtyard areas, kicad can deal with it so i don't see a reason why it should not work as i made it. And we use non rectangle courtyards for other devices as well. (Check my recently merged qfp footprints for an example)

Misca1234:
Sure, let us go with that

PR #897, opened 2018-08-30
Added BatteryHolder_Keystone_1060_1x2032.kicad_mod

5N44P:
added Connfly battery holder DS1092-10 for a 2032 cell.
Battery outline is bigger than the footprint, so i've added it to DWGS.user layer as in other pre existing footprints (e.g BatteryHolder_Keystone_1058_1x2032)

image

Datasheet: https://datasheet.octopart.com/DS1092-10-W6L-Connfly-datasheet-81454616.pdf

------------


5N44P:
I found out also that Keystone 1060TR is VERY similar. I'll compare the measures asap to check for full compatibility. 
If compatibility is confirmed, should i Change the name of the file to reflect the more popular brand and model? 
I think the one i referred to in the PR might be a chinese "compatible" 

https://www.digikey.com/product-detail/en/keystone-electronics/1060TR/36-1060CT-ND/4935558

Misca1234:
> Should we make one part for every manufacturer?
In principle yes, certainly if they differ in measurements.
but
what you could do in this particular case is to add ConnFly as an alias to the already existing 
Keystone symbol, then people can choose which one to use when they generate the fp and bom list.

5N44P:
@Misca1234 thank you for your answer. I have only two doubts: 
> the already existing Keystone symbol
Have i missed it? I can't find the 1060 in the current footprint lib. Should i rename the current and commit here? 
> add ConnFly as an alias 
Is the alias something supported by kicad? Or you mean just adding as a footprint?

Misca1234:
>> the already existing Keystone symbol
> Should i rename the current and commit here?

Sorry, I did not do my background research enough, I just assumed there was a keystone one.
You do not have to rename it, if someone want to add keystone variant the can add it as an alias to yours.
It is up to you, 
if you want to keep the name for the foot print
Keep the name and add keystones as an alias
Change the name to keystone and add ConnFly as an alias to the symbol


> Is the alias something supported by kicad? Or you mean just adding as a footprint?

Yes, for symbols

> So, which one is the original designer?

Who's know, but I have a pretty good idea

5N44P:
All right, i'd prefer to add this as keystone, which seems much more common and the library already has many parts from keystone. 

>Yes, for symbols 

I can't find much information on that. The symbol would be a single cell battery, i'll try to look more into it tonight. 

>Who's know, but I have a pretty good idea

There's  also wurth in the game

5N44P:
I'm closing this since i've renamed everything, including the branch in my fork, and this is no longer valid.
Current proposed PR is #934

PR #903, opened 2018-08-31
Add SODIMM DDR4 260Pins Socket

blazra:
Hi,

adding SODIMM DDR4 260 pin socket.

- It seems that only the TE Connectivity is manufacturing these and I don't know if the footprint is standardized or not so I wasn't sure how to name it.
- Footprint actually fits 2 versions of the socket - 4.0 mm high and 5.2 mm high.
8.0 mm, 9.2mm high sockets and the reverse types have different footprint.
- I am not sure about the anchor position - in the datasheet there is a cross in the middle of the connector body - is that the correct position?

http://www.te.com/commerce/DocumentDelivery/DDEController?Action=showdoc&DocId=Customer+Drawing%7F2309407%7F3%7Fpdf%7FEnglish%7FENG_CD_2309407_3.pdf%7F2309407-1
http://www.te.com/commerce/DocumentDelivery/DDEController?Action=srchrtrv&DocNm=2309409&DocType=Customer+Drawing&DocLang=English&DocFormat=pdf&PartCntxt=2309409-1

sodimm_ddr4_socket


poeschlr:
Sorry for the extremely late review. Your pull request seems to have been overlooked.

I would suggest to rotate the full footprint to result in pad 1 being on the top left corner.
There is no silk layer outline, please add one where there are no pads in the way (outside the body dimension -> outside fab layer)
I found a few measurements that are not quite right. Details see dimensioned drawing.
sodim svg

blazra:
I have rotated the footprint, added silk outline, and fixed the Fab layer.

However, I am not sure about the anchor. I placed it in the center of the whole connector - it looks like it's correct, according to the datasheet.

image

poeschlr:
I think the anchor was ok in the first place. This is because the anchor should be the pick and place origin. Right now the pick and place robot would have no chance to grab the part at the place you use now.

The travis estimation really only works well for symetric parts.

blazra:
That is true. But the datasheet references the "center of the whole connector" to the tape. So if I keep the original "center for pickup", the anchor would represent the position where to grab the part, but there would be no reference to the tape.

In other words. The original center is the correct position where to grab the part, but the robot doesn't know where it actually is. (Unless it uses computer vision or someone sets the offset manually, but in that case it doesn't matter, I guess)

image

PR #908, opened 2018-09-02
Add Valve_Socket_Noval

dsiggi:

        valve_socket_noval
     valve_socket_noval


dsiggi:
Hi,
the travis script says I should add the footprint anchor to pin 1 but that what the klc says is:

> F6.2 - Footprint anchor should be placed in the middle of the component body

poeschlr:
Could you update the screenshot?

From the current screenshot i would suggest to rotate the footprint by 90 degree in clockwise direction.
Pad 1 should be a rectangle or rounded rectangle (the later with 0.25mm radius ratio with maximum 0.25mm radius)
There is no pin 1 marker on silk that is visible after mounting the part.
There is no pin 1 marker on the fab layer.
The circles and numbers on silk are not really necessary. (they will be hidden by the part anyways.) So i suggest to remove them.

Is there a full datasheet? The picture you link does not really dimension the leads. (So i can not really determine if the pad positions are correct or that the drill size is chosen correctly)

I would assume the center of the body is also the center point for the ring of leads. This is not the case in your footprint.
The circle on fab should represent the body size (In this case 25.5mm diameter) You can add the 22mm diameter one in addition to that as a reference point.
The silk outline should definitely be outside the body. So if you want to include the reference circle for the lower part of the body then that silk circle should have a diameter of at least 22.12mm (better still would be 22.22mm. The later would result in a circle completely outside the fab outline)

The courtyard should have a clearance of 0.25mm on all 4 sides. It might be even better to use a circle as the courtyard in this case.

dsiggi:
Here is a new screenshot of the footprint.

valve_noval

dsiggi:
> @dsiggi
> What do you say, do the same with this push as well like the other?

You are right.

PR #910, opened 2018-09-02
Create Potentiometer_Alpha_RD901F-40_Single_Vertical

robkam:
Create Potentiometer_Alpha_RD901F-40_Single_Vertical 
Popular with hobbyist electronics as front panel wiring can be replaced by a PCB.
Pin 1 is round because it's not a polarised component.
Plated through mounting pin holes because less chance of pads lifting during repairs.

http://www.taiwanalpha.com.tw/english2016/p_e_062.htm although a more legible version is at https://www.taydaelectronics.com/datasheets/A-1848.pdf

capture


poeschlr:
Rotate the footprint by 90 degree in counter clockwise direction.
Make pad 1 a rectangle or rounded rectangle (the later with 25% corner size but maximum radius of 0.25mm)
The silk outline should be outside the fab outline (assuming fab outline is drawn correctly to represent the part dimensions)
Add polarity markers on fab and silk layers

robkam:
All the other Potentiometer_THT.pretty pots have a circular pin 1 and none have polarity markers?

robkam:
capture

robkam:
capture

PR #912, opened 2018-09-03
add stocko MKS_16xx connectors

balcer:
Hi, this is may first pull request to kicad repositorium.
Name: Stocko MKS 16xx
Datasheet: https://www.stocko-contact.com/downloads/steckverbindersystem-raster-2,5-mm.pdf
stocko
Script for generating: https://github.com/pointhi/kicad-footprint-generator/pull/169


balcer:
Hi @SchrodingersGat you are right. I fixed the issue. Thanks for the observation.

poeschlr:
The fab outline should represent the nominal body dimensions. (silk should then be outside that outline)

The courtyard rectangle should have a clearance of 0.5mm relative to the body dimension. (for connectors)

screenshot from 2019-01-22 23-57-56


The fab pin 1 marker could be made a bit better i think. Something like this:

screenshot from 2019-01-22 23-51-50

I would also suggest you take a look at some of the other connector footprint generators. (Example JST-XH) We have a lot of useful helper scripts that can be used. (We have for example a central config file that holds KLC specific definitions. These make the scripts much more flexible for future use. This is not a must but would be greatly appreciated.)

PR #915, opened 2018-09-05
Add JST PHD series of connector footprints

evanshultz:
http://www.jst-mfg.com/product/pdf/eng/ePHD.pdf

image

------------


evanshultz:
One question: the datasheet shows 0.5mm square pins with 0.7mm +0.1/-0mm holes in both PHD and PUD datasheets (though PUD doesn't give the pin size) with a caveat that this is a reference. I retained the 0.75mm drill and 1.2mm pads from the PUD script for now. Since the pins are 0.71mm diagonal they should use 0.91mm drill and 1.45mm pads according to the discussion we've had (see https://github.com/KiCad/kicad-footprints/issues/595).

Here is the 8 pin vertical footprint using these larger drill/pad sizes:
image

Anyone against using the larger drill/pad sizes I described above? What about for PUD?

Also, dimensioned drawings of biggest and smallest pin count from both orientations:
image

evanshultz:
@poeschlr 
See my comment above about drill hole size. Should it be larger (and for PUD series also)?

poeschlr:
Drill holes are given in the datasheet. Why overrule JST in this case?

PR #916, opened 2018-09-05
RAC01-xxSGB

Cylindric:
I am adding a RECOM AC/DC converter symbol, footprint and model.

[Official RECOM datasheet](https://www.recom-power.com/pdf/Powerline-AC-DC/RAC01-GB.pdf)

rac01-xxsgb-footprint

[Symbol PR](https://github.com/KiCad/kicad-symbols/pull/896)
[Footprint PR](https://github.com/KiCad/kicad-footprints/pull/916)
[Model PR](https://github.com/KiCad/kicad-packages3D/pull/391)

I don't have Travis but `check_kicad_mod.py` reports no problems.


Misca1234:
Beside exceptions described in the KLC rules, a good rule of thumb is
F.Fab 0.1 width
F.Silk 0.12 with and 0.13 distance from F.Fab
F.CrtYrd 0.05 and 0.12 distance from F.Silk
like this
bild

The distance between pin 1 and 2 is 15.21268, should be 15.24
Pin 4's y coordinate is slightly off, it is -7,59268, it should be -7.62

Misca1234:
If you move pin 3 and 4 to x-pos 28.0 (is 27,94 now) all looks good

Cylindric:
Dammit.  I think that was because the RAC04 is on a 0.1" pitch grid, and I was cleaning _that_ up at the same time.

Misca1234:
Thx for your quick response

Cylindric:
No problem. I'm just hoping I get the workflow sorted a bit at my end so I don't keep spamming silly mistakes into the PRs. Thanks for your patience.

PR #919, opened 2018-09-06
add Amphenol_62684-1x32-1MP_P0.50mm_Horizontal FPC connector

penoud:
add Amphenol_62684-1x32-1MP_P0.50mm_Horizontal FPC connector
------------

https://cdn.amphenol-icc.com/media/wysiwyg/files/drawing/62684.pdf
image
No 


penoud:
I will be up for doing it by script to do the whole familly. But the script I found in Kicad is not suitable. How to create new scripts?

poeschlr:
Take a look at https://github.com/pointhi/kicad-footprint-generator

There are generators for other FFC connectors under Connector_Hirose, Connector_JAE, Connector_Molex and Connector_TE

The closed ones to yours would be the Connector Hirose (FH12) or Connector TE (Series 84952-84953)

---

I would suggest to create the mounting pads as complex pads (either a single polygon primitive that has the full shape or you could also play around with overlaying different primitives. The later could make defining the pad easier.)

evanshultz:
@penoud 
Great! Make a new PR for the footprints and link to the script PR. We can review the footprints then, make any tweaks to them (and the script), and finally merge the footprints. After that we can close this and make a comment in the script PR to confirm that the script is ready to merge too.

In this particular case, if the issue requesting FFC connectors doesn't get a response I'm inclined to close it as well once the script-generated footprints are merged.

I suggest a new PR for the footprints because there's already multiple commits above tweaking a non-script-generated footprint and this started out with just a single so a new PR for the whole family will be more clean.

PR #920, opened 2018-09-06
Bourns PTA series single-gang slide potentiometers.

ppelleti:
[Datasheet](https://www.bourns.com/docs/Product-Datasheets/pta.pdf).

These were generated with the `slide_Potentiometer.py` script on my [slide-pots](https://github.com/ppelleti/kicad-footprint-generator/tree/slide-pots/scripts/Potentiometers) branch.

These slide pots have mounting pins, so they need a new symbol for a potentiometer with a mounting pin.  I am going to submit that PR next...

pta4543

------------


evanshultz:
- Include the datasheet in the footprint description.
- I wonder if there's a better pin 1 style. Perhaps with removing silk? This double silk line is easy to lose since other parts could be placed there which eat away silk. I like to avoid it if there's a good alternative.
- For mounting holes a 1mm oversize is fine with me, but for the 3 pot holes we can use a smaller annular ring. 0.55mm over the drill or 1.75mm, should be OK.
- I wonder if it would be helpful to indicate the travel in silk and/or fab under the part body? I have seen LEDs or other indicators for the "position" or level and placing them would be easier if the start and stop of the knob where obvious.

Dimensioned drawing:
image

ppelleti:
OK, I've made those adjustments.  For the pin 1 indicator, I went with the triangle style I've seen a lot lately.

slide_pot_revised

PR #933, opened 2018-09-11
More pulse footprints

evanshultz:
I have added the transformers from https://productfinder.pulseeng.com/products/datasheets/SPM2007_61.pdf. In some cases, I had previously added a compatible transformer (https://github.com/KiCad/kicad-footprints/pull/645, https://productfinder.pulseeng.com/products/datasheets/P663.pdf) so I simply placed the MPN of the new transformer as a keyword in the existing footprint.

A few of the dimensions were eyeballed, typically where the "body" (the part that sticks up farthest from the PCB) wasn't dimensioned.

image

------------


evanshultz:
The corresponding symbols are submitted at https://github.com/KiCad/kicad-symbols/pull/926.

Misca1234:
Pulse_PA0926NL, should it not be named Pulse_P0926NL ?

PA1323NL
Why is the size given on page 19.0 x 8.6 x 7.6 different on page 5

PE-68386NL is this foot print exactly as PA2001NL ?

pa0264nl is this foot print exactly as pa2004nl ?

evanshultz:
> Pulse_PA0926NL, should it not be named Pulse_P0926NL ?

Yes. I was afraid of mixing up "P" and "PA" so I checked this specifically a couple times. And still miraculously got it wrong. Thank you for catching that. Fixed now.

The dimensions of PA1323NL match on pages 1 and 5, but they're not the dimensions you wrote above. I don't see an issue with this PN. Can you please check this and let me know the issue you found?

For the two you asked if there are exactly the same, yes, Pulse confirmed to me that they are.

Misca1234:
PA1323NL, not sure where I got the 19x from

PR #940, opened 2018-09-14
Adding QFN-76-1EP_9x9mm_P0.4mm_EP3.8x3.8mm

ki5libs:
Adding QFN-76-1EP_9x9mm_P0.4mm_EP3.8x3.8mm
[Datasheet](https://www.marvell.com/documents/bqcwxsoiqfjkcjdjhkvc/)
qfn-76
[3D package PR](https://github.com/KiCad/kicad-packages3D/pull/399)

------------


Misca1234:
@ki5libs 
I saw that you use the scripts to generate the 3D model, nice.
Have you used scripts to generate the foot print as well?

Misca1234:
Yes, the new ones are in
https://github.com/pointhi/kicad-footprint-generator/tree/master/scripts/Packages

> I saw that some recent QFN footprint PRs used another parameter files with different structure.

They might have used the old scripts, which is not recommended

If I understand correctly you have generated this foot print with help of scripts but never pushed the script change (I can not find your push at https://github.com/pointhi/kicad-footprint-generator/pulls)?

The reason for I am looking for the script push is that we only need to check the parameter file, which is easy, instead of take up the foot print itself and then measure all pads and all distances by hand.
So if there is a script file related to the foot print the review only takes a minute or so, but only if the script is pushed as well.

We will always assume that the script do right and no errors (in case the script is doing wrong it can be fixed and then easily re generate the concerned footprints)

ki5libs:
@Misca1234 OK, I've created [PR](https://github.com/pointhi/kicad-footprint-generator/pull/177) for this footprint parameters.
But the last commit to this parameter fie (qfn.yml) was 5 month ago. Looks like all fresh commits work with yml-files in folder "size_definitions". But I couldn't find how to generate footprints using that files...

Misca1234:
When I did my SSOP-20_5.3x7.2mm_P0.65mm

I change the file 
https://github.com/pointhi/kicad-footprint-generator/scripts/Packages/Package_Gullwing__QFP_SOIC_SO/size_definitions/ssop.yaml
and then I executed in
https://github.com/pointhi/kicad-footprint-generator/scripts/Packages/Package_Gullwing__QFP_SOIC_SO/
the line
 python ipc_gullwing_generator.py size_definitions/ssop.yam

which created a directory with the foot prints in
https://github.com/pointhi/kicad-footprint-generator/scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Package_SO.pretty/SSOP-20_5.3x7.2mm_P0.65mm.kicad_mod

Misca1234:
The script is according to the data sheet

PR #942, opened 2018-09-15
Add SMA-J-P-X-ST-EM1 footprint

pointhi:
Datasheet: http://suddendocs.samtec.com/prints/sma-j-p-x-st-em1-mkt.pdf

sma-j-p-x-st-em1

The courtyard extension at the right side is to comply with the rule: `The courtyard should include any extra required clearance for mating connectors`. I assumed a SMA connector with about 8mm diameter like: https://linxtechnologies.com/wp/wp-content/uploads/adp-smam-smam.pdf

In comparison to the existing SMA edge connectors in the library, this one seems to be far more widespread in terms of component size.

------------


Misca1234:
Hi

you have laced pads on both side, 

bild

Is it a good idea to put a via on pads 2 ?

I think you have placed pad 1 wrong, the 0.1 distance should be on the pcb edge side

bild

bild

evanshultz:
Does the courtyard rule apply here since the part body hangs totally off the edge of the PCB?

Why add the `_EdgeMount` suffix? Is there a non-edge mount SMA connector from Samtec with otherwise the same MPN?

Often we mark the PCB edge on the footprint even if it seems obvious. See:
image

@Misca1234 
Pads should be on both sides. Look at the SECTION A-A in the datasheet. The 1.47mm gap is to fit the PCB between the forks of the connector.

Misca1234:
I was not arguing agianst pads on both sides, I was just asking if it was a good idea to add a via to pads 2 so both sides are, from start, connected.

evanshultz:
Gotcha. Good question to ask.

Well, either pad can be used so no via gives flexibility. If the via is in the pad it may be harder (more expensive) to manufacture the PCB or require the user to manually delete the via or instruct the board house to remove it. Some work. If the via is outside of the pads, the footprint is adding copper that the user may not want (at least not in that location).

Leaving the user to connect them seems best to me. Similar to having a DFN with GND on a pin and also the thermal pad, where the user must decide how to handle it. But that's my opinion.

Misca1234:
ok

pointhi:
thanks for spotting the pad issue. I also added a PCB edge marking. For the ```_EdgeMount``` suffix I simply based on the already used naming convention which marks the orientation most of the time for connectors.

What about the courtyard? I'm fine with reducing the dimensions to only incorporate the connector dimensions, and removing the added dimensions for its counterpart.

sma_fp

evanshultz:
IMO the courtyard should only cover the part because it's not on the PCB. However, it's easier to keep a KLC clause consistent rather than add little caveats so leaving it alone with a fixed rule is probably best.

@Misca1234 
Everything I mentioned is fine so I'll let you do any more review and merge.

Misca1234:
Of the above raised points, that is courtyard and suffix, I really do not have an opinion
If 
@evanshultz 
Is ok with it, I will merge it

pointhi:
@Misca1234 @evanshultz would be nice to get it merged if no issues are remaining :)

evanshultz:
I don't like the `_EdgeMount` suffix since I don't feel it's needed. I do see some footprints with it, but most often I don't (especially if there is no non-edge-mounted version) as can be seen in the USB connector footprint repo above.

But dimensionally, it's OK with me:
image

pointhi:
I would stay with ```_EdgeCuts``` for consistency reason. But I can remove it if you like. It is not important as you said

PR #947, opened 2018-09-17
Add HTSSOP-32 with exposed thermal pad on top

LJJS:
Add HTSSOP-32 with exposed thermal pad on top.
For symbols PR https://github.com/KiCad/kicad-symbols/pull/889.

htssop-32

------------

http://www.ti.com/lit/ds/symlink/tpa3126d2.pdf


evanshultz:
- Remove the `-1EP` section because this footprint now doesn't have an exposed pad. This also applies to the symbol using this footprint.
- Can you add the slug dimensions to the description for easy reference?
- Unless you can find just a package drawing, please use the full datasheet link in the description.
- Fab ref des can be 1x1mm text size.
- Please shrink the courtyard to 0.25mm bigger than the body/pads. Now it's 0.3mm on top and bottom.
- We usually use 0.11-0.12mm offset for silk instead of 0.125mm, but I suspect you copied this distance from an old gullwing footprint.
- Pad dimensions from the datasheet are 1.5x0.45mm spaced apart 7.5mm. This footprint does not match.

I am compelled to note that we have a nice gullwing footprint generator using IPC standards, but I'll accept this manually-created footprint also.

We do not currently have a paradigm to indicate the slug size if the slug is up. But unless there is a good idea and consensus quickly, I'll merge this footprint anyway and create an issue. It's not the only footprint that would be affected if we need to make a naming change sometime for this item.

LJJS:
I will have a look at this today and make a new one with the generator.
I found the [package drwaing](https://sigma.octopart.com/62048046/technical_drawing/Texas-Instruments-TPA3116D2DAD.pdf).
Should I use as footprint name something short like "Package_SO:Texas_DAD0032A" or "Package_SO:Texas_TSSOP-32_6.1x11mm_P0.65mm_SlugUp" for more detail?
I like the short one more, because the footprint is easier to find with the information in the datasheet and its not necessary to measure the pads size to find the right one.

evanshultz:
The package drawing looks perfect to me, but it's not from TI. I'd rather have the full datasheet from an "official" URL than a package drawing from a 3rd party URL.

Is this package only used by TI? It seems generic enough that it shouldn't have a vendor name associated. For example, http://rohmfs.rohm.com/en/products/databook/datasheet/ic/power/switching_regulator_system/bd49101arfs-m-e.pdf looks extremely similar. So perhaps we go with `HTSSOP-32_6.1x11.0mm_P0.65mm_SlugUp` but we still need to settle on how to notate a top slug. For package naming I opened https://github.com/KiCad/kicad-footprints/issues/955.

LJJS:
Yeah your are right, I will change that.

I could not find any IC which share the same footprint, but for me both namings are fine.
TI sometimes have some minor differences in pad dimension compared to other footprints.
(Your example have a pin pitch of 0.8mm.)

For my name suggestion see https://github.com/KiCad/kicad-footprints/issues/955#issuecomment-422598553.

evanshultz:
KLC doesn't specify silk offset from fab lines, but our informal standard and scripted value is 0.11mm. This means a 0.1mm fab line just touches a 0.12mm silk line.

Granted, that still means the silk is 0.05mm outside the part body which is (slightly) excessive for SMD footprints according to KLC, which only says so silk under SMD footprints, but it's quite easy to visually check and 0.05mm is inside the placement tolerance for a silk line regardless of the method used.

evanshultz:
@LJJS 
Sorry for the long delay. Are you still here? https://github.com/KiCad/kicad-footprints/issues/955#issuecomment-691848591 finally gives us direction on how to name this footprint, so if you can rename and make any other changes from the above comments I will take another look and then we can proceed. Thanks!

PR #949, opened 2018-09-17
Sharp GP2Y0A41SK0F

serkanxselcuk:
Sharp GP2Y0A41SK0F infrared distance sensor added as virtual footprint.

Datasheet: http://www.sharp-world.com/products/device/lineup/data/pdf/datasheet/gp2y0a41sk_e.pdf

3D package PR is at followeing link: 
https://github.com/KiCad/kicad-packages3D/pull/403

image

image

------------


serkanxselcuk:
@Misca1234 I didn't put NPTH, because there are two options about this 3.2mm hole to mount sensor.

* Screw only
1. to drill 2,5mm hole (NPTH)
1. to thread M3 after PCB manufacturing.
1. mount with screw

* Screw and nut
1. to drill 3.2mm hole (plated TH)
1. mount with screw and hold nut at the other side.

Now user can put any screw hole from hardware library. If we put hole in device library, we must decide which one user have to use. Or we can import both as versions. Which option do you suggest?

evanshultz:
If there's no schematic symbol this shouldn't have a silk ref des, I believe.

Misca1234:
Well, I was just checking the data sheet and was not aware if you knew about "NPTH" or not.
However, I am sceptical to that there is a point to thread a PCB and rely on this thread to keep the component in place.

> Which option do you suggest?

I personally would go for holes, but what are other people thinking ?

evanshultz:
My opinion:
1. This is the least overhead in the library, but there are drawbacks. It's rather hard to place elements exactly at a specific location as there is no "snap to line intersection" or "snap to line midpoint" option in Pcbnew when moving an element. And also what if the user forgets to add holes? Then they will need to drill holes to mount this part in the PCB. But what if there is copper already in this area? We don't have any way to make constrained keepouts in footprints so we can't even help the user not do bad things. Potentially a number of significant drawbacks.
2. & 3. Padstacks can be edited after placing the part, so the user could change from 2.5mm to 3.2mm. But again, it's easy to forget to make this change and it could lead to mistakes.
4. This is based on you considering two options and the screw choices you would use. We would likely need even more than two version of this footprint in the library to suit all potential users. Ugh.

Overall, I go for option 1 as well. But perhaps add text to inform the user to add pads (which will then serve as keepouts for copper.

Misca1234:
@evanshultz 
If I understand your comment you vote for holes already "now"?

serkanxselcuk:
>If there's no schematic symbol this shouldn't have a silk ref des, I believe.

I would contribute Sensor_Distance library when I have time again. Sensors with integrated socket (like this one) would have no pins. There are some distance sensors which would be interesting like GP2Y0D810Z0F or MaxSonar-EZx serie

@evanshultz @Misca1234 
> It's rather hard to place elements exactly at a specific location as there is no "snap to line intersection" or "snap to line midpoint" option in Pcbnew when moving an element.

> But perhaps add text to inform the user to add pads.

Thank you for recommendations. I think text to remember on pads and inform that grid must be changed to 0,5mm to fit hole position would be helpful for 1st option.

Any other ideas?

serkanxselcuk:
@evanshultz 
> Adding similar symbols later is great and we welcome them, but they will use a different footprint since this one has no logic pads. 

Sure, I mean it will get a generic symbol for distance sensors without connection.

>So this footprint will not get a ref des, right?

I think it must still have RefDes, because end status of PCB has this part placed. Placement type is set to virtual because of pick&place process. I don't believe P&P can mount this part automated. And there are some companies which offer mounting complete modules. They do manual montage, wiring and handle housing after PCB-fabrication and soldering.
 
> Right now I can suggest "Add THT pad" where it is shown on Dwgs.User layer. But it doesn't feel great to me. There's surely something better.

Shall we wait a few days before implement this, maybe you or anyone come to mind for a better idea. Or is it your last decision?

Maybe we could place normal THT pad with 1mm hole, 6,1mm diameter size and text on Dwgs.User layer `2,5mm/3mm`. User see this hole is too small for a screw and know it is not adjusted. If user forget to adjust pad hole size, he can drill manual. And place for screw and nut would be keepout anyway.

image


Here is a Board I design using this sensor:
image

serkanxselcuk:
Status now:

image

What think people about this solution?

Misca1234:
I have no objections to the component

@evanshultz 
Are you ok with it now ?

cpresser:
I was looking at virtual components in the library and found this one.
According to KLC F8.1 this should not be virtual. It will get assembled at one point of time. Thus it should be part of the BOM. I suggest changing it to THT.

PR #951, opened 2018-09-18
Add WDFN-8_2.2x2mm_P0.5mm_1EP

cp-aquila:
Add WDFN-8_2.2x2mm_P0.5mm_1EP

------------
To be used with [emi2121mt](https://www.onsemi.com/pub/Collateral/EMI2121MT-D.PDF) common mode filter with esd protection.
Pull request for that symbol will follow soon.

- https://www.onsemi.com/pub/Collateral/511BN.PDF
- Screenshot 1


Misca1234:
Hi
I think you got pad 2 at the wrong y value, according to 
https://www.onsemi.com/pub/Collateral/511BN.PDF

The center of the pin is at e/2 which is 0.25 but the pad is at 0.2, what do you think ?
(Pin 1 is at -0.75 y, which indicate that pin 2 should be at 0.25 with a e = 0.5)

cp-aquila:
> Travis complains that F6.2 is violated because Pad1 is bigger. But actually the center of the footprint is in the center of the package. So I think we can ignore that.

I made a PR to fix this isse in  `check_kicad_mod.py` https://github.com/KiCad/kicad-library-utils/pull/260

Misca1234:
Thx, all looks good , 
however when I look at the other foot print names, I guess you foot print name should be

WDFN-8-1EP_2x2.2mm_P0.5mm_EP0.80x0.54.kicad_mod

What do you think?

cp-aquila:
> WDFN-8-1EP_2x2.2mm_P0.5mm_EP0.80x0.54.kicad_mod
> 
> What do you think?

I think you are right :) According to the KLC that is a more correct name.
A force-push, which updates the name has been done.

Misca1234:
Thx

cp-aquila:
> 
> 
> Thx

Thank your for your work review/maintaining.
I would have done the footprint anyway :)

PR #952, opened 2018-09-18
Add two more SMD solder wire pads

aurel32:
This adds two more SMD solder wire pads of size 3x6mm and 4x8mm, based on the existing 5x10mm one. The aspect ratio is the same.

screenshot_20180918_152923
screenshot_20180918_152848

They violate KLC rules F6.1 and F9.3, but that's also the case of the original 5x10mm pad and it seems normal not to provide a 3D model for a pad:

```
Checking footprint 'SolderWirePad_1x01_SMD_5x10mm':
  Violating F6.1
    For surface-mount devices, placement type must be set to "Surface Mount"
    Footprint placement type set to 'virtual' - ensure this is correct!
  Violating F9.3
    3D model settings
    No 3D model provided for virtual component
```

------------


Shackmeister:
Hi I can see this has both a virtual and an SMD attribute. Did you modify this file manually? It should only have one

aurel32:
@Shackmeister , I tried to minimize as much as possible the difference with `SolderWirePad_1x01_SMD_5x10mm.kicad_mod` that I used as a reference. It also has both virtual and SMD attributes. I guess the virtual one should be removed, correct?

serkanxselcuk:
@aurel32,
>I guess the virtual one should be removed, correct?

I would make it virtual, when I think of pick&place automtion and BOM.

aurel32:
I have just pushed a new version dropping the smd attribute from SolderWirePad_1x01_SMD_5x10mm and also from the 2 newly added ones.

Shackmeister:
please rotate the fab refdes to fit inside the pad.
also the silk designer should be outside the pad

evanshultz:
The silk refdes should also avoid the copper so move it up.

In the keywords and description there are extra spaces, missing spaces, extra commas, and a typo of "measurement".

aurel32:
I do not need this anymore, therefore closing the bug.

PR #953, opened 2018-09-18
Add SIL0010A MicroSiP Package

cp-aquila:
Add SIL0010A MicroSiP Package

Texas_MicroSiP-10-1EP_3.8x3.0mm_P0.6mm_EP0.7x2.9mm_ThermalVia
To be used by LMZM23600

------------
Datasheet: http://www.ti.com/lit/ml/mpds579b/mpds579b.pdf

No 3D-Model


Misca1234:
Hi 

Thanks for you contribution.

A foot print with thermal vias should end with 'ThermalVias' so you should replace '_SMD' in the file name with '_ThermalVias'

Also, with a thermal vias you need to add extra F.paste areas (and if you want a pad on the opposite copper as well)
See this foot print for example, the designer have added the extra F.Paste areas and a pad on the other side as well
Package_SO.pretty/HSOP-8-1EP_3.9x4.9mm_P1.27mm_EP2.41x3.1mm_ThermalVias

bild

cp-aquila:
Hello,

thanks for your helpfull comments.
I did some changes and force pushed them.
- Change name to include "_ThermalVias"
- Added solder-paste only pads
- Added a thermal pad on the B.Cu layer
The Screenshot in the first post has been updated as well.

However, Travis fails since KLC does not allow this pad configuration. ('check_kicad_mod.py' reports the same errors on 'HSOP-8-1EP_3.9x4.9mm_P1.27mm_EP2.41x3.1mm_ThermalVias')

>  Violating F6.3
>    Pad requirements for SMD footprints
>    Pad(s) potentially missing layers
>  Violating F7.4
>    Pad requirements for THT footprints
>    Some THT pads have incorrect layer settings

Here is a 3D-Preview and a 'exploded' view of the pads

Misca1234:
Sometimes travis is not accurate, most of times but not always, we usually check the travis errors to see if the warning/errors are known "anomalies", even if travis report warning/errors components can be merged.

Misca1234:
@evanshultz, @Ratfink, @poeschlr, @antoniovazquezblanco 
(have I forgot someone ?)
Can someone have quick look at these thermal vias and see if all is correct, thermal vias is not my strong side, but the rest of the foot print seems to be correct.

evanshultz:
This one specific example is where our check scripts are weak. They need to be taught so this Travis noise is reduced.

I looked at the footprint as requested and have this feedback:
- The footprint is built NSMD style, which means the copper dimensions are wrong. For example, the copper thermal pad should be 0.8x3.0mm (the mask opening is 0.7x2.9mm). Similarly, the mask and stencil openings for pads 1-10 are 0.65x0.3mm but the copper size is 0.75x0.4mm. You can use the clearance feature to shrink the mask and pad.
- I prefer to see the pad X and Y dimensions swapped instead of rotating pads.
- Paste pads should have only F.Paste layer. Not F.Cu or F.Mask.
- Paste pad height is 0.83mm while the doc shows 0.833mm. While 0.003mm is insignificant, there's no reason not to use the given dimensions.
- It's a bit odd to see paste directly over vias. Usually this will result in a loss of solder in an uncontrolled manner. Certainly plugged or tented vias can be used, but simply avoiding the vias with paste 
- It would be nice to have more than just a link in the description. Give some footprint and/or package info. Include the TI name of SIL0010A.
- To avoid silk under the part is must be 0.11mm+ offset from the fab lines. 0.1mm now.
- Courtyard is more than 0.25mm offset from fab/pads on all sides.
- I'd prefer silk removal to indicate pin 1 as in the middle screenshot. Having a second silk corner is not only liable to be removed but also counter to the latest IPC guidelines.

BTW, I found https://blog.github.com/2012-05-09-introducing-team-mentions/ which one might think would help to ping everybody without having to remember everybody individually. But I think the **KiCad/librarians** is too mixed to make it useful. Only a small subset of the people in that team appear to be involved with PRs.

Misca1234:
@cp-aquila 
Do you want to try to make the foot print with scripts ?
It is fairly easy to generate the foot prints, you just add the parameters from the data sheet into a parameter file and then you use the ready made scripts to generate the foot print itself ?

This means that only the parameter file will be review and not the foot print itself because the assumptions is that the scripts is never wrong.

If you do not want to experiment yourself, I can explain it to you
(here is one example of a push of foot prints with vias 
FP: https://github.com/KiCad/kicad-footprints/pull/958
Parameter: https://github.com/pointhi/kicad-footprint-generator/pull/182/files
)

cp-aquila:
Thanks for the detailed feedback, It helps a lot.

> This one specific example is where our check scripts are weak. They need to be taught so this Travis noise is reduced.

The script could check if the offending pads are all related to the Exposed pad. Then check if all of the paste-only pads are within the bouning-box of the exposed pad. For the vias, a similar check could be made.
I will have a look at this in a dedicated PR in `kicad-library-utils`.

As for the footprint, using a script sounds like a good idea, will do.
* Regarding NSMD vs. SMD I was not sure on how to interpret F6.4. So far I have not seen a SMD part in the kicad-libs, so I choose to keep all clearances at zero and have the user define it within the board.
* Here, for SIL0010A the datasheet does suggest to put the the paste over the vias with the hint that they should be filled, plugged or tented. Paste over vias can be seen on other footprints as well for example `Cypress_QFN-56-1EP_8x8mm_P0.5mm_EP6.22x6.22mm_ThermalVias`. Perhaps we can add a 'paste over vias' rules/recommendation to the KLC to clarify on how the kicad-footprints should handle this?

cp-aquila:
Alright. I did create a parameter file.
The output is this. I also force-pushed the results into the branch for this PR.

Misca1234:
Scripts, nice, All looks good

evanshultz:
@cp-aquila 
We do use NSMD in general as it gives tighter tolerances. I've also never seen a recommendation against it. I tried to find some previous conversation but any decision appears to be at the old repos and I'm not motivated enough to search there. KLC is out of date since we don't specify "NSMD" at the end since that's our default.

The user can't swell/shrink copper, only mask and paste, so building a footprint exactly to the datasheet would require setting the pad size properly as I mentioned above. But using this IPC-compliant footprint ought to work just fine and we will stick with IPC rather than deviate.

Also, we have a simple check to detect paste pads. It could use a bit of refinement to work better but the concept is similar to what you described.

Anyway, thanks!

PR #744, opened 2018-07-12
TerminalBlock: Add Wuerth 691311400102

antoniovazquezblanco:
https://katalog.we-online.de/en/em/TBL_7_62_3114_CLOSE_VERTICAL_PCB_HEADER_69131140010X/691311400102?m=n&sp=https%3A%2F%2Fwww.we-online.com%2Fweb%2Fen%2Fwuerth_elektronik%2Fsearchpage.php%3Fsearch%3D691311400102&_ga=2.253828047.693015018.1531393119-1957336121.1531393119

https://katalog.we-online.de/em/datasheet/6913114001xx.pdf

imagen

Hand made, I had no time to script this, sorry.
------------

Thanks for creating a pull request to contribute to the KiCad libraries! To speed up integration of your PR, please check the following items:

- [x] Provide a URL to a datasheet for the footprint(s) you are contributing
- [x] An example screenshot image is very helpful 
- [x] If there are matching symbol or 3D model pull requests, provide link(s) as appropriate
- [x] Check the output of the Travis automated check scripts - fix any errors as required

PR #70, opened 2017-11-21
Shielding

SchrodingersGat:
Coped from old **Shielding_Cabinets** repo

PR #956, opened 2018-09-19
TerminalBlock: Fix Würth spelling

antoniovazquezblanco:
See https://github.com/KiCad/kicad-footprints/pull/744#issuecomment-422666514

### DIFF
--- a/TerminalBlock.pretty/TerminalBlock_Würth_691311400102_P7.62mm.kicad_mod
+++ b/TerminalBlock.pretty/TerminalBlock_Würth_691311400102_P7.62mm.kicad_mod
@@ -1,10 +1,10 @@
-(module TerminalBlock_Wüerth_691311400102_P7.62mm (layer F.Cu) (tedit 5B582864)
+(module TerminalBlock_Würth_691311400102_P7.62mm (layer F.Cu) (tedit 5B582864)
   (descr https://katalog.we-online.de/em/datasheet/6913114001xx.pdf)
-  (tags "Wüerth WR-TBL Series 3114 terminal block pitch 7.62mm")
+  (tags "Würth WR-TBL Series 3114 terminal block pitch 7.62mm")
   (fp_text reference REF** (at 3.81 -6) (layer F.SilkS)
     (effects (font (size 1 1) (thickness 0.15)))
   )
-  (fp_text value TerminalBlock_Wüerth_691311400102_P7.62mm (at 3.81 5) (layer F.Fab)
+  (fp_text value TerminalBlock_Würth_691311400102_P7.62mm (at 3.81 5) (layer F.Fab)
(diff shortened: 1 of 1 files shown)


antoniovazquezblanco:
Assigned to @Misca1234 as he supervised the previous PR.

pointhi:
Please note that we discussed the naming some time ago, and considered Wuerth as standard (https://github.com/KiCad/kicad-footprints/pull/70#issuecomment-346030376)

The reason is: Würth is the official name, but problematic for users without a German keyboard to enter. Some wrote Wurth which is incorrect. The typical replacement of "ü" is "ue" (https://en.wikipedia.org/wiki/%C3%9C#U-umlaut) which means "Wuerth" would be the "correct" name.

We should add a CI test for this because many people mistype it.

antoniovazquezblanco:
@pointhi We should maybe elaborate a list of manufacturer names :/

evanshultz:
It's hard to balance a reasonably small set of rules or trying to be detailed and catch every little thing but still somehow keep the important rules salient. IMO this falls into the latter.

> I think that is to underestimate the contributers, I don't have a German but a Swedish keyboard and I have no problem to find ü all over the internet or on wikipedia.

Finding a character on the internet is not at all the same as being able to type it quickly and easily. A user with an English keyboard has to perform finger gymnastics to type an umlaut or eszett. And then they have to remember and the code or look it up.

Since American English has been selected for the library, for better or worse, and English has relatively few diacritics, again for better or worse, it could be (**could be**) considered somewhat of a least common denominator language for us.

I think @pointhi is spot on with embedding this into CI. This will help both users and librarians learn and catch these types of things in PRs and local testing, while "documenting" it in some fashion, but it doesn't add to KLC when this comes up so infrequently compared with other parts of KLC. Or pile this type of thing into an appendix to KLC. Oh wow, did I just suggest an appendix?! (@antoniovazquezblanco I assume when you wrote "elaborate a list" you meant KLC.)

Misca1234:
We should remember that ü and perhaps å, ä and ö is just a few additional characters to a-z, many languages, such as Spanish, have other special characters, so either we

- accept the spelling that companies wants to be presented with
or
- invent a translation table for all languages extra characters that is not included in a-z (and no, don't dare to add that f crap rewriting of å and ö to 'ae' and 'oe')

Misca1234:
> Finding a character on the internet is not at all the same as being able to type it quickly and easily. 
> A user with an English keyboard has to perform finger gymnastics to type an umlaut or eszett.

Are you thinking of finding a component in the schematic search engine or create a new component in the library?

evanshultz:
**I think we are in territory where the librarians should come to a consensus, with an executive decision from Rene if necessary, and then move forward together.** Discussion here isn't the proper place to sort out this type of issue anyway.

PR #946, opened 2018-09-17
Add VSON-10 for ROHM BD8314NUV

asukiaaa:
screenshot from 2018-09-17 20-37-01

Datasheet: http://rohmfs.rohm.com/en/techdata_basic/ic/package/vson010v3030_1-e.pdf

Related PRs:
https://github.com/KiCad/kicad-symbols/pull/934#issuecomment-421821180

- [x] Provide a URL to a datasheet for the footprint(s) you are contributing
- [x] An example screenshot image is very helpful 
- [ ] If there are matching symbol or 3D model pull requests, provide link(s) as appropriate
- [ ] Check the output of the Travis automated check scripts - fix any errors as required

PR #958, opened 2018-09-19
THAT Corp and ROHM QFNs

evanshultz:
- http://www.thatcorp.com/datashts/THAT_1580_Datasheet.pdf
- http://www.thatcorp.com/datashts/THAT_5171_Datasheet.pdf
- http://www.thatcorp.com/datashts/THAT_5173_Datasheet.pdf
- http://www.thatcorp.com/datashts/THAT_626x_Datasheet.pdf
- http://rohmfs.rohm.com/en/products/databook/datasheet/ic/power/switching_regulator/bd8314nuv-e.pdf (from https://github.com/KiCad/kicad-footprints/pull/946)

image
image
image
image

------------


Misca1234:
QFN-16-1EP_4x4mm_P0.65mm_EP2.1x2.1mm
http://www.thatcorp.com/datashts/THAT_1580_Datasheet.pdf

lead_len_max: 0.55
it should be 0.45 (ITEM F)

thermal_vias:
    drill: 0.3
it says 0.254 mm in datasheet

evanshultz:
Good catch on lead length. Thank you very much! Fixed now.

It does raise an interesting point that we could be under-dimensioning our footprint names. If one package had a lead length up to 0.55mm instead of 0.45mm, the footprint would be a bit wrong for that package. Without putting lead length and lead width in the footprint name we can't know for sure and CvPcb won't have that information to help the user. That being said, IPC-7351B includes the calculations for lead length but doesn't include this info in their footprint names so I'm mentioning it here just for informational purposes instead of pushing for any changes. I wonder what IPC was thinking? Maybe using the largest/widest pin dimensions is intended to fit packages with smallest pin? Or it was just missed?

Also, I didn't mention it above but it's a bit odd how THAT recommends a 2.1mm Cu pad when their part has a 2.6mm thermal pad. However, that's what they recommend so I made the corresponding footprint.

Regarding the vias, since this is a generic IPC-compliant footprint I didn't bother to pick the via size that the datasheet recommends but rather stuck with the common one I already saw in the YAML file. This should result in fewer drill diameters if one is using script-generated footprints that have a consistent drill size, but I really don't see that being such a concern these days. And having a larger drill means less thermal resistance, though again this isn't likely to be a real concern for THAT1580 as the dissipation isn't that high.

Misca1234:
That would be something to discuss with Rene ^^, I am not qualified to contribute with an opinion in the matter, sorry

asukiaaa:
- The space between via on VSON-10-1EP_3x3mm_P0.5mm is `1.2` not `2`.
- The size of pad is `1x2.1` not `1x2.2`
screenshot from 2018-09-22 12-32-11

evanshultz:
This footprint is built in accordance with IPC's recommendations, which may not match ROHM's. The dimensions are based on the package size with great input from industry and is virtually guaranteed to work well in most manufacturing environments. While there is latitude in the thermal pad size in the IPC doc, and tweaking the IPC inputs can be done, it's an excellent starting point.

We do not know how or why ROHM developed their footprint, and we will rely on IPC footprints unless the vendor provides a strong reason why a deviation should occur. And in those cases, we also have some manufacturer-specific footprints. In this case there was no reason to deviate from the IPC recommendations for the package.

asukiaaa:
Thanks for the comments.
As I wrote on https://github.com/KiCad/kicad-packages3D/pull/401#issuecomment-423716176, I understand that KiCad library basically use same size of pad on package even if the chip maker suggests different pad size.

PR #959, opened 2018-09-20
Add pSemi CSP-16 footprints

evanshultz:
- https://www.psemi.com/pdf/datasheets/pe29101ds.pdf
- https://www.psemi.com/pdf/app_notes/an77.pdf

image

------------


Misca1234:
Why have you named them the same (travis error) ?

evanshultz:
The files were different and I fixed the footprint name inside the file`_Pad0.18mm` one now.

The 3D models are the same because only the pad size is different as mentioned in AN77 I linked above, but the actual component is the same for both footprints.

Misca1234:
All looks good

PR #960, opened 2018-09-20
muRata OKI-78SR added

serkanxselcuk:
DC-DC converter muRata OKI-78SR serie added, which replaces 78xx serie.

Datasheet: https://power.murata.com/data/power/oki-78sr.pdf

image image

image image

3d model PR: https://github.com/KiCad/kicad-packages3D/pull/404

------------


Misca1234:
**Converter_DCDC_muRata_OKI-78SR_horizontal.kicad_mod**
You should rotate 180 degrees so pin is at "upper left corner"
You do not have to add the "inner rectangles"


**Converter_DCDC_muRata_OKI-78SR_vertical.kicad_mod:**
You do not have to add the "inner rectangles"

tips, a good thumb role is
F.Fab be 0.1 wide
F.Silk 0.12, distance 0.13 from F.Fab
F.CrtYrd 0.05, distance 0.12 from F.Silk
Look into the KLC rule for exceptions
so for Converter_DCDC_muRata_OKI-78SR_horizontal it could look something like this

bild

serkanxselcuk:
image  image

@Misca1234 thank you for helpful tips.

Misca1234:
murata seems to have some error on their homepage, I can not download the data sheet, however I can see that you placed the model name "OKI-78SR" inside the F.Fab, you should place the component name outside and the "R%" shall be centred inside the F.Fab area.
(just look at another fp)

When I get hold of the data sheet I can check the rest

evanshultz:
While KLC doesn't specify silk offset from fab lines, our generators use 0.11-0.12mm. Courtyard offset is specified and it's 0.25mm for most parts and 0.5mm for connectors. Not sure which should be used here. Anyway, summing 0.12mm and 0.13mm above gives 0.25mm, but it's an atypical style.

The part name can be inside the fab rectangles. This is done on larger parts where it fits. I recall Oliver writing this a while ago, and nothing to the contrary, but I can't find that comment right now.

Misca1234:
Was we not discussing just recently that no F.silk should be under the component because it is not visible when mounted and therefore is a waste

evanshultz:
KLC only says that silk can't be under SMD parts. THT parts can have silk under them. It may be a waste and I don't feel strongly about it, but as an example here are some THT headers that Rene recently scripted which have silk under the part body:
image

Misca1234:
Then there is no problems to keep it there

serkanxselcuk:
@Misca1234 Hi stephan,

Real dimensions of part is drawn on page 16:
image
1,25 was here still false, I might make a mistake about grids, it must be 1,27.

Drawing on page 17 (which you sent) is a recommendation for THT holes, F.Silk and CrtYrd. I think F.Silk is more matter of taste and a board looks better if distances between F.Silk and real dimension (F.Fab) has an harmonized standard for all parts of board.

Shorter to say
- I will change F.Fab top line to 1,27mm on "...OKI-78SR_horizontal"
- %R in F.Fab layer will be set to center of device at "...OKI-78SR_horizontal"
- **Q:** Should I use F.Silk-F.Fab distance 0.13 or put F.Silk at position of datasheet recommendation?

serkanxselcuk:
Short name in value was not accepted by CI. reverted to long name.

image

image

Misca1234:
> Should I use F.Silk-F.Fab distance 0.13 or put F.Silk at position of datasheet recommendation?

I guess you should stick to data sheet first and if in doubt, you can use 0.13 as a good thumb rule.

Misca1234:
Hi

Page 16 is the physical format, but on page 17 you have the recommended foot print, you should look at page 17 when making the FP but on page 16 when making the 3D model

evanshultz:
> I guess you should stick to data sheet first and if in doubt, you can use 0.13 as a good thumb rule.

Or, as I wrote at https://github.com/KiCad/kicad-footprints/pull/960#issuecomment-425221512, you could use a 0.11-0.12mm silk offset from fab lines which is what has been used in all recent footprint scripts. 0.11mm offset is quite common since a 0.12mm wide silk line will run just next to a 0.1mm wide fab line, thus preventing silk under SMD parts. There's no reason not to keep the same for THT parts as well.

The datasheet doesn't indicate any silk placement, only the part outline which will be fab lines.

Misca1234:
You have named the foot prints
Converter_DCDC_muRata_OKI-78SR_Horizontal.kicad_mod
Converter_DCDC_muRata_OKI-78SR_Vertical.kicad_mod

But the 3D models are named
Converter_DCDC_muRata_OKI-78SR_horizontal.wrl
Converter_DCDC_muRata_OKI-78sr_vertical.wrl

Is there any reason for using 
'OKI-78SR_horizontal' instead of 'OKI-78SR_**H**orizontal'
'OKI-78sr_vertical' instead of 'OKI-78**SR**_**V**ertical'

serkanxselcuk:
> Is there any reason for using
> 'OKI-78SR_horizontal' instead of 'OKI-78SR_**H**orizontal'
> 'OKI-78sr_vertical' instead of 'OKI-78**SR**_**V**ertical'

Not really, I will change them too. Thanks.

Misca1234:
Great super

PR #853, opened 2018-08-21
Add AS1115 TQFN-24 footprints

evanshultz:
https://ams.com/documents/20143/36005/AS1115_DS000206_1-00.pdf/3d3e6d35-b184-1329-adf9-2d769eb2404f

See script info at https://github.com/KiCad/kicad-footprints/issues/851#issuecomment-414472929.

image
image

------------

Thanks for creating a pull request to contribute to the KiCad libraries! To speed up integration of your PR, please check the following items:

- [ ] Provide a URL to a datasheet for the footprint(s) you are contributing
- [ ] An example screenshot image is very helpful 
- [ ] If there are matching symbol or 3D model pull requests, provide link(s) as appropriate
- [ ] Check the output of the Travis automated check scripts - fix any errors as required

PR #965, opened 2018-09-24
Add LED_SMLVN6

asukiaaa:
I add a footprint of full color led.

Datasheet: https://www.rohm.com/datasheet/SMLVN6RGB1U

Related PRs:
https://github.com/KiCad/kicad-symbols/pull/998
https://github.com/pointhi/kicad-footprint-generator/pull/185

screenshot from 2018-09-24 22-32-27


Misca1234:
Hi

Are you sure about this, the solder pattern described on page 1 have pads that is of the size 
1.45x0.8 and 1.45x0.6, yours are all 0.8x0.5

What do you say about rename the foot print into 'LED_Rohm_SMLVN6RGB1' instead, just to be somewhat future proof against new components from ROHM ?

asukiaaa:
Thank you for the comments.

I updated as recommended solder pattern.
I renamed package name as `LED_ROHM_SMLVN6`because last page of datasheet shows the package name as `SMLVN6`.

screenshot from 2018-09-26 19-40-23

Misca1234:
Nice, you done it with scripts

All looks good

asukiaaa:
Thank you for merging.

evanshultz:
The pin 1 mark in both silk and fab layers is not to the standard now. Fab should have a beveled corner, not a square corner and a bevel. The silk can be removed around pin 1 instead of extended along pin 1 as shown in the IPC 7351C docs. For a example, see https://github.com/KiCad/kicad-footprints/pull/853 which is script-based and was just merged.

Misca1234:
Something like this ?

bild

Misca1234:
> 1. Rounded rect pads should have 25% radius up to 0.25mm max.

It is 25%
bild

> 2. The fab text could probably be a bit bigger and still fit within the fab rectangle.

Fixing it


> 3. The courtyard should be 0.25mm from the part body (fab lines) or copper, so at least on the left and right sides it appears too far away above.

It is 0.2, according to KLC F5.1 section 2


4. Silk is usually 0.11-0.12mm offset outside the fab lines, but KLC doesn't specify this.
5. The newest DFN/QFN generator would put a silk line along the bottom of the part body and the right half of the top of the part body. See https://github.com/KiCad/kicad-footprints/pull/865. The newest gullwing one extends the top line to the leftmost edge of the pad (adding a dogleg if necessary). See https://github.com/KiCad/kicad-footprints/pull/849. Either way is unlike the above image.


bild

evanshultz:
KLC F5.1 is about silkscreen. Courtyard is covered in F5.3 and it states 0.25mm for this footprint.

Misca1234:
then we are about here

bild

asukiaaa:
Thanks for the information.
How about this?
screenshot from 2018-09-27 21-36-59

asukiaaa:
I committed to this branch but github and TravisCI does not detect it.
Should I create new PR?
https://github.com/asukiaaa/kicad-footprints/commit/7d40cc0364a2c89bc7a4c7580e28185ef40a6ddd

Misca1234:
@asukiaaa 
> I committed to this branch but github and TravisCI does not detect it.

It is because the branch have already been merged


> Should I create new PR?

If you want, otherwise we can take the one I created, but if you want to do the push you need to update the master, create a new branch, make the changes and then push it.
Let me know what you want to do (take mine or yours)

evanshultz:
- The fab refdes may be able to be just a bit bigger and still fit in the fab rectangle.
- This isn't codified in KLC, but somewhere we discussed using the IPC recommendation of keeping silk away from pads by the line width. That would be 0.12mm and it looks closer than 0.12m now. (KLC says 0.2mm but that was just a random number chosen.)

asukiaaa:
Thanks for reviewing.
I update text size of Fab.Ref and set clearance of F.SilkS larger than 0.15, then created new PR.
https://github.com/KiCad/kicad-footprints/pull/982

PR #967, opened 2018-09-24
Mate n lok 350211 1

nickysn:
Added footprint the TE Connectivity MATE-N-LOK 350211-1 connector. This connector is commonly used as the power connector for ATA drives and 5.25-inch floppy drives.

https://www.te.com/usa-en/product-350211-1.html
https://www.te.com/commerce/DocumentDelivery/DDEController?Action=showdoc&DocId=Customer+Drawing%7F350211%7FU5%7Fpdf%7FEnglish%7FENG_CD_350211_U5.pdf%7F350211-1

image

------------


Misca1234:
How come you have named it to 
TE_MATE-N-LOK_350211-1_1x04_P5.08mm_Vertical
instead of 
TE_MATE-N-LOK_1_350211-x_1x04_P5.08mm_Vertical

F.CrtYrd distance to F.Fab should be 0.25, you can keep the rectangular shape

The F.Silk should be on the outside of the F.Fab line

bild

nickysn:
What does the 1 before 350211 in your suggested name mean?
TE_MATE-N-LOK_1_350211-x_1x04_P5.08mm_Vertical
According to the KLC, the connector name format is:
[Series]_[Subseries]-[Pins]_[MPN]_[Pin layout]_P[Pitch]_[Modifiers]_[Orientation]_[Options]
Series is TE
Subseries is MATE-N-LOK
Pins should be 4, shouldn't it?
MPN is 350211-1 or 350211-x, because this footprint also applies to the 350211-2 and 350211-3.
So, after adding the "x" and the pin count, it becomes:

TE_MATE-N-LOK-4_350211-x_1x04_P5.08mm_Vertical

Is that the correct name?

Misca1234:
I was just wondering because your name is different form all the others, but I am not the person to decide what is right or wrong in this case ?

nickysn:
Regarding F.Silk and F.Fab: I've currently done F.Silk to follow exactly the drawing from the datasheet. I've drawn F.Fab smaller and with a much simplified shape. Now I understand that I should make F.Fab exactly like the drawing and F.Silk slightly bigger, is that correct?

poeschlr:
Sorry for the late answer. Yes the outline on fab must represent the nominal dimensions of the part outline. (It is allowed to simplify it for very complex shapes.)
The silk outline should be fully outside the part outline such that it is visible after assembly.

nickysn:
Ok. I made F.Fab match the nominal dimensions of the part and F.FilkS slightly bigger (0.15mm distance to F.Fab). I also made F.CrtYrd to be 0.25mm away from F.Fab. However, KLC F5.3 states:

"4. Unless otherwise specified, clearance is 0.25mm"
"6. Connectors should have a clearance of 0.5mm, in addition to the clearance required for mating of connector"

Since this is a connector, shouldn't F.CrtYrd be 0.5mm away from F.Fab?

Another thing unresolved is what should be the correct footprint name?

poeschlr:
thanks

PR #968, opened 2018-09-24
Fix BarrelJack Wuerth Mounting Pin pad name

skarcha:
Renames pad number 3 to MP, to match pad name on symbol Barrel_Jack_MountingPin

barrel_jack_mp_symbol
barrel_jack_mp_footprint

And this this the fixed one. No more errors loading netlist in Pcbnew.

barrel_jack_mp_footprint_new

------------

### DIFF
--- a/Connector_BarrelJack.pretty/BarrelJack_Wuerth_6941xx301002.kicad_mod
+++ b/Connector_BarrelJack.pretty/BarrelJack_Wuerth_6941xx301002.kicad_mod
@@ -31,7 +31,7 @@
   (fp_line (start 6.2 5.5) (end 5 5.5) (layer F.CrtYd) (width 0.05))
   (fp_line (start 6.2 0.5) (end 5 0.5) (layer F.CrtYd) (width 0.05))
   (fp_line (start -4.6 -1) (end -2.5 -1) (layer F.SilkS) (width 0.12))
-  (pad 3 thru_hole oval (at 4.8 3 90) (size 4 1.8) (drill oval 3 0.8) (layers *.Cu *.Mask))
+  (pad MP thru_hole oval (at 4.8 3 90) (size 4 1.8) (drill oval 3 0.8) (layers *.Cu *.Mask))
   (pad 2 thru_hole oval (at 0 5.8) (size 4 1.8) (drill oval 3 0.8) (layers *.Cu *.Mask))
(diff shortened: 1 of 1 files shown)


Misca1234:
Good catch

Perhaps we should change the symbol instead?
The pins are numbered 1,2,3 in the data sheet, which means the foot print is more "accurate".

skarcha:
Thanks!

Perfect, I totally agree. I could do it later today.

poeschlr:
Every manufacturer uses different pin numbering schemes for such parts. We can either have one generic symbol and use that pin numbering scheme for all footprints or we need a specialized symbol for every part.

My suggestion is to go with the former and accept this fix. (assuming pin 3 really is only used for adding mechanical strength. If it is connected electrically to anything then a different symbol might be better suited.)

skarcha:
@poeschlr It makes sense to me. To have a generic symbol with mounting point for mechanical strength named as MP and have footprints with MP or not, depending on whether it is for mechanical strength.

You decide, guys. I'm a newbie designing my first circuit and board... :angel: 

Thanks for your time!

evanshultz:
For this particular part, a different symbol is needed. Here is the circuit from the datasheet (http://katalog.we-online.de/em/datasheet/6941xx301002.pdf):
image

You should use the `Barrel_Jack_Switch` symbol:
image

This does continue our conversation about using pin names rather than numbers when it helps to make one symbol mate with numerous footprints...

Misca1234:
poeschlr want to change the foot print
evanshultz want to change the symbol 
?

skarcha:
Well, in my case I only need a barrel jack with a mounting point. What footprint should I use with it?

evanshultz:
**Librarians: DO NOT MERGE THIS PR!!**

Pin 3 is connected electrically on this component and it should not be named MP.

@Misca1234 
???? Rene specifically said if pin 3 is only mechanical. And I'm showing that it is not. No disagreement.

The symbol I showed above is different than the symbol in the first post, and I've showed the correct one for the actual part.

@skarcha 
If you use the symbol I posted above it will work with the existing footprint. You can leave pin 3 floating if you don't need it.

poeschlr:
Closing this as @evanshultz discovered that the current footprint is correct if used with the symbol intended for it.

skarcha:
@evanshultz Thank you! I'll use it. And thank you all for your time and great work! :+1:

evanshultz:
@skarcha 
I hope your first PCB works out well! We're happy to have an invested community and hopefully we can see more contributions from you in the future!

PR #969, opened 2018-09-25
Add SOT-416

fauxpark:
Looking at adding the [BAV70 series of dual diodes](https://assets.nexperia.com/documents/data-sheet/BAV70_SER.pdf), there are two packages the libs are missing. I adapted this one from the SOT-23 footprint.

Datasheet: https://www.nxp.com/docs/en/package-information/SOT416.pdf

------------


Misca1234:
Hi

You should add the data sheet link, https://www.nxp.com/docs/en/package-information/SOT416.pdf, to the description

fauxpark:
Yeah, I saw most of the SOT packages had no datasheet and just went with that. Fixed.

fauxpark:
I missed shrinking the width down from the SOT-23 footprint's size. Also fixed.

As for the courtyard: The KLC seems vague/incomplete on the subject of clearances. For the courtyard X offsets I'm using padX+(padW/2)+0.15 (these two footprints are smaller than 0603 so they get 0.15mm clearances). Possibly also need to +0.2mm for the pad clearance?

Not so sure what to do with the courtyard height either. If I do topsilkY+(topsilkT/2)+0.15 I get 1.17 but the value I put is 1.16...must have hit the 5 key instead of 6.

poeschlr:
@fauxpark wrote:
> As for the courtyard: The KLC seems vague/incomplete on the subject of clearances. For the courtyard X offsets I'm using padX+(padW/2)+0.15 (these two footprints are smaller than 0603 so they get 0.15mm clearances). Possibly also need to +0.2mm for the pad clearance?

The courtyard follows IPC suggestions.
The courtyard clearance is not in any way influenced by silk features or by pad clearance settings (especially not by the huge default kicad settings)

From the [IPC (pre release draft) definition](http://www.ipc.org/committee/drafts/1-13_d_7251WD1.pdf) it self:
**Courtyard** The smallest polygon area that provides a minimum electrical and mechanical clearance (courtyard excess) around the combined component body and land pattern boundaries.

**Courtyard Excess** The area between the polygon circumscribing the land pattern and the component, and the outer  boundary of the courtyard. The courtyard excess may be different in the x-and y-direction.  

**Courtyard Manufacturing Zone** The area that provides a minimum electrical and mechanical clearance (courtyard excess) around the combined component body and land pattern boundaries

[Found a nice picture from pcb library expert.](https://www.cskl.de/fileadmin/user_upload/ipc_standard_pcb_library_expert_courtyard_contour_following.PNG) Sadly in german so here the translation of the terms used:
- Courtyard Fläche -- magenta -> Courtyard Area (The same as Courtyard in the IPC definition)
- Courtyard Begrenzung -- outer edge of magenta area -> Coutyard boundary. This is what we draw onto the courtyard layer.
- Bauteile & Pads (maximale Fläche) -- light green -- This is the combined area taken up by pads and by the part body.

The light blue stuff is the manufacturing area. Not directly supported as a separate entity in kicad. It showcases that the courtyard areas of different parts can not overlap but denote exclusive areas for the part in question.

fauxpark:
Here's an example of what confused me:

F5.1 says:
>Silkscreen must not be placed over pads or areas of exposed copper
>
>Clearance between silkscreen and exposed copper elements must be at least 0.2mm or more than the pad mask expansion (whichever is greater).

Easy. I know what to offset the silk from. But F5.3 simply states:
>Unless otherwise specified, clearance is 0.25mm

Clearance from what? Pads? Fab lines? Whichever is greater?

Now I understand what the first paragraph is saying. I initially skimmed over it assuming it was simply a description of what the courtyard is used for - that is what the previous two (silk and fab) do. They don't contain any constraints of this sort. I should have read more carefully, but maybe this particular rule could be worded a little more explicitly?

Misca1234:
Your foot print look like this now

bild

the pad clearance is going just outside the F.CrtYrd, either you move the F.CtrYrd to "outside" the pad clearance, or you change the pad clearance a little bit 

bild

fauxpark:
Hm. I thought that this footprint was smaller than 0603 (and thus the courtyard clearance would be 0.15mm) but it turns out it is about the same size if not slightly larger. I will add 0.1mm to each side.

Misca1234:
Thx

PR #974, opened 2018-09-27
Added small DFN package

SchrodingersGat:
New DFN-8 package for Microchip gate driver:

http://ww1.microchip.com/downloads/en/DeviceDoc/MCP14A0303_4_5-Data-Sheet-20006046A.pdf

This is copied directly from an existing DFN footprint and only the EP size has been changed.

image


Misca1234:
Hi

Here is the link to this specific package 
http://ww1.microchip.com/downloads/en/DeviceDoc/8L_TDFN_2x3_MNY_C04-0129E-MNY.pdf

All microchips packages can be find here
http://www.microchip.com/quality/packaging-specifications

Should not this package also be in a thermal vias variant, or it is not necessary?
"For best soldering results, thermal vias, if used, "

I think the latest fashion for pads is rounded rectangles
bild

Should not the pads be 0.25x0.85 (X1 and Y1 in the drawing) ?
Should not the pad be sized 1.6x1.5 (sadly only max values are given in the data sheet)

evanshultz:
I submitted the scripted version of this footprint at https://github.com/KiCad/kicad-footprints/pull/1005.

Misca1234:
@SchrodingersGat 
Evan have already a scripted version of this foot print including a thermal vias variant.
I suggest we go with that push and close this without merge.

SchrodingersGat:
Ok I'm happy with that :)

PR #975, opened 2018-09-27
Added Fairchild Power33 package

SchrodingersGat:
6-pin dual MOSFET package from Fairchild.

Datasheet: https://www.fairchildsemi.com/datasheets/FD/FDMC8032L.pdf

image


Misca1234:
Hi

The x-size for pad 1, 2, 5 and 6 is set to 0.55, in data sheet it is 0.52
The x position for pad 1, 2, 5 and 6 is set to 1.4, in data sheet it is 1.39
The y position for pad 1 and 6 is set to 0.97, in data sheet it is 0.975 (0.98)

Is possibly to move the FootPrint Text reference REF** slightly up so it wont interfere wit hthe F.Ctryrd ?
Can you shorten the left "tag" of Pin 1 marker in F.Silk so it wont "dip into" pin 1 area ?

bild

evanshultz:
@SchrodingersGat 
Oliver, are you going to come back to your open PRs?

- There are single and dual Power33 packages, so can you include this in the footprint name and description?
- I would remove "trenchfet" from the keyword since this relates to the die inside and not the package.
- http://www.onsemi.com/pub/Collateral/AN-9057.pdf.pdf shows a stencil pattern. (Thought seemingly a different pad design, too.)
- Fab lines are 0.12mm. (Why is Travis not catching this?!)
- Silk offset is only 0.1mm, which means silk is under the part body. Use a 0.11mm+ offset.
- It looks like the fab refdes can be a bit bigger.
- Courtyard is a bit too wide.
- Pins 1/4/5/8 are too wide (0.55 vs 0.52). [Above]
- Those same 4 pins are all off by 0.005mm vertically. While this tolerance can't be held on actual PCB fabrication there's no reason not to make the footprint exactly to the drawing. [Above]
- Corner radius should be 0.25mm max.
- Can you extend the lower silk line, at least on one side, to better define the package outline?
- If a DFN package has an EP, we number the pads around the perimeter and then go to the underside pads. Here, that would means pads 1-4 are on the outer edge and then pads 5-6 are underneath. I'm not sure what the best way is since the datasheet shows only 8 pins so there's a lot of conflict.

SchrodingersGat:
I have made adjustments as suggested. Although I'm not sure what you mean by the corner radius? Where has it exceeded 0.25mm?

Also I appreciate the desire for accuracy but a super strict adherence can be a barrier to contribution. e.g. the KLC specifies that the Fab lines can be between 0.1 and 0.15mm 

I've also updated the pinout on the symbol associated with this footprint

evanshultz:
- Pads 2 and 3 should be 0.325mm on the Y location.
- Corner radius hasn't exceeded 25%. IPC 7351C is indicating 25% radius up to 0.25mm maximum. Your fixed corner radius of 15% does not meet that spec:
image
- I think instead of Fairchild this should be an ON Semi footprint.
- Is Pacakge_SON the right location? How did you chose this folder? http://www.carsem.com/pdf/ds_mlp.pdf mentions both QFN and SON while https://en.wikipedia.org/wiki/Quad_Flat_No-leads_package#Variants and http://eesemi.com/mlp.htm just mention QFN. I'm not sure but would appreciate more information.
- Would you be willing to add the paste pads from the link I sent above?

Indeed KLC does allow a range of fab line width but recommends 0.1mm. Not using 0.1mm generates a warning and is atypical in the library. If there's a reason to deviate from the recommended value that's certainly OK with me, though. Is there a reason to not use  the recommended line width?

evanshultz:
Thanks! All looks good.

I see we have other Fairchild footprints already, so if we do a clean up to ON Semi sometime in the future this is just one more.

PR #977, opened 2018-09-27
Added SOT-1123

SchrodingersGat:
Teeny tiny transistor package:

Example datasheet: http://www.onsemi.com/pub/Collateral/NST3906F3-D.PDF

image


Misca1234:
Hi

You have set the pad X-size to 0.35, it should be 0.34 ?
If you want you make the f.silk width to 0.1 (KLC F5.1:5)
Is not F.Silk a little close to the pads, < 0.2 mm (KLC F5.1:2) ?

Misca1234:
@poeschlr 
So all is good with you then (if so i merge) ?

poeschlr:
well other than he fab reference being a bit tiny (as mentioned above)

i also did not check any measurements.

Misca1234:
@SchrodingersGat 
If you make the REF** slightly biggar as Rene suggest I can merge it

SchrodingersGat:
hey thanks for the feedback guys, sorry it has taken ages to respond.

image

Sufficient?

poeschlr:
I checked the footprint more closely. The copper pads are a bit too large and therefore the outside to outside dimension does not agree with the datasheet.
The courtyard is too far away on the right side.

The silk to copper clearance is a bit small (0.05mm) should be 0.1 (assuming silk line width is 0.1mm)

sot

poeschlr:
@Misca1234 i will make a pull request fixing your merge that was a bit too early.

Misca1234:
Ok, thx

PR #979, opened 2018-09-27
Footprints for CUI SJ1-3533NG audio jack.

ppelleti:
[Datasheet](https://www.cui.com/product/resource/digikeypdf/sj1-353xng.pdf).

I am submitting two footprints: one as specified in the datasheet, and one with circular holes.

These footprints are meant to be used with the symbol `AudioJack3_Ground` from `Connector.lib`, which has pin numbers T, R, and S.  Since there is no pin 1, I have not included a pin 1 indicator.

(Strangely, all of the *other* audio jack footprints in the library have numeric pin numbers, even though the schematic symbols use letters.  I am not sure how that is supposed to work.)

sj1-3533ng

sj1-3533ng-circular

------------


ppelleti:
> In the data sheet, for 3533, it says "NP" for pin 4 and 5, does this really mean "No Pin", or the pin is actually there but not used ?

Pins 4 and 5 are not physically present:

p18_4018 audio jack

> Place the F.CtrYrd line 0.25 from the F.Fab line, right now it is a little far out

[KLC F5.3](http://kicad-pcb.org/libraries/klc/F5.3/) states "Connectors should have a clearance of 0.5mm, in addition to the clearance required for mating of connector"

> I think the circular holes are to small, the pins are 1x2 => hypotenuse = 2.23 mm and the diameter is only 2 for the hole

The pins are not 1x2.  The dimension 1x2 is in the drawing labeled "PCB LAYOUT", so 1x2 is the dimension of the holes.  In the drawing above, the long dimension of the pins is given as 1.5mm.  (I don't think the short dimension of the pins is specified, though.)

audio-jack-datasheet

The pins fit in the holes with plenty of clearance:

p18_4024 audio jack

Misca1234:
Ok, great

Issue #266, opened 2018-01-17
Footprints with complex pad shapes would benefit from the use of polygon pads.


poeschlr:
Polygon pads result in better paste and mask handling then overlapping pads.

The downsides are:
- not compatible with version 4. (But we already have ~~some~~ **many** libs that include rounded rectangles. This already results in invalid libs.)
- ~~currently not handled by stepup which makes checking footprint correctness a lot harder.~~

Edit: Adding list:
- [x] SOT-89s https://github.com/KiCad/kicad-footprints/tree/master/Package_TO_SOT_SMD.pretty
- [x] Texas_X2SON-4_1x1mm_P0.65mm.kicad_mod (https://github.com/KiCad/kicad-footprints/pull/521)
- [x] Infineon MLPQs and PQFNs at https://github.com/KiCad/kicad-footprints/tree/master/Package_DFN_QFN.pretty
- [x] Package_DFN_QFN.pretty: QFNs with odd-shape corner pads (https://github.com/KiCad/kicad-footprints/issues/658; implemented in script)
- [x] Jumper.pretty (Fix: #716)
- [x] RF_Shielding.pretty (open PR at https://github.com/KiCad/kicad-footprints/pull/890)
- [x] Coilcraft LPS4018 (https://github.com/KiCad/kicad-footprints/pull/257)

evanshultz:
If for no other reason than I would probably forget if I didn't write it down, here are a few footprints (or footprint families) that would benefit:
- SOT-89 at https://github.com/KiCad/kicad-footprints/tree/master/Package_TO_SOT_SMD.pretty
- https://github.com/KiCad/kicad-footprints/blob/master/Package_SON.pretty/TI_X2SON-4_1x1mm_P0.65mm.kicad_mod (this one also needs the prefix changed from TI_ to Texas_)
- Infineon MLPQs and PQFNs at https://github.com/KiCad/kicad-footprints/tree/master/Package_DFN_QFN.pretty

pointhi:
Lets start a list of libraries which use the old system and could profit from polygon pads:

* **Jumper.pretty** Fix: https://github.com/KiCad/kicad-footprints/pull/716
* **RF_Shielding.pretty**
* ... (please add other libraries you know)

evanshultz:
List started above.

evanshultz:
[...]

PR #972, opened 2018-09-26
Add FC source files

evanshultz:
I'm developing FreeCAD source files for a few footprints because that seemed like the best way to go. However, I'm not having complete success. I've included the WIP FC files.

Here is the footprint info:
- `Infineon_PG-LLGA-5-1`: https://www.infineon.com/cms/en/product/packages/PG-LLGA/PG-LLGA-5-1/
- `L_Wuerth_WE-TPC-3816`: http://katalog.we-online.de/pbs/datasheet/7440310047.pdf

Currently I'm running into a couple issues:
1. I cannot find how to add the labels for ref des and value. I looked in many WBs, Googled, and tried to copy them from the demos but I do not see how to add this.
2. The polygon pads aren't coming through. (Other layers seem to work OK.) I've added an anchor pad inside the boundary of the polygon outline and this works if I make a simple pad shape (like a rectangle) but for a complex pad shape it doesn't. I am successful generating footprints using the polygon pad demos included with KSU so I believe the issue is something I'm doing wrong.
3. I do not see how to set the pad number. This is not happening with any of the demos (all pads are numbered `#`) so perhaps this is a limitation of KSU?

Note that the paste layer sketch of `Infineon_PG-LLGA-5-1` still has 1 DOF which I haven't chased down yet. I'll build out the other paste pads once I take care of this. However, I don't think that's material to my questions above. All other sketches are fully constrained. That being said, I haven't used a MCAD tool since a decade ago in school so any tips and tricks or best practices are welcome!

From looking at the demos and the other source FC files in this repo, it appears some "post-processing" is required in KiCad or a text editor to add datasheet links in the description, pad numbers, etc. I'm only seeing that part of the footprint can be built in FC right now but perhaps I simply don't understand.

FC screenshots:
image
image

Pcbnew screenshot:
image

@easyw
Pinging you since surely you're able to help. :)

------------

Thanks for creating a pull request to contribute to the KiCad libraries! To speed up integration of your PR, please check the following items:

- [ ] Provide a URL to a datasheet for the footprint(s) you are contributing
- [ ] An example screenshot image is very helpful 
- [ ] If there are matching symbol or 3D model pull requests, provide link(s) as appropriate
- [ ] Check the output of the Travis automated check scripts - fix any errors as required

PR #981, opened 2018-09-27
Fixed a bunch of outdated pad styles

evanshultz:
**DO NOT MERGE YET AS ALL ITEMS BELOW ARE NOT CHECKED OFF**

This is a big one, so grab a cup of coffee and sit down in a comfy chair. Or... you can trust me and merge with a quick review. :)

I set out to address footprints listed at https://github.com/KiCad/kicad-footprints/issues/266, and I found a few more with outdated pad designs so I fixed them up too.

Changed multiple overlapping pads to a single custom pad:
- [ ]  Inductor_SMD/L_Coilcraft_LPS4018 (also moved silkscreen from under part body)
- [ ]  Inductor_SMD/L_Wuerth_WE-TPC-3816
- [ ]  Package_DFN_QFN/Infineon_MLPQ-16-14-1EP_4x4mm_P0.5mm
- [ ]  Package_DFN_QFN/Infineon_MLPQ-40-32-1EP_7x7mm_P0.5mm
- [ ]  Package_DFN_QFN/Infineon_MLPQ-48-1EP_7x7mm_P0.5mm_Pad5.15x5.15mm (rename?)
- [ ]  Package_DFN_QFN/Infineon_MLPQ-48-1EP_7x7mm_P0.5mm_Pad5.55x5.55mm (rename?)
- [ ]  Package_DFN_QFN/Infineon_PQFN-22-15-4EP_6x5mm_P0.65mm
- [ ]  Package_DFN_QFN/Infineon_PQFN-44-31-5EP_7x7mm_P0.5mm
- [ ]  Package_DFN_QFN/Linear_DE14MA (overlapping pins had different pin numbers(!) and updated copper and fixed 3D model name)
- [ ]  Package_DFN_QFN/ST_UFQFPN-20_3x3mm_P0.5mm
- [ ]  Package_TO_SOT_SMD/HVSOF5 (and fix pad X position)
- [ ]  Package_TO_SOT_SMD/LFPAK33
- [ ]  Package_TO_SOT_SMD/LFPAK56 (I didn't touch the other LFPAK56 because they're non-standard and perhaps should be removed?)
- [ ]  Package_TO_SOT_SMD/PQFN_8x8 (this needed a few dimensional fixes too)
- [ ]  Package_TO_SOT_SMD/SOT-89-3 (no referenced datasheet so picked the closest one and updated the footprint to match)
- [ ]  Package_TO_SOT_SMD/SOT-89-3_HandSoldering
- [ ]  Package_TO_SOT_SMD/SOT-89-5
- [ ]  Package_TO_SOT_SMD/SOT-89-5_HandSoldering
- [ ]  Package_TO_SOT_SMD/TDSON-8-1 (nearly all dimensions were wrong so this got fixed too)
- [ ]  Package_TO_SOT_SMD/Vishay_PowerPAK_SC70-6L_Single

Removed pin number from paste pads:
- [ ]  Package_TO_SOT_SMD/ATPAK-2
- [ ]  Package_TO_SOT_SMD/ROHM_HRP7

Merged multiple adjacent pads into a single one:
- [ ]  Package_DFN_QFN/DFN-S-8-1EP_6x5mm_P1.27mm
- [ ]  Package_DFN_QFN/Linear_UGK52_QFN-46-52
- [ ]  Package_DFN_QFN/Microchip_8E-16
- [ ]  Package_DFN_QFN/Microsemi_QFN-40-32-2EP_6x8mm_P0.5mm (also fixed line widths)
- [ ]  Package_DFN_QFN/Mini-Circuits_DL805 (and removed silk to mark pin 1)
- [ ]  Package_DFN_QFN/Texas_RUM0016A
- [ ]  Package_DFN_QFN/Texas_R_PUQFN-N12
- [ ]  Package_DFN_QFN/Texas_VQFN-RGR-20
- [ ]  Package_DFN_QFN/Texas_VQFN-RGW-20
- [ ]  Package_DFN_QFN/Texas_WQFN-MR-100_3x3-DapStencil (and updated silk pin 1 mark)
- [ ]  Package_DFN_QFN/Texas_WQFN-MR-100_ThermalVias_3x3-DapStencil (and updated silk pin 1 mark and added missing vias from doc and rename to put _ThermalVias at the end?)
- [ ]  Package_DFN_QFN/UDFN-4_1x1mm_P0.65mm (and updated silk and moved courtyard to 0.25mm clearance and added datasheet)
- [ ]  Package_TO_SOT_SMD/Infineon_PG-HSOF-8-1 (basically a total do-over)
- [ ]  Package_TO_SOT_SMD/Infineon_PG-HSOF-8-1_ThermalVias (basically a total do-over)
- [ ]  Package_TO_SOT_SMD/Infineon_PG-HSOF-8-2 (new footprint)
- [ ]  Package_TO_SOT_SMD/Infineon_PG-HSOF-8-2_ThermalVias (new footprint)
- [ ]  Package_TO_SOT_SMD/Infineon_PG-HSOF-8-2_ThermalVias2 (new footprint)

Other changes:
- [ ]  Package_TO_SOT_SMD/TDSON-8-1_HandSoldering: Removed since the X dimension was +/-3.5mm and the fixed reflow one is +/-3.325mm

I left alone QFNs with odd-shaped corner pads and/or D-shaped pads since I believe Rene's fancy generator script can (or could) make them.

Dimensioned drawings and screenshots, where relevant, to follow.

------------


evanshultz:
Obviously a ton of Travis errors, mostly because these are old footprints. I'll work through them.

evanshultz:
Coilcraft LPS4018

https://www.coilcraft.com/pdfs/lps4018.pdf

image

image

KSU appears to only import the first primitive of a custom pad so double the numbers on the left side for pad heights.

evanshultz:
http://www.analog.com/media/en/package-pcb-resources/package/pkg_pdf/ltc-legacy-dfn/05081731_C_DE14MA.pdf

image

The dimensioned drawing isn't even worth doing since what I changed are pads 4 and 9 and KSU doesn't import the custom pad shapes (yet?).

evanshultz:
https://www.rohm.com/datasheet/BU52014HFV/bu52004gul-e

image

image

I fixed the courtyard after dimensioning this.

evanshultz:
https://www.fairchildsemi.com/datasheets/FD/FDMT80080DC.pdf

image

image

I fixed the courtyard after dimensioning.

evanshultz:
http://ww1.microchip.com/downloads/en/DeviceDoc/3L_SOT-89_MB_C04-029C.pdf

image

image

I kept the existing body width of 5mm even though this Microchip datasheet shows 4.5mm.

evanshultz:
https://www.infineon.com/cms/en/product/packages/PG-TDSON/PG-TDSON-8-1/

image

image

evanshultz:
http://ww1.microchip.com/downloads/en/DeviceDoc/MIC550x-300mA-Single-Output-LDO-in-Small-Packages-DS20006006A.pdf

image

image

I'm not sure what is the correct courtyard spacing for this. Should it be 0.25mm like other DFNs? 1mm like BGAs? Either way, it's not equal on all sides so something is wrong.

evanshultz:
https://www.infineon.com/cms/en/product/packages/PG-HSOF/PG-HSOF-8-1/

https://www.infineon.com/cms/en/product/packages/PG-HSOF/PG-HSOF-8-2/

image

image

image

evanshultz:
That's all the footprints that need dimensioned drawings, I think. Next I'll show screenshots for all footprints (save the Wuerth one that isn't finished) in the order they're listed above unless the changes was simply making one big pads instead of a matrix of pads.

image

image
image
image

image

image

image

image

image

image
image

image

image

image

image

image

image

image

evanshultz:
That leaves the following questions
1. Should `Package_DFN_QFN/Infineon_MLPQ-48-1EP_7x7mm_P0.5mm_Pad5.x5x5.x5mm` footprints have the `_Pad*` removed from their 3D model name? I think so.
1. Should the same footprints use a different nomenclature? I don't know why they got merged this way but the `_Pad*` should be `_EP*`. Change it?
1. Remove LFPAK56 with vias since they're not from the datasheet and seem like a specific implementation? All vias are way outside the footprint boundary. It seems to me this should be done on the board or a local library but not in our official library.
1. Should SOT-89 footprints shrink the body size to match the datasheet? I chose one with the biggest pads that were similar (but still a bit smaller) than the footprint in our lib which had no datasheet.
1. Rename `Texas_WQFN-MR-100_*3x3-DapStencil` to something better? Perhaps just `Texas_S-PWQFN-N100` or `Texas_RKM-100` (we should really figure out how we are going to name these weird TI footprints) and of course the `_ThermalVias` version?
1. Courtyard clearance for `UDFN-4_1x1mm_P0.65mm`?

I wrote a short script to search libraries for the footprint used so that can help us track down some of the potential impact of the items above.

evanshultz:
I just pushed the last footprint I'm aware of that belongs in this group of footprints.
image

image

This is now ready for review. Please see the questions above.

evanshultz:
Just commited changes to the Package_SO lib. Didn't notice those before. In general I just kept them the same as before only merging pads. The list in the first comment is updated.

image
image
image

Misca1234:
@KiCad/librarians 
> This is a big one, so grab a cup of coffee and sit down in a comfy chair. 
> Or... you can trust me and merge with a quick review. :)

Anyone oppose to trust him ?

poeschlr:
The main thing i would suggest to check is that all pads have zone connect set to solid instead of none (for polygon pads the default is none so it could easily happen. There sadly is no option to select thermals for polygon pads.)

evanshultz:
@poeschlr 
So what you write only applies, to polygon pads, right?

I disagree with having a solid connection on all pads because that will make soldering difficult, but if the only choices are Solid and None I understand why Solid must be chosen. Sigh.

Please confirm the below and I'll update.

@Misca1234 
That should work for footprints with only custom pads.

However, most of these footprints have only a few custom pads and so I believe we will then use the pad-specific connection setting seen here:
image

We would not want Solid connection everywhere if it doesn't have to be, so we can do it per-pad on many footprints.

evanshultz:
Oops. I didn't mean to submit that comment above yet.

During some quick experimentation, the per-footprint settings appear to have no effect if there are only custom pads. Only the per-pad settings seem to have any effect.

@poeschlr 
So this means every single custom pad in this PR needs the pad-specific zone connection setting changed to Solid, right? And this should be the default for all custom pads in the library with the current behavior of KiCad?

evanshultz:
@poeschlr @jkriege2 @Misca1234 @antoniovazquezblanco @DanSGiesbrecht 
And before merging, I would like to answer the questions posed at https://github.com/KiCad/kicad-footprints/pull/981#issuecomment-425599992.

evanshultz:
Any and all: please see and respond to the questions mentioned above.

@poeschlr 
Please confirm you want the per-pad zone connection set to Solid as I described above.

evanshultz:
There are a number of open items above that I'd like help to answer.

@poeschlr 
Can you please confirm the change above? I want to make sure I'm understanding.

evanshultz:
@poeschlr 
Another poke.

evanshultz:
@poeschlr 
Please review my question above about zone connection.

poeschlr:
For custom pads we only have the option solid. (And none but i feel that one will do more harm than good.) This is a known restriction of custom pads and the reason why v6 will get a chamfered pad.

evanshultz:
Thanks for answering. All custom pads should be set to Solid connection now except for the four that GitHub has trouble figuring out (they are unchecked above).

That leaves the questions I posed at https://github.com/KiCad/kicad-footprints/pull/981#issuecomment-425599992.

evanshultz:
@poeschlr (and anyone else interested in contributing)
Can you please look at the questions I raised at https://github.com/KiCad/kicad-footprints/pull/981#issuecomment-425599992? I'd like to get this PR ready for a review and those need answer first. Thanks!

poeschlr:
To 1: Yes. (but the answer to 2 makes this question redundant.)
To 2: Without looking at the footprint i would still guess that yes it should be EP.
To 3: yes the current thermal vias version seems strange. I would replace it with a better one that does not increase the pad area.
To 4: Yes any footprint must match the dimensions given in the datasheet assigned to it (exception: If we use industry standards to create the footprint. Then the input to the calculation must match the dimensions given in the datasheet.)
To 5: Texas_S-PWQFN-N100 would fit other footprints. If it has an exposed pad then the size of it should also be included.

evanshultz:
Thanks for answering!

1 & 2. The 5.15x5.15mm datasheet actually shows a 5.55x5.55mm EP, so I removed the 5.15 one and changed `Pad` to `EP`. No symbols are currently using this footprint, but that's because they were left out of the v5 library migration. While it doesn't depend on this PR in any way, I submitted https://github.com/KiCad/kicad-symbols/pull/1395 to re-add them.
3. Removed.
4. Updated using dimensions from IPC-SM-782 page 96, which also matched both referenced datasheets (2.5x4.5mm).
5. Renamed to `Texas_S-PWQFN-N100_EP5.5x5.5mm`. One symbol was affected and https://github.com/KiCad/kicad-symbols/pull/1394 will fix it. Since the symbol PR above is substantial I broke this one out separately.

What about item 6?

evanshultz:
@poeschlr 
Can you check on item 6, referenced above? The courtyard offset was 1mm, but I don't see why it shouldn't be 0.25mm.

poeschlr:
If the udfn footprint does not give a specific clearance in the datasheet then i would suggest 0.25mm clearance. (It is not a bga footprint but more like a lga or qfn. both of these classes have 0.25mm clearance set in the IPC standards)

evanshultz:
Great. That's what I thought. Courtyard updated:
image

So... anything else?

evanshultz:
@poeschlr 
Can this be merged now?

evanshultz:
Note to self: In addition to the linked PRs above, https://github.com/KiCad/kicad-symbols/pull/1395 is using the renamed Infineon MLPQ-48 footprints.

evanshultz:
@poeschlr 
Do you see anything stopping this from being merged?

poeschlr:
There seems to be a merge conflict now. Could you look into that?

evanshultz:
OK. Branch conflicts fixed and Travis is happy.

I only touched the few QFNs mentioned above. I couldn't know or expect somebody would do so much replacing of existing QFNs/QFNs in the last, gosh, ~4.5 months since I created this PR. I don't see anything that went wrong so I believe this is still good to merge.

evanshultz:
@poeschlr 
Be warned that is this is merged at least one symbol PR needs to be merged and there are several other PRs that will then be "unlocked". If you merge this I'll go around to the other PRs and make a note there.

evanshultz:
@poeschlr 
You've done most of the answering and advice here. Is this able to be merged?

poeschlr:
I am not prepared to merge footprints that have not been checked in detail. I also do not currently have the time (or energy) to check 50 plus footprints in one go. 
Right now i suspect this will require a few weeks of slow and tedious work. (Meaning very low chance of this being done for v5.1 which is likely the source of your frustration.)

For the future please limit manual changes to the footprints to a handful of parts per pull request. Scripted changes for the most part are a bit different with this requirement. But this also only applies if the generated footprints are highly similar. If they are completely different then we can not simply only check one or two extremes and we are back to the same requirement as for manual changes.

---

My reasoning behind the requirement for thorough checks stems from my experience of converting footprints over to the script. I discovered that a few years ago the library team only checked contributions against the KLC not against the datasheet. Nearly all these old parts have slight errors (especially when they include exposed pads.) I do not want to repeat the same mistake and leave the library in a similar state for future librarians.
This is compounded in this case from my knowledge of how bad the polygon pad drafting is in kicad. -> Makes it likely to make subtle mistakes.

evanshultz:
I included dimensioned drawings above to hopefully bootstrap the review process. Is there something more contributors can do? (Splitting changes up into multiple PRs is the same amount of work, just distributed.)

evanshultz:
Remove the Infineon QFN needed by https://github.com/KiCad/kicad-symbols/pull/1395 and popped it into a PR all by itself at https://github.com/KiCad/kicad-footprints/pull/1414.

myfreescalewebpage:
Conflict due to merging of #1459 to be fixed @evanshultz

myfreescalewebpage:
@evanshultz looking at the git diff, I'm not sure of the deletion of the two footprints we have discussed...

evanshultz:
Here it is in renaming of these files in the web editor:
image

And I checked the files in this branch again and the `DapStencil` one don't exist.

Am I missing something?

myfreescalewebpage:
Following the integration of the other PR, the new footprint just simply exist.
In the current PR, we should just see that the DapStencil footprint are removed, not renamed, else you will overwrite the new footprints.

evanshultz:
@antoniovazquezblanco 
While the Infineon footprint has been merged, the Wuerth footprint is at https://github.com/KiCad/kicad-footprints/pull/981 and has not been merged.  So I think it's premature to merge this. Once I wrap up a couple things I will break apart #981 for easier merging.

poeschlr:
@evanshultz is this still relevant or was it replaced with the other pull requests?

evanshultz:
I believe all changes here should be in separate PRs, but I kept this open as a 'meta' bug to track progress.

myfreescalewebpage:
@evanshultz I think @poeschlr asks if the PR is not actually fully fixed ? The checkboxes in the first comment are all checked. It's difficult to know the remaining task actually on this.

jneiva08:
@evanshultz i need to make footprint for Wuerth TPC with another size
in the description you have put, "StepUp generated footprint"
How can use StepUp to make?

evanshultz:
@jneiva08 
I am no expert but I got help from @easyw at https://github.com/KiCad/kicad-footprints/pull/972. Hopefully that documents my struggles enough for you to follow along.

PR #983, opened 2018-09-28
Added Relay Fujitsu FTR-F1

Cylindric:
I am adding a Fujitsu FTR-F1 Relay (DPST and DPDT variants) symbol, footprint and model.

[FTR-F1 Datasheet](https://www.fujitsu.com/downloads/MICRO/fcai/relays/ftr-f1.pdf)

The model was created using Fusion360, exported to STEP and converted using FreeCAD StepUp tools.

FTR-1A
ftr-1a-footprint

FTR-1B
ftr-1c-footprint

[Symbol PR](https://github.com/KiCad/kicad-symbols/pull/)
[Footprint PR](https://github.com/KiCad/kicad-footprints/pull/)
[Package PR](https://github.com/KiCad/kicad-packages3D/pull/410)
[Package-Source PR](https://github.com/KiCad/kicad-packages3D-source/pull/)


Cylindric:
Fixed the rotation, and a bit of SilkS costs nothing so added the pin-1 marker to it too.

ftr-1a-footprint

evanshultz:
KLC F5.2 clause 4 (http://kicad-pcb.org/libraries/klc/F5.2/) recommends the fab ref des text size to be 1mm. Here it is 1.5mm but there is no reason to deviate from the recommendation. This PR should not have been merged with a larger-than-1mm text size. While there is an allowable text size range that goes above 1mm, you will see this is not done in the library and 1mm or smaller is used in practice.

Cylindric:
Aah, I made it larger due to the large footprint, as per the KLC 5.2.4c "Size of text should be scaled to match component size".

There are _so_ many old non-compliant parts in the libraries it's hard to get a feel for the current conventions if they differ from the KLC. What's the process for updating the KLC? It seems strange to have quite specific guidelines that are then expected to be ignored.

Happy to submit a minor update PR to change the text if you want.

evanshultz:
Indeed. Some repos have a lot of old footprints, others are better.

The KLC section of the website is at https://github.com/KiCad/kicad-website/tree/master/content/klc.

Cylindric:
I've only submitted a few parts so far, and those have needed a fair amount of revision - pretty sure I'm not yet qualified to suggest alterations to the KLC, but can make a beginner's observations ;)

poeschlr:
@Cylindric wrote:
> There are so many old non-compliant parts in the libraries it's hard to get a feel for the current conventions if they differ from the KLC.

There is only a small fraction of footprints not compliant with KLC. Most offenders have been fixed during the transfer process to the new repo. If you work on a particular lib that seems very bad look at another one as guidance.

> KLC 5.2.4c "Size of text should be scaled to match component size".

The problem is more that we typically use an upper limit of 1mm for text sizes. Even if KLC allows for up to 2mm. (I think that part of the rule is there for legacy reasons and can be changed.)

The lower limit of 0.1mm also seems a bit excessive to me.
relevant issue: https://github.com/KiCad/kicad-website/issues/341

Cylindric:
No worries, I'll happily submit a minor patch to change that text size, and will make a note to myself for future reference.

From a beginner's point of view, after submitting a few parts that needed a lot of fixes, I must say you folks are crazy patient and fast to reply - many thanks for helping to create such a great library of parts!

Misca1234:
No problem, we are here for you

PR #985, opened 2018-09-28
Add Keystone 2462 2xAA battery holder

SeanDS:
This is the Keystone 2462 2xAA battery holder. It is based on the Keystone 2468 2xAAA part already in the library, but with updated dimensions based on the datasheet drawing.

[Datasheet](https://www.keyelco.com/product-pdf.cfm?p=1027)
[Product page](https://www.keyelco.com/product.cfm/product_id/1027)

image


Misca1234:
Hi

Thx for your contribution

The "pig" holes should be 3.30 in diameter and with the distance 14.99
The pin holes should have the distance 14.99
Distance between pig holes and pins should be 27,18

The overall width should be 59.51
The overall width should be 32,92

Just move them a little and all will be good

SeanDS:
Hi @Misca1234, I guess 32,92 is the overall *height*, not width? And, which outline should I make equal to the overall width and height: the F.Fab layer, the F.SilkS or the F.CrtYd?

poeschlr:
The fab outline should represent the exact dimensions of the part.
The silk outline should be just outside of that (inside silk is ok for tht parts if it aids in manual manufacturing)
The courtyard is determined by the required clearance for the particular part (0.25mm in this case). It measured relative to the body it self (so the fab outline). If pads are outside the body dimension then these must also be taken into account for generating the courtyard (Not the case here)

SeanDS:
Thanks, I think I've fixed it now.

Misca1234:
Great  Thanks

Just one single thing left, the distance of the F.CrtYd line 

Left x cord -2,84
Top y cord -9.22
Right x cord 57,17
Bottom y cord 24,21

then we all done

PS
To next time, checkout repository 

in directory
https://github.com/pointhi/kicad-footprint-generator/tree/master/scripts/Battery

You will find a python file and a .yml file, these two files are used to generate the type of foot print you have made.
You add the parameters to the .yml file and execute the python file and then the foot print is automatically made.
Now you just push the .kicad_mod file as before and the modified .yml file to
https://github.com/pointhi/kicad-footprint-generator/tree/master/scripts/Battery

In this way you just add the parameters as stated in the data sheet and dont have to worry about F.Fab, F.Silk, FFab etc.
And we will only review the .yml file (we trust the python script)
DS

Misca1234:
Perfect, thanks a lot

SeanDS:
Done!

Thanks for the instructions regarding the battery script - I didn't realise the scripting had been improved so much for KiCad 5! I am used to the basic set of scripts for IC footprints in KiCad 4.

Since this is already scripted, should I instead add this to the Battery YML file and delete the footprint I made?

Misca1234:
> Thanks for the instructions regarding the battery script - I didn't realise the scripting had been improved so much for KiCad 5! 

The script support is very nice

If you are adding soic, dfn,qfn.son FP, use this one, 
https://github.com/pointhi/kicad-footprint-generator/tree/master/scripts/Packages
For BGA
https://github.com/pointhi/kicad-footprint-generator/tree/master/scripts/Package_BGA

Actually most 3D models can be generated as well
https://github.com/easyw/kicad-3d-models-in-freecad/tree/master/cadquery/FCAD_script_generator

Misca1234:
> Since this is already scripted, should I instead add this to the Battery YML file and delete the 
> footprint I made?

For me it is ok that it is hand made, if you want to make a script based that is up to you, however if you will make additional batteryholder, I recommend you use the script

PR #987, opened 2018-09-30
Added Amphenol USB micro-B connector 10118193-0001LF

tmineno:
Drawing: https://cdn.amphenol-icc.com/media/wysiwyg/files/drawing/10118193.pdf
Specsheet: https://cdn.amphenol-icc.com/media/wysiwyg/files/documentation/gs-12-641.pdf
untitled

------------


poeschlr:
Is this intended for the version with or without flange?

You used non plated through holes for the mounting pads but you did not set the pad size equal to the hole size. Either use plated if you want a copper pad or set hole and pad size to equal. (I would suggest the use of plated through holes and assigning the same pin number as for the other body contacts.)

Pad 1 marker is missing from the fab layer.

The edge mark should be moved to the user drawings layer plus a comment on the user comment layer.

The land patter looks good in general but the body outline (fab layer) is not quite right. Also the courtyard outline is not correct as it should have a clearance of 0.5mm all around (rounded to nearest 0.01mm grid)

usb svg

PR #988, opened 2018-10-02
Button_Switch_THT: Fix courtyard layer in EG1271

antoniovazquezblanco:
Courtyard was placed in the bottom layer instead of the top. Changed it.


evanshultz:
Good catch.

antoniovazquezblanco:
Thanks!

PR #989, opened 2018-10-02
Add Infineon PG-LLGA-5-1

evanshultz:
https://www.infineon.com/cms/en/product/packages/PG-LLGA/PG-LLGA-5-1/

The drawing appears to show mask removed under the entire part body, so I made a polygon on F.Mask. Am I misinterpreting the footprint drawing?

FreeCad source file at https://github.com/KiCad/kicad-footprints/pull/972.

Cu + Mask:
image

Paste:
image

I don't know why the Cu donut is offset when importing the footprint but at least the dimensions are correct.
image

------------


evanshultz:
I'm not sure why Travis thinks the footprint is not centered, but all pads are within the fab outline and the fab outline is centered so this is bogus.

easyw:
@evanshultz 

> I don't know why the Cu donut is offset when importing the footprint but at least the dimensions are correct.

this is a bug in ksu reloading the fp inside FC .. I need to fix it 😉

evanshultz:
@easyw 
Thanks for the confirmation. Would you like me to make any issues at https://github.com/easyw/kicadStepUpMod or are you tracking everything?

poeschlr:
The center calculation only takes pads into account not fab outlines. It might be beneficial to change that.

easyw:
> Thanks for the confirmation. Would you like me to make any issues

it should be fixed now [PR](https://github.com/easyw/kicadStepUpMod/commit/a99ba74cb973243c8072a9799413dea5b4025687)

poeschlr:
It seems this package is specialized for microphones. Meaning the Sensor_Audio lib might be a better fit.

Why is there a polygon on the mask layer?
https://www.infineon.com/dgdl/Infineon-Board+Assembly+Recommendations-Open+Sensors+Leadless-P-v04_00-EN.pdf?fileId=5546d462602a9dc8016045cdeb5f1a47#page=8 shows soldermask like normal.

Even if soldermask needs to be removed it should be done using a mask only pad instead of a polygon.

evanshultz:
Thanks for reviewing! The mask polygon is there because I mis-interpreted the footprint drawing. Fixed now by adding F.Mask t o all 5 pads and removing the polygon, and I also footprint moved to the library you mentioned above. Anything else?

poeschlr:
Looks good, thanks

PR #990, opened 2018-10-03
Add 44- and 64-pin dual row QFN

fauxpark:
Used by several Atmel MCUs. Not sure how to name them, it seems to be referred to in the datasheet as "DRQFN".
Pad sizes are again just nominal lead size so they probably need adjusting to conform to IPC specs.

Datasheet: http://ww1.microchip.com/downloads/en/DeviceDoc/Atmel-8272-8-bit-AVR-microcontroller-ATmega164A_PA-324A_PA-644A_PA-1284_P_datasheet.pdf#page=646

Datasheet: http://ww1.microchip.com/downloads/en/DeviceDoc/Atmel-8074-8-and-16-bit-AVR-Microcontroller-XMEGA-B-ATxmega64B3-ATxmega128B3_Datasheet.pdf#page=70

------------


evanshultz:
So you made these by hand, right? Not scripted?

Instead of the datasheet as documentation, could you use the Microchip packaging info? It also includes footprint recommendations. For the 64 pin on, it's either http://ww1.microchip.com/downloads/en/DeviceDoc/64L_VQFN_7x7_Dual_Row_[SVB]_C04-21420a.pdf or http://ww1.microchip.com/downloads/en/DeviceDoc/64L_VQFN_7x7_Dual_Row_[SUB]_C04-21420a.pdf. Can you figure out which one (both say ZXC so I gave up at that point) and find the right 44 pin one? Your silk style is perfect and we should be able to use rounded rect pads in all locations (IPC 7351C is replacing oblong/oval and D-shape with rounded rect) but it may take seeing it built to know if this will maintain at least 0.2mm clearance between all pads. We can dig into the heel and toe requirements from IPC for this type of package if we need to.

Also, could you include a thermal vias version of each? A grid of 0.3mm drill (0.6mm annular ring) vias about 1mm apart (3x3 for the 2.6mm EP and 5x5 for the 4mm EP) should be good.

To confirm, symbols using these footprints are not in the lib yet, right? I don't see them but I could just be missing them.

poeschlr:
That is sadly not supported by my scripts right now. It is not really easily included in one of them.

fauxpark:
Yes, these are made by hand. I'll have a look at those footprints and the thermal vias.

I haven't added any of the symbols that use these footprints yet, but here they are:
```
ATmega164A-MCH     44MC
ATmega164PA-MCH    44MC
ATmega324A-MCH     44MC
ATmega324PA-MCH    44MC
http://ww1.microchip.com/downloads/en/DeviceDoc/Atmel-8272-8-bit-AVR-microcontroller-ATmega164A_PA-324A_PA-644A_PA-1284_P_datasheet.pdf

ATmega165A-MCH     64MC
ATmega165PA-MCH    64MC
http://ww1.microchip.com/downloads/en/DeviceDoc/atmel-8285-8-bit-avr-microcontroller-atmega165a_pa_325a_pa_3250a_pa_645a_p_6450a_p_datasheet.pdf

ATmega169A-MCH     64MC
ATmega169PA-MCH    64MC
http://ww1.microchip.com/downloads/en/DeviceDoc/atmel-8284-8-bit-avr-microcontroller-atmega169a_pa_329a_pa_3290a_pa_649a_p_6490a_p_datasheet.pdf

ATmega169PV-8MCH   64MC
ATmega169P-16MCH   64MC
http://ww1.microchip.com/downloads/en/DeviceDoc/doc8018.pdf

ATxmega128B3-MCU   64P2
http://ww1.microchip.com/downloads/en/DeviceDoc/Atmel-8074-8-and-16-bit-AVR-Microcontroller-XMEGA-B-ATxmega64B3-ATxmega128B3_Datasheet.pdf
```

fauxpark:
Here are the _ThermalVias footprints:

evanshultz:
- The naming bothers me. We need some clear way to designate that these are not normal QFNs but instead have two rows of pins. JEDEC uses "Multi Row" ("Dual Row" in this case) so perhaps we can append `_DualRow` before the EP size? Putting it there means a trailing asterisk on the footprint filter would be sure to catch the `DualRow` footprints but would still allow flexibility for EP size and the choice of a footprint without thermal vias. @poeschlr @Ratfink 
- I don't know about the center pad pin number. We usually increment it, even if the pads aren't just numeric (see `Texas_WQFN-MR-100_3x3-DapStencil`), so perhaps B31 would be valid. But that just seems weird to me. And `Nordic_AQFN-73-1EP_7x7mm_P0.5mm` does set a precedent for `EP`. It may be, however, that neither are a good path to follow. @poeschlr ?
- I rather dislike rotated pads but it's not in KLC so you can leave them if you wish.
- The radius of the thermal pad isn't shown, but IPC 7351C drafts show 0.25mm max so let's use that.
- One thermal pad corner isn't beveled near pad A1 but I think that should be OK.
- Can you make the QFN-64 inner row D-shaped pads as shown in the doc? You will need to use a custom pad shape.
- QFN-64 via spacing is wrong.

image

image

fauxpark:
The datasheets call these "DRQFN" (most of the time). The symbol descriptions can then use "..., DRQFN-64" etc.

---

> Can you make the QFN-64 inner row D-shaped pads as shown in the doc?

AFAICT there is sufficient clearance between the A and B pads, as you mentioned above:

> we should be able to use rounded rect pads in all locations (IPC 7351C is replacing oblong/oval and D-shape with rounded rect) but it may take seeing it built to know if this will maintain at least 0.2mm clearance between all pads

---

> QFN-64 via spacing is wrong.

I increased the spacing a little because it seemed the paste pads were too small for the thermal pad size.

poeschlr:
The dual row suffix is a good idea.

evanshultz:
Oh yes. You mentioned DRQFN above. Sorry that I'd forgotten. It seems to be widely used when I Google it so that works for me.

I'm not sure why Microchip would use oval for the inner row on the -44 but not on the -64. Odd. But if we're going to deviate from what Microchip says and try to stick with IPC we should use 25% (0.25mm max) corner radii. Right now the radius is so big the pads are oval. So would you prefer to match Microchip or IPC (at least on pad shape)? If we're sticking strictly with Microchip's footprint we will need to use that prefix to the footprint name.

With nine 1mm paste pads coverage is ~54%. With your 1.15mm ones it's ~71%. So if we're not making this per Microchip this deviation is fine.

evanshultz:
Right. Good catch. Well, oval pads it is then. I'd rather have the clearance and make sure the part can solder without bridging. I suspect users would appreciate that too.

Can you wrap up the other questions I mentioned above? Thanks!

fauxpark:
@evanshultz I fixed the EP radius and added _DualRow to the footprint names. Was there anything else?

evanshultz:
I was able to clarify with an IPC official guy that round rect or square is OK for thermal pads, it depends on the manufacturer recommendations. Since the doc shows rounded rect, and without other guidance, I'm OK with this radius change. Thanks!

There are two things after taking another look:
1. As it appears the pads are identical in shape if they're oval versus round rect with a maximum radius. If you don't mind, I'd rather use oval since that's more compatible with older versions of KiCad and easier to visually read and understand the footprint file. (Both may not matter but I don't see a drawback.)
1. I was good with naming the footprint DRQFN. It's more concise which helps to show more info in the UI, and I think even with a silly *QFN footprint filter it will be OK since the pad count filter in CvPcb plus the symbol's footprint filter makes any other QFN of the same size unlikely (for a single row of 64 pins it will likely be a larger package). So if you want to change to the other nomenclature it's OK with me.

Lastly, I'd like to make this a manufacturer-specific footprint. I'm thinking DRQFN is not standardized so can you put `Microchip_` at the front of these footprints? We can always make a change if it turns out this is package is common and there is broad agreement about how to make the footprint. But at this point, FWIW, it's not the right time for that. OK with you?

fauxpark:
Done. I'm not sure how useful the switch to oval pads really is though, seeing as the thermal pads still use roundrect, and on the ThermalVias version the radius is required for the paste pads to not intersect with the vias.

evanshultz:
Hmm. You're right. I had forgotten that other pads were shapes required by v5. Well, I suppose either way is fine. Do you have any other thoughts/concerns or is this ready to merge?
image

evanshultz:
Thanks! Now to add the symbols?

PR #992, opened 2018-10-03
Add footprint for XP Power JTE series dual output DC to DC converter.

stambaughw:
Add footprint for XP Power JTE series dual output DC to DC converter.

This is one oddity with this footprint.  The [datasheet](https://www.xppower.com/Portals/0/pdfs/SF_JTE06.pdf) starts the pad numbering at 2 so I followed it accordingly.  I didn't see anything in the library docs about this.  I set the footprint anchor point to pin 2 because setting it to where pin 1 would normally be struck me as not very friendly but I'm not opposed to changing it.

Model pull request https://github.com/KiCad/kicad-packages3D/pull/414.

------------

xp-power-jte-series-footprint


poeschlr:
I assume this is intended for the metal case version? (This would mean the part number would need a suffix -M added)
The metal version has round pins with 0.5mm diameter. So the 0.7mm drill size would be correct for it.
The plastic version has square pins with side length 0.5mm -> diameter would be ~0.707mm. So it would need a drill size of ~0.91mm.

The courtyard clearance should be 0.25mm (+/-0.005)

For reference the dimensioned drawing created from this footprint:
screenshot from 2018-10-03 20-33-02

stambaughw:
It's the plastic case version (no -M suffix) at least that is the 3D
model I made for it.  I'm assuming that 0.25mm courtyard clearance is
from the nearest physical feature (body or pad) on a given side.  I'll
fix the pad drill and the courtyard clearance shortly.  Thank you for
reviewing this.

poeschlr:
Yes the courtyard is relative to either body or pads. In this case only the body needs to be taken into account as the pads are fully covered by it.

stambaughw:
I fixed the pad drill and it seems to leave a rather small solder area.
 Is there a recommended oval pad size for a given drill diameter for
through hole components?

poeschlr:
We only have a minimum requirement in the KLC (shorter length is 0.3mm larger than drill size. This ensures an annular ring >= 0.15mm). 

But we are already discussing if we can give more advice in this issue: https://github.com/KiCad/kicad-footprints/issues/595

stambaughw:
I'll leave the pad size as is.  If we want to change it later, we can
update it at that point.

evanshultz:
Not exactly.

The 0.15mm annular ring is for THTs without a lead and so it doesn't apply here. The issue you linked covers leaded holes from IPC 2221 where we decided to use Level B by default.

That means with this drill of 0.91mm, the annular ring should be 1.51mm minimum. Bigger would also be possible on a case-by-case basis, and your 1.5x2.5mm oval holes should be just fine.

poeschlr:
@stambaughw thanks for your contribution. Everything looks ok now.

PR #993, opened 2018-10-04
Add new TDK Buzzer

Cylindric:
I am adding a TDK PS1240P02BT Buzzer symbol, footprint and model.

[Datasheet](https://product.tdk.com/info/en/catalog/datasheets/piezoelectronic_buzzer_ps_en.pdf)

The model was created using Fusion360, exported to STEP and converted using FreeCAD StepUp tools.

buzzer-footprint

Symbol PR - no symbol - compatible with the generic buzzer in Device.lib
[Footprint PR](https://github.com/KiCad/kicad-footprints/pull/993)
[Package PR](https://github.com/KiCad/kicad-packages3D/pull/415)
[Package-Source PR](https://github.com/KiCad/kicad-packages3D-source/pull/169)


Misca1234:
You have named the 3D model different form the foot print, 
you don't think it would be troublesome ?

Misca1234:
What do you think of doing the pin 1 marking like this ?

bild

Misca1234:
> Symbol PR - no symbol - compatible with the generic buzzer in Device.lib

I guess you assume people will look for a foot print associated with the desired component first and then understand that they should use a generic symbol and then connect the two when they us the "CvPcb" ?

Cylindric:
> You have named the 3D model different form the foot print,
> you don't think it would be troublesome ?
I think that might have been more due to me renaming the part to be more specific. The Model ended up not having all the dimensions in the name. Happy to update the model PR with the same exact name.

> I guess you assume people will look for a foot print associated with the desired component first and then understand that they should use a generic symbol and then connect the two when they us the "CvPcb" ?
Not quite. I assume that people will look for a symbol first, and then determine specific component devices later. I can't imagine that every schematic is drawn with the precise final physical part fully determined. Isn't that why there's a separate stage to associate footprints with symbols?

I guess I'm not sure I understand the question.

We do it with other "trivial" components. We don't want a separate (identical) symbol for every possible manufacturer part do we? I could add a "TDK_PS1240P02BT" alias to the buzzer if you like. It's going to be a device class that has a lot of specific options for the same symbol though. Surely every two-legged piezo-type buzzer out there is going to have the same symbol? Not convinced it adds much value. Don't people (I'm no pro) design the schematic first, and then determine specific components further down the line? "I need a buzzer. I'll find the exact correct one later". 

If not, why are symbols and footprints separate? Is the intention to have an exact correlation with Symbol <-> footprint <-> model, and only only allow grouping for different packages by the same vendor (DIP/SOP/etc)?

evanshultz:
Different ECAD tools work different ways, but one clearly-conveyed strength of KiCad is to abstract the symbols from the footprints. Using a generic symbol is the way to go here, just as we would do with resistors.

As long as the generic buzzer symbol's footprint filter will catch this footprint, we're good. And since the symbol `Buzzer` in `Device.lib` has a footprint filter of `*Buzzer*`, and this footprint contains the word "Buzzer", all should be well.

Cylindric:
> You have named the 3D model different form the foot print,
I've added a commit to the package to make the name match the footprint.

> What do you think of doing the pin 1 marking like this ?
I like it. I've added a commit to the footprint to make it match this suggestion.

Misca1234:
Nice, just missing one small detail, 
Add the url to the description

From
Buzzer, D12.2mm height 6.5mm
to
Buzzer, D12.2mm height 6.5mm, https://product.tdk.com/info/en/catalog/datasheets/piezoelectronic_buzzer_ps_en.pdf

Cylindric:
Hah, I always forget _something_ ...

Misca1234:
Great thanks

PR #996, opened 2018-10-05
Added BUS_PC

nickysn:
This adds BUS_PC: an 8-bit ISA bus card edge connector, similar to BUS_AT (16-bit ISA). In fact, it was created by cloning BUS_AT and removing the extra short card edge connector that extends the 8-bit ISA bus to 16-bit.

------------


poeschlr:
We would need some sort of document that gives dimensions for this part. Otherwise we can not review it. This document should then also be linked inside the footprints description field.

nickysn:
Ok, I'll try to find something.

nickysn:
Will this one do? Pages 68..73 contain a mechanical specification.
https://archive.org/details/bitsavers_intelbusSpep89_3342148/page/n67

This one is for the edge card connector that gets inserted into the 62-pin connector on figure 11.1.1

evanshultz:
What we would like to have is a document that shows the copper lands that fit into that type of connector/header. To verify the footprint and ensure it should work well with a standard interface we need to know the length, width, spacing, etc. of the copper pads.

For example, https://cdn.amphenol-icc.com/media/wysiwyg/files/drawing/10025026.pdf shows (I didn't check anything myself) PCIe Gen3 copper lands on page 2. Something like this is what we want.

Sadly, I wasn't able to find anything at Farnell, DigiKey, or Mouser. I will look again if I have time but I hope you're able to find something.

Sadly, none of our "golder finger" footprints have a document attribution.

evanshultz:
What about the last page of https://www.hirose.co.kr/Upload/Product/Catalog/e54801112.pdf? It appears the right (longer) section is the 62-pin 8-bit bus? This is a Korean website but at least it's direct from the mating header vendor.

I looked and looked and looked and that's all I could find. Maybe my Google-fu is just failing me right now...

nickysn:
Yes, this is correct, the right section is the 62-pin 8-bit bus. The 16-bit bus connector was intentionally designed so that an 8-bit card would also fit inside it, because the bus is backwards compatible.

nickysn:
Ok, I updated the footprint, according to the Hirose document. If you think it looks good, I'll also update BUS_AT.

image

nickysn:
Just after adjusting this to the Hirose document, I found yet another drawing from a different manufacturer: https://www.te.com/commerce/DocumentDelivery/DDEController?Action=showdoc&DocId=Customer+Drawing%7F645169%7FW%7Fpdf%7FEnglish%7FENG_CD_645169_W.pdf%7F645169-3
Which one should we follow?

nickysn:
Oh, and here's yet another one. Page 1-35 of the IBM 5150 Technical Reference:
http://www.minuszerodegrees.net/manuals/IBM_5150_Technical_Reference_6322507_APR84.pdf

evanshultz:
Thanks for the research!

The IBM one is poorly-dimensioned compared with the others, though it is obviously a rather important source of information for the ISA bus.

Let's see how they compare (might as well look at both 8-bit and 16-bit footprint requirements):

--- | IBM | Hirose | TE
--- | --- | ------ | --
Finger width | 1.524 | 1.6 | 1.27
Finger height | 7.62 | 10 | 6.98
8-bit total pad span | --- | 43.18 | 43.18
8-bit notch width | --- | 47.75 (reference) | 47.9
16-bit total pad span | 76.20 | 76.20 | 76.20
16-bit notch width | 81.026 | 81.03 (reference) | 80.92
Pad-pad distance across slot | --- | 10.16 | 10.16
Slot width | --- | 5.46 | 5.44
Slot depth | --- | 10 | 10.16
Total pad span | --- | 129.54 | 129.54
Total width | --- | 134.24 | 134.26
Bevel angle | 90 (no bevel) | 30 - 45 (0.5 width) | 45 (0.64 height)

The finger height includes the undimensioned backoff on the Hirose drawing to result in the total distance from the PCB edge to the pad top.

Assuming I gather the dimensions correctly, which you should check, they match quite well except for the finger dimensions and bevel. So here's how I'm seeing those items:
1. I think adding a bevel is worthwhile.
1. The contact point is ~4mm from the PCB edge on both Hirose and TE datasheets and the depth the PCB fits into the connector is ~7.5mm so the 10mm Hirose fingers seem excessive and consume PCB space for no benefit.
1. I don't see the width of the contact specified but since there can't be other parts around the fingers I don't see why wider fingers wouldn't be more robust at no detriment to PCB cost or reliability.

Please consider my thoughts on the above and then suggest a reconciliation for the differing measurements above. I have little experience with this type of card edge connector so if you know what works what would be great to incorporate.

evanshultz:
@nickysn 
Will you be getting back to this PR to finish it off?

nickysn:
I hope so. I was busy with other stuff, but I'll try to find some time to finish this.

antoniovazquezblanco:
@nickysn are you still interested in this?

Thanks!

nickysn:
I wanted to take measurements from several of the ISA cards that I have, but I'm currently on vacation, so I can't do it right now. I'll take a look at this, when I come back.

antoniovazquezblanco:
Thank you!

PR #997, opened 2018-10-06
Add Diodes SOT-553

jneiva08:
Datasheet: https://www.diodes.com/assets/Package-Files/SOT553.pdf

footprint

dsot553

------------


Misca1234:
You should make the pads as rounded rectangles

bild

Misca1234:
Great thanks

evanshultz:
Why was this a manufacturer-specific footprint? It's a standard footprint and should have the `Diodes_` prefix. This is a sticking point for merging https://github.com/KiCad/kicad-symbols/pull/1048.

Misca1234:
There is already a SOT-553 in the library that is slightly different

evanshultz:
The existing SOT-553 footprint has no documentation so we have no idea where it came from or why it's built that way. IPC gives us a way to develop standardized footprints from package drawings that removes any variation between vendors using a standard package.

I don't find the reference/standard/source SOT-553 drawing. Does anyone know what organization originated it? It may be JEDEC MO-293 variation UAAD-1; I could be mistaken but I think they match.

@poeschlr Shall we standardize now instead of having two SOT-553 footprints and proliferating what should be unnecessary manufacturer specificity?

PR #998, opened 2018-10-06
Add Vishay PowerPak 1212-8

jneiva08:
Datasheet: https://www.vishay.com/docs/72597/72597.pdf

Screenshot:

pp1212-8
------------


evanshultz:
Nice dimensioned drawing. Thanks!

Would you mind using a single custom pad shape for pin 5 rather than overlapping pads? You can see examples where I've done this for other footprints at https://github.com/KiCad/kicad-footprints/pull/981.

jneiva08:
Have changed to custom shape.

evanshultz:
- Since there are single and dual versions add `_Single` to the end.
- Can you put some general search terms in the description like the other PowerPAK footprints?
- Also include https://www.vishay.com/docs/71656/ppak12128.pdf in the description.
- Please add keywords like other PowerPAK footprints.
- The fab lines should be 3.05x3.05mm since that's the body size ignoring the leads (like SOIC footprints). You should clear the silk under the side leds by at least the silk line width (0.12mm).
- Two dimensional issues: Right side pad total distance is only 2.385mm and width of the big section of pin 5 is not 1.725mm:
image

jneiva08:
Hope i didn't miss anything. This time tried to make pad 5 with less points.

size

evanshultz:
This looks great. Thank you!

PR #999, opened 2018-10-06
Texas WSON-6 DQK

jneiva08:
Datasheet: http://www.ti.com/lit/ds/symlink/csd16301q2.pdf

Screenshot:
footprint
size

------------


Misca1234:
All looks good but I have one question, why do you have an extra pad under 7 and 8 ?


bild

jneiva08:
It was supposed to be "Recommended Stencil Pattern" (Pag.10 from datasheet)

I think what i'm doing wrong. I will fix it

Misca1234:
> It was supposed to be "Recommended Stencil Pattern" (Pag.10 from datasheet)

I am not sure about that, no other FP is using that "constructions", beside those who have thermal vias
but, as you said, page 10 indicates it


BTW, do you intent to make a 3D model as well?
If not, I made one

bild

jneiva08:
fell free to make on :)
for me 3D parts is in another league

Misca1234:
Naaa, this one is particular easy 
https://github.com/KiCad/kicad-packages3D/pull/416

You just add the parameters of the model to a file and then let scripts + FreeCad create the 3D model
https://github.com/easyw/kicad-3d-models-in-freecad/blob/master/cadquery/FCAD_script_generator/QFN_packages/cq_parameters.py

for this one you create the model like this (I use this particular version)
...../.../../../FreeCAD_0.17.11223_x64_dev_win/FreeCAD_0.17.11223_x64_dev_win/bin/FreeCAD.exe main_generator.py Texas_WSON-6_DQK

evanshultz:
- DQK seems to be a TI-specific type of WSON-6 so perhaps `Texas_DQK` would be best? TI package nomenclature is rather confusing but they told me the 3-letter name is the most differentiating.
- Silk to copper clearance looks a bit tight. It should be the silk line width or greater.
- Most commonly I see ref des above the footprint and value below it.

> I am not sure about that, no other FP is using that "constructions"

Many other footprints use exactly this method for paste pads and this is our standard convention when the paste pads are not a uniformly scaled and concentric shape of the copper pad, as is the case here.

Misca1234:
Thx

PR #1006, opened 2018-10-10
Add Jack_3.5mm_Switronic_ST-005-G_horizontal to Connector_Audio

asukiaaa:
Datasheet: http://akizukidenshi.com/download/ds/switronic/ST-005-G.pdf

screenshot from 2018-10-10 20-18-35


Misca1234:
There is a decision to not using numbers on pins on but the enumeration used in the symbol
Have you checked the symbol numeration contra your numeration ?

asukiaaa:
Thank you for the comment.
This footprint may be one of AudioJack3_Ground, so I update numbers to support it.

screenshot from 2018-10-15 21-26-00

screenshot from 2018-10-15 21-21-08

Is it OK to use alphabet to number of footprint?
If not, should I create a symbol for this footprint?

Misca1234:
Well, there is a trend within KiCad for very common symbols to be a generic type and when running CvPcb to match the generic symbol to a specific foot print.

You can test this your self by adding the generic symbol to a sheet and then run CvPcB to see if your foot print appear, if so, all good

Misca1234:
The measurements is in accordiance to the data sheet
If you want you can remove the little circle at pin T

bild


@KiCad/librarians 
According to KLC Pin 1 should be marked in F.Silk, however these types of FP's does not have a "Pin 1" and beside, it is polarized so I guess there is no need for a Pin 1 marker?

asukiaaa:
> If you want you can remove the little circle at pin T

Some audio jack footprints have a circle near the first pin.

screenshot from 2018-10-17 19-41-59

screenshot from 2018-10-17 19-41-37

> I guess there is no need for a Pin 1 marker?

In case that the answer of the question is yes, I agree to remove the circle and change R pin as circle.
Otherwise I want to keep the circle on F.Fab layer.

evanshultz:
The case about footprint pin names is https://github.com/KiCad/kicad-footprints/pull/1014. This has not been approved and merged yet so we're still in a holding pattern over numbers vs names. (The symbols were accepted but the footprints don't match yet so it's not closed.)

I don't find a pin 1 (or pin T or pin "whatever") mark to be useful for parts which have only one possible (usable) insertion. This is also not been generally approved but I'm in favor and would happily remove all markings from the footprints in the PR I mentioned above if this is widely accepted.

@poeschlr @jkriege2 @Ratfink @antoniovazquezblanco @DanSGiesbrecht
Thoughts on the above? This also ties into https://github.com/KiCad/kicad-symbols/pull/592, I feel.

asukiaaa:
Thank you for updating other footprints.

I share a script to create this footprint.
https://github.com/pointhi/kicad-footprint-generator/pull/205

Is there anything to do for merging this footprint?

evanshultz:
Accepting using letters instead of numbers is the blocking point.

asukiaaa:
Does it mean to rollback to first commit version?

https://github.com/KiCad/kicad-footprints/pull/1006#issue-221712938

In that case, is a symbol needed for this footprint?

evanshultz:
A correct symbol with TRS pins exists. We need to merge all footprints with letter pin numbers. So leave this PR as is and we'll get to it once the other librarians are above to review. Thanks for your contribution.

asukiaaa:
Thank you for the explanation.
I understand that.
I wait review and merge.

poeschlr:
@evanshultz could you double check the pin number assignment? I am not sure if T and S should not be exchanged.

---

Nearly everything of the footprint is ok otherwise. Only the courtyard clearance on the right side needs to take the pads into account.

Screenshot from 2019-06-01 16-38-17

asukiaaa:
Thank you for the feedback.
I update clearance of lines on F.CrtYd.

Screenshot from 2019-06-02 15-48-39

poeschlr:
Why has the fab and silk layer been changed? These where ok.

asukiaaa:
Thank you for the feedback.

> Why has the fab and silk layer been changed? These where ok. Edit: and the NPTH pads also moved.

I thought that you wanted me to updating clearance but it was misunderstanding.

> Only the courtyard clearance on the right side needs to take the pads into account.

I still unclear what you want me to do.
Do you want me to change shape of pad R to Circular from Rectangular?

evanshultz:
@poeschlr 
Checking the datasheet to the image at https://github.com/KiCad/kicad-footprints/pull/1006#issuecomment-498004705, the footprint pin names are correct.

asukiaaa:
I'm still unclear but I force push to drop a commit to add clearance.
The footprint backed to previous version that you showed a picture.

poeschlr:
See the drawing i added. There is a dimension marked in pink. This dimension should be 0.5mm

asukiaaa:
> Why has the fab and silk layer been changed? These where ok. Edit: and the NPTH pads also moved.

I understood what you said.
Updating my environment caused changing result of dividing integer by running script.
I got same result with [using float instead of integer](https://github.com/pointhi/kicad-footprint-generator/pull/205/commits/267e3eb06b7ff2e1501396822b16632e1ad72f88).
Sorry for confusing.

I change clealance of CrtYd as 0.5mm.

Screenshot from 2019-06-11 23-04-59

poeschlr:
For the future please avoid force pushes. (I now need to double check everything again as i can not check if the pad and fab changes of the forced push where properly reverted. Without force push i could have simply checked the commit range from when it was ok the last time round.)

---

The courtyard clearance in y direction is not correct.
Screenshot from 2019-09-01 21-11-36

asukiaaa:
Thank you for the feedback.
I updated position of horizontal line on F.CrtYd layer.

> For the future please avoid force pushes.

Sorry for wasting your time.
I will not use force push for reviewed branch.

poeschlr:
Looks good now, thanks

poeschlr:
Damn i missed the fact that you did not add a description other than the datasheet. Could you do that, the part is then ready to be merged.

asukiaaa:
Thank you for checking.
I updated description.
How about it?

poeschlr:
you also removed the reference on fab, add it back please

asukiaaa:
I think that the reference on fab is there.

Screenshot from 2019-09-08 14-36-46

I'm not shure why the position of reference was changed on kicad_mod on commit https://github.com/KiCad/kicad-footprints/pull/1006/commits/f93f827e3c6c6693e4374152ce1604ab6e2cc9ed but the result seems same.
(Maybe updating kicad or os caused this change.)

Screenshot from 2019-09-08 14-37-48

poeschlr:
Oh i kind of misread the diff then. thanks

asukiaaa:
Thank you for checking and merging.

PR #1008, opened 2018-10-10
Fixup 3 dmodel path

Misca1234:
Fixup of the 3D model paths and missing 3D model tag

Relay_DPDT_AXICOM_IMSeries_JLeg
Added ${KISYS3DMOD}/

Texas_SWRA117D_2.4GHz_Left
Texas_SWRA117D_2.4GHz_Right
Added 3D model tag

------------


Misca1234:
Not sure if we should fix the rest as well

evanshultz:
https://github.com/KiCad/kicad-footprints/pull/1008/commits/8fafdc1d11c4086b1cbed08b5c94d44afd6fa97f looks fine, but there are a lot of other changes here. Perhaps you can merge your branch with upstream and then only submit that one commit?

Misca1234:
Hmm, I thought I did that just before I createdthis branch due to the Lair Technology incident

Misca1234:
$ git merge master
Already up to date.

Misca1234:
My knowledge about git is not that deep

Shackmeister:
@Misca1234 I believe it's ok not to have any 3D reference on virtual footprints which doesn't doesnt make any sense to have a model on

evanshultz:
Should these antenna to be virtual? Perhaps having a 3D model is nice so the user can see the antenna in the 3D viewer? Just thinking out loud...

evanshultz:
Aaaahhhh. I was under the impression that virtual components didn't get a 3D model. It just means there's not a physical component that goes on top of the footprint. So the copper of the antenna still shows up and that's all I was thinking about. Thanks.

Yes, indeed I think they should be custom pads. Not necessarily in this PR but put it on the list...

PR #1011, opened 2018-10-11
Molex ffc10

gcormier:
image

https://www.molex.com/molex/products/datasheet.jsp?part=active/0528921033_FFC_FPC_CONNECTORS.xml

3D Model : https://www.molex.com/molex/part/partModels.jsp?&prodLevel=part&partNo=528921033&channel=products

------------


gcormier:
Only outstanding issue is the 3d model.

image

evanshultz:
What is the issue with the 3D model?

You built this by hand, right? We have a script that should be able to build the entire family in one go.

gcormier:
I did indeed. Not sure on how to contribute 3d models.

evanshultz:
Submitting 3D models isn't required as long as the footprint references one so that if and when a 3D model is added it will appear seamlessly in the 3D viewer.

We haven't accepted 3rd party 3D models in general because of licensing. Molex we know can't be used.

Note that I didn't review anything, just chiming in with this info.

gcormier:
Did you want to see any more changes to this?

poeschlr:
The outline of the keepout area should be on dwgs layer. Plus add text describing it onto the comments layer.

The keepout does not line up to the pads correctly.

The courtyard clearance must be 0.5mm all around.

A few body dimensions are a bit wrong.

ffc svg

gcormier:
caef7f2 fixes a few.

[x] Outline moved to appropriate layer
[x] Courtyard clearance should be 0.5mm now

KLC says comments are optional - I can put in "KEEPOUT" but it seems redundant?

For the wrong dimensions do you mean the 10ths of a millimeter? (ie 4.56 instead of 4.5)

The keepout is a pain. Do you want the actual line to line up, or the actual edges of the keepout area?

poeschlr:
All measurements simply need to agree with the datasheet. If the datasheet gives 5 significant digits then the footprint must be made to agree with these. (Luckily most manufacturers at most use 2 significant digits. But i had a footprint just today where a few measurements where given with 4)

This is true for the fab layer, pads and also for keepout areas. The courtyard layer must be rounded to the nearest 0.01 grid point. Silk i don't really care about. Draw it on any grid you like just make sure you have clearance to pads and to the body itself. (meaning fab line)

gcormier:
Okay, I'll try to give it another look this weekend.

Did you have a tool that output the dimensions in your initial review?

poeschlr:
Even better. I have a tutorial for it: https://forum.kicad.info/t/how-to-check-footprint-correctness/9279

gcormier:
I would appreciate your review on 1c70c68 Fab layer before I spend more time on the rest. Please ignore the silkscreen layer off to the side, I hope to nearly re-use it and tweak the offset from the Fab layers.
image

I redid the footprint from scratch using Fusion360 to dimension everything.

image

poeschlr:
The fab outline should be closed. Only the silk outline needs to avoid pads

gcormier:
Here it is.

image

gcormier:
Validated the measurements in Freecad and they've gotten off again.

Just gonna leave this here for future reference if I decide to give another go. The output still needs a lot of work from what is created but perhaps it's a better start with proper grid and coordinates. Someone can probably script the whole batch in python for the whole series which is above my head..

```
Building new KiCad FP Wizard footprint with the following parameters:
Footprint Wizard Name:        FPC (SMT connector)
Footprint Wizard Description: FPC (SMT connector) Footprint Wizard
Pages: 2
Pads
	n: 10
	pitch: 0.5mm
	width: 0.3mm
	height: 1.4mm
Shield
	shield_to_pad: 2.65mm
	from_top: 3.25mm
	width: 1.6mm
	height: 2.2mm
```

Issue #1012, opened 2018-10-11
audio jack symbols use letters; footprints use digits


ppelleti:
The footprints in `Connector_Audio.pretty` have numeric pin numbers, while the corresponding library symbols have alphabetic pin numbers.  This means they don't work together.

I first noticed this when I filed #979, but it also recently came up [on the forum](https://forum.kicad.info/t/connector-audio-stereojack-3-5mm-odd-pin-numbering/13038).

evanshultz:
Thanks for bringing this up.

You are correct in the forum that using letters was most consistent across all manufacturers, provides some intelligence about the pin's use, and also differentiates between the connector pins that go to different jacks (the jacks connected to the cable and inserted into the connector) with combo connectors.

See https://github.com/KiCad/kicad-footprints/issues/191. I have made, I think, new footprints that will match the symbols and will also update any existing ones when I push them. I've been (slowly) gathering feedback and consensus before submitting that PR. It was a mistake to leave this hanging in the v5 libraries and cause this user issue. Apparently I should prioritize getting the footprints wrapped up.

Issue #191, opened 2017-12-22
Missing audio connectors


evanshultz:
https://github.com/KiCad/kicad-footprints/blob/master/TODO.md doesn't include (or doesn't include many) various professional audio connectors:
- 1/4" / phono
- XLR
- speakON

Some of these footprints exist in the old library but are outdated. I'll update and expand this collection of audio-specific connectors.

poeschlr:
Nice catch

terjeio:
@evanshultz I want to make 3D-models for Pro-Signal PSG01550 and PSG01551 RCA/phono sockets, I have footprints for them already but most likely not KLC compliant so needs to be updated. Are these in the list of those you want to add or should I keep them on my todo list?

evanshultz:
Go ahead. I'm going to be adding only Neutrik and Amphenol pro audio connectors (XLR, speakON, etc.) to start because they're generated considered "standard" or "reference" and most other brands of connectors are compatible. Thanks for the proposed contribution!

evanshultz:
@poeschlr @jkriege2 
Question on pin numbers for these audio connectors that don't have pin numbers that are sequential integers starting at 1. Here are some examples (see the PDF drawing):
- http://www.neutrik.com/en/speakon/speakon-chassis-connectors/nl8md-v
- http://www.neutrik.com/en/audio/plugs-and-jacks/1/4-vertical-jacks/nj6fd-v
- http://www.neutrik.com/en/audio/plugs-and-jacks/jacks/combo-a-series/ncj6fa-v

Shall we give them normal audio conventions (and have a corresponding symbol) which I'm used to? That would mean the following pin numbers (in order from the list above):
- 1+, 1-, 2+, 2-, etc.
- T, T+N, R, R+N, etc.
- 1, 2, 3, T, R, S

If we use this pin numbering, where should the symbol be centered? Most of them will not have a pin numbered simply `1`. Do we use the closest pin number to 1? Or center the body geometry?

jkriege2:
For the (switching) audio-jacks, there exist generic symbols, e.g.
![2018-01-19 07_31_44-symbol library editor - connector_specialized d__kicad_kicad-symbols_connector_](https://user-images.githubusercontent.com/11967523/35137661-dd48d862-fcea-11e7-8284-6524f553db49.png)
So if possible I would make the footprints compatible to that symbol. 

For the XLR- there are also generic symbols XLR3-XLR5, so I would make the compatible to thise, e.g.
[...]

PR #1014, opened 2018-10-12
Audio jack fix

evanshultz:
Fixes https://github.com/KiCad/kicad-footprints/issues/1012.
Related to https://github.com/KiCad/kicad-footprints/issues/191#issuecomment-359740503.

The intent here is the fix the pin numbers so CvPcb will at least find the footprints; I didn't check dimensions or do other work.

I couldn't find a datasheet for PJ311 but hopefully someone can; if so I'll be happy to include it here.

**Since this is a bug, can I get a quick review?**

image

Datasheets:
Jack_3.5mm_Ledino_KB3SPRS_Horizontal
https://www.reichelt.de/index.html?ACTION=7&LA=3&OPEN=0&INDEX=0&FILENAME=C160%252FKB3SPRS.pdf
Jack_3.5mm_Neutrik_NMJ6HCD2_Horizontal
NMJ6HCD2, TRS 1/4in (http://www.neutrik.com/en/audio/plugs-and-jacks/m-series/nmj6hcd2
Jack_3.5mm_PJ320D_Horizontal
http://www.qingpu-electronics.com/en/products/WQP-PJ320D-72.html
Jack_3.5mm_PJ320E_Horizontal
http://www.qingpu-electronics.com/en/products/WQP-PJ320E-177.html
Jack_3.5mm_PJ31060-I_Horizontal
http://www.china-bsun.com/Product48/1577.html
Jack_3.5mm_QingPu_WQP-PJ398SM_Vertical_CircularHoles

------------


evanshultz:
As 5.0.2 is going to be released soon, and I assume we can tag and update the libs, can I get a quick review here so this can be included?

evanshultz:
Can I get a review on this? Otherwise the symbols aren't able to match with these footprints.

evanshultz:
Can someone review and merge this?

evanshultz:
I found product pages from manufacturers that appear to match these footprints and updated them.

Can I please get someone to review this and push it to fix this bug? The 5.0.2 release is being planned and I would like this included since our current symbols don't match any of these valid footprints!

Misca1234:
I can check it tomorrow night

evanshultz:
Thanks for the review. There was no documentation for any of these connectors, so perhaps there are other datasheets which apply. If that is the case, then please help me to find them. The documentation I found matched the footprints and so I made them match the documents I found.

NMJ6HCD2:
I think you are right. Here is the drawing:
image

I fixed it:
image

PJ31060-I:
Here is the relevant snippet of the website showing the pinout and schematic:
image

I don't see any issue and I don't know why you're comparing this to the Ledino connector.

PJ320D:
image

I have corrected the footprint's pinout (swapping S and R2 pins):
image

If there is another issue please point it out. I don't see why the Ledino connector is relevant.

PJ320E:
image

Pin 5 is a dummy pin so it should not have a pin number:
image

Misca1234:
> I don't see any issue and I don't know why you're comparing this to the Ledino connector.

The reason is that almost all jackets have schematics the look like this

bild

Like the 
https://cdn-reichelt.de/documents/datenblatt/C160/KB3SPRS.pdf

For the C160/KB3SPRS it is numbered like this
bild

In your make over, you have named pin one to 'T', when other symbols have the same thingy in their drawings, 
bild

Should they not also be renamed into 'T'  ?

evanshultz:
The pin numbering means nothing and can vary from vendor to vendor. It's the part of the jack (on the cable) it connects to that is important, and we're moved to numbers to abstract that into a generic symbol (and of course the pin naming needs to match on the footprints).

Take a look at https://en.wikipedia.org/wiki/Phone_connector_(audio). The last image you posted above is of the Sleeve, so it should be "S".

For jacks which pass through when no cable is inserted, we use "*N" terminology where the asterisk is "S" or "T" as this is common. The "N" stands for "normalling".

So, in the middle screenshot you sent, here is the pinout we would use:
1. S
1. R
1. T
1. RN
1. TN

Does that clear it up?

evanshultz:
Is this ready to merge now?

evanshultz:
@Misca1234 @antoniovazquezblanco 
Are either of you able to take a look at this?

antoniovazquezblanco:
@Misca1234 you had a look at this. Do you feel anything else is missing? I feel quite lazy about starting a review on this right now :sweat_smile:

evanshultz:
Guys. It's almost been a week since Wayne tagged 5.0.2 and if we're to release a new set of libs this should be fixed. It's already been reported by users.

@Misca1234 
Please review the above as your highest priority task in the next day.

@antoniovazquezblanco 
If there's no action would you please take a look? It's not too much to review.

poeschlr:
I really do not like rushing things just to get a release out the door. 
If the symbol side of this has been merged after 5.0.1 then we have the option to re tag 5.0.1 as 5.0.2 (meaning no lib update for this minor release). This would give us time to do a proper review of everything and release it with 5.1.0.

If however 5.0.1 already includes the new symbols then we are forced to include this PR as well. (It would then fix an existing bug within 5.0.1)

poeschlr:
I can not really see a symbol fitting the following footprints:
- Jack_3.5mm_Ledino_KB3SPRS_Horizontal (nearest fitting are AudioJack3_Ground_SwitchTR or AudioJack3_Switch but both have one more pin than the footprint)
- Jack_3.5mm_PJ31060-I_Horizontal (There is no symbol for switched jacks that have a pin R1)
- Jack_3.5mm_QingPu_WQP-PJ398SM_Vertical_CircularHoles (Nearest fitting would be AudioJack2_Switch but that one has a second switch an therefore the pin SN is too much)

As i predicted there are simply too many different options to fill everything with generic symbols. So in this case i would leave the footprints that do not have a valid symbol alone. (And later create new fully specified symbols that can then use the numbering scheme as it is in the datasheet of these parts.)

robkam:
BTW
#1069 also makes a few changes/corrections to Jack_3.5mm_QingPu_WQP-PJ398SM_Vertical_CircularHoles (been waiting for this to get reviewed/merged since August)
#1073 adds a footprint for the same jack with oval holes and a couple more for the stereo version

evanshultz:
So is there something overlooked here or can this be merged? I'll submit a PR for the missing symbol (T and R1 normalled but not R2 or S) in the next 24 hours (probably in around 12 hours).

poeschlr:
I checked the current master again. The symbols look different than in your screenshot. One example is the AudioJack2_Switch symbol:
screenshot from 2018-12-03 18-35-36

My guess is that either something went wrong during the merging process or you forgot to push something.

evanshultz:
@GrantM11235 
Thanks for the info.

I really can't say for sure because there was no documentation associated with this footprint. Other connectors in the library matched with QingPu so my best guess was that someone submitted multiple footprints from that same company, but it was only a guess.

Also, the QingPu does not have a suffix after `PJ320D` (notice the `-A` and `-B` suffixes at https://datasheet.lcsc.com/szlcsc/Korean-Hroparts-Elec-PJ-320D-4A_C95562.pdf and `-X` at https://datasheet.lcsc.com/szlcsc/XKB-PJ-320D-X_C319111.pdf), but again this is not conclusive.

Pins named 1-4 don't tell us anything about the desired pinout either, so that's no help.

The earliest history I can see on GitHub for this footprint is at https://github.com/KiCad/Connectors.pretty/commits/master/PJ320D_3.5mm_Jack.kicad_mod, and it doesn't include the original import nor any vendor info. Maybe someone else can find more history?

It would be best to include the vendor name and the full MPN, but we do not want to change footprint names right now during the 5.0.x cycle. I think doing that when available, and adding these other variants too, would give users the coverage they desire since it appears the same physical connector can have different internal connections (I didn't check all dimensions carefully but they look the same).

(As an aside, this is why abstracted symbols with named pins works well, because different pinout variations can all use the same symbol. At least that's how I see it.)

evanshultz:
@poeschlr 
Alright. I'm working on that symbol now, but I don't see that stops this from getting merged if there are no other issues. Do you find anything else that needs to be changed?

poeschlr:
We are in this mess exactly because somebody did not require simultaneous changes to both symbols and footprints before merging. I am not going to make the same mistake with the same lib again. This one can be merged as soon as we can check it against symbols and then both should be merged at the same time. 
Another option would be to split this PR into the parts that are already ok and the ones that are not. The QuinPu one might also be better left alone for now as we will need additional time to check it properly.

evanshultz:
Here is the original footprint:
image

I would naively assume the pin numbers mean the following:
1 = T
2 = R1
3 = R2
4 = S

This actually matches the XKB and Korean Hroparts if the pin numbering above is correct, but again I don't see any way to know for sure. I'm happy to update this PR to include either of those and the QingPu, both footprints including full manufacturer name and MPN, so we are covered either way. However, does that mean the existing footprint is removed? Or it's kept with the pinout as described above and then renamed later as either XKB or Hroparts?

Required symbol PR is posted at https://github.com/KiCad/kicad-symbols/pull/1207.

Now, here are the symbols in master and the aforementioned PR required for these footprints:
image

And here they are in CvPcb (one is showing both potential options from this PR, but otherwise they are already matched):
image

So if the symbol PR is merged as well, I think there is nothing missing and all footprints are usable. Is there something still broken?

GrantM11235:
QingPu has another connector [WQP-PJ320G](http://www.qingpu-electronics.com/en/products/WQP-PJ320G-256.html) which looks like it uses the exact same footprint, but the pinout matches the XKB and Korean Hroparts connectors.

I emailed QingPu to ask for clarification.

poeschlr:
You might really not have the current version of the master evan. The AudioJack2_Ground_Switch has a gnd pin in there. One without ground is even missing after merging your symbol pull request.

To be more precise: The symbols for J2 and J3 look different than in your screenshot

Here is what i have in the lib instead:
screenshot from 2018-12-03 20-21-48

poeschlr:
I think the major points are now fixed.
Thanks for your effort.

Everything else can be fixed at a later time.

Misca1234:
Sorry for not following up, I was kinda 100% on the valve issues

McColson:
Hello,
I think the NMJ6HCD2 isn't a 3.5mm jack but a 6.35mm.

evanshultz:
@McColson 
Yep, and my PR at https://github.com/KiCad/kicad-footprints/pull/1145 will fix that.

PR #1016, opened 2018-10-15
Add barrel connector w/out silks overhanging

ElectricRCAircraftGuy:
Add barrel connector with reduced silkscreen so that when panelizing boards it does *not* overhang onto adjacent panels. 

This is the footprint I started with:

image

I then copied it and made this new part:

image

Which looks like this with F.CrtYd and F.Fab layers hidden:

image

Notice that now you can put the connectors right up to the edge of a board without any silkscreen hanging off of the board, for panelization purposes. 

Note that I considered calling the new part `BarrelJack_Horizontal_Panelize` or something, but felt simply `BarrelJack_Horizontal_ReducedSilkS` was more descriptive. 

NA

I kept the 3D model tied to the original part, since they are the same: `BarrelJack_Horizontal`

Here is the output of the footprint checker:

```
$ check_kicad_mod.py BarrelJack_Horizontal_ReducedSilkS.kicad_mod -vv
Checking footprint 'BarrelJack_Horizontal_ReducedSilkS':
  Violating F7.3
    Pad 1 should be denoted by rectangular pad
    Only pad 1 should be rectangular
  Violating F9.3
    3D model settings
    3D model name is different from footprint name (found 'BarrelJack_Horizontal', expected 'BarrelJack_Horizontal_ReducedSilkS'), but this might be intentional!
```
All errors are handled appropriately. Here are my responses to the remaining errors:

F7.3: this ignore: pad 1 is properly denoted with rectangular pads while the others are rounded. Same as the part I copied *from*.
F9.3: this is intentional. I am keeping this part's 3d model tied to `BarrelJack_Horizontal`.

As you can see, I've done my homework. Let me know if you have any feedback. This part is ready to accept.

------------

DONE!


ElectricRCAircraftGuy:
A good panelization tool in KiCad that automatically removes overhanging silkscreen from all footprints would solve this problem for all footprints now and forever. However, meanwhile, I still think this is a good addition to the common library to help others. Once this panelization tool is made (https://bugs.launchpad.net/kicad/+bug/1782927), then just go through the whole library with grep and in one single commit it wouldn't be hard to delete all footprints ending in "_reducedSilkS" or whatever. Meanwhile though, I think this is the right approach for the general library, as the KiCad panelization tool could take years to be implemented. 

@poeschlr ?

evanshultz:
It is true that it will probably take a very long time to be implemented in KiCad.

There may be a nice Python solution once 5.1 is released and the Python interface is back in Pcbnew. Surely there will be a number of folks wanting to add this, or update/improve existing panelization tools.

Your description of an automatic silk clipping tool that operates on each board in the panel, and does not require special footprints, is the right approach IMO.

I still feel strongly that this is a poor precedent for the official library. The distance you want the connector to hang off the PCB may be unique to only you and all other users would like different, and varying, dimensions. This is far from a universal solution for all panelized boards, and it only applies to panelized boards. You haven't specified the distance on the PCB or off the PCB in the footprint name. And the phrase "reduced silk", to me, doesn't describe the footprint's use. But even if it was named well, I would not want to merge this. I appreciate the contributions you've made to the library so far, and hope you will make more contributions, but I think this footprint belongs in your personal footprint library.

poeschlr:
A better solution would be to have a feature in kicad that removes the silk outside of the board. I could not find an existing wishlist bug report so i created one: https://bugs.launchpad.net/kicad/+bug/1804992

---

Now about the part at hand. I would in be prepared to allow such parts in the library. But we really need to define them well. Your maximum inset idea might be the best solution to be honest as it will cut down on the number of possibly needed footprints.

I would however remove your dashed line as it does not really make sense if the board edge is at any different place.
(The body dimension is still documented on the fab layer and also kind of on the courtyard layer. So i see no problem here.)

PR #1018, opened 2018-10-16
Add ON Semiconductor VCT28 packaging footprint to Package_DFN_QFN

dogtopus:
Datasheet: http://www.onsemi.com/pub/Collateral/601AE.PDF

Preview:
vct28

Aka. case 601AE. Yet another weird QFN variant. Used by several chips by ON Semiconductor (e.g. LC717Ax0AR series)

Currently no 3D model fits this footprint. The 3D model reference is however present and properly formatted as per KLC F9.3.7.

------------


Misca1234:
@dogtopus 
I guess you made this by hand, there are scripts for making these foot print as well the 3D model, would you be interested in making this (and perhaps the 3D model) by script instead by hand ?
(just in this particular case you have to make the pads a little shorter to guarante the distance between them, the chamfer between pin 1 and 28)

dogtopus:
Good to know there is a script for it. Any pointers?

For the spacing, do I only need to care about the chamfer or is it better to shrink all the pins to width, for example, 0.18mm?

dogtopus:
Another question: If I redo the footprint in scripts, should I send the PR to the FP generator instead? Do I still need to send a copy of the generated file to here?

Misca1234:
Yes, it require two pushes

1. The kicad_mod generated foot print is pushed to this repository as usually
    (a screen shoot and a link to data sheet as well)
2. You make a push to the script repository as well
3. Add a link to the script push in the foot print repository push
4. Add a link to the foot print push in the script repository push

If you use the "normal well known scripts" based on .yml files we will normally not review the .kicad_mod file but only the data in the .yml file => way faster 
The reason is that we normally trust the "normal well known scripts" to be correct (if they are not, then the bug is fixed and new foot prints can be re made in a second or so)

Misca1234:
See this for an example 
https://github.com/KiCad/kicad-footprints/pull/1013

dogtopus:
Porting done, will resubmit it in another PR.

PR #1023, opened 2018-10-16
Add Lodestone VTM inductor mount footprints

evanshultz:
http://www.lodestonepacific.com/CatKpdf/VTM_Series.pdf

I've only added the first one because I'd like feedback before I continue.
1. I made up the `L_Mount` prefix since this is a mount used for inductors. Is there a better convention?
1. I left off a suffix on the PN because the inductor winding can be the leads (-0 suffix) or the pins can be installed in the base (-4 suffix). Either way, the holes are in the location. But since the lead diameter is only known on the mount with leads, should I make it specific to the -4 version?
1. The pin locations are because these are almost always wound from corner to corner so the physical part can be installed either way and it still works. This means pin 1 and 2 need to be at opposite corners for our standard inductor symbol to work. It's unlike the pinout in the datasheet but I think this is best.
1. Is this the right way to represent the pins since they do need to be visible to the PCB designer but they don't touch the PCB and so small parts can fit under them?
1. Similarly, I believe the courtyard should not envelop the pins.

image

If I can get the above answered I'll move this to center on pin 1 and then add the other 3 PNs. Thanks!

------------


poeschlr:
The land pattern looks ok in general.

I am not so sure we should design this to have pin numbers fitting a symbol not really intended for it. It might be best to add a specialized symbol for these parts. (One with all 4 pins)
We might however decide to have only one such symbol so we should really take care that the most probable usecase is possible.

screenshot from 2019-01-16 22-57-50

evanshultz:
Thanks for looking!

I can't recall ever seeing one of these bobbins used where the only winding (or first winding, if there are two) aren't wound in opposite corners as I've done above. And I've seen and used this bobbin a lot. That allows the part to be placed in the PCB in either possible orientation and have the desired electrical connection. Our inductor symbol has pin 1 and pin 2, so this matches. And the `*_Coupled` and `EMI_Filter_LL` will work as the windings use pins 1-2 and 3-4. I'm not quite understanding if you have an alternate proposal, so could you please add some details?

That was question 3. What about the other 4 questions?

evanshultz:
I've pushed all 4 version moved so pin 1 is at the origin. Other feedback? Answers to the above questions?

evanshultz:
image

Not easy to measure the end of the round as you've done above with the normal dimensioning WB we usually use, so I bailed. "38.101" is bogus because the pins are correctly placed at 38.1mm.
image

poeschlr:
Looks good. Thanks

evanshultz:
Thanks. I'll assume the above questions were not answered because everything is fine and this is a reasonable precedent.

PR #1027, opened 2018-10-17
Add LT/ADI BGA/LGA-133 uModule footprints

evanshultz:
https://www.analog.com/media/en/technical-documentation/data-sheets/4637fc.pdf

image
image

------------


evanshultz:
- Should these be `Analog_` instead of `Linear_`? The datasheets haven't changed so I left them as LT.
- Should I indicate on the package where the pad is on top? I can see it being helpful for location in some applications, but the datasheet indicates it's not all that useful as a heatsink so perhaps not.

image
image

Misca1234:
Have you been thinking of making these with the scripts ?
At least Linear_BGA-133_15.0x15.0_Layout12x12_P1.27mm

We should be able to modify the BGA script to make 
Linear_LGA-133_15.0x15.0_Layout12x12_P1.27mm.kicad_mod
as well

evanshultz:
Actually, I didn't even think of it. But I should have.

Since there are still many hand-built footprints that could be included in a script I hope this is OK to merge. You are correct that we should go through the library and migrate hand-built footprints to scripted ones at some point.

Misca1234:
> Should these be Analog_ instead of Linear_? The datasheets haven't changed so I left them as LT.

Go with the data sheet

I can check the manually ones, but hint, hint 
```
BGA-133_15.0x15.0_Layout12x12_P1.27mm:
  description: "Analog Devices (Linear Tech), 133-pin BGA uModule, 15.0x15.0x4.92mm, http://www.ti.com/lit/ds/symlink/msp430fg6626.pdf"
  pkg_width: 15.0
  pkg_height: 15.0
  pitch: 1.27
  pad_diameter: 0.75
  mask_margin: 0.035
  paste_margin: 0.000001
  layout_x: 12
  layout_y: 12
  row_skips: [[], [8, 10], [8], [7], [8, 11, 12], [10], [10], [10], [], [], [], []]
```

I will checkt he squared one tonight, i will also check the python script to see if they can handle squared pads, if not, I will add the possibility

evanshultz:
OK. Thanks.

It cannot do square pads. I made the last commit to the script so I'm a bit familiar with Clay's work. I also just realized that it doesn't check that the length of the row_skips <= layout_y and the length of each element in row_skips is <= layout_x. That would be one way to validate input data and to help catch potential user input errors. It also doesn't use the nice stream feature for input data like Rene's DFN/QFN generator and its spawn. I was thinking to make that change and move this generator with rest of the "Package_" generators once I got a few more things off my plate, but if you want to do any of it that would be great too. Go for it. :)

Misca1234:
No, I saw that too, but I think this script is fairly old, I guess it pre-dates the discussion of what "input method" should be used, please go ahead and modify the script with checks.

Just add a paramter which can be set to "circle", "rect and "roundrect", then we should cover many situations like this

bild

Misca1234:
Ok, then I modify the script to handle circle, rect and roundrect pads.

Do you want to wait for the change to be merged (bga.py) and re generate your foot prints
or do you want to merge these in this push as they are?

Misca1234:
Here is the script push 
https://github.com/pointhi/kicad-footprint-generator/pull/207

Misca1234:
@evanshultz 
The modification have been merged, you can generate the rect/roundrect pads now

evanshultz:
Thanks for updating the script!

Since the script shouldn't generate footprints any different than these I'm still happy to merge them as is.

Misca1234:
Then we go with this

Issue #900, opened 2018-08-31
No radial non-polar cap footprints


evanshultz:
All radial cap footprints are polarized. We need non-polar footprints too.

evanshultz:
THT caps:
Footprints: https://github.com/KiCad/kicad-footprints/pull/1019
Script: https://github.com/pointhi/kicad-footprint-generator/pull/200

evanshultz:
SMD caps:
Footprints: https://github.com/KiCad/kicad-footprints/pull/1030
Script: https://github.com/pointhi/kicad-footprint-generator/pull/206

evanshultz:
All are merged now. Yay!

PR #1030, opened 2018-10-18
Add nonpolar SMD electrolytic caps

evanshultz:
As noted at https://github.com/KiCad/kicad-footprints/issues/900, we didn't have non-polar electrolytic cap footprints.

http://kicad-pcb.org/libraries/klc/F3.3/ specifies something different than the existing footprints names:
"SMD Electrolytic Capacitor": I kept the existing convention even though it's not what's shown in KLC. The example shows the X and Y dimensions at the PCB (most often these are the same) and optionally the height. Our existing footprints use the cylindrical body diameter and height. Since these packages are normally named like our existing footprints and lead pitch doesn't vary much or at all, I think a KLC update is best. Also, KLC doesn't consider these caps can be nonpolar. Perhaps something like `[C/CP]_Elec_[Manufacturer]_[Diameter]x[Height]_[X]x[y]_[Modifiers]_[Options]` with `Manufacturer`, `X`, `Y`, `Modifiers`, and `Options` being optional?

SMD footprint changes (there are script changes so some do not apply to new footprints):
1. Clean up descriptions and keywords
2. Simplify YAML format to eliminate redundant and repetitive data
3. Default to rounded rect pads
4. Support calculating pad sizes per IPC-7351B

I have only shown one cap for each diameter even if there are multiple heights:
image

Script PR: TBD

FYI: I plan to push new polar footprints with pads but no name changes soon.

https://docs.google.com/spreadsheets/d/1SMeC7jJJgnXdY1HykWLle9lsqvj6r4VNmIYQRUr0mLM/edit?usp=sharing shows my spreadsheet calcs used for verification against IPC-7351.

------------


Misca1234:
All **fp_circle (center 0 0) (end** corresponds to the parameter given the name in the file name, so ok

Are there any objections to these foot print?

Misca1234:
I saw that push as well, ok, then we wait with this one until the scripts are in place

Misca1234:
Thx

PR #1032, opened 2018-10-19
SMD electrolytic caps to round rect pads

evanshultz:
Polar complement to https://github.com/KiCad/kicad-footprints/pull/1030.

http://kicad-pcb.org/libraries/klc/F3.3/ specifies something different than the existing footprints names:
"SMD Electrolytic Capacitor": I kept the existing convention even though it's not what's shown in KLC. The example shows the X and Y dimensions at the PCB (most often these are the same) and optionally the height. Our existing footprints use the cylindrical body diameter and height. Since these packages are normally named like our existing footprints and lead pitch doesn't vary much or at all, I think a KLC update is best. Also, KLC doesn't consider these caps can be nonpolar. Perhaps something like `[C/CP]_Elec_[Manufacturer]_[Diameter]x[Height]_[X]x[y]_[Modifiers]_[Options]` with `Manufacturer`, `X`, `Y`, `Modifiers`, and `Options` being optional?

SMD footprint changes (there are script changes so some do not apply to new footprints):
1. Clean up descriptions and keywords
2. Simplify YAML format to eliminate redundant and repetitive data
3. Default to rounded rect pads
4. Support calculating pad sizes per IPC-7351B

I have left these alone as much as possible but they could be consolidated and have pads calculated according to IPC. Perhaps later when it's a good time to break things and streamline the library and conform to KLC?

I have only a few footprints for reference:
image

Script PR: TBD

------------


SchrodingersGat:
Thanks @evanshultz

PR #362, opened 2018-02-10
Add Resistor Stackpole CSS/CSSH current sense resistors

strombom:
CSSH2728
https://www.seielect.com/catalog/SEI-CSS_cssh.PDF

image

Is there a resistor footprint script somewhere that I can add those resistors to?

------------

Thanks for creating a pull request to contribute to the KiCad libraries! To speed up integration of your PR, please check the following items:

- [x] Provide a URL to a datasheet for the footprint(s) you are contributing
- [x] An example screenshot image is very helpful 
- [ ] If there are matching symbol or 3D model PRs, provide link(s) as appropriate
- [x] Check the output of the Travis automated check scripts - fix any errors as required

PR #1034, opened 2018-10-20
R stackpole

Misca1234:
This is an overtake of push
https://github.com/KiCad/kicad-footprints/pull/362

Stackpoles shunt resistor serie
https://www.seielect.com/catalog/SEI-CSS_cssh.PDF

This push only contain the CSS0805 item becaosue evanshultz  and Shackmeister had concerns about the orignal push

@evanshultz @Shackmeister 
This is an initial copy of the original one but you had comments about the Kelvin connections, please elaborate

bild

bild

------------


evanshultz:
The resistor part itself doesn't have Kelvin pins on it, just 2 pins. So that part is fine with me. If the user wants to create Kelvin pads like are shown in the datasheet, using the net tie components near the resistor pads is the best option of which I'm aware.

I didn't review the footprints just chiming in that I'm fine merging with just 2 pads.

Misca1234:
The generic symbol look like this


bild

Misca1234:
Is your idea to create a new shunt symbol for this serie and then add the kelvin path manually during the PCB work?
If not, would you not agree that, if the ambition is to use the generic symbol then the FP should have 4 pins?

evanshultz:
No. I'm saying the datasheet only shows two pins and two pads. So I would accept making the KiCad footprint have only two pads and using the generic R symbol (not the R_Shunt one).

The user can add a Kelvin connection if they want to on the board.

The symbol you show above is for something like https://www.ohmite.com/assets/docs/res_rw1_rw2.pdf, which has 4 pads. That's not the part in this PR, IMO.

Misca1234:
How do you want to do with the file name ?
The original pusher named them like this

R_Stackpole_CSS0603_XX-YYm.kicad_mod

where XX and YY was the resistance span, I guess it was to get dedicated 3D models for each fp, do you want to resuse the names as well.

evanshultz:
I would have suggested by thickness since that also would nicely capture the 3D model differences. It works nicely for all but CSSH2512, but there using thickness is ugly.

Perhaps a `_Pad<x size>x<y size>mm` suffix after `R_Stackpole_CSS[H]<body size>`?

Ugh. I don't see a good way to do it.

Misca1234:
So CSSH2512 would be something

R_Stackpole_CSSH2512_Pad3.05x3.68mm
R_Stackpole_CSSH2512_Pad2.19x3.68mm
R_Stackpole_CSSH2512_Pad2.79x3.68mm

evanshultz:
Sadly that's the best idea I have. Do you have another suggestion?

Misca1234:
Naaa, for me either way is ok, they will be scripted anyway,
Let me script away

Misca1234:
It wont work
For example 
CSS2010 is in two dimensions, where the physical height differ but the pads are of different size, so perhaps the original idea with the resistance range as suffix is better ?

evanshultz:
Not quite. CSS2010 is fine. The break between resistance for pad size and height match, so the height can be used to divide them. Same with CSS2512.

CSSH2512 is the problem because it's not clean breaks between pad size and height.

The bottom of page 3 and top of page 4 have pad sizes while page 2 and the top of page 3 have dimensions. Compare the breaks in the pad size table with changes in height (H column) for dimensions. You'll see the divisions match as I've stated except for CSSH2512. That's why it's messy.

Misca1234:
Where was we on this one?

Should we only have two pads in the foot print despite the symbol have 4 pins ?

Should we add the height as well to the foot print name so CSS0805 will be

R_Stackpole_CSS0805_Pad1.80x2.18_H0.30mm

evanshultz:
I think only two pads as I said above, since this part has two pads. The user can use net ties if they want to Kelvin connect to it.

Thinking about this again, if the user is placing a component they will be thinking about the resistance value, right? So maybe that is the best way to go since it will be most helpful to the user, I think. How about you?

Misca1234:
So back to 

R_Stackpole_CSS0603_XX-YYm.kicad_mod

where XX and YY is the resistance span,

?

evanshultz:
Pretty much. Looking at the current SMD resistor footprints, I think just `R_Shunt_Stackpole_CSS0603` would be fine for the versions which only have one recommended footprint. For other ones, like the 2010 size, we could use `R_Shunt_Stackpole_CSS2010_R0.001-0.003` and `R_Shunt_Stackpole_CSS2010_R0.0031-0.1`. I suppose the `R` prefix is understandable but perhaps Rene wants to provide some input on this. I'm also not sure why the `Shunt` part is needed since I don't feel it adds anything useful.

poeschlr:
I took the liberty of assigning @evanshultz and marking this as pending changes. Hope this is ok for everyone.

PR #1035, opened 2018-10-20
ECS-2520MV oscillator footprint

Shackmeister:
other pin 1 style ideas are very welcome, I think these parts quite hard.
Datasheet:
https://www.ecsxtal.com/store/pdf/ECS-2520MV.pdf

ecs_crystal

------------


cpresser:
There is already a similar crystal in the libs: "Crystal_SMD_2520-4Pin_2.5x2.0mm"
It is basically the same device, but with slightly different pad geometry. 
You could copy its pin1 style, or just use that footprint :)

poeschlr:
The courtyard clearance on the top side is smaller than all others. Could you fix that?
screenshot from 2019-01-01 14-30-14

Shackmeister:
@poeschlr fixed

poeschlr:
Thanks

PR #1036, opened 2018-10-20
Add RJ45_BEL_SS74301-00x

cpresser:
Add RJ45_BEL_SS74301-00x
Shielded vertial RJ45 Connector with two LEDs (Yellow/Green)

- https://belfuse.com/resources/drawings/stewartconnector/dr-stw-ss-74301-001-ss-74301-002-ss-74301-005.pdf
- https://www.digikey.de/product-detail/de/stewart-connector/SS-74301-002/380-1207-5-ND/2257495
- no 3D model
- ~~The LED drill positions are 1mil off compared to the data-sheet so they can be on a 50mil grid. That is well within the specified tolerance (6mil) for the drill position.~~

------------


cpresser:
The footprint-name is not perfect. I did put an 'x' as the last char, but only the versions -001, -002 -003 -005 -007 exist.

Another thing that could be changed is the pin numbers of the LEDs. I did match them to the symbol 'RJ45_LED_Shielded'. There the LED on pins 12-11 is close to pin1.
Thats not reflected in this footprint.

poeschlr:
Please add a pin 1 marker on the fab layer as well.

The courtyard clearance must be 0.25mm all around (rounded to the nearest 0.01mm grid)

screenshot from 2019-01-18 20-36-40

I have forgotten to mention that if a datasheet specifies both imperial and metric dimensions then we prefer the use of the metric ones. (In this case the metric ones are rounded to two significant places.) It is ok for this footprint. Just something for your next contribution. (The original footprint you made was not exact for either measurement system. Which was the reason for my previous comment.)

cpresser:
Thanks for you patience, I am a slow learner :)
How did you make that preview? That looks really helpful.

cpresser:
No, the last 3 commits have all been made on the same machine, no kicad update in the meantime.
The kicad version used is 5.0.2+dfsg1-1 
* 288641675846d06773dd274fa7a289c9a13eab03 
* 3178d7198ed0613689ee6d51c1310ebff92cab6d
* cfcd79aa3ab22dc6694c1e936a5807f9f4cccfad

poeschlr:
Ok it is not kicad that screws up here but github. (A better difftool shows it much clearer what really changed.)

The only thing that seems strange is that 4 courtyard lines have been moved to after the fab lines.
Is it possible that you deleted some lines and created new ones instead? (for example by duplicating from the other side?) If this is the case then i would assume that kicad orders the lines in the file by the order the feature has been added in the footprint editor.

Again this is not a problem for your pull request just personal interest.

cpresser:
I might have used the duplicate feature. I did draw some additional lines on the edge of pads to extract their coordinates. Of course, I deleted them one I was done with it.
Same for pins (shield); duplicated and changed their diameter so I can see the 0.25mm distance from the copper.
So in conclusion, this is most likely just my weird workflow.

poeschlr:
The courtyard is not correctly closed on the top right corner.

And add the orientation suffix (Vertical) to the footprint name

cpresser:
Sorry about the courtyard; I am working on 
I am working on https://github.com/KiCad/kicad-library-utils/issues/277 right now and did not properly remove my test-stuff.

poeschlr:
Thanks

PR #1039, opened 2018-10-22
Solder pad ring is too small for hand soldering

ferdymercury:
Increase solder pad size from 1.26 to 1.75, necessary for hand soldering.

https://www.bourns.com/docs/Product-Datasheets/3339.pdf

Please also add the 3D model:
https://www.bourns.com/engineering/3339/3339P.stp
```
  (model ${KIPRJMOD}/Potentiometer/3339P.stp
    (offset (xyz 0 2.5 6.5))
    (scale (xyz 1 1 1))
    (rotate (xyz 180 0 90))
  )


```
------------


ferdymercury:
Thanks for the swift reply. What do you mean by wrong parameters?

I just downloaded the step file from the manufacturer and adjusted the location so that it fits in the footprint. It looks correct to me in this screenshot:

image

ferdymercury:
I committed the new file with the HandSoldering suffix, as suggested.

evanshultz:
Confirmed that the only change between the existing footprint and this one is the pad size increased to 1.75mm diameter. Thanks!

PR #1040, opened 2018-10-22
First footprint of a PCB ferrite bead

ferdymercury:
https://assets.lairdtech.com/home/brandworld/files/28C0236-0JW-10.pdf
https://www.bourns.com/products/magnetic-products/j.w.-miller-through-hole-ferrite-beads-emi-filters

I designed the 3D model also in Inventor (.ipt) and exported to .stp

image

------------


ferdymercury:
image

3D model attached.
[Ferrite3Dmodel.zip](https://github.com/KiCad/kicad-footprints/files/2501866/Ferrite3Dmodel.zip)

ferdymercury:
I uploaded the 3D models and did the corresponding PR.

easyw:
@ferdymercury 
> I designed the 3D model also in Inventor (.ipt) and exported to .stp

have you designed the model from scratch, or you just rearranged the models coming from bourns repository?
https://www.bourns.com/resources/design-tools/engineering-files-list?folder=fb2000_series
I see that the ferrite body is exactly the same of the bourns ferrite body...
If your design is original, this model can be merged, but in this case, please include also your inventor 3D model to the 'source' repository at this PR
https://github.com/KiCad/kicad-packages3D-source/pull/174

ferdymercury:
I designed the wires myself, but took the ferrite core file from the manufacturer. Sorry. I will redesign that part now.

ferdymercury:
Part is now fully redesigned from scratch.

ferdymercury:
Updated footprint screenshot:
image

evanshultz:
Since inductors are not polarized they do not need a square pin 1.

ferdymercury:
Changed

ferdymercury:
image

ferdymercury:
@easyw is there any further work needed from my side?

poeschlr:
The courtyard outline is not closed. This is a problem for the new courtyard check of the DRC.

It might be better to approximate the courtyard using straight lines in this case (getting arcs like this to form a closed outline might be near to impossible with the restrictions of kicad.)

The fab outline is not quite right.
screenshot from 2019-01-01 11-43-47

easyw:
Hi @poeschlr 
would you mind to test this one? (I used FC to create a closed path)
[LairdTech_28C0236-0JW-10.zip](https://github.com/KiCad/kicad-footprints/files/2719339/LairdTech_28C0236-0JW-10.zip)

poeschlr:
Not quite. I fear kicad might do a bit of rounding here. (Both yours and the one by @ferdymercury  might work as long as the devs leave the endpoint tolerance at the current value of 0.01. I would however really prefer it if our lib does not depend on such an undocumented value.)

easyw:
Later today I could try to discretize the arcs to poly-lines and see if that may lead to something that kicad may like...

easyw:
ok, here the try using a poly-line coming from the discretization of arcs...
[LairdTech_28C0236-0JW-10-discretized.zip](https://github.com/KiCad/kicad-footprints/files/2719442/LairdTech_28C0236-0JW-10-discretized.zip)

poeschlr:
Man that are many lines. We might want to talk to the devs what the costs of these are for the DRC algorithm.

easyw:
but does it work for original DRC issue?

poeschlr:
Ok i thought about it a bit. Right now we really don't have the manpower to invest too much time into this. (Nearly 400 open pull requests)
I would really prefer the courtyard of this part to be either a simple rectangle or an octagon. This is easy to check for us maintainers, should not be too heavy on DRC. And maybe most importantly it is relatively easy to create using kicad internal tools. (when compared to the use of arcs)

I will however ask the devs to take a look at this pull request and give input.

poeschlr:
@easyw for some reason your last answer was not there when i made mine. Yes it would solve the issue of relying on the tolerance for detecting closed outlines. It is however even harder to check correct correct courtyard clearance.

thomask77:
> Since inductors are not polarized they do not need a square pin 1.
That's not entirely true - if you're designing a switching power supply, the switching node should be the better shielded terminal, i.e. the inner wire in a wound wire inductor. So a pin 1 marking may be very important.

ferdymercury:
I simplified the footprint to avoid all these problems.

image

poeschlr:
I like your solution in general. (The silk layer looks really nice.) But i would really like the outline of the lower part still to be included on the fab layer. (The white or light gray part in your 3d model screenshot)

ferdymercury:
What about now?

image

poeschlr:
The fab layer and courtyard clearances are still slightly wrong (correct values are added to the dimensioned drawing)

screenshot from 2019-01-18 20-06-05

ferdymercury:
changed, thanks.

poeschlr:
Looks good, thanks.
screenshot from 2019-01-18 23-47-46

PR #1041, opened 2018-10-22
Mounting holes for a PCB to DIN rail adapter

ferdymercury:
New footprint for mounting a PCB to a DIN rail with M3 screws. I think it might be very useful for the community.

------------


ferdymercury:
image

ferdymercury:
Linked to https://github.com/KiCad/kicad-symbols/pull/1095

ferdymercury:
I uploaded the models and did the PRs.

Misca1234:
Do you have a data sheet link to this one ?

pointhi:
You are missing the mask layer (look at the 3d-image)

ferdymercury:
Datasheet is https://www.phoenixcontact.com/online/portal/us?uri=pxc-oc-itemdetail:pid=1201578&library=usen&tab=1

ferdymercury:
@pointhi you mean the silk layer? Is it mandatory? I thought it is not needed, as the component attaches from the bottom.

pointhi:
@ferdymercury, no I mean the holes. Solder mask is not removed from the holes

ferdymercury:
Got it, thanks. Updated now.

ferdymercury:
Screenshot of updated footprint
image

myfreescalewebpage:
@ferdymercury I think this is ready to be merged. Thanks for the job !

Cheers,
Joel

evanshultz:
A few questions about this footprint:
- We have built all TO-220 footprints that are "laying down" on the top side and they can be flipped if needed. That would seem to work here as well instead of building it so the part, by default, is on the bottom. I don't know any rule but perhaps we should discuss if "part on Front" should be the standard for the library.
- There is no outline on any *.Fab layer.
- Part body is 43x10mm, while the courtyard is on top of the silk layers at 42.5x10mm. Fab should indicate part body bounds with the silk being on or outside and the courtyard being outside.
- The drill holes are 3.2mm diameter but the Phoenix drawing shows 2.75mm. Was this enlargement intentional?
- I assume the circles around the holes are to indicate the screw head size an a clearance from the screw head. Is there an existing way to note this in our footprints? I can recall mounting screws and other similar examples but I didn't check out those footprints to see if we've been consistent or smart about it.
- The description is only a link.
- I would suggest `DINRailAdapter_PhoenixContact_1201578` as the naming unless the hole count is important. It seems the MPN fully specifies the number of holes so unless the drill size was intentionally changed then perhaps `_M3` or `_M3Holes` could be a suffix. That puts the important part up front.

I'm not familiar with this part so forgive me if there's something detailed that I don't understand.

myfreescalewebpage:
Hello @evanshultz 

> We have built all TO-220 footprints that are "laying down" on the top side and they can be flipped if needed. That would seem to work here as well instead of building it so the part, by default, is on the bottom. I don't know any rule but perhaps we should discuss if "part on Front" should be the standard for the library.

I don't know any rule, but "front" of this part is probably the head of the screws, bottom is the piece of plastic. That is how the submitter indicating it was used and after looking at some application I agree.

> There is no outline on any *.Fab layer.

There is. Or maybe you would like an additional rectangle ? But probably on B.Fab layer in this case ?

> Part body is 43x10mm

Interesting, spec do not indicates the same value !

> The drill holes are 3.2mm diameter but the Phoenix drawing shows 2.75mm. Was this enlargement intentional?

Yes. The drilling in the plastic body is smaller than the screw size (M3) to get it fixed. And the drilling in the PCB is larger than the screw size (still M3) to get it mounted easily.

> I assume the circles around the holes are to indicate the screw head size an a clearance from the screw head. Is there an existing way to note this in our footprints? I can recall mounting screws and other similar examples but I didn't check out those footprints to see if we've been consistent or smart about it.

That's similar to mounting holes I think.

> The description is only a link.

Not understood. In my Kicad window:

capture

> I would suggest DINRailAdapter_PhoenixContact_1201578 as the naming unless the hole count is important. It seems the MPN fully specifies the number of holes so unless the drill size was intentionally changed then perhaps _M3 or _M3Holes could be a suffix. That puts the important part up front.

Probably yes. We don't know how this kind of symbol will evoluate and which one will be submitted at the moment, that's very new as discussed also with you at https://github.com/KiCad/kicad-footprints/issues/1245.

Cheers,
Joel

myfreescalewebpage:
@ferdymercury for your information, we have decided at https://github.com/KiCad/kicad-footprints/issues/1245 to separate the footprint in a new library. PRs are https://github.com/KiCad/kicad-footprints/pull/1252 and https://github.com/KiCad/kicad-packages3D/pull/498.

Joel

PR #1042, opened 2018-10-22
New 2 pad testpoint for Multimeter

ferdymercury:
Easier access with a Multimeter compared to the other smaller 2-pad testpoint accessible


ferdymercury:
image

poeschlr:
Sorry for the late response.

Could you move the silk reference to the top, fab value bottom slightly outside the part outline?
Also move the fab reference to center of the footprint.

ferdymercury:
image

Changed, thanks.

PR #1046, opened 2018-10-24
Fix Molex PicoBlade 5304x part numbers

dimhoff:
The Molex PicoBlade 53047 and 53048 footprints used incorrect part
numbers in there names. This fixes the filename, module name,
description and 3D model name.

This also fixes the incorrect datasheet URL for the 53048 footprints.

------------


evanshultz:
@dimhoff 
Thanks for using KiCad and having the initiative to report a potential issue. I see this is your first commit and we rely on a strong community to grow and fix the official library.

The datasheet, https://www.molex.com/pdm_docs/sd/530470610_sd.pdf, confirms this. Yes, the PNs are wrong.

These footprints are all generated by a script. Did you fix the script or modify these by hand? The scripts are at https://github.com/pointhi/kicad-footprint-generator/blob/master/scripts/Connector/Connector_Molex/conn_molex_picoblade_tht_top.py (53047) and https://github.com/pointhi/kicad-footprint-generator/blob/master/scripts/Connector/Connector_Molex/conn_molex_picoblade_tht_side.py (53048).

evanshultz:
The 53048 script also has the wrong datasheet. A suitable one would be https://www.molex.com/pdm_docs/sd/530481110_sd.pdf.

Shackmeister:
@dimhoff 
If you would be up for it, the 3D models will also need some renaming afterwards :)
The script can be found here:
https://github.com/easyw/kicad-3d-models-in-freecad/tree/master/cadquery/FCAD_script_generator/molex

dimhoff:
@Shackmeister 
I don't think there are models for these connectors. Only for the SMD variants.

I have made model for these connectors in FreeCAD, and wanted to submit them to the KiCad library. Would it be a problem that I used FreeCAD directly instead of a script like this? (see: https://github.com/dimhoff/di_kicad_library/blob/master/packages3d/Connector_Molex.3dshapes/Molex_PicoBlade_53047.FCStd)

Shackmeister:
@dimhoff ah sorry, didnt notice it was the THT variants :)

Handmade models are very welcome, but scripted are preffered, since they are much easier to maintain.
If you have the time and courage, I would recommend you to try scripting them :)
It's not to hard and (mostly) fun

evanshultz:
@dimhoff 
The scripts generate the footprints, which then go into a separate PR in this repo.

In this case, since the changes you made are identical to the script changes it should be OK to merge this as is. I prefer to merge scripted footprints, though, as that validates the script output.

Please confirm you ran the scripts and that nothing else is amiss. If you want to push the scripted footprints for the ones in this PR it's fine too.

dimhoff:
@evanshultz 
I validated the scripts vs. my changes through diff. Output is identical apart from timestamp and newline at end of file.

So I think keeping this PR will be fine.

evanshultz:
Great! Thank you for catching this mistake and spending the time to correct it!

PR #1048, opened 2018-10-26
Flip WLCSP-6 footprint

florolf:
The current WLCSP-6 package numbers bumps from A to C top to bottom and
1 to 2 right to left. However, almost all manufacturers[1][2][3] use a
left-to-right numbering scheme. Mirror the footprint to handle this
common case, but also keep the original as a "mirrored" version to
handle the exceptions (such as the Panasonic AN26032A[4]).

[1] https://www.analog.com/media/en/package-pcb-resources/package/pkg_pdf/wlcspcb/CB_6_2.pdf
[2] http://www.onsemi.com/pub/Collateral/FAN5361-D.PDF
[3] http://ww1.microchip.com/downloads/en/DeviceDoc/MIC9416x.pdf
[4] https://media.digikey.com/pdf/Data%20Sheets/Panasonic%20Semiconductors%20ICs%20PDFs/AN26032A.pdf

------------


florolf:
FWIW, I realized that I didn't look at the Panasonic datasheet properly, they use a different scheme altogether. I can't find any chip which actually uses the "mirrored" variant, so I dropped it.

poeschlr:
The footprint is currently used by the symbol for IP3319CX6

The current footprint has the pin numbering in accordance with that part. (See datasheet https://assets.nexperia.com/documents/data-sheet/IP3319CX6.pdf)

PR #1050, opened 2018-10-26
Adding Runtron RM065/063 device

christianlupus:
Copied potentiometer Triwood RM065 from old archive and made it compatible with current requirements

------------

- [x] Provide a URL to a datasheet for the footprint(s) you are contributing:
  - http://www.dipmicro.com/?datasheet=RM065.pdf
  - https://github.com/KiCad/Potentiometers.pretty/blob/master/Potentiometer_Triwood_RM-065.kicad_mod
- [x] If there are matching symbol or 3D model pull requests, provide link(s) as appropriate: no 3D model provided


Misca1234:
Hi

Thanks for rescue an component
Do you have a data sheet link to this one, the only I found was to other companies than triwood

But as starter I suggest the following
Rotate the fp so pin 1 get "up in the left corner"
bild

move "out" silk outside the pad area, make sure to use the appropriated distance
bild


When you find the data sheet link, add it to the text "Document link:"
"Potentiometer, horizontal, Trimmer, RM-065 **http ......**"

christianlupus:
Hi,

> Thanks for rescue an component
> Do you have a data sheet link to this one, the only I found was to other companies than triwood

For me, this is true as well. To be honest, I ordered a few trimmers in China and needed a matching footprint. So I went rescuing the triwood as it seemed to match.

- [x] Rotated fp so that pin 1 os upper left corner
- [ ] Silk outside of pad area
- [x] Removed clutter in the center
- [ ] Datasheet
- [ ] Make Travis happy ;-)

Regarding the silk topic: What is a appropriate distance for the clearance? 0.1mm?

Many datasheets of similar components (https://www.mouser.de/datasheet/2/54/3318fk-1159133.pdf, http://www.runtron.com/down/PDF%20Datasheet/Carbon%20Film%20Potentiometer/RM065%20RM063.pdf) are using metric values. Should I rename the FP to something with a datasheet and then stick to this?
As these seem quite the same, how is this reflected best in the library (like "this FP is both valid for Runtron and Bourns")?

evanshultz:
First, thanks for contributing to KiCad!

The potentiometers are, unfortunately, a special case. They are created from the script at https://github.com/pointhi/kicad-footprint-generator/tree/master/scripts/Potentiometers and https://github.com/pointhi/kicad-footprint-generator/blob/master/scripts/tools/footprint_scripts_potentiometers.py.

The script was written with the pots at the wrong rotation. It should have pin 1 at top left as @Misca1234 mentioned above, but since our script is silly it would be OK to leave the footprint as it was. (Writing a new pot footprint script with more power and flexibility is a longer-term project for me.) But rotating it is fine as well. Just filling in the history.

http://kicad-pcb.org/libraries/klc/ are the official library rules and http://kicad-pcb.org/libraries/klc/F5.1/, clause 2, specifically says 0.2mm. However, this hasn't been updated and we've been allowing IPC 7351C draft recommendation of the line width (so just 0.12mm). Either one is an acceptable silk-pad clearance.

Mouser, DigiKey, and https://www.bourns.com/docs/technical-documents/obsolete-parts/3318FK_obsolete.pdf?sfvrsn=1574000b_1 show that Bourns 3318 is obsolete. So probably that's not the best choice. I would use Runtron's datasheet in the documentation and to update this footprint. Just stick with that one vendor; it appears there are small differences between those parts so we will leave it up to the user to figure out of the footprint matches the particular part they want to use.

I would like to suggest using the script instead of hand-modifications, but the script doesn't support this exact footprint shape (one side is square and one side is rounded) so it match the Runtron datasheet you'll need to modify by hand.

christianlupus:
OK, I changed the FP such that the pins are in metrical system completely and changed the name to be Runtron. I added by the way the vertical version afterwards.

I would ask someone to look over it and to give me some comments in case anything is missing still. Thank you.

christianlupus:
Shame on me for overseeing that the reordering messed up with the pin 1 in upper left corner rule...

evanshultz:
No problem! Thanks for making all of these changes!

- You have the PNs flipped: RM063 is horizontal and RM065 is vertical.
- Please center the silk ref des.
- The fab drawing should match the outside dimensions of the part, but there are differences:
image
- There are also some dimensional issues (I note these drawings are poorly dimensioned so it may be hard to get correct dimensions or you may refer to the Bourns or another datasheet):
image
- The radius of the circles has excessive precision.
- The courtyard should be 0.25mm outside the extents of the pads and the body. If the fab lines are correct, there is inadequate clearance here:
image
- It's not required, but I prefer to NOT have rotated pads if you don't mind setting their rotation to 0 degrees (it serves no purpose, especially for square and round pads).
- The datasheet shows 1mm pads, but yours are 1.2mm. Also, the annular ring can be 

I know some of these pics are zoomed, but I'm trying to capture the item in detail and I think if you look at the footprints it will be clear. Please ask if you would like further clarification.

christianlupus:
OK, I just wanted to ask what the next steps are. I think I managed to address all issues but in case anything is still missing, I can still modify things.

evanshultz:
Thanks for checking in. Someone needs to find time to review your changes. I'm rather busy today but I will plan to get to this tomorrow.

evanshultz:
Thanks for making these updates! Sorry for the delay. There are a few things I noted below and then we can get these footprints merged. I'm using http://www.runtron.com/down/PDF%20Datasheet/Carbon%20Film%20Potentiometer/RM065%20RM063.pdf.

Both:
- Silk should go to 0.12-0.2mm of pads. Right now the silk is not contiguous and far from pads.
- The pad diameter can be 1.55mm instead of 1.8mm.

Vertical:
- The 3.8mm dimension on the datasheet is 4mm on your part. The 3.8mm is going to the end of the leads, not the flat back part of the potentiometer body.
- The other side of the body should be curved, not beveled.
- The diameter of the knob is slightly smaller than the body.

Horizontal:
- It looks like the back of the body goes to about the middle of the leads, or thinner than it is now.
- Is there a reason the rearmost pin (pin 2) has a bevel on the fab layer? Where do you see this?

evanshultz:
Thanks! These are looking really nice now!

A sketch like this is probably the best we can do when the datasheet is missing so many dimensions. In some cases we may find another vendor which has a better datasheet. This same method could be done for the width of the rear lead on the horizontal footprint and the diameter of the knob, but this is OK.

- Courtyard should be 0.25mm from the body or pads. On the bottom of the horizontal one, for example, that is 3.525mm. But it should also be on a 0.01mm grid, so either 3.52mm or 3.53mm would be acceptable. However, it is now 3.6mm. Can you please review all courtyard offsets?
- A fab line should be all the way across the bottom of the vertical one. And then draw horizontal lines on silk on both sides of pin 2.
- Silk should be offset from, and outside of, the fab lines. This isn't specified in our library guidelines but 0.11mm is typical (this results in the edges of the fab and silk lines just touching).
- You may add short silk lines sticking down towards pin 2 and more silk around pin 3 on the horizontal one, but this is optional.

Just to capture where it is now, here are current screenshots:
image
image

christianlupus:
> A sketch like this is probably the best we can do when the datasheet is missing so many dimensions. In some cases we may find another vendor which has a better datasheet. This same method could be done for the width of the rear lead on the horizontal footprint and the diameter of the knob, but this is OK.

I tried to do similar things on-the-fly without creating PNG files out of it. The diameter of the horizontal knob should be correct. The rear pin I reshaped also a bit but in my opinion this is only a wild guess. Here the projection does not allow to measure exactly.

Your other tasks come in here:

- [x] Courtyard should be 0.25mm from the body or pads. On the bottom of the horizontal one, for example, that is 3.525mm. But it should also be on a 0.01mm grid, so either 3.52mm or 3.53mm would be acceptable. However, it is now 3.6mm. Can you please review all courtyard offsets?
- [x] A fab line should be all the way across the bottom of the vertical one. And then draw horizontal lines on silk on both sides of pin 2.
  I hope this was done as suggested/intended. If not, please tell me.
- [x] Silk should be offset from, and outside of, the fab lines. This isn't specified in our library guidelines but 0.11mm is typical (this results in the edges of the fab and silk lines just touching).
  This will put the silk lines on the 0.01mm instead of the 0.1mm grid. I did this but was curious if this would break any other requirement....
- [ ] You may add short silk lines sticking down towards pin 2 and more silk around pin 3 on the horizontal one, but this is optional.
   See the images. Is this what you thought of?

horizontal
vertical

evanshultz:
Thanks for sticking with this. Very close now!

To answer your question about the horizontal pot footprint, yes. That is the silk I was talking about. Short horizontal lines by pin 2 could also fit.

Regarding grids, only the courtyard is on a grid. Fab and silk elements should be exactly placed and there is no grid requirement.

Just one item left: The courtyards are still not quite correct. Here are the dimensions:
Horizontal:
X: -1.03mm/6.03mm (this applies to both footprints
Y: -2.35mm/3.53mm
Vertical:
Y: -1.55mm/6.03mm

Once the courtyard is adjusted I think this is ready to merge. :)

christianlupus:
Dear @evanshultz I am not sure, if your values of Y are correct. I get different values when taking the most prominent features on the Fab/Cu layers into account. I was mislead: the silk layer are *not* to be taken into account.

evanshultz:
Certainly my numbers could be wrong. Sorry if I did make a math mistake.

Rather than recalculating, let's try another method in case seeing is best. Here is a dimensioned drawing of the latest commit. I'm showing just the fab, top copper, and courtyard layers. All clearances should be 0.25mm but they're all a bit too big:
image

Are you able to see the courtyard clearances being a bit too big?

christianlupus:
@evanshultz You are measuring from outer side fab to inner side countyard or differently? Because if I enable the clearance of 0.25mm on the pin 2 of the horizontal pot, it mostly touches the countyard line. So it is pure empty space of 0.25mm minimum as I tried to reach it (my understanding of clearance).

evanshultz:
I am measuring all nominal dimensions of lines. Copper will be the edges.

Let me take a few screenshots in KiCad so you can point out if I'm not doing something right.

We'll start with the vertical one and look at the left side courtyard.
image

Let's find the leftmost side of pin 1. It is 1.55mm / 2 = 0.775mm.
image

So the courtyard line should be at 0.775mm + 0.25mm = 1.025mm.

Since the courtyard must be on a 0.01mm grid, we can choose either 1.02mm or 1.03mm for the left-side vertical courtyard position. I choose 1.03mm.

Going back to the footprint, the line is actually placed at 1.05mm:
image

This 0.02mm difference is shown in the dimensioned drawing above, where I measured 0.275mm. Moving the courtyard line to 1.03mm could then result in a distance of 0.255mm, which is as close as we can get with the 0.01mm grid requirement for courtyard lines (if I measured either 0.245mm or 0.255mm I'd accept it).

Does that explain things? I may have made a mistake above so please let me know if I did. If not, hopefully that explains how to measure the courtyard offset for the KiCad library.

christianlupus:
OK, then our understanding of clearance is sort of different. I put a (temporary) clearance of 0.25mm on the 1st pad.
explain1

Now I move the left vertical countyard line such that it lies both on the 0.01mm grid as well as its shape does not intersect with the clearance defined in the first step.
explain2

Compared with your explanations: The -1.025mm equals the right most vertical line (ideally thin) that can be drawn. Thus I need to shift the countyard left by half of the line width 0.5mm/2 = 0.025mm. Thus I used -1.025mm-0.025mm = -1.05mm.

Here I am unsure, what the KLC wishes:
- Countyard's middle line is offset by 0.25mm
- Countyard's inner line is offset by 0.25mm

I am ok with both versions and I will for sure adopt to whatever version is "correct". I am going to change the values to be offset by the middle line but not yet pushing. Then you can confirm that middle line is correct and we can proceed.

poeschlr:
The center line of the courtyard outline is what counts (For DRC). It only has a thickness such that users can visually see it.
This means the clearance is between the center of the outline features and the edge of the pad.

christianlupus:
OK, in this case @evanshultz is perfectly right and all I need to do is correct the distances. This was my fault. Sorry for this long discussion. I already pushed a commit (6ed18da) with the updated clearances of the courtyard.

evanshultz:
Thanks!

No problem. Good that we got it figured out and perhaps KLC could be more clear since we offset silk to match edge-to-edge but the courtyard line is on the center.

Just the uppermost courtyard needs brought in on both footprints:
image

christianlupus:
I assume for the fab lines the same is true: Clearance is measured from the middle of the line, right? I will adopt.

evanshultz:
This looks great! Thanks for sticking it through this journey and I appreciate you contributing! Hope to see more from you in the future! :)

PR #1058, opened 2018-10-31
Converter_DCDC XP_POWER IA, IH, ITX and ITQ series

ferdymercury:
Symbol at https://github.com/KiCad/kicad-symbols/pull/1086
https://www.xppower.com/pdfs/SF_IA.pdf
https://www.xppower.com/pdfs/SF_IH.pdf
https://www.xppower.com/pdfs/SF_ITX.pdf
https://www.xppower.com/pdfs/SF_ITQ.pdf

image

------------


ferdymercury:
Why that? See here:
image

Misca1234:
not sure, I was not thinking, however, the 3D model I did, 
https://github.com/easyw/kicad-3d-models-in-freecad/pull/238

was correct ;)

ferdymercury:
all 3d models are now scripted at https://github.com/easyw/kicad-3d-models-in-freecad/pull/239

poeschlr:
I did a quick check of the footprints.

The courtyard outlines need rework. Make sure you have 0.25mm clearance to the part bodies (center of fab line to center of courtyard line)

The fab outline should be the nominal dimension of the body. Some dimensions are given as max (well most case dimensions), but the tolerances are given as well. Meaning you can calculate the nominal size as maximum minus the tolerance.

ferdymercury:
Do you think we could script these several components? But I would need help with the folder structure and a python example @pointhi can you create the folder structure for DCDC converters in your script repo?  Thanks!

ferdymercury:
Maybe we also need to change here the max dimensions to nominal dimensions https://github.com/easyw/kicad-3d-models-in-freecad/pull/239

poeschlr:
Regarding folder structure: You can simply add new folders within a pull request. No need for somebody else to get involved.
regarding examples: take a look at any tht connector footprint. for example jst-xh.
This should give you everything you need. (Main difference is that the connectors have more detailed silk artwork and equally spaced pads. You might need to create some form of data structure that encodes which pins exist in a package and what their numbers are.)

ferdymercury:
Just did the scripted version. Please review before add the generated fp files here. Thanks.

myfreescalewebpage:
Hi @ferdymercury @poeschlr 

"SIP" is not required in the name of the footprint, this is what "S" just before stands for. Also, what means "THT" ?

Cheers,
Joel

ferdymercury:
I removed SIP from the filename now https://github.com/pointhi/kicad-footprint-generator/pull/252/commits/31547421a1a036f76ce0004d0188dc4a31855ab8

poeschlr:
The typical way we do things with scripted footprints is that we first merge the footprint and after that @pointhi  gets the go ahead to merge the script from us library maintainers. (He only looks at the script closely if you add something to the core library. Otherwise it will be judged by the produced footprints.)

ferdymercury:
Oh, I see. I have updated them here now.

poeschlr:
All your pin 1 marks on the fab layer are on the top right corner. Even for parts that have two pins in the top row. (There the mark is closest to a different pin)

You did not answer my question about how you determined the hole size.

The IAxxxxS and IA48xxS footprints have their outline positioned wrong compared to the pins.

The courtyard clearance should be 0.25mm not 0.5 (0.5 is only for connectors)

xp svg

ferdymercury:
> 
> 
> I will take a look. The hole sizes, I just used something that is comfortable for hand soldering.

Yes, I did answer, was some time ago.

Let me know if I should make it smaller.

poeschlr:
Oh i did miss that it seems.

The main reason as to why i ask is because the pin position tolerance is quite large (0.35mm) so we might need to use comparably larger holes. I will take a look at this after some time soon. (Maybe even today depending how long i take to finish the review on the other things i have already started.)

ferdymercury:
> The courtyard clearance should be 0.25mm not 0.5 (0.5 is only for connectors)
> 
> xp svg

Ok, I can change that. But the case tolerance is 0.5mm, so there might be a problem if the courtyard is too tight?

ferdymercury:
Note added https://github.com/pointhi/kicad-footprint-generator/pull/252/commits/e094e4202d742f7de84b4b7ca6c4e85c18c11337

ferdymercury:
fixed pin marker https://github.com/pointhi/kicad-footprint-generator/pull/252/commits/68957b3dd0de75e98ecfa46e26f6b1493e00e933

ferdymercury:
fixed IAxxxxS and IA48xxS first pin position https://github.com/pointhi/kicad-footprint-generator/pull/252/commits/d4922e422d718c5048960302b7325c3c2ca5e9aa

ferdymercury:
3D models are ready here https://github.com/KiCad/kicad-packages3D/pull/440/commits/4546bdb494026dfcba284f908cc00baaafecd5d0

poeschlr:
Looks good now. Thanks

PR #644, opened 2018-06-08
Add ST PowerSO-20/36 footprints

Salamandar:
Add ST PowerSO-20/36 footprints.

* The SlugDown footprints are based on the datasheet of the component (thus the dimensions of the slug).
* The SlugDown_ThermalVias footprints are based on the suggested PCB footprints.
* The SlugUp are the same footprints without slug (obviously).

Datasheet for the footprints: http://www.st.com/content/ccc/resource/technical/document/technical_note/group0/89/b8/8f/9b/ed/7b/41/56/DM00454055/files/DM00454055.pdf/jcr:content/translations/en.DM00454055.pdf

capture du 2018-06-10 23-03-13
capture du 2018-08-20 13-42-44
capture du 2018-06-10 23-05-20
capture du 2018-08-20 13-41-38

---
Thanks for creating a pull request to contribute to the KiCad libraries! To speed up integration of your PR, please check the following items:

- [x] Provide a URL to a datasheet for the footprint(s) you are contributing
- [x] An example screenshot image is very helpful
- [ ] If there are matching symbol or 3D model PRs, provide link(s) as appropriate
- [x] Check the output of the Travis automated check scripts - fix any errors as required

PR #1060, opened 2018-11-01
Add ST PowerSO-20/36-SlugUp/Down-ThermalVias footprints

Salamandar:
Add ST PowerSO-20/36-SlugUp/Down-ThermalVias footprints
This replaces the #644.

- Datasheet : https://www.st.com/content/ccc/resource/technical/document/technical_note/group0/89/b8/8f/9b/ed/7b/41/56/DM00454055/files/DM00454055.pdf/jcr:content/translations/en.DM00454055.pdf

Screenshots: 
ST_PowerSO-20_SlugDown_ThermalVias
ST_PowerSO-36_SlugDown_ThermalVias


Salamandar:
I don't know if I should keep mask on vias…

poeschlr:
Sorry for the long delay.

Is it on purpose that there are vias only connected to the bottom pad? (I would assume the intention would be to have copper there but maybe covered with soldermask) meaning leave the current large top pad as mask only and add a copper only pad that covers the area of the current pad plus a bit more to fully cover the vias.

The courtyard clearance in x direction is a bit wrong.
The 36 pin version has a slightly wrong body outline.

screenshot from 2019-01-01 00-31-19
screenshot from 2019-01-01 00-28-22

PR #1062, opened 2018-11-01
add Mini-Circuits XX112 SO package

lorem-ipsum:
This adds a SO variant used by Mini-Circuits and called XX112 in their naming system ([datasheet](https://ww2.minicircuits.com/case_style/XX112.pdf)). The vias are not fully specified in the datasheet, so I've chosen their locations. I extended the name to follow the regular naming scheme, however `Mini-Circuits_XX112` should already be unique.
screenshot_20181101_173725


DanSGiesbrecht:
Closed and reopened PR to get the CLA assistant to re-check.

poeschlr:
This could be scripted. @lorem-ipsum would you be willing to look into that?
The scripts can be found here: https://github.com/pointhi/kicad-footprint-generator/tree/master/scripts/Packages/Package_Gullwing__QFP_SOIC_SO

lorem-ipsum:
I tried building this footprint with the generator, but I am facing some difficulties. When building the footprint based on the mechanical dimensions of the housing the created pads do not match the ones specified by Mini-Circuits. As they already specify exact dimensions I would prefer sticking to those (+ rounded rects). Is there a way to specify the pad sizes instead of the dimensions of the mechanical parts?

poeschlr:
The mini circuits document is either from 2012 or 2013 (depending on which date format is used)

This is before the introduction of the modern industry standards used by the generator script. As this is a manufacturer specific footprint i would accept it if it is build after the manufacturer specification. But like other mini circuits contributions it would then need a mention of the case number and of the land pattern drawing number in the footprint name.

lorem-ipsum:
Ok, makes sense. Should I then add both? The standard compliant, script generated version using the name used so far and the manufacturer specific?

poeschlr:
Is this intended for a generic symbol or for a specialized one.
If it is for a specialized one then adding both options could make sense. (one is set as the default footprint the other reachable via the footprint filters as an alternative) For use with generic symbols it might be best to choose the manufacturer specific options named such that users understand for which part it is valid.

lorem-ipsum:
This footprint is used for three RF switches from Mini-Circuits (overview [here](https://ww3.minicircuits.com/WebStore/Switches.html)). It might be used for other Mini-Circuits components, but I haven't found any.
I did now add the standard compliant variants based on the script ([generator PR](https://github.com/pointhi/kicad-footprint-generator/pull/268)) and renamed the previous version to `Mini-Circuits_XX112`. I would set the later as default in the symbols.
screenshot_20190112_124920

poeschlr:
I added some comments to your script PR.

poeschlr:
Added further comments to the script. (clarifications)

myfreescalewebpage:
@poeschlr @lorem-ipsum just seen this thread, but not read all coments, just a question, why not the same strategy of other Mini-Circuits footprint proposed by @aewallin ? Example: https://github.com/KiCad/kicad-footprints/pull/1136 and many others.
Why several footprint and not only `Package_SO.pretty/Mini-Circuits_XX112_LandPatternPL-217.kicad_mod` ? Datasheet of a device using this footprint: https://ww2.minicircuits.com/pdfs/KSW-2-46+.pdf
Joel

poeschlr:
Because this one is a generic SO package wheras the others are specialized packages only produced for and by Mini Circuits.

myfreescalewebpage:
ok

lorem-ipsum:
As this has been hanging around for some time: Whats the verdict now?

`Mini-Circuits_XX112` would be most consistent with other components. In theory this could be a generic SOP package, but most likely it will be unique for Mini-Circuits (taking all the tolerances into account). This then also should go into `RF_Mini-Circuits.pretty` and I would take a look at the land patterns.

poeschlr:
What was hanging. I thought i made it clear that this is a  normal so part and will therefore be handled like a normal so part.

lorem-ipsum:
In the current state it does not fit with the rest of the library. I can delete the ´Mini-Circuits_XX112´ copy or move it to the ´RF_Mini-Circuits.pretty´.

poeschlr:
I am confused. The screenshot looks like this is a completely normal SO footprint. The datasheet looks like this is a completely normal SO package. Meaning how does this not fit into the SO library?

lorem-ipsum:
Fine by me. But this PR adds the footprint once with the generic name and once with the specialised `Mini-Circuits_XX112` name. That one should either go away or be moved to the special Mini-Circuits library.

poeschlr:
I would remove the mini circuits version but add info about it to the description. (I am not sure if i have a clean way to do this. For now you can misuse the datasheet field or invest time into adding something nicer.)

lorem-ipsum:
I've dropped the Mini-Circuits XX112 variant and added a tiny comment into the description by misusing the datasheet part in the generator.

poeschlr:
Looks good, thanks

PR #1065, opened 2018-11-02
Add XP Power ISU02 series DC to DC converter SMT footprint.

stambaughw:
Here is the footprint associated with the [symbol pull request](https://github.com/KiCad/kicad-symbols/pull/1116)
------------

- [x
xp-power-isu02_-md-footprint
] Check the output of the Travis automated check scripts - fix any errors as required


stambaughw:
Here is the [datasheet link](https://www.xppower.com/Portals/0/pdfs/SF_ISU02.pdf) I forgot to attach.

evanshultz:
- Please provide a useful description before the datasheet URL. Something like `2 Watt, Regulated Single & Dual Output, 1500 VDC Isolation, 19.0x17.0x8.7mm)`.
- The datasheet, keywords, and 3D model reference are for another part.
- Add a pin 1 mark on the fab layer.
- Silk should not be under the part body for SMD parts, so please offset the silk from the fab lines by 0.11mm (this is the usual amount).
- The courtyard should be +/-9.55mm wide.
- Pads should be 2.45mm wide, not 2.54mm.
- I like to see unrotated text since, which affects the fab layer ref des, if you would.
- Would you mind to horizontally center the ref des? Not required, but typical.

stambaughw:
I noticed that it a lot of the newer footprints are anchored on the
center of pin 1.  Do you want me to do the same for this footprint.

On 11/8/2018 4:02 PM, evanshultz wrote:
>   * Please provide a useful description before the datasheet URL.
>     Something like |2 Watt, Regulated Single & Dual Output, 1500 VDC
>     Isolation, 19.0x17.0x8.7mm)|.

I'm assuming that the datasheet url really doesn't belong here so I will
just add the description.

>   * The datasheet, keywords, and 3D model reference are for another part.

Yes, but it is the 3D model specified has the same dimensions or at
least it's really close.  Does it really make sense to create another 3D
model just because the name doesn't match?

>   * Add a pin 1 mark on the fab layer.

What would you recommend here?  I don't see any common pin one indicator
on the other DC to DC converter footprints and I don't see anything in
the KLC that indicates what this should be.

>   * Silk should not be under the part body for SMD parts, so please
>     offset the silk from the fab lines by 0.11mm (this is the usual amount).

Did I miss calculate the offset?  How did you arrive at 0.11mm.  I
though I used an offset of 1/2 of the line width from the nominal part
outline.

>   * The courtyard should be +/-9.55mm wide.

OK.  I'm guessing change is due to the fact that I used the wrong pad width.

>   * Pads should be 2.45mm wide, not 2.54mm.

Good catch.  I actually used this footprint on a prototype board and I
thought the pads seemed a bit long.

>   * I like to see unrotated text since, which affects the fab layer ref
>     des, if you would.

By unrotated do you mean uncheck the "Keep upright" option in the text
properties dialog?

>   * Would you mind to horizontally center the ref des? Not required, but
>     typical.

Will do.

evanshultz:
It's always nice to be able to antagonize you a bit. :)

1. THT footprints have origin on pin 1. SMD ones have origin in the center, so this one is fine.
1. Do keep the datasheet URL and put it after the description.
1. We do use unique 3D models. We have a generator so this is simple.
1. A notch on the pin corner on the fab layer will be fine. See clause 2 of http://kicad-pcb.org/libraries/klc/F5.2/.
1. 0.11mm offset, using a 0.1mm fab line and 0.12mm silk line, means the line edges just touch. 1/2 either line width will result in an overlap.
1. Perhaps. Courtyard should be 0.25mm from the extent of the pads or body.
1. When text is rotated upside down (180 degrees) it is automatically unrotated for display even though the footprint still includes a useless ` 180` in it. So removing that was my request.

stambaughw:
Hey Even,

No problem with the antagonizing. ;)  I really want to get this right.
I use a lot of odd ball parts in my designs and I've made so many
symbols and footprints over the years that I figured I might as well
contribute them to the libraries rather than maintain my own copies.
Someone else might find them useful.  I'm sure I will get this right at
some point.

1. OK.
2. Will do.
3. Not as simple as one would hope.  This is particularly true of the DC
to DC convert script.  The pins on this package are gull wing but the
converter script creates flat pins that extend across the width of the
body which doesn't look very natural.  I haven't had the time to look at
the script so maybe someone can do that for me.  I will add a new 3D
model based on the one I originally used to the script and submit a pull
request to the model library.
4. Will do.
5. I wasn't aware the overlap with the fab layer was an issue.  I was
merely trying to make sure the silkscreen outline was as accurate to the
actual part dimensions as possible but I have no issue changing this.
6. Got it.
7. I didn't even notice that there was a superfluous 180 rotation.  I
think this may have been a copy an paste error so I'm guessing that this
is also an issue with the footprint I copied from.  I just don't
remember off the top of my head which footprint that was.

stambaughw:
I pushed the changes you requested and here is the [pull request for the 3D model](https://github.com/KiCad/kicad-packages3D/pull/456).

evanshultz:
Thanks!

Can you do two more things?
1. Get rid of the dual space in the description and add the Wattage?
1. Add vertical silk lines on the sides between the pads. I noticed this before but forgot. Sorry.

stambaughw:
I made the requested changes.  I am somewhat concerned about the silkscreen on the sides of the footprint.  I use automated optical inspection (AOI) at work and some times the silkscreen between pads confuses the AOI inspection of solder joints.  It's trivial to modify the footprints for this but it is something to keep in mind when generating footprints.

evanshultz:
Thank you. There is still a double space in the description. Alternatively, copying exactly the text I pasted above and appending the datasheet URL should work.

Noted about the silkscreen. I've not run into that problem but I can imagine it happening if the silk line is short.

stambaughw:
I pushed the fixes.  Sorry about the double space.  I also added the
dimensions to the description and fixed some keywords that were
incorrect due to copying the original footprint.

I've only seen the AOI issue with short silkscreen lines and the
rectangle (body outline) we used to have in our SOIC footprints.  I just
wanted to be sure our librarians are aware of AOI issues.  I've never
heard anyone comment on it before.

evanshultz:
Thanks Wayne! You mentioned many more symbols and footprints so I'll do the symbol for this one next and hopefully see more of your contributions when time allows.

PR #1066, opened 2018-11-03
Add Raspberry Pi Zero footprint using Through-hole pin socket

diegoherranz:
To use a Raspberry Pi Zero ("W" model too) as a daughter board.
Pin header would be soldered on the Raspberry Pi Zero and the main board uses a pin socket.
No specific schematic symbol at the moment. A standard Pin socket can be used in the meantime.

Reference image:
pi-zero-header-w-536x4081

Screenshot:
rpizero

3D render:
rpizero_3d

Notes about dimensions:
- Official Mechanical drawing:
  https://www.raspberrypi.org/documentation/hardware/raspberrypi/mechanical/rpi_MECH_Zero_1p2.pdf
- Dimensions of the PCB are pretty well defined on the drawing, not a problem.
- Connectors:
    - 1x 40-pin header: well defined in the drawing, not a problem.
    - 2x micro USB, 1x mini HDMI: location around the outline is defined, but not their size nor how much they stick out which I have measured on a real unit.
    - 1x camera FFC: not in the drawing so measured on a real unit.
- Please note that it's all mirrored left-right against the drawing since the components on the Raspberry Pi Zero would face down so that pin 1 matches pin 1.

Considerations taken:
- Origin at pin 1.
- Fab shows the real dimensions of the PCB but not the external connectors.
- Courtyard uses 0.25 mm clearance all around except around the connectors where uses 0.5 mm clearance.
- Silkscreen just outside of the module (half-width out) and not showing the shape of the connectors.
- Socket copied from PinSocket_2x20_P2.54mm_Vertical. It still contains all the layers from it (Fab, Silk, etc.) which I think that can be useful when designing a board (given the module will be raised up you can place components underneath).
- 3D model of the socket rotated 90 degrees. I know that KLC says no rotation or offsets different from zero but I think this makes sense in this case.

I've tried to stick to the KLC conventions but given it's not a "conventional" component it's not always straight forward.

Let me know if you want any changes.

Thanks.

------------


evanshultz:
Would this be best as a template (https://github.com/KiCad/kicad-templates) rather than a footprint?

diegoherranz:
This footprint allows you to use the Raspberry Pi Zero as a component on your board, as opposed to a template which would allow you to design an accessory board to plug onto the raspberry. It may be a subtle difference but I think that they are completely different things. Let me explain two examples:

If you are designing a small board which effectively becomes a HAT or add-on board for the Raspberry Pi Zero, then yes, I think that a template would absolutely be the right thing to use.

However, if you are designing a much bigger board where the Raspberry Pi Zero will simply be yet another component, I think that a footprint is the way to go.

Thanks!

fauxpark:
> No specific schematic symbol at the moment

There is already a symbol: `Connector:Raspberry_Pi_2_3`, but its footprint filters are only configured for the generic pin headers/sockets.

diegoherranz:
> There is already a symbol: Connector:Raspberry_Pi_2_3, but its footprint filters are only configured for the generic pin headers/sockets.

I hadn't seen that one. Good! I guess we could add another footprint filter entry to that one.
Thanks.

evanshultz:
OK, now I got it and that makes perfect sense to add the footprint this way. Thanks for explaining.

Let me take a closer look now that I understand what's going on.

diegoherranz:
Thanks for the thorough review.

> I'm thinking it would be nice to have a simplified 3D model of the whole RPi Zero so that it appears in the users' board and can help with collision detection. Would you be opposed to that instead of just showing one header?

Absolutely not. I didn't add a model for the RPi Zero because my 3D CAD skills aren't great, but if someone creates it, it would obviously be the best solution.


> I'm thinking just calling this `Raspberry_Pi_Zero_Socketed_Reverse` or something similar would be good. This would indicate there is a socket and the board is upside down. Do you have another idea?

I called it `Raspberry_Pi_Zero_PinSocket_2x20_P2.54mm_Vertical` because I was planning an adding another version in the future using a SMD socket instead of Through-Hole, but doesn't need to be that verbose. Maybe `Raspberry_Pi_Zero_Socketed_TH_Reverse` to allow for a future `Raspberry_Pi_Zero_Socketed_SMD_Reverse`? Not sure about `Reverse`, though. The natural location for the header on the RPi zero is facing up as shown on the first picture above. So the natural placement on the RPi Zero has to be upside down. That's what feels as natural to me and the `Reverse` makes me think that we've done something different to that. But anyway, not a strong opinion.

> KLC says pin 1 should be in the upper left for THT footprints, so please rotate 90 degrees CCW.

OK, will do as soon as I get some time, probably next week.

> I would like fab and silk lines to go around the connectors hanging off since, if the RPi Zero is projected onto the PCB, they will describe the consumed board area. Perhaps start with the rounded rectangle and then add the "warts" sticking out for the connectors.

I originally did that on the Fab layer, but then I changed it when I saw [F5.2](http://kicad-pcb.org/libraries/klc/F5.2/) which says "Outline should be simplified and not display complex features" and the examples shown there do not enclose the whole package so maybe the simplified Fab is OK? Happy to change it if you think otherwise.

Regarding Silk, I doubted a lot whether to draw the connector or not, so not a problem, will do.

> Because the courtyard must be closed, it will be better to avoid arcs. We have had trouble with that in the past, so can you make the courtyard only use orthogonal lines?

Yes, will do. Maybe we can avoid these problems in the future by adding checks on kicad-library-utils to make sure it is a close path?

Thanks!

evanshultz:
3D model:
I'm sure some skilled person will be able to whip up a 3D model, so let's name it like the footprint and then when a model does exist it will all be ready.

Naming:
First, we should definitely use "THT" and "SMD" since that matches what's used elsewhere. Next, I wrote "Socketed" but for brevity just "Socket" ought to be sufficient. And last, I was hoping "Reverse" would indicate the Zero's CPU is facing down (towards the motherboard) since it would be possible to solder it either way. I'm open to other words I just didn't have anything better at the time. I like where you're headed, so can you come up with a better term to fully define the orientation and placement of the Zero?

Fab:
Especially since the outline of all the connectors can vary a bit, as some aren't specified or the exact vendor could change, I'd like to for sure have the Zero board outline on the fab layer. This clause of KLC very much applies where there can be dramatically too much detail (like importing a DXF drawing as a footprint), but here I think this will work well.

Then silk can probably go just around the connectors themselves since some connector will be there and the silk is only a rough guide.

Is that OK with you?

Courtyard:
Sadly, KiCad will try to close slightly open ends and I understand that algorithm varied a bit from release to release during v5 development. Even if the vertex was exactly on top of each other KiCad could choke when importing a footprint. Simplified courtyards was our way to ensure compatibility for now. IPC 7351C will recommend contoured courtyards, and KiCad may be stable, so having more fancy courtyards (like arcs) is probably a nice update that can be done in the future.

diegoherranz:
I made the changes discussed on this PR.

Regarding the name, I finally went for `Raspberry_Pi_Zero_Socketed_THT_FaceDown`.
I found `Socketed` more clear than `Socket` since the former conveys the idea of module+socket.
Regarding the `Reverse` suggestion, I went for `FaceDown` since I believe that one to be absolutely clear and it's not much longer.

Thanks.

evanshultz:
1. The keywords has a typo of "through".
1. Let's put `_MountingHoles` at the end of the footprint like others in this lib. (Not sure how you could safely get away without mounting holes since there's only a single inline-pin connector, but...)
1. Can you add a link the the mech drawing in the description?
1. The silk lines don't appear to be offset enough to be outside the fab lines. Please either offset them by 0.11mm or put them right on top of the fab lines.
1. Thinking about this again, wouldn't the user want the THT header to be a component on the BOM along with the Pi Zero, and thus two separate parts in the design with unique ref des? Maybe the position of the header should be shown but the holes not included? This is a bigger issue because many of the footprints in this lib have THTs for headers, so I'll merge as it is but this raises a question to be answered for this library.

diegoherranz:
> The keywords has a typo of "through".

Fixed

> Let's put `_MountingHoles` at the end of the footprint like others in this lib. (Not sure how you could safely get away without mounting holes since there's only a single inline-pin connector, but...)

Yes, you need the mounting holes otherwise or it would be too weak. Done. 

> Can you add a link the the mech drawing in the description?

Done

> The silk lines don't appear to be offset enough to be outside the fab lines. Please either offset them by 0.11mm or put them right on top of the fab lines.

The silk lines are 0.12mm wide and they are offset by 0.06mm which means that the whole silk line is outside of the component. I think that that meets [F5.1 (3)](http://kicad-pcb.org/libraries/klc/F5.1/) ("For SMD footprints, silkscreen must be fully visible after boards assembly (no silkscreen allowed under component)".

Happy to offset it more, no problem at all. Just let me know.
I think we should modify the KLC on this respect to specfiy which offset we prefer.  E.g. "Recommended offset against the fab lines is 0.12mm". That would help help making all the library more consistent.

> Thinking about this again, wouldn't the user want the THT header to be a component on the BOM along with the Pi Zero, and thus two separate parts in the design with unique ref des?

Can that be done at the schematic-layout? I mean having 2 parts with an unique ref des?

> Maybe the position of the header should be shown but the holes not included? This is a bigger issue because many of the footprints in this lib have THTs for headers, so I'll merge as it is but this raises a question to be answered for this library.

I see where you're going, and the way it is at the moment would require tweaks at the BOM level, but I think that is easier and less prone to mistakes than aligning the header and the raspberry footprint manually (everytime by every user). I guess ideally we would want something like a single component on the SCH/LYT that ends up being 2 components on the BOM (but that's not supported and may not be easy to support). Maybe a quick and dirty way is renaming the component for BOM purposes as "HEADER xxxx (whichever header desidred) + Raspberry Pi Zero". Even if that is a single line on the BOM still, at least you see that it is two things.

Thanks!

evanshultz:
Silk line offset:
Yes, good point, though you reference the SMD clause and this is a THT footprint. I think I wrote about that somewhere else. For THT footprints, silk can be under the part body according to KLC. The standard offset which is used in all the modern scripts is 0.11mm, but you're right that 0.06mm does work for a SMD footprint. Just bringing it up for consistency but I'll merge as is. KLC does need an update here.

Header + Pi Zero outline:
I've always done it as separate parts (by "design", yep, I did mean the schematic which then drove the board) so that I had unique items on the BOM. But since all the other module footprints are like this already let's go with it. I opened https://github.com/KiCad/kicad-footprints/issues/1408 to discuss and track this issue.

So are we ready to merge this? Your last commit looks OK to me. Since we were discussing things above I don't want to jump the gun.

diegoherranz:
Of course, it is THT. I'm stupid, hehe!
For THT makes much more sense to match the fab line (no offset). Let me fix it in a sec.

diegoherranz:
Re-reading your comment, do you mean 0.11mm for everything? I can use that instead of no offset.
Where does 0.11mm come from by the way? Thanks.

fauxpark:
I assume that is (silk / 2) + 0.05.

poeschlr:
Silk to fab offset is not overly critical. It should be outside of the nominal body size but still near enough to provide guidance should the pcb be hand populated.
And it should maybe look good as well. One option is to have silk and fab lines just touching. (The famous 0.11mm offset. My scripts all use this offset.) 
If i create a footprint by hand then i use a reasonable grid and draw silk near to the fab outline. (I typically use 0.05mm for this task. If i am overly motivated i might however go done to 0.01mm)

The crtl key is a vital help here as it restricts lines to multiples of 45 degrees. And the snapping support that is added to current nightly builds.
Another helpful tool is the sketch or outline mode. (shows lines as a thin outline. Found in the left toolbar)

diegoherranz:
Cool, I understand it better now.
I'll go for 0.11mm for consistency, and as discussed, it would be good to add this recommendation to the KLC.
I need to fix the 3D model path as well since I updated the footprint name but not the model!

Nice about the outline mode, I hadn't used it before. Very useful.
How does the snap work? I'm using nightlies.

Thanks!

poeschlr:
I don't really want to add a fixed value to be honest. Using exactly 0.11 clearance is easy for scripted parts. Handdrafted footprints really need a lot more tolerance. My suggestion really is: Draw silk such that it is easy to create but still looks good and gives the maximum viable benefit to the user of the footprint.

diegoherranz:
I'll leave this as is then and simply fix the 3D model.

evanshultz:
Thanks!

PR #1068, opened 2018-11-05
Power integrations e dip 12 b

Misca1234:
Replaced the data sheet link from a retailers url to the manufactures url

------------


evanshultz:
I notice that the package in the datasheet is called `eDIP-12` without the `B` suffix. Should that be fixed as well? Searching for "eDIP-12B" online I'm not sure it was ever a proper package name so I'm a bit confused. This would require updating `LNK457V` and `TOP264VG` symbols in `Regulator_Switching.lib` so I'm OK leaving it out now and making an issue, but I do think we should capture that our footprint name doesn't match the vendor's name.

Why was the 3D model repositioned? I don't see that a 3D model exists so shouldn't the 3D params go to the default?

Misca1234:
I do not know why 3D model was repositioned, I was only changing the data sheet, let me check

Misca1234:
This whole foot print is weird, it is essential a DIP but they have numbered the pins in the opposite direction compared to the 99.99999999% otehr DIP's out there.
But they number the pins on the two other packages in the "correct" direction.

Misca1234:
In the master the 3D model is positioned like this

  (model ${KISYS3DMOD}/Package_DIP.3dshapes/PowerIntegrations_eDIP-12B.wrl
    (at (xyz -0.2 -0.1751968504 0))
    (scale (xyz 1 1 1))
    (rotate (xyz 0 0 0))

evanshultz:
Yes, it is a weird package.

I'd prefer to stick with the standard 3D params in KLC, which is what you've done in the latest commit. Are you OK for me to merge?

Misca1234:
Sure, lets do it

PR #1069, opened 2018-11-06
Fixes for consistency with other footprints

robkam:
was PR #828
mono with circular holes QingPu_WQP-PJ398SM

mono circular holes qingpu_wqp-pj398sm


robkam:
one

poeschlr:
Is this still relevant or has some other pull request made this obsolete?

robkam:
> Is this still relevant or has some other pull request made this obsolete?
It's still relevant. Except there was another PR to resolve TN, S, T vs. 2, 1, 3. I didn't follow what was decided.

poeschlr:
As this is an audio connector i would assume the letters are what should be used.

robkam:
With the letters there's no mismatch to the symbols, but the data sheet uses numbers.

poeschlr:
The decision was made to use numbers for this part. This is because we are not prepared to have atomic parts for audio connectors.

poeschlr:
Oh dammit. Sorry i got confused in my last response. We should use letters.

poeschlr:
Thanks

robkam:
Thanks! :-)

Issue #1067, opened 2018-11-05
Wrong SMD footprint for MeanWell IRM-02-XX


omogenot:
Hi,

There is a mistake in the footprint ACDC-Conv_SMMeanWell-IRM-02-XX. Looking at the data sheet (and the physical unit), there are only 2 x 7 pins (numbered up to 24) whereas the footprint in the library has 2 x 8 numbered up to 26.

evanshultz:
Thanks for posting this potential issue.

Including the datasheet and a screenshot would be helpful so someone can quickly verify.

The datasheet (https://www.meanwell.com/Upload/PDF/IRM-02/IRM-02-SPEC.PDF) shows:
![image](https://user-images.githubusercontent.com/1936989/48013782-0640a080-e11d-11e8-9c91-c20776120966.png)

Here is the footprint:
![image](https://user-images.githubusercontent.com/1936989/48011653-2752c280-e118-11e8-8e32-67e01cc8d4f0.png)

Yep, sure looks seriously messed up to me. And the IRM-03 SMD version appears to be equally hosed.

For completeness, here is the schematic symbol for the SMD version:
![image](https://user-images.githubusercontent.com/1936989/48013825-21131500-e11d-11e8-9852-fc57df4faacf.png)

Fortunately, the schematic symbols looks good.

So it looks like pin 1 wasn't put in the right place, and the direction of pin increment was wrong, and the skipped pins were wrong, which all lead to only pin 13 being in the right place (and that just accidentally worked out). So this converter is totally unusable now. How very sad.

In addition to fixing the pin numbering, the fab and silk lines will need to be modified to correctly note pin 1's location.

It does not appear that these footprints are scripted, so I suppose they were handmade. Are you willing to submit fixed footprints?

Note: How did https://github.com/KiCad/kicad-footprints/issues/749 and https://github.com/KiCad/kicad-footprints/pull/750 get reviewed and merged but this serious bug remained??

omogenot:
Hi.
We're not looking for who made a mistake, I am the only one never making any :-) Don't worry.
I made the foot print mod, are you interested in it?

[ACDC-Conv_SMMeanWell-IRM-02-XX.kicad_mod.zip](https://github.com/KiCad/kicad-footprints/files/2550320/ACDC-Conv_SMMeanWell-IRM-02-XX.kicad_mod.zip)

antoniovazquezblanco:
[...]

PR #1071, opened 2018-11-06
Fix MeanWell IRM-02 SMD pad numbering

evanshultz:
From https://github.com/KiCad/kicad-footprints/issues/1067.

No image since it's well-documented in that issue and this is a simple change. Also updated the datasheet.

(Just FYI, I think this was copied from the IRM-03 where the pinout is correct; it may have been missed that the IRM-02 pinout is different and therefore can't be a direct copy.)

------------


antoniovazquezblanco:
Thank you!

PR #1072, opened 2018-11-06
Rename SIP-3 Packages in Sensor_Current

jneiva08:
Rename SIP-3 Packages in Sensor_Current to include size and pitch

* Allegro_SIP-3 to Allegro_SIP-3_4.09x1.52_P1.27mm
* Diodes_SIP-3_AmmoPack to Diodes_SIP-3_4.1x1.5mm_P2.65mm
* Allegro_SIP-4 to Allegro_SIP-4_5.21x1mm_P1.27mm

This change was because of the following Symbol PR [#1048](https://github.com/KiCad/kicad-symbols/pull/1048)

------------


evanshultz:
I just noticed this. Sorry that I missed it yesterday as it would have cleared up for me what you were doing.

As I mentioned in the symbol thread, I would prefer no vendor prefix if the packages are compatible across manufacturers (but then body size and shape need to be captured in the footprint name) or making unique footprints for each vendor even if the footprints are otherwise the same. Is that OK? Do you have some other ideas?

Also, I assume these footprints are here because some other symbol needs them. Did you check all symbols to see if this renaming would break any footprints association? If so, the affected symbols will need to be updated to match the footprint renaming.

jneiva08:
If you agree, i will close this PR and will create a new on only with Diodes_*_P2.65mm and create  Diodes_*_P1.27mm. Later can create another only for Allegro_SIP-3/4

evanshultz:
Either modifying this PR or create a new PR is OK with me as along as we're both planning for the same changes to be merged.

I think the best path, both considering a requirement to not remove any footprints currently being used, is to simply add `Diodes_SIP-3_P2.65mm` and `Diodes_SIP-3_P1.27mm`. Is that what you're saying above?

jneiva08:
Added missing footprint for hall switch from diodes.

the footprint looks:
diodes_1 27mm

It's identical from Allegro SIP-3, but with Diodes size :)

jneiva08:
@evanshultz have changed name of existing Diodes_SIP and added the new footprint

evanshultz:
For reference, datasheet is at https://www.diodes.com/assets/Datasheets/AH1806.pdf.

Sorry that I missed this earlier.
- 0.74mm is from the package front to the center of the pads above, but in the datasheet it appears to be the front of the pins. The pin thickness is 0.38mm nominal, so I think the 0.74mm above should be 0.93mm. Can you please check and confirm?
- Pull back silk from pads by 0.12 to 0.2mm. Some places are too close and others are too far away.
- Shrink the ref des on the fab layer a bit so it fits fully inside the body outline.
- In the drawing above distance J isn't dimensioned. On the 1.27mm pitch one it's only 2.51mm.
- IPC-2221 Level B says these 0.38mm square pads should have 0.7374mm drill holes (0.2mm over the lead diameter), so your 0.8mm holes are OK. But the annular rings should be 1.35mm minimum (lead diameter + 2 * 0.15mm + 0.25mm) while your pads are 1.1x1.6mm. Can you make the smaller dimension a bit bigger? Optionally, you could slightly shrink the drill size to get a bit of relief in the annular ring minimum size.

When you dimension, if you don't mind, including the courtyard offset helps to quickly check. Thanks!

jneiva08:
@evanshultz have made the following changes:
- Inserted an offset of 0.19mm in the body,
- Changed the F.SilkS,
- Reduced the letter size fo %R to 0.7, fits better inside of body,

diodes_1 27

The size of J is defined on the datasheet as 2.62mm.
If you increase the size of the pad in x to 1.35 mm it will touch the adjacent pin, the pitch is 1.27 mm, even if reduce the drill hole to 0.75mm.

On Diodes_*_P2.65, can make the offset of 0.19mm, and change the pads to 1.35mm.

evanshultz:
The 1.27mm one now looks good. Thanks! I guess let's just leave all the pads alone.

Can you make the same changes to the P2.65mm one? Make sure to pull back the silk here.

jneiva08:
Updated footprint with the same points of 1.27mm

diodes_2 65

evanshultz:
Thank you. I meant in general and not just in this one place. Sorry if that wasn't clear.

poeschlr:
A hacky way to visualize the silk to pad offset is by setting the footprint clearance temporarily to the clearance value you want to achieve. (Example 0.2mm)

The same trick can also help with designing the courtyard area for footprints where the pads are outside of the body.

Just don't forget to set the clearance back to the normal value after you used this trick.

jneiva08:
@evanshultz sorry for only do it now, have moved a little all F.SilkS near of pads, hope it is all fine

poeschlr:
Did you check that the original footprint is not currently used by any symbol in the library? If it is then please update the footprint field of the affected symbols.

jneiva08:
The added footprint was intended for AH1806-P, the symbol still not merged.
The Diodes_SIP-3_AmmoPack was renamed to include the pitch.

evanshultz:
@jneiva08 
Sorry for my delay in this.

There are two areas where the silk is a bit close. I've used the hacky trick Rene mentioned above, with the clearance set to 0.12mm.

2.65mm pitch:
image

1.27mm pitch:
image

If you can adjust the silk a bit more I think this is ready to merge and then we can wrap up the symbol.

jneiva08:
@evanshultz Did you update?

Have opened the footprints in PCBnew and have Clearance by default in 0.2mm and the footprints looks:

1.27mm
sip-3_1 27

2.65mm
sip-3_2 65

evanshultz:
I did intend to update, but somehow I failed. All looks well so please accept my apologies for letting this PR slip. Thank you!

jneiva08:
@evanshultz Thanks for you time. Do not worry about it.

PR #1074, opened 2018-11-09
Fix Kingbright APFA3010 pad numbering

xxv:
Flip pins 1 and 2 as per [the datasheet for APFA3010LSEEZGKQBKC](http://www.kingbrightusa.com/images/catalog/SPEC/APFA3010LSEEZGKQBKC.pdf).

0xd31a028e


xxv:
Err, that's not right. I'll re-submit.

PR #1075, opened 2018-11-09
Fix Kingbright APFA3010 pad numbering

xxv:
Flip pins 4 and 3 as per [the datasheet for APFA3010LSEEZGKQBKC](http://www.kingbrightusa.com/images/catalog/SPEC/APFA3010LSEEZGKQBKC.pdf).

0xd31a028e


evanshultz:
Good catch! Thanks for noticing this error and pushing a fix.

Actually, our centering algorithm is a bit weak and moving the footprint center would shift the footprint on users' boards if they update from the library, so do you mind moving the footprint back where it was (only fix the pinout)?

FYI, I verified the symbol is already correct.

xxv:
I pushed the reverted commits.

evanshultz:
Great. Thank you for this fix.

I see you've posted a few times but this is your first PR. If you're not aware, you can find the library guidelines at http://kicad-pcb.org/libraries/klc/. I hope you'll continue to contribute to the KiCad library!

xxv:
@evanshultz Thanks for the useful link and super helpful feedback! I'll definitely be contributing whenever I can :smile:

PR #1076, opened 2018-11-09
extend selection of DIN41612 connectors

lorem-ipsum:
This selection was generated by an updated script using the
kicad-footprint-generator. This current selection was built by checking
the catalogs of ept, ERNI and Harting and collect the most standard
series.
Added in this commit were: B (was already present), C, D, E, F, Q, R
Still missing standard series are:
  - H11, H15 (vertical and horizontal are different from each other and
    require an updated script, only H11 horizontal is present)
  - M (has a lot of variations which needs to be tracked).
There are also many manufacturer specific variants which might follow
(some are very easy to add to the script, others not so much).

The new script was made to copy the style of the existing scripts, just
adding the increased flexiblity to generate all series. Changes to the
generated footprints compared to the existing B series are:
- a1 pin is a roundrect instead of rect
- vertical connectors got their silk ref moved to the side. They are
  often used on backplanes, so there might be no space on top.
- courtyard of the horizontal connector is simpler on the connector
  side. This space if off the PCB anyways and there needs to be room to
  attach the connector.
- names are now always using two digit numbers for rows and pins.
- Q and R series vertical have slightly more drawings, trying to
  symbolise their inverted connector nature (vertical is male, while
  usually they are female).

There are two series where the names contain the rows which are
occupied. These are due to connectors only ept sells (E and F series
2x16). The standard is to leave the center row empty, ept also sells a
variant where the top row is empty. To distinguish those only the
non-standard has the rows added to the name.

As a general remark to these connectors. The naming of the pins and rows
is very strict, based on the location relative to the mounting position.
At least in vertical this always holds true, horizontal got some
flexibility on the spacing. The origin is always located at the a1 pin,
even if it is not present. This ensures that connectors can be switched
without having to change everything. This was already the case for the
existing connectors.

There are no real datasheets (as far as I could see the standard is not public). I worked of the catalogs of [Harting](https://www.harting.com/sites/default/files/2017-11/17_7232_DIN_41612_Katalog_D_proof_2017-01-13_verlinkt_0.pdf), [ERNI](https://www.erni-x-press.com/de/downloads/kataloge/englische_kataloge/erni-din41612-iec60603-2-e.pdf) and [ept](https://www.ept.de/DIN41612.pdf). Some example screenshots below, but there are way to many connectors to display them all. The generation script can be found [here](https://github.com/lorem-ipsum/kicad-footprint-generator/tree/din41612/scripts/Connector/Connector_DIN), I will post a PR once it is clear that the style is fine.

screenshot_20181109_084445
screenshot_20181109_084526
screenshot_20181109_084555
screenshot_20181109_084621screenshot_20181109_084423


evanshultz:
This is fantastic! Adding more of these connectors was on my long-term list of tasks, but now there's no need with your quality footprints!

- Please add a datasheet or a few to each footprint so we have attribution. Since we use English for the KiCad library please use English documentation.
- In the description, would you mind to swap the text describing the rows and pins so that the number of rows comes first? Row first is (I believe) the most common way to describe connectors.
- http://kicad-pcb.org/libraries/klc/F3.6/ specifies that no leading zeros should be used in the number of rows.
- I prefer the original silk ref des placement where the silk is right side up by default. I understand what you're saying, but the silk ref des usually must be moved anyway after component placement.
- Am I missing seeing a pin 1 mark on the fab layer?
- How are female and male demarcated? For example, type B/3 is available in both M and F genders in a vertical footprint, but I don't see male and female footprints. Since the 3D model is tied to a specific footprint (at least for now, and it will always have a default one), we should separate by gender.
- Please offset silk from fab lines by 0.1mm so the line edges just touch.
- Make the silk run around the outside body of the connector as much as possible (only avoid pads).
- For 1mm drill holes, IPC 2221 specifies 1.55mm annular rings for Level B. You can keep slightly larger pads if you like, just letting you know IPC says slightly smaller is OK.

Checking the script is better/easier than reviewing each individual footprint. Let me spend some more time looking at the script and footprints, and in the interim it's fine to go ahead and submit the script PR.

lorem-ipsum:
I started the PR for the footprint-generator and fixed the trivial parts.

About the comments:
- There is currently no pin 1 marker on the fab layer, I always find it difficult to find an appropriate style. Do you have any suggestions?
- Female / Male are so far not demarcated as they share the same footprint. I was not thinking about 3D models. I will go through them once I have time and generate both variants.
- I assume your silk screen running around the body comment was targeting the horizontal connectors. I would rather stick to the current style, maybe extend it a bit in the center towards the pins. The first pin row is already blocking most of the housing shape and I would rather not go around the pins. One could draw the full line if the `a` line is not occupied.
- I reduced the copper diameter a bit. As some of these connectors have a lot of pins to layout the extra space might come in handy.

evanshultz:
Likely a suitable pin 1 mark on the fab layer would be adding two lines to form a triangle against an existing line on the layer. Here is an example:
image

I was mentioning silk for both horizontal and vertical. As shown above, tucking closer to the pins (avoid pads by the line width, or 0.12mm) looks nice. Small segments between pads isn't necessary. But what about running all around the extents of the header? Are there situations where this header isn't always hanging off the PCB? For vertical ones, silk will help avoid inserting the part incorrectly by 180 degrees. The last screenshot in your first post would otherwise have little way to show which is the correct insertion. If the physical part doesn't clearly show pin 1 even one rectangular pin isn't helpful.

Also, in the images at the top I see a couple footprints without any rectangular pin.

lorem-ipsum:
Thanks for the helper for the fab layer, I will integrate that.

About the silk layer: As far as I know all of the horizontal connectors extend below their mounting plane on the PCB and must be placed on the edge, but I will double-check that. The vertical connectors do already have a slight silk painting for the connector outline, I moved that out a bit, but didn't post a new screenshot. It now looks like this:
screenshot_20181113_231104
I can run this around as far as possible instead of having only the start of it, but it won't work for all of the connectors due to pin positioning.

evanshultz:
Alright. Thanks for the update. If you would check into the horizontal connectors requiring a board cut and provide as much silk on vertical connectors as possible that's all my questions for now. Please make those changes and push another set of footprints and I'll review them in depth. Thanks again!

evanshultz:
@lorem-ipsum 
Will you be getting back to this PR soon?

lorem-ipsum:
I am quite busy right now due to the end of year season, but I hope to find some time around christmas to work on this again.

lorem-ipsum:
I extended the silk marker for the vertical connectors to include the full outline where possible, but I personally preferred the previous solution. As a general note: it is not possible to place the connectors in the wrong orientation (without damaging the connector) as the mounting holes are placed asymmetrically.
I also checked the horizontal connectors, they are all designed to be placed at the board edge.

evanshultz:
A few notes after taking a look at the footprints in the latest commit:
1. Should there always be a round rect pin? The lowest letter-number combination? So if `a1` isn't present then it will be `a2`?
1. There is text smashed together in the description. For example, `DIN41612_B2_2x16_Horizontal` includes `2 rows16 pins wide`. Perhaps a missing space or comma?
1. I'm not seeing a fab pin 1 mark. Did you get to add that?
1. Some of the silk pin 1 marks are offset. See `DIN41612_E_2x16_Vertical`:
image
1. Still missing female and male variants.
1. Outer silk lines can be closer to fab lines. They are 0.2mm now but can be 0.11mm. This is optional.
1. I like your suggestion at https://github.com/KiCad/kicad-footprints/pull/1076#issuecomment-438167574 to extend the silk lines in towards the outer pins in the row on horizontal footprints. Can you do that?
1. I would think the silk lines showing connector orientation (around the pins) could be placed on top of the fab lines. They're indicating exactly the same thing and so no offset is needed.
1. On the vertical connectors, there are overlapping fab lines above and below the pins:
image
1. By the vertical connectors not being assymetrical, you mean how to X position of the 2.8mm mounting holes on `DIN41612_R2_3x16_Vertical` are not inline with the B row, right? So trying to insert the connector 180 degrees around would result in a 0.6mm mis-alignment and it would be obvious to any normally-functioning human that something is wrong.

evanshultz:
@lorem-ipsum 
Did you have a response for my comments above? Also, and sorry if this sounds pushy, but do you have time to work on this soon? Some other DIN 41612 PRs have popped up recently and if these nice footprints can be get merged we should be halfway to eliminating future issues or requests. So that all adds up to: thank you for taking the time and effort to get so far already!

agalakhov:
BTW, does somebody work on DIN 41617 as well?

evanshultz:
@agalakhov 
Not as far as I'm aware.

lorem-ipsum:
2nd try as github ate my comment yesterday. I've integrated your feedback for the existing connectors. The selection is not complete yet, but I would like to get those done first :-) About your comments:
1. I am not really sure. Currently only a1 is made a square, if it does not exist then there is none. As the series uses this strict naming scheme I thought it would make sense. But I can also change that
2. fixed
3. added them. I used an arrow for the vertical ones and a simple bar for the horizontal ones as I found no nice spot for an arrow like this:
Screenshot_20190522_081859
Screenshot_20190522_081833
4. basically the same as 1. they always point at a1 even when it is missing.
5. These are all that exist in the standard. Usually vertical is female and horizontal is male. Only exceptions are the R and Q series where this is flipped. There exist non-standard extensions which I would like to add later on, but they usually come with slightly different outlines.
6. moved them closer (was easy to do)
7. I moved them as close as I could before they started colliding with the pins 32. The ones at top I kept the same to keep it symmetric.
8. changed
9. fixed
10. yes, thats exactly what I mean. Nothing one can not "fix" with force, but it should be enough for regular users.

lorem-ipsum:
It is still complaining about the fab markers you mentioned in your comment as 9th. They no longer collide, but they touch. This is due to the way the script works by first drawing the standard outline and then adding the extra markers. Is this something I have to fix or is it OK the way it is now?

evanshultz:
Thanks for continued work on this! You have put in some real time and done some fabulous work!

I'll just answer items which aren't closed above:
3. The "pin 1" fab mark can collide with the pads so you can do an arrow inside the body outline for all footprints. Also see below.
4. See question below.
7. Maybe I'm missing this but it seems that the silk lines we discussed are missing. Here is what I see that doesn't match the hand-drawn proposal you made above:
image
And the 2x8 version can have even longer silk lines than the footprint I just showed above:
image

Regarding your comment above, that was fixed in the check script. It didn't handle multiple lines in the same direction with common endpoints correctly. You can see all CI tests pass now. That being said please take a look if there are places where a single line could be used instead of multiple smaller lines.

@poeschlr 
1. Can you please verify this? Should there be a rounded rect (or unique-shape) pad on the "lowest" pad if `a1` is not present?
4. Similarly, should the "lowest" pin be marked (such as `a2`), if `a1` isn't present? Right now the mark is at the normal `a1` location even when that pin isn't present. Compare `DIN41612_B3_2x5_Female_Vertical` and `DIN41612_B3_2x10_Female_Vertical`.

lorem-ipsum:
Hi @evanshultz,
thanks for the review. I will fix the silk lines, I was always thinking about the vertical variants and completely forgot the horizontal ones.

About the pin 1 fab marker: Should I change this to an arrow or is the bar fine? Even if it can collie with the pins I am not sure where to place to look "good".

evanshultz:
Thanks!

For fab marks, let's use arrows everywhere and try to place it inside the part body closest to the relevant pin. We'll wait for Rene's response to get more info about that.

evanshultz:
@lorem-ipsum 
KiCad 5.1.3 will be released relatively soon. If we're able to get this done in a week or two it can probably fit into that release. If not, that's OK too, but it would be great to finally get all your hard work in the library.

lorem-ipsum:
I've now extended the silk markers for the horizontal connectors and changed their Fab markers to arrows:
Screenshot_20190714_112208

There is still the open question on what to do when pin a1 is not existing.

evanshultz:
Thanks for the continued work here!
1. Please verify silk marks are outside the part body so they're visible after assembly.
1. Fab marks should be within the part body so they won't get covered up by other components. They can overlap with the pins. So the arrow above can scoot down inside the part body.

@poeschlr 
Should the "lowest" present pin be marked, and not empty space if pin a1 is removed? Yes, I think.

evanshultz:
@poeschlr 
Can you please review and chime in on the pin marking question above?

evanshultz:
@poeschlr 
Can you please answer the question above I've asked of you?

lorem-ipsum:
As I've now used those a bit in some projects: Independent of which pin in styled as pin 1 I think that the pin 1 position (0,0 of the coordinates) should stay at the a1 position. The DIN41612 standard gives very clear definitions on where to place the connector and there a1 is relative to that. This allows to swap the connectors easily while maintaining the correct placement of the connector on the PCB.

poeschlr:
I agree with @lorem-ipsum here.

evanshultz:
@poeschlr 
I don't think we're talking about the same thing.

I have been asking you to comment on the silk marking of pin "1". I believe the "lowest" pin present on the footprint should be marked, so pin a1 if present, but pin a2 if there's no a1, etc. I have been talking about silk markings.

The comment above by @lorem-ipsum is about the footprint origin, a different question.

I suppose you could have been saying the keep the footprint origin at the phantom a1 location even if a2 is the "lowest" footprint _and also_ put the silk mark there, but it's not clear. The origin could stay at the a1 location, even if pin a1 isn't present, while the silk mark could move to pin a2.

Can you clarify?

evanshultz:
@poeschlr 
See above. Still need input.

poeschlr:
Ok i missed that question. The two options are: Tread the lowest pin as the "pin 1" (different pad style, plus markers on silk and fab) or do not include any markings (It does not make sense to mark a phantom pin)
Of these two i would prefer treating the lowest pin as pin 1.

evanshultz:
@lorem-ipsum 
So the two comments I made at https://github.com/KiCad/kicad-footprints/pull/1076#issuecomment-511236819 should be resolved, in addition to marking the "lowest" pin that exists on fab and silk layers. The footprint origin can stay at a phantom pin but please make sure a comment indicates this situation in the script. Thanks!

lorem-ipsum:
I've now changed the script such that it always marks the lowest pin (with roundrect, on fab and on silk). It now looks like this:
Screenshot_20190905_102807
Screenshot_20190905_103522

evanshultz:
Thanks again!! I've restarted the numbering:
1. I'm not sure if I was unclear, but the pin "1" silk mark has moved. See https://github.com/KiCad/kicad-footprints/pull/1076#issuecomment-503298742 where it's near pin a1. Now the footprint has the silk mark by pin b1, and it should move back near pin a1.
image
2. For a footprint like this, pin "1" is in the lower left of the footprint. (It is in the upper left of the pads.) While I fear this is painful, and I should have noticed it before, please rotate the footprints.
image
3. Also regarding the screenshot above, when there are missing pins the silk at the bottom can go closer towards the pins. And short silk lines can go just in front of the "a" row of pins, between the columns of pins, along the edge of the body.
4. Vertical connectors have odd geometry on the fab pin "1" mark. The horizontal ones look more triangular with a 1mm base and 0.7mm height, but the vertical ones are odd. Could they use the same geometry as the horizontal connectors?
image
5. Issue with the pin "1" mark placement (all three vertical F footprints):
image
6. `DIN41612_F_2x16_RowsZB_Male_Horizontal` and `DIN41612_F_3x16_Male_Horizontal` have only 0.24mm air gap between the pins. This is really tight and the voltage rating of the connector is much greater so something seems amiss to me. Please check the pin locations on those footprints. (As the vertical versions have greater air gap maybe the horizontal footprints need corrected?)
7. Let's put a `_THT` suffix on all footprints because there are SMD DIN 41612 footprints. At least, Mouser shows that https://www.erni.com/en/products-and-solutions/detail/204763/ is SMT and a DIN 41612 connector.

chschlue:
Should IEC 60603-2 at least be mentioned somewhere?
Although these connectors are commonly called that, DIN 41612 has actually been revoked years if not decades ago.

evanshultz:
@lorem-ipsum 
As you can see above, these footprints are going to be useful. Thanks again!

According to https://en.wikipedia.org/wiki/DIN_41612, `IEC 60603-2` is the updated version of DIN 41612 (who knew??). Thanks for pointing that out, @chschlue ! Unless somebody objects, perhaps keep the footprint naming as it is but add this newer IEC standard in the description? Is that OK with you?

chschlue:
Fine by me.

To be honest, I wasn't aware of that either. My standards library told me while trying to check out DIN 41612 to do a review :)

lorem-ipsum:
IEC 60603 is already mentioned in the keywords, I can also add it to the description. About the other comments:
1. I moved the marker down there as it is guaranteed to be visible after assembly. The previous location next to a1 is covered for some connectors. I would still prefer the previous location though.
2. It is on the top left, this is only a confusion due to 1. 
3. Will try to integrate it.
4. will change it
5. Oh, thanks for spotting this. Will make sure to fix it.
6. I will double check the spacing.
7. I don't think the linked connector is part of this standard, but I will add a THT suffix anyway. It does not hurt.

evanshultz:
@lorem-ipsum 
Thanks! When you have generated new footprints and uploaded them I'll take another look.

evanshultz:
@lorem-ipsum 
5.1.5 is coming so if you're able to spend time to make the updates you mentioned above we could get these footprints into the next release of KiCad. :)

evanshultz:
@lorem-ipsum 
5.1.5 is already tagged but there will be a 5.1.6. Since v6's release is probably quite some time in the future I'd love to see this make it into the library before 5.1.6 so that users will have these footprints available. Do you think you'll be able to make the updates discussed above in the next couple months?

garlick:
Exactly what I need!  Thanks for the hard work here!

lorem-ipsum:
sorry for the long quiet time. I am currently busy with a lot of other things. I want to get this (and my other PRs) forward during the holidays.

lorem-ipsum:
So I think I got everything fixed by now.

About 3: I changed the drawing routine for the silk screen lines to use the keepout functionality found in tools (and accelerated it a bit). Now every horizontal connector has silk between the pads, even those with all pins:
Screenshot_20191224_161205

Number 6 was indeed a mistake, the holes are smaller now as they should be.

garlick:
Hi, is it possible to include the H11 power (15 amp) connectors?

The straight female backplane connector aligns with the regular row a and c, pins 2, 5, 8, 11, 14, 17, 20, 23, 26, 29, 32 (row a and c are connected to the same pin).
https://www.erni.com/fileadmin/import/products/assets/DC0002065.PDF

The right angle male board connector is perhaps a bit odd as its pins align with the same numbered columns, but it has a single row more distant from board edge than regular row c.
https://www.erni.com/fileadmin/import/products/assets/DC0002583.PDF

However, I do not want to slow down this PR as it contains a lot of great stuff!    Just thought I would ask in case it's easy to add to script inputs.  I will derive from existing footprints if unavailable.

lorem-ipsum:
I also want to add other connectors, especially H11 and H15 which I am also using in some projects. However I first would like to get the style of the symbols done in this PR before adding them as they will require more and more special rules in the script.

evanshultz:
Thanks for the continued work on this! These are a huge addition to KiCad and please accept my apologies for just getting to this now.

I'll start new numbering:
1. Even if it's under the part body I prefer the silk mark near the lowest-numbered pin as well for horizontal connectors with multiple rows of pins. Which sounds like what you want too. Can you move the silk triangle? Sorry if we confused each other.
1. Pad-to-silk spacing is a bit tight. See the short silk lines between pins mentioned above (just over 0.15mm) and the silk pin 1 mark (~0.12mm). I assume this would also catch the clearance between row A and the silk line above on DIN41612_Q2_2x16_Male_Horizontal_THT has 0.2mm. What is the clearance now? Can you target 0.2mm spacing everywhere?
1. On horizontal connectors you're welcome to flip the fab triangle around so it's inside the body edge just like vertical connectors. The triangle doesn't need to point towards the pin. All up to you. I don't have a preference. Just letting you know either way is fine.
image

And here are some questions because I'm not that familiar with DIN 41612, I wasn't able to find the standard online, and I would like to know your thoughts so I can understand:
1. Are row names in the standard? Do row names vary between different manufacturers even if the standard gives row names?
1. On a family like Type F, where the rows are Z, B, and D, how do we know which pin is the lowest-number pin since row B is in the middle?
1. I notice you start each row at 0, but at https://www.erni-x-press.com/de/downloads/kataloge/englische_kataloge/erni-din41612-iec60603-2-e.pdf#page=145 the pin numbers go from 2 to 32 with only even numbers. Is this in the standard?
1. Similarly, DIN41612_D_2x16_Male_Horizontal_THT and DIN41612_D_2x8_Male_Horizontal_THT have half the number of pins but the corresponding pin isn't in the same location. Does the standard cover this?

chschlue:
B, C, D, E:


F, G:


M:


U, V:


H:

evanshultz:
@lorem-ipsum 
Do you expect to have time to wrap this up in the near future? It would be great to get these great footprints merged!

Please check out the info posted by @chschlue and ask if you have any questions. It looks like pin numbering might need a bit of tweaking.

lorem-ipsum:
Sorry for the long silence. About your points: I increased the clearance of the silk screen to .2.

About the naming: I just realised that I broke the pin naming somewhere while integrating the changes. I've uploaded a new version, it follows the pattern @chschlue has posted. I hope this fixes most of the confusion, sorry for that.

The only open question remaining from my side is which pin to mark on the silk and fab layer. Always the a0 pin, even if it is not existing (thats how it was in the beginning, also used by the few existing footprints), or the lowest numbered existing pin (which is the current status).

evanshultz:
@lorem-ipsum 
Thanks for the update! We still have time to get this into 5.1.6 and I'm willing to put in the time to get this done if you are.

The silk-to-pin clearance is just less than 0.2mm on some footprints, like `DIN41612_Q2_2x16_Male_Horizontal_THT` (oddly, on other similar horizontal connectors it is >0.2mm). I measure ~0.17mm.
image

Also here on `DIN41612_D_2x8_Male_Horizontal_THT`:
image

I checked pin numbers and row names against the above images and have the following questions:
1. Is it intentional there is not full-stuffed type D? With a row/column spacing of 2.54mm and 2.6mm pads it couldn't possibly work. I'm not asking you to add any more footprints, but just want to check that the existing footprints are correct.
1. Type F: It looks like only 3 letters are supported but the spec above shows 4. For example, `DIN41612_F_2x16_Female_Vertical_THT` has columns `z` and `d` but why is the latter not `f`? Is that correct according to the spec when some letters are not used? The footprints appear to have three equally-spaced rows/columns but I'd expect four equally-spaced rows.
1. Types Q and R are not above. @chschlue could you help us out again?

Type F also has a few things I'm not sure about:
- The origin of `DIN41612_F_2x16_Female_Vertical_THT` is not inline with column `z` while it is on the horizontal version. Is that an error?
- And on that same connector shouldn't `z2` be the origin?
- Spacing between rows/columns isn't consistent. Is that correct? `DIN41612_F_3x16_Female_Vertical_THT` is 5.08mm between columns but the horizontal version is 2.54mm between rows.

Regarding the pin marking, I agree with Rene's comment at https://github.com/KiCad/kicad-footprints/pull/1076#issuecomment-518946862 that the lowest-numbered pin that is present should be marked. So `a1` instead of `c1` on `DIN41612_R2_3x16_Male_Horizontal_THT`, for example.

And another one for you, @chschlue : does the spec say anything about how to name letters if they aren't fully stuffed? For example, for types with three letters is removing the middle letter and only using the 'lowest' letter the only possible variants? Can only the two 'biggest' letters be used? For four-letter types are these any rules or suggestions for de-stuffing?

lorem-ipsum:
I increased the clearance further, I guess the issue was that I have .2 clearance to the point defining the line. This neglects the extend of the silk line which can still reach into the clearance. I increased it by this size, hopefully it is good now. How can I get the view drawing this .2 line around the pad? I couldn't find any setting for it in the footprint editor.

The connectors I added were based on the connectors I found in the catalogues of the manufacturers listed in the first post. I don't know which ones of those are standard and if there are more which are standard (but if they don't produce them you will have a hard time buying them). Harting lists parts as extension to the norm. They follow the same structure, but I read it that they are not defined in the official standard.
* In those I did not find any fully stuffed D. As it is for higher currents I assume this is on purpose.
* Q and R are also from the catalogues. They follow C and B, but Male and Female are switched between vertical and horizontal.

About the row numbering: It is supposed to be a grid with 2.54mm spacing. Columns are numbered from 1 to 32 and row starting at "a" with increasing digits as they are needed. Some connectors skip rows to get larger clearance between the rows (e.g. the C series). A few add a row in front of the "a" row, this one is then labelled "z" (so -1 in the alphabet). The position of the a1 pin is fixed relative to the mounting holes. This is at least for the vertical connectors where the placement is important as they are used in backplanes. For horizontal ones it is fixed in terms of height above the PCB, but this is not really relevant for the footprint. Also for those the row spacing can be different as the spacing only has to match on the connector front.
* So for the F series with three rows you mentioned the labelling starts one early at "z", skips "a", has "b", skips "c" and has "d".

evanshultz:
I set the Pad Clearance in the Footprint Settings dialog to get the ring around the pad of a desired oversize:
image

I don't see any silk change on `DIN41612_Q2_2x16_Male_Horizontal_THT` compared to what I showed, above. And on `DIN41612_D_2x8_Male_Horizontal_THT` the round pads are OK but the round rect is like this:
image

Regarding letters, thanks for the information. As one example, the screenshot from @chschlue above shows Type D has letters `z`, `b`, `d`, and `f`. That doesn't match with the above but I'm not sure what is the right way to go. I am more inclined to use a spec rather than datasheets if it comes down to that. Do you see this also and can you reconcile it? Maybe I'm missing something obvious?

chschlue:
> * Q and R are also from the catalogues. They follow C and B, but Male and Female are switched between vertical and horizontal.

True. Same for S, T, U and V.
60603_t2_1
60603_t2_2

Give me a minute or two to read up on the other open questions.

chschlue:
> 1. Is it intentional there is not full-stuffed type D? With a row/column spacing of 2.54mm and 2.6mm pads it couldn't possibly work. I'm not asking you to add any more footprints, but just want to check that the existing footprints are correct.

D (and its inverted counterpart S) has contacts in rows `a` and `c`, even-numbered columns only.

> 2. Type F: It looks like only 3 letters are supported but the spec above shows 4. For example, `DIN41612_F_2x16_Female_Vertical_THT` has columns `z` and `d` but why is the latter not `f`? Is that correct according to the spec when some letters are not used? The footprints appear to have three equally-spaced rows/columns but I'd expect four equally-spaced rows.

The drawing I posted earlier is a little misleading by itself.
F (and U) is triple-row mechanically, with contacts specifed in even-numbered columns and rows `z`, `b` and `d` _or_ `z` and `b` only, bottom row `d` being empty.
G (and V) has four rows `z`, `b`, `d` and `f`, all stuffed, even columns only.
Does that make sense.

> 3. Types Q and R are not above. @chschlue could you help us out again?

See my last post.

> * The origin of `DIN41612_F_2x16_Female_Vertical_THT` is not inline with column `z` while it is on the horizontal version. Is that an error?
> * And on that same connector shouldn't `z2` be the origin?

Apart from that I suggest calling this one `DIN41612_F_2x16_RowsZD_Female_Vertical_THT` and renaming the other one `DIN41612_F_2x16_Female_Vertical_THT` because `DIN41612_F_2x16_RowsZB_Female_Vertical_THT` is the one that's actually standardized.

> * Spacing between rows/columns isn't consistent. Is that correct? `DIN41612_F_3x16_Female_Vertical_THT` is 5.08mm between columns but the horizontal version is 2.54mm between rows.

Don't forget that vertical contacts directly relate to footprint holes while the horizontal connectors allow for variations saving PCB space.

> And another one for you, @chschlue : does the spec say anything about how to name letters if they aren't fully stuffed? For example, for types with three letters is removing the middle letter and only using the 'lowest' letter the only possible variants? Can only the two 'biggest' letters be used? For four-letter types are these any rules or suggestions for de-stuffing?

Many of the connectors ERNI and Harting manufacture are simply out of spec (there is no such thing as an IEC 60603 single row B-type, for example, nor any half- and third-length connectors at all).
There are connectors like like a five-row E with only it's outer rows `a` and `e` stuffed, `b` to `d` missing or the F variant mentioned above.
Generally, the spec is meant to be exhaustive and contains no provisions for "logical" extensions.
Does this answer the question?

evanshultz:
Well... I'll go with what @chschlue says on the parts and naming. Apart from the one naming change above does everything else look OK to both of you?

Apart from that, I think only the thing that really needs to be done is moving the silk pin 1 marker. If the silk-to-pad clearance can grow a bit to meet 0.2mm everywhere that's a bonus.

lorem-ipsum:
So. I think I found the part I missed when increasing the clearance, now it looks ok for me. I also moved the pin 1 marker on the horizontal footprints back to its previous position. It always points on the lowest numbered pin.

Screenshot_20200429_095552

evanshultz:
The pin 1 markers look great. Thank you!

Silk-to-pad clearance is also good. It's close but great on `DIN41612_D_2x8_Male_Horizontal_THT` and a bit larger on `DIN41612_Q2_2x16_Male_Horizontal_THT`. But both will be OK.

Can you please address the footprint naming suggestion from @chschlue at https://github.com/KiCad/kicad-footprints/pull/1076#issuecomment-615732560?
```
Apart from that I suggest calling this one DIN41612_F_2x16_RowsZD_Female_Vertical_THT and renaming the other one DIN41612_F_2x16_Female_Vertical_THT because DIN41612_F_2x16_RowsZB_Female_Vertical_THT is the one that's actually standardized.
```

@chschlue 
Since you seem to have the specs and may be more familiar with the whole range of parts than me, would you mind taking a quick look to see if I'm a bonehead?

Lastly, because I apparently want to cause you problems and never merge this, it appears the horizontal connectors have a much larger outer silk offset then vertical ones. Vertical ones are nice and tight at 0.11mm, while horizontal are much farther away at 0.26mm. It doesn't make the footprints unusable, and KLC doesn't have a specific value for this distance, but 0.11mm would be nice if you could use that for all footprints. Compare, for example, DIN41612_B2_2x8_Female_Vertical_THT vs DIN41612_B2_2x8_Male_Horizontal_THT:
image
image

Thanks again, @lorem-ipsum ! 5.1.6 is planned for a couple weeks and it I'm very hopeful and excited these footprints will make it!

chschlue:
I have the specs and I meant to take look but forgot about this PR again.
I should be able to get at it some time this weekend.

lorem-ipsum:
I did increase the size of the mounting holes to 2.85mm and performed the rename.

About the F series: should it be for all versions? I checked ept, Harting and erni. ept and erni use large holes (1.6mm) only for the vertical connectors, not for the horizontal, Harting does not use them at all. I changed it for the vertical ones to the 1.6mm size, for the horizontal I think the space will become to limited when using 1.6mm holes. Also no one seems to produce them.

Also the gaps in the silk screen are assymetric around the holes. I checked my code and couldn't find any reason for this. Currently I am suspecting some issues in `addPolyLineWithKeepout`, but I could be wrong. I checked the pins and couldn't find any where it extends into the 0.2mm tolerance, so I would keep it at this for the moment and try to fix this later.

chschlue:
Sry, you're right.
1.6mm is only for vertical F, 1mm is correct for horizontal.

lorem-ipsum:
Ok, good. Thank you very much for spotting this.

evanshultz:
This all looks good to me, after looking at the diff and a quick visual check in KiCad. @chschlue ? Any comments?

chschlue:
LGTM

evanshultz:
@lorem-ipsum 
Thanks so much for your huge contribution here! Many other users had asked for these footprints and now the library includes a big range of them plus an updated script upon which more can be built. I really appreciate all your time and expertise to make this happen!!

garlick:
I'm super excited to use these!

aewallin:
great work!
what would be needed for H15? I am looking at plug-in 3U powersupplies and they seem to use H15 quite often, sometimes called "power DIN" maybe?

evanshultz:
@aewallin 
The script will be merged soon and then it will be a starting point to work from. @lorem-ipsum mentioned at https://github.com/KiCad/kicad-footprints/pull/1076#issuecomment-568875676 that H15 will be a likely next addition.

PR #1081, opened 2018-11-10
Create DFN-14-2EP_3x4mm_P0.5mm_EP1.78x1.65mm_EP1.07x1.65mm.kicad_mod

ferdymercury:
https://github.com/KiCad/kicad-symbols/pull/1133
https://www.analog.com/media/en/package-pcb-resources/package/pkg_pdf/ltc-legacy-dfn/05081731_C_DE14MA.pdf

image

------------


ferdymercury:
Attempted a script version too.

myfreescalewebpage:
@poeschlr do you think it is possible to script this kind of footprint ?!

Joel

poeschlr:
Such footprints are not scriptable at this point in time. (I do not expect them to be fully scriptable anytime soon to be honest)

poeschlr:
Regarding name: Maybe add the manufacturer as prefix to mark it as something different. We could also include the drawing number but i feel this is a bit overkill.

ferdymercury:
Added prefix as suggested

myfreescalewebpage:
@ferdymercury you have added LTC_ in the file but not renamed the file itself.

I would like to suggest, based on the package sheet, name `Analog_DE14MA-2EP_3x4mm_P0.5mm_EP1.78x1.65mm_EP1.07x1.65mm` for this footprint.

Joel

ferdymercury:
fixed, thanks

ferdymercury:
Fixed, thanks for the review

myfreescalewebpage:
@ferdymercury Paste on the whole pad 15 and 16 is too much. If you look at other DFN packages, DFN-8 for example, there are only some small pad just for the Paste layer. This reduce a bit the quantity of Paste. Example below.

capture

ferdymercury:
updated, thanks

myfreescalewebpage:
@poeschlr ping ? Thanks!

myfreescalewebpage:
@poeschlr no sure you have see my above ping ? Thanks!

evanshultz:
The Travis warnings are normal when there are paste or mask pads over a copper pad (so the paste and mask pads don't have the copper layer turned on). Our check script is not yet smart enough to realize this footprint is designed properly.

`DE14MA` already captures this package, so the weird double EP part can be removed. Finally, this footprint isn't even needed because https://github.com/KiCad/kicad-footprints/blob/master/Package_DFN_QFN.pretty/Linear_DE14MA.kicad_mod exists. It doesn't have the vendor name change since Linear's acquisition by ADI, but it still works. And I cleaned up the footprint at https://github.com/KiCad/kicad-footprints/pull/981 already (but didn't touch the name). Renaming is not required at this time and any symbols using this footprint would need to be assessed before making a change to the footprint name.

myfreescalewebpage:
Thanks so much @evanshultz for this response, so the footprint is not required. @ferdymercury sorry to ask you working on this, the footprint was already present ! It is called Linear_DE14MA indicated by @evanshultz . I will update the symbol PR to ask you the modifications, but at this moment, we can close this one.

ferdymercury:
Thanks for noticing!

PR #1084, opened 2018-11-10
Add RAYEX ELECTRONICS L90 Series 30A Relay

Vlad-23:
Add RAYEX ELEC. L90 Series (compatible with TE Connectivity T90, Omron G8P and other) footprints.

Datasheet: https://a3.sofastcdn.com/attachment/7jioKBjnRiiSrjrjknRiwS77gwbf3zmp/L90-SERIES.pdf

Matching symbols:  KiCad/kicad-symbols#1134

Screenshots (updated): 
relay_spdt_rayex_elec_l90
relay_spdt_rayex_elec_l90s
relay_spst_rayex_elec_l90a
relay_spst_rayex_elec_l90as
relay_spst_rayex_elec_l90b
relay_spst_rayex_elec_l90bs

------------
All major issues are fixed now. I need this type of relay in my project and was surprised that it's not already there, so, decided, why not to contribute and share it. But I can't draw my own 3d model. So is it ok to add 3d model downloaded from 3dcontentcentral (https://www.3dcontentcentral.com/secure/Download-Partner-Model.aspx?partner=TraceParts&name=INDUSTRIAL+RELAY+30A&lang=en&catalogid=7433&id=439097&parturl=https%3a%2f%2fwww.tracepartsonline.net%2fws%2f3dcc%2fproduct.aspx%3ffwsid%3d3DCC%26product%3d10-12112013-123511%26htb%3d1%26lang%3den) or similar? If not, maybe you can merge without 3d model or maybe somebody can draw it? Please take a look, if there any other issues, let me know. 
Thanks for your time.

------------


Vlad-23:
All major issues are fixed now. I need this type of relay in my project and was surprised that it's not already there, so, decided, why not to contribute and share it. But I can't draw my own 3d model. So is it ok to add 3d model downloaded from 3dcontentcentral (https://www.3dcontentcentral.com/secure/Download-Partner-Model.aspx?partner=TraceParts&name=INDUSTRIAL+RELAY+30A&lang=en&catalogid=7433&id=439097&parturl=https%3a%2f%2fwww.tracepartsonline.net%2fws%2f3dcc%2fproduct.aspx%3ffwsid%3d3DCC%26product%3d10-12112013-123511%26htb%3d1%26lang%3den) or similar? If not, maybe you can merge without 3d model or maybe somebody can draw it? Please take a look, if there any other issues, let me know. 
Thanks for your time.

Updated Screenshots: 
relay_spdt_rayex-l90
relay_spdt_rayex-l90s
relay_spst_rayex-l90a
relay_spst_rayex-l90as
relay_spst_rayex-l90b
relay_spst_rayex-l90bs

5N44P:
@Vlad-23 there's no need to add a 3D model and you shouldn't include a model not made by you for licensing reasons. 
You only need to go to the footprint property and insert the path where the correct model would be, if there was one. 

In your case it should be something like 
```${KISYS3DMOD}/Relay_THT.3dshapes/Relay_SPDT_RAYEX-L90.kicad_mod```
and similar for the other footprints. 

Just add that the correct path for every footprint and the travis error will disappear

Vlad-23:
@5N44P thank you. Now it's fine.

Need somebody to manually restart travis check for KiCad/kicad-symbols#1134 there shouldn't be any issues.
I think now it's ready to merge.

poeschlr:
The datasheet does not really give any numbering scheme for these. So might it be better to use standard relay contact numbering?

One option would be: A1,A2 for the coil, 11 for the common pin, 12 for the normally closed side and 14 for the normally open side. The second common pin is a harder nut to crack. (I am not aware of a standard that contains something like this.) Maybe use 11a and 11b.

The standard i use would even change the pin 11 to pin 13 for the type where you only have a normally open contact instead of double throw.

---


Your drill diameters do not agree with the datasheet.

The courtyard clearance must be 0.25 all around (rounded to nearest 0.01 grid)

At least two dimensions are not really specified in the datasheet. Where did you get the info on them?

relay svg

Vlad-23:
@poeschlr thank you for the review.

1. Courtyard clearance should be correct now.
2. I have applied contacts numbering scheme as you suggested. But I don't really care about numbering in this case, it just should match with symbols. I also think that changing contact 11 to 13 would add a mess, but if it's needed I can renumber it this way also. Just want to keep numbering simple.
3. Dimension that are not specified in the datasheet, I have measured by myself. But you can also check it on other relays of the same type, they have pretty much the same dimensions.
    - https://www.te.com/commerce/DocumentDelivery/DDEController?Action=showdoc&DocId=Data+Sheet%7F1308242_T90%7F0214%7Fpdf%7FEnglish%7FENG_DS_1308242_T90_0214_T90.pdf%7F6-1393208-8)
    - http://www.symmetron.ru/suppliers/omron/files/pdf/omron/G8P.pdf
    - http://images.100y.com.tw/pdf_file/29-SLA-12VDC-SL-C.pdf
4. I have made drill diameters slightly bigger because I wanted to fit easily other manufacturers relays with thicker contacts (please check datasheets mentioned above). Just wanted to make this more versatile. I really thing that hole shouldn't be really tight. When it's soldered it don't make much difference.

So thank you for your time and please check this again now.

poeschlr:
The footprint is specialized to one component. If you want it to fit bigger leads then choose that component as the base for your footprint. (Part specific footprint must precisely match the datasheet assigned to them.)

And yes i am aware that having pin 11 change to 13 creates more work. The standard simply is defined that way. (I do not agree that it creates a mess. Both the symbol and footprint will be specialized for the NO vs SPDT part anyways. I bet if we do not follow the standard fully i will get angry mails about that.)

One strange result of following the standard however is that it makes it impossible to use a symbol for a NO relay on the footprint of a SPTD one while the NC symbol can still be used with it. (I would however advice against using the NC or NO symbol on a SPDT footprint anyways. Meaning i am kind of sad that NC pin numbers do not change compared to the SPDT version making it possible for users to use such a hack. Again something where i expect an angry mail similar to "why can i do this with NC but not with NO? followed by some swear words and name-calling.")

Vlad-23:
@poeschlr done, as you suggested. Please check updated screenshots. For L90AS I've changed 11A and 11B to 13A and 13B, not sure is it acceptable to you. Maybe you can provide a link to your standards, just curious. Drill holes diameters also corrected to match the datasheet.

Rules are rules! Who am I to argue with that. ;-D
Thanks for your time.

P.S. Maybe you can also review matching symbols? Or you work only with footprints? Maybe you could help with that also, I'm not sure symbols are all perfect without bugs.

poeschlr:
I just noticed you have silk on top of the fab outline. The silk outline must be outside of the nominal body dimensions (But not too far out. This is done to have it useful for manual assembly and also visible after assembly.)

The dimensions of the footprint look good.
Screenshot from 2019-03-24 16-53-41

Vlad-23:
I've checked few presented relay footprints and made the same way. Check it out please.

poeschlr:
Thanks

PR #1085, opened 2018-11-13
Connector: Add B6A-PCB-45 from IHI connectors

antoniovazquezblanco:
https://lugsdirect.com/PDF_Webprint/B6A-PCB-45-XX(-X).pdf

https://lugsdirect.com/PDF%20Documents/NOTESonPCBFootPrintLayouts(SMT-SMD-Mode).pdf

screenshot from 2018-11-13 10-15-41
screenshot from 2018-11-13 10-16-05

------------


poeschlr:
The small pads within the large one should only have paste assigned (No copper, no mask no pad number)

Is there a suggested footprint for this part somewhere (How did you determine the pad size? How did you determine how much paste is needed?)

The courtyard clearance is a bit wrong.
The width of the part body is wrong.

screenshot from 2018-12-31 23-18-48

antoniovazquezblanco:
> The small pads within the large one should only have paste assigned (No copper, no mask no pad number)

Fixed

> Is there a suggested footprint for this part somewhere (How did you determine the pad size? How did you determine how much paste is needed?)

The only documentation I've found is the one linked above. A suggestion from the manufacturer in that docs: The typical edge fillet on the IHI Solderable terminal in SMT mode is about .025” rad. (varies with amount of solder/paste used) so we suggest that the copper pad showing through the masking would be .050” total wider than the physical contact footprint to accommodate the .025” all around.
Also I have sent the existing footprint to manufacturing and received no bad feedback on the amount of paste used. I am waiting for some boards to arrive and test the result. Any wisedom from you more experienced guys is very very welcomed!

> The courtyard clearance is a bit wrong.

Courtyard is 0.5mm from the nearest SilkS line. Should I ignore SilkS for the courtyard?

>  The width of the part body is wrong.

Fixed

poeschlr:
Courtyard area does not take silk into account at all. It only cares about the nominal body dimensions and pads. (As we use nominal body dimensions for the fab outline this can be used as a reference.)

poeschlr:
You mentioned that you have send this to production and are willing to test it. Did you come around to doing that?

antoniovazquezblanco:
Due to space constraints in the design I've sent into production, pad surface was reduced below manufacturer recomendations and a special soldering job was required for the component. The result was good but the manufacturing process was a bit faulty. I could not test this final design myself and I am not sure if I will be able to do so in a future version of this device.

poeschlr:
Thanks for the feedback. I think we can use the footprints as they are.

antoniovazquezblanco:
Thank you very much!

PR #1091, opened 2018-11-16
Pulls/hanrun hr911105a

karlp:
Adds a footprint for HanRun HR911105A magjack.

HanRun's own website is not always very good, pdfs widely available. http://www.kosmodrom.com.ua/pdf/HR911105A.pdf
symbol/3d model linked in a separate commit, but is unlicensed for distribution, so only provided for context and information.
export-hr911105-model
screenshot from 2018-11-16 21-50-39

------------


karlp:
(it would be nice if all those travis checks were available to peruse _before_ you file and get failures.  It's _very_ unclear how to get your way to the very bottom of the klc page, after you even find that that's what's being referred to)

What's the rationale for having parts with holes having zero poitn on "pad 1" but surface mount parts on the centroid?! that jsut seems completely arbitrary

What's the point in having step models be supported if travis is rejecting them?
What's the point in having support for positioning models if travis is going to reject them?

The rest I can fix.

karlp:
updated and fixed all.

For F7.3, it's simply not appropriate to make pad1 rectangular here.
the 3d model checks are not relevant, they're in the second UNLICENSED commit, only included for context.

karlp:
The objection to a square pin is just that a) it's a polarized footprint, there's no chance of mixing anythign up, and b) it takes up more space in a tight area.  

I'm completely ok with a square pad for symmetrical footprints, but it seems completely counterproductive for such a strongly keyed footprint.  I'd suggest the original amphenol footprint is wrong, not something that should be copied here.

myfreescalewebpage:
@karlp I disagree with you, pin 1 is usually indicated with a square pin for many parts, connectors and other, even if the part is not symmetrical.

Then, KLC requires pin 1 to be rectangular, that's a wanted practice for this library: http://kicad-pcb.org/libraries/klc/F7.3/ so the Amphenol connector is conform to our guidelines.

poeschlr:
Square pin 1 will make debugging a board with this connector much easier.

One additional note: the pads look awfully small. You might want to increase them to get a bit larger annular rings.

poeschlr:
You currently have non valid 3d model settings. The footprint should not use rotation or offset settings. The 3d model (should it ever be provided) is to be aligned correctly using a CAD program.
The model must be set to wrl file ending. KiCad does replace the ending with step should it need it.

Rotate the footprint by 180 degree such that pad 1 ends up nearest to the top left corner.
Add pin 1 markers on fab and silk layers. (Again not only for ensuring correct insertion of the part but for debugging purposes. But also to keep the rules simple -> easier for us maintainers.)

The fab outline must represent the nominal dimensions of the part. The silk outline must be placed outside it. (But not too far away such that it still is useful for aligning the part correctly.)

The silk outline should not go across pads. In fact there should be 0.2mm clearance between the pads and silk.

A courtyard outline must be provided with 0.5mm clearance (for connectors) relative to pads and nominal body size (whatever results in the larger area) This courtyard outline is allowed to be contoured for the part. (But a simple rectangle will also be accepted.)

rj svg

karlp:
This isn't abandoned, it's just pending the symbol work: https://github.com/KiCad/kicad-symbols/pull/1146

poeschlr:
There is a change request for the footprint that was not fulfilled in a year ;). If that is not the definition of abandoned then i don't know what is.

karlp:
The last comment on it a year ago was someone asking you for clarification, "@poeschlr can you have a look to this PR to get a decision, particlularly at my comment #1146 (comment) ? Thanks." 
That would seem to be abandoned by you, not by me.

But yes, this is abandonded now. I'm not spending more time on these hoops, the upstreaming value proposition is far too slim.

PR #1092, opened 2018-11-16
Pulls/hlk pm01

karlp:
(commit messages are in commits, you probably want to drop the second one, that handles the unlicensed 3d model)

export-hlk-pm01
screenshot from 2018-11-16 22-23-25

------------


poeschlr:
The 3d settings are not correct for the official library. You need to setup the footprint for zero offset, no rotation and scale 1.

The datasheet link in the description does not work (I get "service not available")

I used this to check the part. But it would be preferable if there is a datasheet on the manufacturers website. It also does not include details about the lead diameters. So i can not check if the hole diameters are correct.
https://img.filipeflop.com/files/download/Datasheet-HLK-PM01.pdf

The courtyard clearance should be 0.25mm all around
The outline dimensions do not really agree with the datasheet i found.
ac_dc svg

The pin 1 markers you used do not quite agree with our normal standards. Look at other footprints in the same lib for inspiration on how you can make one. Especially `Converter_ACDC_MeanWell_IRM-10-xx_THT`.

karlp:
1. The 3d model is completely unlicensed, and only included for context, as indicated in the commit message, if you'd prefer I not include this, no problem, I can drop those commits in future PRs
2. I cannot do anything about hlk having a website outage. hlk doesn't/didn't offer any pdfs on their website, or I would have linked to them, I can make no claims about the accuracy of any arbitrary pdf you found.
3. I find it difficult to follow your pin 1 marker guidance.  I can definitely move the f.fab layer arrrow inside as is often done, but the silk straight line is very inconsistent.  I copied from the meanwell IRM 02 / 03, but the meanwell 05/10/20 have a different style altogether, which is correct?.  The KLC guidelines for F5.2-2-c just say, "For connectors, a small arrow indicator drawn next to pin-1 should be used" which is extremely open to interpretation.  Please update the guidelines if you have you hard rules on this, there's very little I can do ahead of time with such limited guidance.
4. clearance, I can change that from 0.21 to 0.25, sure, I missed that magic number in the KLC earlier, but I can see it now.

poeschlr:
Please correct the 3d model settings.
Correct the courtyard clearance.

The website is back up. It has more info meaning i found that not only the outline but also drill sizes are wrong. (hole size should be lead diameter plus 0.2mm)
ac_dc svg

pin 1 marker:
your marker is on the top of the outline right now. Move it to the left side like in the one i linked. (This moves it much closer to the actual position of pin 1 and is more closely in line with industry standards)

karlp:
ok. a) fixed dimensions, I don't know how those were wrong, I presume from using maths operations in the windows, with a non-zero origin, anyway, they were wrong, thanks for catching that.
b) dropped 3d models, it was a separate unlicensed commit for context only, oviously it was a distraction instead of helpful, I won't do it again :)
c) courtyard and pin sizes changed per requirements.

Thanks for bearing with this, I've got say it's been incredibly time consuming to try and get this right.  Are there any tricks or helpers for making the courtyard exactly 0.25 outside an existing box?  It's incredibly tedious resetting origin points and editing lines continuously.  I very much _want_ all my parts upstreamed wherever possible, but it's wildly time consuming at present.  Certainly it will get faster with practise, but I feel like I must be missing easy helpers.

poeschlr:
The 3d model settings of the footprint still need to be filled out such that it is possible to add one later without needing to change to footprint.
(It is not required to add the models them selfes. Only fill out the footprint settings.)

> Are there any tricks or helpers for making the courtyard exactly 0.25 outside an existing box?

Yes the move exactly tool. (crtl + m)

karlp:
What do you mean, I have to fill in a fixed file name?  that's... helpful?

If move exactly is the best tool, then there's still room for improvement....

karlp:
How does one even put in a file that's not real?  Am I expected to edit the file directly to insert that? I can't see anyway of puttin in some blank template in the UI at least.

karlp:
I believe I've done this correctly now?  I would _strongly_ suggest that kicad itself simply attempt to find a file with the same basename if no explicit part is specified.  Further, given that a git commit is required to add the model file _anyway_ it seems an unnecessary complication to try and fill in the name for that file ahead of time.  It's a line of text in a structured text file, git will work it out.

Thanks for the links, some interesting stuff.  And yes, the snap to endpoint is very useful, I don't think I would have managed if I wasn't on nightlies :)

karlp:
andddd one more time.  wrl? really? we support step, like everyone uses, but the footprint _must_ be wrl?  What's the purpose of all the step files in the repo then?

karlp:
and again!  I _reallllly_ wish the https://github.com/KiCad/kicad-library-utils repo, at _least_ the "check my filez y0" part was included the footprint library directly.

poeschlr:
Why should sourcecode be mangled in between data? That would just create a maintainance nightmare. It would also make reuse of the source code harder.

karlp:
I dunno, why are "checkpatch" tools ever included in the repo for the patches?  *looks at every other opensource project that applies checks on patches*

karlp:
it could be a git submodule even if that's easier to accept.

poeschlr:
wrl is used as the file ending in the footprint because wrl is simply better for rendering within kicads viewer. (It makes nicer pictures.)
The step exporter (both kicad internal and stepup) replace the file ending automatically.

The best solution would have been not to include any file ending at all. But we did not have the luxury to start from a clean slate.

poeschlr:
Now lets turn this around. What would be our benefit of moving the script into this repo? How is the result superior to having two repos? There needs to be a lot of benefit for us to invest the required time to change this.

I can tell you a few important downsides:
- The script repo has common assets used for different places.
- If somebody wants to reuse the scripts for something else they do not really want to download the footprint repo. (Or worse if the footprint repo includes the scripts in it, then the 3d model repo would need to do the same. That one is a few GB in size.)
- Separating the script repo means we have the issues and pull requests separated. (Working on the footprints requires a completely different skill-set compared to working on the scripts)

Submodules to me are not really a solution. Keeping the submodule pointers pointing to the correct commit is extra work. (for us maintainers to keep the footprint repo pointing to the correct script commit. And for users who need to update the script part anyways. So where is the benefit to using git pull in a separate local repo?)

karlp:
process points aside, I don't believe there are any remaining technical/style issues with this PR?

poeschlr:
Looks good now.

For reference the updated dimensioned drawing of the footprint.
screenshot from 2018-12-20 21-33-38

Issue #1045, opened 2018-10-23
Fiducial inclusion in .POS file


thomasonw:
After  migrating to KiCAD v5, the board houses have been noting that Fiducials are not included in the .POS file.  I am able to point them to the .RPT file for the details, but is there a reason why they are not included in the .POS file as well?

I have noted that by manually editing the footprints and selecting 'Surface Mount' vs. 'Virtual' as the Placement Type will force the including of Fiducials in the .POS file.  Should be the default in the fiducial footprints?

(Perhaps also related --  [https://github.com/KiCad/kicad-symbols/pull/653](https://github.com/KiCad/kicad-symbols/pull/653) )

evanshultz:
So the behavior changed between v4 and v5? If so, that may not be a library issue. (A library change may be the solution, but perhaps not the source of the issue.)

I suggest you post at https://forum.kicad.info/ and/or https://launchpad.net/kicad for "how to" and possible bugs. This repo is only about the library, and we don't know your issue involves the library.

thomasonw:
Actually, I do not know.  I skipped over v4   

I do know V3 (2013-07-07) did indeed include added fiducials in the .POS files.

thomasonw:
@evanshultz  -- OK, comment back from launchpad:  _"I believe this is a library issue these are made as virtual footprints, but should be SMD if they should show up in the pos file."_

[https://bugs.launchpad.net/kicad/+bug/1799601](https://bugs.launchpad.net/kicad/+bug/1799601)

antoniovazquezblanco:
I made it work by editing every fiducial to be locked (in order to be able to update my board without them dissapearing) and surface mount (to make them available at the pos file) in my PCB.

![screenshot from 2018-11-22 10-00-16](https://user-images.githubusercontent.com/304193/48892110-8107fc00-ee3d-11e8-9c91-3541825882c7.png)

evanshultz:
OK. So you didn't find another solution but to change their type to SM.

antoniovazquezblanco:
Yes, that is it.

antoniovazquezblanco:
Should we change this in the library?

evanshultz:
Yes, otherwise users will run into this issue again and again. I created https://github.com/KiCad/kicad-footprints/pull/1104 and would like to see it merged for 5.0.2.

PR #1104, opened 2018-11-22
Updated fiducials to Surface Mount

evanshultz:
Fixes https://github.com/KiCad/kicad-footprints/issues/1045. Also see https://bugs.launchpad.net/kicad/+bug/1799601.

Fiducials do not show up in the POS (pick n place) report because they are Virtual type and to show up there they must be Surface Mount type. This is an apparent regression because, while the type has never been correct in KiCad v3 they appeared in the report (we don't know about v4). At any rate, they should be in the POS report so at least for now this is the solution.

Changes:
- Made all Surface Mount type
- Removed old syntax ("~" for pins without number, "(at 0 0)")
- Simplified keywords
- Removed duplicated German from description
- Removed excessive precision
- Set line widths as specified in KLC
- Converted double Cu circles to a single one

Some of the copper are drawings and not a pad, but I think the chance of an issue is extremely small so I left it alone. At a later date these could be converted to a custom pad shape if it's important.

Also, some of these footprint names are not intuitive and just poor. But... can't change them now.

------------


evanshultz:
A second fab layer ref des has been added.

Other CI errors:
1. I must assume the silk fiducials work now so I'm loath to reduce silk line widths because the camera may have a harder time to locate them.
1. I've also not added a 3D model reference but I can if asked since I suppose there is something that could be shown in the 3D viewer.
1. Is a courtyard required on fiducials that won't need rework? Maybe to maintain placement clearance?
1. On some footprints there is nothing on the fab layer but I think that's OK as they don't need anything.
1. Silk/pad intersection is the silk on exposed fiberglass (the pad is removing mask) so that's fine.

antoniovazquezblanco:
Fixes #1045.

Explanation: Using keyword to close the issue when this is merged.

antoniovazquezblanco:
I had a look at this.

1. Agree
2. Agree
3. I would add a courtyard to tell the user to avoid placing anything inside the required perimeter for the camara to detect it.
4. Agree
5. Agree

Some of those fiducials are massive btw.
Could you please explain to me why there are fiducials in bottom layers with silk refdes in the top ones? Shouldn't bottom layers use mirrored top layer fiducials? Why do those exist?

Thanks

evanshultz:
OK. I'll add courtyard and then we can merge if all else looks good.

evanshultz:
Courtyard is added. I could add a fab circle or square to ones with no fab drawing so that the user, looking at the fab layer, would be able to see something was in that area. But I think it's a minor thing. These look fine in the 3D viewer as they are and our check script should be smarter to know if a pad does not have copper it's OK to have silk there (but we can ignore for now). Anything else?

I have no idea why all of these fiducials exist or why they're built as they are. I lean towards not changing things, but I agree with your questions and can't explain any of it. I've always used much smaller fiducials that are just a copper dot with a slightly bigger mask opening around it, with all elements on one side; I then mirror the fiducial when I needed one on the bottom side.

Note that IPC 2221A recommends a circular fiducial clearance that has a radius three times a circular fiducial (copper? silk?) radius. The minimum is double the radius. This is according to SMEMA (see Figure 5-6). I don't have access to 2221B so I can't say if things have changed.

antoniovazquezblanco:
Merged as is. Opening an issue to be resolved in the future about those rare fiducials.

Thank you very much!

PR #1106, opened 2018-11-24
Murata BNX025 EMI-filter

aewallin:
Footprint for Murata BNX025 EMI-filter

1. datasheets:
PDF Catalog, page 223/245
https://www.murata.com/~/media/webrenewal/support/library/catalog/products/emc/emifil/c31e.ashx?la=en-us
recommended land-patter on catalog page 229/245
note same land-pattern for BNX022, 23, 24, 25 - I am not sure how to best name the footprint - it matches those four Murata filters.

futher specs e.g. here:
https://www.murata.com/en-us/products/productdetail.aspx?cate=luNoiseSupprFilteBlockType&partno=BNX025H01%23

2. screenshot attached.
filter_bnx025

3. No symbol or 3D model is available yet, AFAIK.

4. check_kicad_mod.py runs with only the 3D-model error (F9.3)
------------


aewallin:
symbol pull-request:
https://github.com/KiCad/kicad-symbols/pull/1168

poeschlr:
A quick review (only looked at the screenshot):
- The pin 1 marker is a line parallel to the pad if the pad is outside the part outline. (See any soic footprint for inspiration)
- Add a pin 1 marker on the fab layer (example a chamfer on the nearest corner)

aewallin:
pin1 indication on silk is now a line.
pin1 indication on f.fab is a chamfer.
silk now has better outline of component
pad shape offsets set to zero (traces will snap to pad center)
not sure of courtyard is OK or not? review needed.
bnx025

poeschlr:
Is it on purpose that you have more silk to fab clearance on the right side than everywhere else (At least the screenshot looks like that)

poeschlr:
Regarding courtyard: If you are motivated than you can redesign it to hug the part more closely. This will be a bit more work required from your side and also increase the load on the DRC checker.
The payoff are fewer false positive violations in designs with high part density.

aewallin:
I don't plan on improving the courtyard manually. 

If I find time I would work on a better auto-courtyard script that can follow the outline of a component.

aewallin:
close and re-open in order to get travis to run (it was stuck)

poeschlr:
The outline on the fab layer does not quite agree with the datasheet. It is also not centered.

How did you determine the pad sizes and positions? (There is no suggested footprint in the datasheet)

aewallin:
suggested land-pattern, page 227 of the first link above. Suggested GND-areas/vias not drawn in the footprint..
bnx_landpattern


I will check the f.fab dimensions next.

aewallin:
after revision, some dimensions:
bnx_dims

the need for GND-zones and GND-vias according to the recommended land-pattern can be discussed... but the suggested Vias are quite far away from the footprint in Y-direction IMO.
Pad clearance could maybe be increased as per land pattern suggestion?

aewallin:
Also Murata has models BNX 022,23,24,25. Two of them 22+23 are lower in height 3.1mm, while two are higher 24+25 are 3.5mm.
So we could name this BNX02x (where x=2,3,4,5 is assumed) - but in principle two different 3D models would be required.

poeschlr:
I think we could add two footprints here. (If you are motivated) One as is now plus one with the zones pre defined and vias added. (The courtyard area would still stay relative to the current footprint.)

For the record i double checked your dimensioned drawing in freecad and it looks good for the land pattern.
The courtyard clearance is at 0.5mm instead of 0.25mm

aewallin:
Thanks for the comments!
- I changed the courtyard clearance to 0.25mm
- I started on a "_GNDVias" version (is that name OK?)
- The via hole diameter is not given in the datasheet, 0.8mm is a guess
- The via position are not given, from the drawing and symmetry I am guessing one via is centered on the CG-pad and the others are on a 4.5mm grid.
- I can continue with the GND-zones if I first get some feedback on this sketch. How far to extend the GND-zones is a bit arbitrary (no dimensions in datasheet).
bnx_gndvias

poeschlr:
I would say GNDVias is a bit overly specific. (Paints us in a corner if we ever have a package that needs similar handling but instead of using the ground pin it is on some other pin.)

If the large zone and vias are only there for thermal reasons then i would go with ThermalVias like we always do.
If the zone (and vias) is there to manage current flow or for EMC shielding then we might want to think of a different name.

---

via diameter and positions are good guesses. I will accept anything here that does not look completely out of place. (I can only hold you to the datasheet. If no dimensions are given then you are free to choose.)

---

Yes silk should be removed on top of vias as it might very well be the case that users do not have the luxury of getting them tented.

---

the extend of the zones is another thing where i would except nearly anything. I think a good endpoint is such that the vias are inside the zone you define by a reasonable margin.

aewallin:
Here's a variant named "ThermalVias" with some approximation of the datasheet GND-plane.
I'm not sure if it's worth to reproduce the exact shape drawn in the top right corner of the datasheet figure.

pad2 has different suggested clearance values in X and Y directions. Hopefully specifying the larger clearance (Y downward 0.8mm) does not give DRC errors in X-direction (left) where the clearance-zone overlaps wit the top pad4.
thermalvias

poeschlr:
The clearance around pad 2 is too large. One gets a DRC error with the footprint alone as it is now.

poeschlr:
Looks good to me. Thanks for another great contribution.

PR #1112, opened 2018-11-25
OCXO OH300

aewallin:
ConnorWinfield SMD OCXO model OH300

Datasheet: http://www.conwin.com/datasheets/cx/cx282.pdf

Screenshot:
oh300

Symbol PR: https://github.com/KiCad/kicad-symbols/pull/1174

------------


poeschlr:
Thanks for this contribution.

Sadly i can see some issues with it that need to be fixed.
The datasheet calls out a keepout area for traces on the "top layer" (copper layer nearest to the part)
It also calls for a copper zone in the layer below it.

Kicad does not directly support adding keepout areas to the footprint. So we can only add notes to it to indicate this. Read http://kicad-pcb.org/libraries/klc/F4.5/ for more details.

The pin 1 indicator on the fab layer is missing. I suggest a chamfer on the nearest edge.

Rotate the footprint by 90 degree clockwise to get pin 1 to the top right corner.

The courtyard clearance should be 0.5mm for crystals (+/-0.005 mm to get it on 0.01mm grid)

There are also some dimensions wrong. See dimensioned drawing for more details.

oscilator svg

aewallin:
- missing the keepout (I don't know how to draw this)
- KLC errors on silkscreen intersecting bottom GND-pad
oh300

poeschlr:
We can not add a bottom pad to get the result explained in the application note. We do not know if this footprint is used in a two layer or 32 layer pcb. This is why a note on the user layers is the best option.

What part of the linked KLC (tutorial) is unclear to you regarding the keepout? Maybe we can help you out there.

aewallin:
regarding the tools: it seems the way to draw on Dwgs.User or make comments on Cmts.User is to first add the graphic/text on some other layer and then edit the properties with "E" and change the layer to Dwgs.User or Cmts.User?
tools

aewallin:
The footprint now has a zone drawing on Dwgs.User and a note about keepout on top and GND on 2nd layer.
oh300

poeschlr:
We typically do not use circles as pin 1 markers. Make a chamfer on the corner nearest pin 1.

Rotate the part by 90 degree clockwise to get pin one to the top left corner.

The courtyard clearance is not yet quite right in y direction.

screenshot from 2018-12-05 19-02-09

aewallin:
rotated, chamfered f.fab, fixed courtyard clearance:
oh300

poeschlr:
Looks good, thanks.

PR #1114, opened 2018-11-26
Fix 12BH611-GR pad numbering

### LINKED ISSUE #1113: Footprint 12BH611-GR: wrong pins
I used 12BH611-GR as a footprint for a Device/Battery. 

12BH611-GR, from a top view, having the snap on the right side:

- Pin 1 is used for the negative side (upper)
- Pin 2 is used for the positive side (lower)

But in order to work with Device/Battery, it should be:

- Pin 1: positive side (lower)
- Pin 2: negative side (upper)

I went ahead and changed it in my local copy, and verified that everything worked the way it should after swapping the pin numbers.

frarees:
Fixes #1113


evanshultz:
Thanks!

We need pin 1 at the origin, so can you please move the footprint? There are a number of line widths that CI caught which you can fix if you like but are not required.

frarees:
Went through those. Didn't know about those rules, but Travis helped out 👍

evanshultz:
Cross-posting.

I can merge this if you like, but if you care to fix line widths that would be nice. You could also unrotate pin 1.

Note that there is no 3D model so moving the footprint isn't a problem for our library.

frarees:
Going to fix the rotation of pin 1. What is the issue with line widths?

evanshultz:
Here is the current footprint:
image

Great! Thank you!

There are actually still a couple things that would be nice to do:
- Move the value below the footprint.
- Move the silk ref des above it.
- Put the fab ref des in the middle (currently it's under the silk ref des).

Would you mind also making those fixes? Certainly the pinout is the critical one so if that's not possible I will merge as is.

evanshultz:
Fabulous! Thank you!

PR #1115, opened 2018-11-26
Added Toshiba isolator SOIC footprints

evanshultz:
Scripted footprints needed by https://github.com/KiCad/kicad-symbols/pull/1177. See lots more info at https://github.com/KiCad/kicad-symbols/pull/884.

- https://toshiba.semicon-storage.com/info/docget.jsp?did=11791&prodName=TLP185
- https://toshiba.semicon-storage.com/info/docget.jsp?did=14111&prodName=TLP185(SE
- https://toshiba.semicon-storage.com/info/docget.jsp?did=12858&prodName=TLP291-4

Since the outstanding PR above already is using the footprint without indicating skipped pads I left the footprint name alone. I suspect it's specific enough with the 4.55mm width that this is OK.

Script PR: https://github.com/pointhi/kicad-footprint-generator/pull/227.

image

------------


antoniovazquezblanco:
https://toshiba.semicon-storage.com/info/docget.jsp?did=12858&prodName=TLP291-4 suggest the following land pattern:
screenshot from 2018-11-27 10-32-22

I have modified pad 8 to use the suggested measurements. Those look very odd to me. See below:

screenshot from 2018-11-27 10-32-37

I am not sure about wether to follow manufacturer suggestions or not in this case.
The other two datasheets do not specify a recommended land pattern and I am ok with your footprints.

evanshultz:
These are all built using an IPC 7351B generator so we disregard any recommended footprints. We could make a vendor-specific footprint in this case, but I'd prefer sticking with the scripted versions.

antoniovazquezblanco:
Agree with you

PR #1117, opened 2018-11-28
Add D52M

DanSGiesbrecht:
Adding the "M8" form-factor D52M ANT module by Garmin Canada (Dynastream Innovations).

frontside

backside

------------

  - [Datasheet URL (warning - download)](https://www.thisisant.com/assets/resources/Datasheets/D00001687_D52_Module_Datasheet.v.2.3.pdf)
#### Footprint:
footprint
#### Copper pads:
copper

copper-datasheet
(datasheet, page 22)

#### Solder paste:
paste

paste-datasheet
(datasheet, page 23)

  - **Violating F6.2**: Travis-CI thinks that the center of the pads represents the center of the part. This is incorrect. The anchor is correctly placed at the center of the module.
  - **Violating F9.3**: A 3D model has not been created. I plan on contributing this in the future.


evanshultz:
- A 3D model still needs to be referenced so it will show up in the 3D viewer automatically once it's added.
- https://www.thisisant.com/assets/resources/Datasheets/D00001687_D52_Module_Datasheet.v.2.3.pdf opens in a tab for me. I'd be happy with that instead of a product page link.
- Can you add a description in the description instead of just a link? Maybe move some keywords.
- Change the text in the keepout to "Copper keepout" as mentioned at http://kicad-pcb.org/libraries/klc/F4.5/.
- The round pads should not have the paste layer turned on. And you can turn off the paste margin parameter for each pad.
- I don't see anything about the mask design so I guess it should be round for self-centering, but how did you determine their size? It's a bit odd to have the square paste pads overlaying the copper-mask barrier...
- The mask pads could be actual pads instead of polygons. That's a bit more compact representation but if this way makes it to the Gerbers (please verify for me!) it's OK do it this way.
- Not required, but you could move the fab ref des from behind the pads.
- How did you come up with the name of this footprint? I would have thought `Dynastream_D52M`, and then optionally the dimensions you added, would be good.
- Dimensioned drawing:
image

DanSGiesbrecht:
Hi @evanshultz 

Thank you for the notes. Sorry about the delay getting to this. Note that the datasheet link has been updated to "Garmin", rather than "Dynastream". Here's the updated [link](https://www.thisisant.com/assets/resources/D00001687_D52_Module_Datasheet.v.2.3_(Garmin).pdf).

I'll change the name to: `Garmin_M8_9.8x14.0mm`, and work through your other suggestions as well, once I have some free time.

DanSGiesbrecht:
@evanshultz thanks again for the review. The following changes have been made:

- Added 3D model file (does not exist yet).
- Updated link.
- Elaborated description.
- Changed keep-out text.
- Turned off SMD pad paste layer, removed paste margin.
> I don't see anything about the mask design so I guess it should be round for self-centering, but how did you determine their size? It's a bit odd to have the square paste pads overlaying the copper-mask barrier...
- I agree that it's a bit odd to have square paste over soldermask, however, it's manufacturer-recommended for this part:
  - The solder paste design comes from the datasheet (figure 16, page 23).
  - The manufacturer recommends a 1:1 ratio of module LGA pad to SMD footprint pad (both 30 thou), and no soldermask expansion. I left the soldermask expansion as 0 (default), so that designers may set it to be whatever their board-house requires for good soldermask alignment. Unfortunately, the 1:1 ratio is not described in the datasheet.
> The mask pads could be actual pads instead of polygons. That's a bit more compact representation but if this way makes it to the Gerbers (please verify for me!) it's OK do it this way.
- The solder paste polygons do successfully transfer to the gerbers. :smile:
> Not required, but you could move the fab ref des from behind the pads.
- If it's alright, I'd like to leave the fab REFDES in the center, for clean assembly drawings.
- Fixed the footprint name.

I've been wondering: how do you generate those dimensioned footprint drawings??

DanSGiesbrecht:
@evanshultz please hold on merging until the naming of the footprint is resolved: https://github.com/KiCad/kicad-symbols/pull/1184#issuecomment-461114465

evanshultz:
Rene posted the process to dimension footprints this way at https://forum.kicad.info/t/how-to-check-footprint-correctness/9279.

evanshultz:
My only remaining question is about the mask design. If Garmin says not to use any mask expansion, then you can make it 1nm but do it at the footprint level instead of per-pad. We don't want to board house to swell the mask if the footprint shouldn't have this done, though 1.5mm pitch is big enough to probably not get into trouble with a typical swell. Can you also link here to the doc that explains this for historical reference?

DanSGiesbrecht:
Good call, I'll update this for the footprint (rather than per-pad).

Unfortunately, there isn't a reference document which explains this. The knowledge of mask-expansion is based on internal conversations with the module manufacturer. I can request this to be added in a future datasheet update.

DanSGiesbrecht:
@evanshultz renamed the file to align with the symbol PR conversation, and changed the global footprint soldermask expansion to 0.001mm. I think this should be good-to-go now. Thank you for all the input.

evanshultz:
You should be able to go to "0.000001". That's 3 more sig digs than you have now and the limit of KiCad's internal precision. I am pretty sure it is 1nm but I could be wrong. The way is it is so small as to be within the tolerance of manufacturing, so going smaller won't actually help, but if we use the smallest value that KiCad can represent (instead of some tiny but intermediate value) that tells us the intent is to make the mask exactly the same size as the pad. We have done this elsewhere in the library.

So... long winded but I think it makes the most sense to go to the limit. OK for you?

DanSGiesbrecht:
Done!

evanshultz:
Thanks!

PR #1122, opened 2018-11-29
Added USB A vertical connector

kajusK:
Added a vertical USB A connector made by Molex.
Datasheet: https://www.molex.com/pdm_docs/sd/1050570001_sd.pdf

2018-11-29-181858_1920x1080_scrot


aewallin:
Unofficial review:
- most other USB-footprints have an unbroken and continuous F.Fab outline 
- pin1 is indicated on F.Fab with a small arrow. square vs. round pin1 pad seems to vary...
- dimensions look OK to me

poeschlr:
I took a closer look. I found a few more things to be considered.

The courtyard clearance should be 0.5mm (rounded to nearest 0.01mm grid point)
The fab outline should represent the nominal body size. (without flange it is 13.1x5.72mm, or with flange it would be 14.63x7.25mm. I would say that the outline should be without flange in this case. (The flange is so far away from the board that it really does not influence anything on it. You can however place a second outline on the drawings layer plus a note on the comments layer telling the user that this is the maximum size of the flange.)

screenshot from 2018-12-14 00-32-20

---

Regarding pin 1 markers. Here my suggestion for a better fab outline and a pin 1 marker on silk. Make the fab outline continuous as already requested. And move the silk outline closer to the fab outline.
(green for the fab outline, pink for the silk outline.)

screenshot from 2018-12-14 00-39-54

kajusK:
Thank you for the feedback, I've updated the dimensions to match the size of the connector without flange, made F.Fab unbroken and fixed the courtyard clearance.

2018-12-31-100718_1920x1080_scrot

kajusK:
Any update on this? Is there anything blocking this from merging?

poeschlr:
The pin 1 marker on the fab layer is still missing. (See my green line in the picture i showed above)

You rounded the y size. Do not do that.
The courtyard clearance is not quite right. (Must be 0.5mm rounded to nearest 0.01mm grid meaning it must be somewhere between 0.495 and 0.505.)

screenshot from 2019-02-02 09-41-49

kajusK:
Hi @poeschlr,
thank you for the comments, I've updated the dimensions, clearances and added the pin 1 marker.

poeschlr:
Silk is now extremely close to the pads. Could you ensure about 0.2mm clearance (allows the use of cheaper fabs.)

Everything else looks good.
screenshot from 2019-02-09 12-51-27

kajusK:
@poeschlr Thanks for checking the footprint, the clearance was updated.

kajusK:
Is there anything else blocking this PR from merging?

poeschlr:
Sorry for the wait i was busy at work. Looks good now, thanks.

PR #1124, opened 2018-12-02
Connector PCB Edge: create Mini PCIe sockets

myfreescalewebpage:
Creation of Mini-PCIe cards and sockets according to Mini-PCIe specifications (https://s3.amazonaws.com/fit-iot/download/facet-cards/documents/PCI_Express_miniCard_Electromechanical_specs_rev1.2.pdf).

Screenshots:

half_card
full_card
half_socket
full_socket
dual_socket

Half card and full card have an Edge.Cuts layer added by hand. No KLC found about this subject if this was permitted or not. In this case it is very useful to get the correct size of the card. This can also be useful for other Arduino board etc if a KLC can status on this king of footprints with a specific cutting size of the PCB.

The sockets are centered on the socket itself and not on the footprint, which include also card space. Please let me known about the policy in this case and I will fix it if required.

Also, I have no 3D models for all these footprints.


myfreescalewebpage:
Hi @herostrat thanks for review, here are some responses to your points.

> Wouldn't adding the cards into the templates be a better option?

I personally never uses the templates (and I do not know how to use them because they are not available in Kicad menus - I think many people don't know they exist!) What about if you have a board with multiple form factors (minipcie, hald size minipcie, arduino shield etc) ? You will restart your design every time while a single footprint to change is so easy.However, I don't know the Kicad policy on this subject.

> I personally would prefer a footprint of the socket itself, and - if the maintainer see fit - add the half, full and dual socket ass an addition.

I can also add the socket footprint itself, alone, keeping the others footprints ? This permit to easily change just generating a new netlist and import in Kicad PCB editor. Complete footprint is an easy way to not think about the distance between the socket and the end of the mini pcie card where the clips are added.

> The areas for the mounting solution on the sockets should be keepout areas shouldn't they?

No, you solder a clip, example below.

712wvmp0bal _sl1000_

I should also propose versions with screws (example below) but waiting for the review of this ones at the moment.

myfreescalewebpage:
Screenshot of the socket only footprint just added to the pull request:

socket

myfreescalewebpage:
Renamed footprints according to discussion at https://github.com/KiCad/kicad-symbols/pull/1196

evanshultz:
Let me try to summarize what I think is going on here. Could be wrong or missed something, so please help me get this all sorted out so that a real review can commence.
1. This repo should not include anything with edge cuts. Those are for templates.
1. Here was can review and merge footprints that will go onto a larger board. I see the maximal options being just the connector, the connector and clips, the connector and screws, and just the card outlines. There are many flavors of these that could be included, shown on pages 25-31 of the linked doc.
1. These are not card edge connectors because they are for holding a PCB above a bigger PCB. So they should go into, I guess, a new library? `Connector_Card` is already taken so we could do `Connector_PCI` if there will be enough PCI/PCIe connectors to support a new library but not so many they need to be broken out by manufacturer. This library would then include other PCI connectors that may be strewn about into other libs at the moment. Or perhaps there's a higher-level library name that's better, something like `Connector_CardEdge` or `Connector_InternalCard` or `Connector_ExpansionCard` (???) which could also include https://github.com/KiCad/kicad-footprints/pull/879? Note that this item, especially, is up for review and suggestions are welcome.
1. Looking at DigiKey, I see several vendors for the connectors. Do you know if the connectors and clips/screws are truly universal (all meeting the spec) so we don't need any manufacturer-specific footprints but just one standard one?
1. DigiKey also shows me we have several variations to consider: number of positions, bay/row arrangement, etc. These would all be unique footprints.
1. I do not like this unofficial doc, and from the symbol it appears we can't access a free and official one, so what shall we do? We could link to the PCI-SIG page of the document, which gives the official link and includes the document name. Any enterprising users could go from there.

@myfreescalewebpage 
Can you please read the above, confirm if I've got it right, and then list what footprints you'll be including here? We can sort out the footprint names and then have a direction to get this done. You can also remove the templates from this PR and create another PR for them in the templates repo if you like.

myfreescalewebpage:
Hello @evanshultz thanks for looking at my PR! I try to answer below for each point:

1. _This point is for cards only._ I would like this rule (Edge.Cuts) to be reviewed. In this case it's very convenient to have a connector and associate a footprint that is a card. Moreover, you can change the form factor (half/full card) during the design, which is not possible with a template. This footprint is - according to me - very easy to use compared to templates. Half and full card dimensions are normalized.

2. Not understood, can you say it differently ?

3. I'm open to change the library name. Currently the Connector Edge contains some PCI/ISA... connectors so I think it's the best place but we can think more globally to this question.

4. The proposed footprints are based on the MiniPCIe spec (even if I agree the link is not nice) and thus they have to be considered 'generic'. I do not know if all manufacturer respect exactly the dimensions of the spec, probably there is some differences on some part of the connectors which are not functional. My PR is intended to provide a generic footprint, and I'm not against specific footprint in the future of course :)

5. I don't see which parts you are considering. This PR is Mini-PCIe only, they have always 52 positions. Mini-PCIe is different from the other PCI Express connectors.

6. We can do like on the symbol, simply removing the link if you want. I don't know if the official spec is provided by the PCI-SIG website.

Cheers,
Joel

evanshultz:
1. I should be more clear. Edge.Cuts are allowed as part of the footprint, but not to define the footprint. That is for a template. I understand what you're saying, the templates are expressly used for this purpose. If you want to submit all the footprints without Edge.Cuts and then "abuse" them (just a joke) as templates that's fine. But a footprint and template are different due to their purpose. Please move further discussion on this topic to https://github.com/KiCad/kicad-footprints/issues/1307 so that things are kept together as things are always up for debate and revision.
2. I'm looking to confirm that the pages of the doc I mentioned are what you want to submit, and I noted the greatest possible footprints I saw in the doc that could be submitted.
3. Sure. I just happened to click on the Samtech ones that were "golden fingers" and didn't realize any there were plastic headers into which "golden fingers" fit. So I thought it might be best to start a new lib since there are several contributions that could fit. But... I now see they're all together. We can always pile on more and then split later if that's easiest/best so I'm fine continuing. Thanks for pointing this out.
4. Great. If there are known incompatibilities that's good to know, but if this is one standard that is generally being used verbatim then all is good.
5. If I go to the circled product category on DigiKey and then select "PCI Express Mini Card" as the Card Type, I see the following. That made me think there may be multiple variations we should plan for now instead of making one footprint generic which will cause us issues later.
image
6. They are at https://pcisig.com/specifications/pciexpress/. Of course, they're restricted.

myfreescalewebpage:
1-  Will read and react. Thanks for the link!

5- I wasn't aware of the 76 positions version. I have found https://www.te.com/commerce/DocumentDelivery/DDEController?Action=srchrtrv&DocNm=8-1773459-7_EXPRESS_MINI_CARD_QRG&DocType=DS&DocLang=English&DocFilename=ENG_DS_8-1773459-7_EXPRESS_MINI_CARD_QRG_1011.pdf just now but I do not found many more information... What do you suggest ? Renaming these footprints (we should also considering renaming the symbol in this case: https://github.com/KiCad/kicad-symbols/pull/1196) ? Or keeping like that and the future 76 positions version should be named something like "BUS_PCI_Express_Mini_DisplaySocket" (I do not know if there is half/full version but it seems not).

6- Yes restricted....

evanshultz:
5. If these are part of the spec then we should note it somehow. How about just copying PCI-SIG's nomenclature? DMC (display mini card) is also used in JAE docs so perhaps that is relatively official ("display" does not get any hits in the doc you linked). That gives the following (note the capitalization change in the first word):
- `Bus_PCI_Express_Mini_Full_Clip`
- `Bus_PCI_Express_Mini_Half_Clip`
- `Bus_PCI_Express_Mini_Dual_Clip`
- `Bus_PCI_Express_Mini_Full_DMC_Clip`
- `Bus_PCI_Express_Mini_Half_DMC_Clip`
- `Bus_PCI_Express_Mini_Dual_DMC_Clip`
- `Bus_PCI_Express_Mini_Full_Screw`
- `Bus_PCI_Express_Mini_Half_Screw`
- `Bus_PCI_Express_Mini_Dual_Screw`
- `Bus_PCI_Express_Mini_Full_DMC_Screw`
- `Bus_PCI_Express_Mini_Half_DMC_Screw`
- `Bus_PCI_Express_Mini_Dual_DMC_Screw`

We could also use pin numbers (`52Pin` or `76Pin`) instead of `DMC`.

I'm not saying you have to contribute all of them, but does that cover all of them?

https://github.com/KiCad/kicad-footprints/issues/1408 is also related, in that the connector, clip, and form factor (including screw locations) could be separated. Essentially, a PCIe card is sorta like a module (or a space where one can go).

myfreescalewebpage:
@evanshultz looks very good for me, actually the PR covers:
- `Bus_PCI_Express_Mini_Full_Clip`
- `Bus_PCI_Express_Mini_Half_Clip`
- `Bus_PCI_Express_Mini_Dual_Clip`

I have planned to propose:
- `Bus_PCI_Express_Mini_Full_Screw`
- `Bus_PCI_Express_Mini_Half_Screw`
- `Bus_PCI_Express_Mini_Dual_Screw`
but I was waiting for the review of this PR to do it.

My PR also propose
- half and full cards (to be discussed templates or not, I have also read https://github.com/KiCad/kicad-footprints/issues/1408#issuecomment-463400664 very interesting !!)
- 52 pins socket only, with no clip, no screw. I propose to rename to `Bus_PCI_Express_Mini` if this is OK for you, the corresponding 76pins version will be `Bus_PCI_Express_Mini_DMC`

If the last comment above is ok for you I can make the changes right now.

Joel

evanshultz:
I'm not sure. I see the connector as a schematic symbol and one element in the BOM, with the card as a second BOM item. My feeling is to have all the items by themselves and then have some way to align the pieces as desired. I wrote about this more somewhere else but I don't know what y'all think and if it's the best way to do this in KiCad (maybe the tool behavior drives the library?).

I'm fine with the <blank> and `DMC` naming you mentioned above. I think what we wrote matches.

myfreescalewebpage:
> I'm not sure. I see the connector as a schematic symbol and one element in the BOM, with the card as a second BOM item.

Well, in case of the card, you do not have any connector... that's purely the PCB it self which is a connector.

For now, I will separate the cards in another PR we can let in standby to focus on the others. I will rename according to your previous proposition too and we should also think to the new library name (if we want to separate the, also thinking to #1404).

myfreescalewebpage:
Just put again the first comment, updated, to keep the history visible (new PR for the cards is #1481):

Creation of Mini-PCIe sockets according to Mini-PCIe specifications (https://s3.amazonaws.com/fit-iot/download/facet-cards/documents/PCI_Express_miniCard_Electromechanical_specs_rev1.2.pdf).

Screenshots:

socket
half_socket
full_socket
dual_socket

The sockets are centered on the socket itself and not on the footprint, which include also card space. Please let me known about the policy in this case and I will fix it if required.

Also, I have no 3D models for all these footprints.

myfreescalewebpage:
@evanshultz any news here ? Thanks

evanshultz:
@myfreescalewebpage 
What news were you looking for?

myfreescalewebpage:
@evanshultz is this PR ready or not ready ? Thanks!

evanshultz:
1. Shouldn't the origin be at the center of the entire footprint instead of the center of the connector? See the CI errors. Note that it would easiest to review if you make any other changes, review them until we're happy, and then finally move the footprints. No need to re-review the entire thing.
1. Can you beef up the description so it's not just a URL?
1. Can you link the URL to the relevant page? Pages 24-26 and 28, I believe.
1. Right now, the silk is rather far from the connector's fab outline. And I don't know what size that should be. Can you supply a connector datasheet so I can check at least the connector body size? And maybe it's best to snuggle the silk a bit closer to the body and cut it out around pads.
1. Are the bottom left and right 3.2x2.3mm pads for ground, shield, or something else? If they can have any connection, they need a pad number (or name).
1. Should the mounting keepout areas just be fab lines instead of pads since multiple options to secure the card can exist?
1. The 1.6mm and 1.1mm holes are Aperature pads. Should they be NPTHs?
1. There is a horizontal line on Dwgs.User through the x-axis that seems unnecessary.
1. Because the origins align, I placed all footprints on top of each other and dimensioned them. The only issue I see is that all courtyard offsets (shown in red) are too big:
image

myfreescalewebpage:
@evanshultz thanks for your review.

1. I'm not sure, I think having the origin at the same place for the 4 footprints is nice too if the user decide to change it during the design. Also, it's more logical to have the center at the center of the connector itself. We can decide when all other points are fixed as you indicated.

2. Sure, just done.

3. How do you do that ?

4. The footprint is clearly a standard example based on mini pcie spec, not a specific one but here is an example of this kind of connector: https://www.te.com/global-en/product-2041262-1.html

5. Pads for mechanical only

6. Not bad, this is more logical with the spec and permit various options. In this case maybe we should rename again the names of the footprint ? Let me known before i make the modifications.

7. Yes, changed.

8. Yes, removed.

9. Which values to you suggest ? I have set a little bit large to avoid any issues during mini-pcie card insertion.

evanshultz:
1. OK.
3. Put `#page=28` (as an example) at the end of the URL.
4. I understand. But you drew the fab lines as they are for some reason. How did you pick that body size? And once that's settled, maybe snuggle the silk closer?
5. OK. I'd rather see them named so the user isn't forced to have floating copper pads and could instead embed them in a zone. How about naming them `MP` (the symbol will need a corresponding pin).
6. I think so. Maybe just remove the trailing `_Clip`? Thanks! I suppose we use the keepout style mentioned at https://kicad-pcb.org/libraries/klc/F4.5/.
9. As these are connectors I would still go with 0.5mm unless we have reason to do something else. Do you know of any relevant requirements or guidelines?

myfreescalewebpage:
1. --
2. --
3. -- Added at page 11 (beginning of mechanical specifications in the doc)
4. I don't remember exactly why the fab line is like that. I can reduce if required. https://s3.amazonaws.com/fit-iot/download/facet-cards/documents/PCI_Express_miniCard_Electromechanical_specs_rev1.2.pdf#page=32 indicate the minimal distance between two cards is 31mm; I can arrange so that two cards separated of 31mm do not overlap, what do you think ?
5. Why not `MP`. The symbol is currently generic card/socket, so it will need to be separated, one symbol with MP for the socket, and another one without MP for the card itself. Is it ok for you ? I have proposed the symbol myself in the past so it's ok for me.
6. Will do, yes it's great like in KLC F4.5
7. --
8. --
9. Linked to point 4

evanshultz:
3. I think it might be best to link to the page for each card size, as I mentioned above. Easiest to jump right from the footprint to the info describing the footprint dimensions then. What do you think?
4. Sounds good. What body size did you use (and what MPN)?
5. Oh yes. We're on the same page.
9. I see why that supports the spacing in the doc, but I'm not sure that works for the footprint. Allowing footprints to space apart by 31mm results in the MP pins crashing (they are 31.6mm pad edge-to-pad edge). I'm still in favor of the courtyard extending 0.5mm around the body and pads like we usually do for connectors. There will be a courtyard crash in this case but I don't see a way around that. I think we want to accept courtyard encroachment violations for cards placed next to each by 31mm and catch errors for any other parts invading the courtyard space. Your thoughts?

myfreescalewebpage:
Should now have fix major part of the issues. Body size of 9x30mm, silkscreen distance 0.2mm (I have to fix to not overlap the pads right ?). CrtYd 0.2mm of the MP and 0.5mm of the other part of the footprint.

New screenshots:

socket

dual

full

half

myfreescalewebpage:
MP pads renamed "MP".

myfreescalewebpage:
@evanshultz waiting your position on this design now, I know I still have silkscreen lines to fix to not overlap the pad, but it is easier to fix at the end if the other dimensions are correct. Thanks.

evanshultz:
9.25x30mm is the size of the TE connector in the datasheet. Is that a common size?

Why set the courtyard offset this way? Is there a reason for two different values?

myfreescalewebpage:
> 9.25x30mm is the size of the TE connector in the datasheet. Is that a common size?

No it's more a "maximal" size. This one from Amphenol for example is only 8.4x30mm: https://www.mouser.fr/datasheet/2/18/10123921-1303073.pdf

> Why set the courtyard offset this way? Is there a reason for two different values?

Which "two different values" do you mean ?

evanshultz:
Sorry for not being clear. By "two different values" I meant vertical and horizontal courtyard, but that only applies to the connector footprint. The ones with mounting locations have three different courtyard offsets. I wrote above that I think we should offset the courtyard all the way around by 0.5mm, like other connectors, and explained why above. Did you have another idea?

myfreescalewebpage:
> Did you have another idea?

Not really. I just made a screenshot to take a decision:

Capture

The distance is 0.2mm around the MP pads on the bottom, left and right sides. Do you mean 0.5mm at this place ? It's currently 0.5mm on the top of the footprint.

If yes, it's ok for me.

evanshultz:
If all courtyard offsets can go to 0.5mm that would be great. 0.5mm outside of the body or pads, whichever is greater.

myfreescalewebpage:
@evanshultz thanks, just fixed the distance, 0.5mm all around the device.

myfreescalewebpage:
@evanshultz what do you think now ? Thanks

myfreescalewebpage:
@evanshultz oing! I think we are now closed to add this to the lib :)

evanshultz:
@myfreescalewebpage 
Apologies for the delay. I see just one thing left: The silk outline is a rectangle that runs over pads. Please break up the lines to clear all pads.

The symbol does not have a MP pin, which means those two mechanical pins are floating. Could you update the symbol to add a MP pin?

myfreescalewebpage:
@evanshultz no problem. Thanks.
Symbol fix at https://github.com/KiCad/kicad-symbols/pull/2518 (I don't know how the MP pin should be defined)
Footprint lines a little bit simplified to not overlap the pins.
Joel

myfreescalewebpage:
Capture

evanshultz:
Thanks! There could be silk lines just inside of the MP pins. It's a minor thing, but if you add that then let's merge.

myfreescalewebpage:
Yep ! Done :)

Capture

myfreescalewebpage:
@evanshultz ping! So close to the final :)

evanshultz:
Yay! Thanks for all the work on this!

myfreescalewebpage:
Thanks to you also for all this work of review!!! Amazing :)

Issue #1087, opened 2018-11-13
TerminalBlock_Phoenix without mounting pad names


tolot27:
According to [KLC F2.1](http://kicad-pcb.org/libraries/klc/F2.1), connectors with mounting pads should be named `[count]MP`. The mounting pad of many connectors in `TerminalBlock_Phoenix.pretty`, especially, the PTSM SMD series, are not named at all.

The link in the `README.md` to https://github.com/pointhi/kicad-footprint-generator/scripts/TerminalBlock_Phoenix is also outdated.

antoniovazquezblanco:
Naming scheme has probably changed since the addition of that components. It would be awesome if you could submit a PR that fixes this.

Thank you very much @tolot27

tolot27:
Maybe, someone can comment on the linked issue. I'd like to continue

tolot27:
I'm about to create a PR but first I have to verify one thing. Currently, only connector symbols with one mounting pin such as `Conn_01x05_MountingPin` are available. The Phoenic-Conntact PTSM SMD series has two mounting pins.

If I use the `Conn_01x05_MountingPin` with a fixed connector (TerminalBlock with both mounting pads named `MP`) I get no warning after update of netplan. Then the two mounting pins are connected, which is not necessary.

Shall I create connector symbols with two mounting pins first?

![image](https://user-images.githubusercontent.com/1370732/49318997-1587f380-f4fb-11e8-8834-879548a44b43.png)

tolot27:
Why do we need symbols with mounting pads at all? For ground pads it is obvious to me, but not for mounting pads.

Shackmeister:
See discussion here https://github.com/KiCad/kicad-footprints/is

tolot27:
Maybe, check scripts could be improved somehow to test for mounting pads. However, that would require a specification for validation.

tolot27:
What I still do not understand is that according to [KLC F2.1](http://kicad-pcb.org/libraries/klc/F2.1) the naming of mounting pads should be `[count]MP`. But most if not all connectors changed with #422 do not have numbered mounting pads. Are there any footprints with numbered mounting pads or is the KLC outdated?

tolot27:
Ah, sorry, I mixed things up. I thought the name of the pads should be prefixed with a unique number. Now I understand that the package name was meant.

PR #1125, opened 2018-12-02
name mounting pads of PTSM SMD connectors with `MP`

tolot27:
name mounting pads of PTSM SMD connectors (`TerminalBlock_Phoenix.pretty`) with `MP`

fix for #1087 

Generator script will follow later.


poeschlr:
Thanks.

Note to self: the tht version of these is scripted. The SMD version is not. (Meaning changing this here does not require an update to the script repo.)

PR #1126, opened 2018-12-02
Mini-Circuits PL-005 land-pattern

aewallin:
Mini-Circuits land-patter PL-005 used for e.g JTOS-series VCXOs.
Notes:
- GND-vias without Mask-layers cause KLC violations in the scripts.
- GND-zones not drawn, expected that user draws a GND-plane under symbol.

land-pattern: https://ww2.minicircuits.com/pcb/98-pl005.pdf
component: https://ww2.minicircuits.com/pdfs/JTOS-50+.pdf
package: https://ww2.minicircuits.com/case_style/BK276.pdf

pl-005

JTOS symbol PR: https://github.com/KiCad/kicad-symbols/pull/1199

no 3D model.

------------


aewallin:
re-name, as discussed in #1148

aewallin:
after changes:
pl005

poeschlr:
You have the courtyard on the bottom layer.

You defined the large ground area for other similar parts. Is there a reason this is not done for this one?
Your via positions do not quite agree with the suggestion from the datasheet. 

I suspect that you might not find space for the pin 1 marker if you move them to where they need to be you. So maybe switch to a different style. Might i suggest the one we use for DFN footpirnts. For that simply remove the top silk everywhere on negative x (meaning the top horizontal line goes from the current right corner till the center line. No vertical line on the left corner.)

The body width does not quite agree with the datasheet.

screenshot from 2019-01-01 20-44-25

aewallin:
I'm not sure what is the best way forward with the GND-zones, given that the footprint-editor does not support zones.

About 5 rectangular SMD pads are required in this case. Is it worth the complication, or can we assume the user of the footprint has a GND-zone on the top-layer anyway?

poeschlr:
You can use polygon pads. Might even be possible using freecad and stepup: https://forum.kicad.info/t/kicad-stepup-the-sketcher-for-footprint-generation/9582

aewallin:
Now with GND-pads defining the GND-zone. Coyrtyard moves a bit out also in X-direction.
pl005
pl005_3d

poeschlr:
Thanks

PR #1127, opened 2018-12-02
Mini-Circuits land-pattern PL-052

aewallin:
Mini-Circuits land-pattern PL-052

KLC-violationd due to missing Mask-layers on the pads used to create the GND-zone and GND-vias.

component, example1: https://ww2.minicircuits.com/pdfs/AMK-2-13+.pdf
in package: https://ww2.minicircuits.com/case_style/CD542.pdf
component, example2: https://ww2.minicircuits.com/pdfs/ADE-6.pdf
(in package CD637)
land-pattern: https://ww2.minicircuits.com/pcb/98-pl052.pdf

pl-052

no symbol PR yet.

------------


aewallin:
@poeschlr Here's a land-pattern used by two components in different cases (see links above).
The cases CD542 and CD637 differ only in height, as far as I can see.
- how should this land-pattern be named?
- can a footprint have two separate 3D models assigned? (choose correct one for rendering)
- OR should this one be split into two separate footprints for the different cases?

poeschlr:
A footprint can have multiple 3d models. However there is no way to control which one should be shown. (The union of all models is shown)
I would suggest to include the height in the footprint name (other mini circuit footprints already follow this example.)

poeschlr:
All looks good here. Thanks.

screenshot from 2019-01-01 22-07-51

PR #1132, opened 2018-12-03
Add Mini-Circuits land pattern PL-247

aewallin:
land-pattern: https://ww2.minicircuits.com/pcb/98-pl247.pdf
component: https://ww2.minicircuits.com/pdfs/LPF-B0R3+.pdf
package: https://ww2.minicircuits.com/case_style/HZ1198.pdf

KLC check gives error about the footprint anchor, but I dont know how to move the footprint by 0.000635mm.

pl-247

used by e.g. https://github.com/KiCad/kicad-symbols/pull/1212

------------


aewallin:
after revisions:
pl247

serkanxselcuk:
Hello @aewallin ,
you can move items exactly, if you select item(s) "right click>Move exactly" or shortcut key "Ctrl+M".
image
 In this menu, you can give exactly distance in X or Y direction you want to move.

poeschlr:
The large pad looks quite different to how it is shown in the datasheet. (It should be larger)
The clearance to the normal pads is also not quite right.

screenshot from 2019-01-02 00-40-31

aewallin:
Here I would prefer leaving the GND SMD-pads out completely. Just define a 0.032" clearance for the signal pads, maybe a keepout-zone as per the 0.138" dimension in the spec - and rely on the user providing a GND-zone everywhere else.
What do you think?
(different SMD pad clearance-values in X/Y directions would be nice in this case..)

aewallin:
from the land-pattern drawing I estimated that the GND-zone overlaps the f.fab outline of the package by 40mils. The GND-zone pads are now revised:
pl247_dims

myfreescalewebpage:
@aewallin same problem here already discussed on other Mini-Circuits footprint.

Joel

aewallin:
@myfreescalewebpage I don't see any issues with the coordinates of any pads or the dimensions. They are even numbers in units of mils.
The dimensions pointed out by @poeschlr above should be corrected in 7526fa3

poeschlr:
Thanks

PR #1133, opened 2018-12-04
Add Mini-Circuits land-pattern PL-079

aewallin:
Before further review:

- [x] re-name to include case name
- [x] description with case name and land pattern
- [x] 3D model to case name

land-pattern: https://ww2.minicircuits.com/pcb/98-pl079.pdf
component: https://ww2.minicircuits.com/pdfs/SYPD-1+.pdf
package: https://ww2.minicircuits.com/case_style/TTT167.pdf

pl-079

------------


Misca1234:
@aewallin 
You pushed a number of similar foot prints, have you thought of doing them with a script instead ?
I can help you if you do not know how

aewallin:
Thanks for the suggestion.
Do you have links to some (python?) example-scripts?
Is there a tool or workflow that allows rapid visual feedback when writing the scripts?

poeschlr:
https://github.com/pointhi/kicad-footprint-generator is used to generate most of our scripts.

I am however not sure if it will really make your work much easier. Even thought all of these are for a similar function, they still look quite different. (Scripting is not made easier by having parts with a similar function. It is really only helpful if the resulting footprints look similar.)

I have however not looked at all your contributions so i could be wrong with my current assessment.

poeschlr:
Regarding your part. From the screenshot i can already tell you that the courtyard needs to move outwards. The clearance is measured relative to either body or pads. Whatever results in the larger area (in a particular direction.) In this case the courtyard clearance would be measured releative to the pad outline in all directions.

aewallin:
cortyard should now be ok.
circle pin-indication from datasheet. is another f.fab pin1-indicator still needed?
do SMD parts need rotation so pin1 is up?
pl-079

Misca1234:
I am pretty sure the script 
https://github.com/pointhi/kicad-footprint-generator/scripts/general/StandardBox.py 
should be able to pull off the trick

@aewallin 
if you look into 
https://github.com/pointhi/kicad-footprint-generator/Battery/BatteryHolder.py

You can see how the StandardBox is used.
In StandardBox.py you define the with and height and where smd/tht/npth "pins" are placed, then the StandardBox.py will "crawl" around and create the silk and courtyard correctly.

aewallin:
after revisions:
pl079

myfreescalewebpage:
Hello @aewallin 

Had a look to the footprint, the dimension and position of the pads and Fab layer are not exactly the same as given in the datasheet. This is always some few mils, but is it really wanted ?

Cheers,
Joel

aewallin:
I checked the pads, but I cannot find issues - could you explicitly show what is wrong?
pl079
(from https://ww2.minicircuits.com/case_style/TTT167.pdf)

myfreescalewebpage:
Yes for example, looking at pin 4:
- P = 1.52mm in the datasheet, 1.524mm on your footprint
- L = 1.78mm in the datasheet, 1.778mm on your footprint

Also positions have 4 or 5 decimals... not so easy to review...

aewallin:
In units of mils I don't see any issues with coordinates of pads or dimensions.
If mini-circuits cannot get a multiplication of 25.4 between mils and mm right then which unit shall we use? I have used mils because they are more even numbers in most cases for these parts.

aewallin:
close and reopen for CLA

aewallin:
the need for explicit GND-zone pads can be discussed. Here I have assumed the user has a GND-zone on the top layer anyway, and it will fill the right areas, excluding the clearance areas on the signal pins.

myfreescalewebpage:
> In units of mils I don't see any issues with coordinates of pads or dimensions.
If mini-circuits cannot get a multiplication of 25.4 between mils and mm right then which unit shall we use? I have used mils because they are more even numbers in most cases for these parts.

Understood ! You are editing and looking in mil, I'm reviewing with millimeters ! The multiplication/division can be exact because it is not exact in the datasheet too, it is rounded !

There is : http://kicad-pcb.org/libraries/klc/G1.9/

> Metric units are preferred, where appropriate
Linear dimensions should be specified in millimeters

Joel

aewallin:
IMO issue is in the Mini-Circuits datasheet. It's obvious they design their parts and dimensions in mils, and then convert to metric. If they are unable to do the multiplication by 25.4 right then how am I supposed to guess how and when to round to mm? Is there something in the KLC about significant digits when rounding to mm? 

random rounding this way or that just makes this footprint more wrong compared to how it is now.

myfreescalewebpage:
I have only found http://kicad-pcb.org/libraries/klc/F7.7/, in which it is indicated that drill size are rounded to 0.01mm. This is also how Mini-Circuits has rounded the values.

@poeschlr for this kind of parts, how can we handle this ?

My main opposition to the inches referential is that the distances/sizes in the kicad_mod file are in Metric referential. Thus, this give very strange numbers, such as "4.3815" on some pads for this device.

Joel

aewallin:
a few dimensions here:
dimensions

poeschlr:
The rounding of drill sizes only applies to such sizes that are calculated from rectangular lead diameters. (This is because the sqare root will always produce strange values.) If a drill size is called out in the datasheet then that one has to be used with the same number of significant digits.

---

I would add the same requirement as with other similar footprints. I would add the pre defined copper zone. And add a second footprint without zone and vias.

myfreescalewebpage:
@poeschlr the question is more about the unit to use, inch of millimeters ? The kicad_mod files are using metric, so actually the footprint txt file contains values with lots of digits...

poeschlr:
We prefer metric. But we can only use metric if the datasheet gives metric measurements. The land pattern definition for this part does not give metric measurements.

myfreescalewebpage:
@poeschlr the datasheet does (https://ww2.minicircuits.com/pdfs/SYPD-1+.pdf) but not the land pattern file you are right

myfreescalewebpage:
@poeschlr I understand you are not agains keeping this will imperial measures because the manufacturer clearly used this. Okay for me.

@aewallin what about comment of @poeschlr:

> I would add the same requirement as with other similar footprints. I would add the pre defined copper zone. And add a second footprint without zone and vias.

Joel

aewallin:
I suggest a separate issue is created for the mini-circ footprints without LandPattern/GND-zone/GND-vias.
I will add the GND-zones here.

aewallin:
I have added pads defining the GND zone:
pl079

myfreescalewebpage:
@poeschlr @aewallin I'm not confident with this version. Soldering or un-soldering the device will need to heat the whole GND plane. When making a ground plane on a PCB, the GND pins of the devices are not "mixed" with the ground plane, there is a small clearance with no copper and some small connections between the pads and the plane.

As shown on this image for example

910b36c6d36afb1ed884067fcb53d7ea6741166fc3c_406x500

aewallin:
on 5.0.2, placing e.g. SYPD-1 symbol, assigning this footprint, connecting the GND-pin, and having a GND-zone drawn on the pcb I get this. I don't see any issues.
Your pad3 looks like pad connection set to "thermals", not "solid" as it should be for a solid GND plane.

for users who are worried about the soldering the "plain" version of the footprint is a good idea - one can the add as much vias or gnd-zones as wanted - and not follow the manufacturer land pattern recommendation...

sypd1

poeschlr:
@myfreescalewebpage these footprints are likely designed by mini circuits with reflow soldering in mind. This process does not really need thermals as it heats the whole board evenly.

There is a tradeoff between having options for rework and having proper EMC or thermal management. I would assume the engineers at mini circuits are aware of that.

I therefore assume they choose the suggested footprint for a reason. Well knowing that it not really suited to be reworked or soldered by any other process than reflow. (Under reflow soldering i mean a proper regulated reflow process. Such a process regulates the heat in a precise profile to ensure proper soldering even without thermals.)

myfreescalewebpage:
And without the need to repair the board later too !

Have just checked the dimensions/positions/layers, okay for me. @poeschlr any clues ? Else we can merge.

Joel

poeschlr:
Looks good to me. Thanks @aewallin

PR #1138, opened 2018-12-04
Fix pinout of PJ320D audio jack

GrantM11235:
This is a followup to #1014.

I sent an email to Wenzhou QuinPu because I suspected that there was an error on the [WQP-PJ320D product page](http://www.qingpu-electronics.com/en/products/WQP-PJ320D-72.html) that showed the wrong pinout. They sent me [this datasheet](https://github.com/KiCad/kicad-footprints/files/2644866/WQP-PJ320D.pdf) which shows a different pinout which matches two other PJ320D manufacturers ([XKB](https://lcsc.com/product-detail/Audio-Video-Connectors_XKB-Enterprise-PJ-320D-X_C319111.html) and [Korean Hroparts](https://lcsc.com/product-detail/Audio-Video-Connectors_Korean-Hroparts-Elec-PJ-320D-4A_C95562.html)) as well as a connector I own from an unknown manufacturer.

I have suggested to Wenzhou QuinPu that they should update their product page.

This is the correct pinout:

PJ320D at commit 46d25b5

------------


evanshultz:
Thanks for this investigation! If you can sign the CLA I will merge.

Issue #1140, opened 2018-12-05
Phoenix PTSM footprints with mounting pads miss the -1MP in the footprint names.


poeschlr:
More details see: https://github.com/KiCad/kicad-footprints/issues/1087

We changed the pad numbers but forgot to rename the footprints.

PR #1142, opened 2018-12-05
PSTM -1MP in SMD package names

tolot27:
rename package names of TerminalBlock_Phoenix_PTSM to include `-1MP` for SMD connectors

fix for #1140


poeschlr:
Thanks

PR #1143, opened 2018-12-05
Add Panasonic EVM3Y SMD potentiometer

MatthewStephenson:
Panasonic EVM3Y SMD potentiometer

Datasheet: [https://industrial.panasonic.com/cdbs/www-data/pdf/AOI0000/AOI0000C51.pdf](https://industrial.panasonic.com/cdbs/www-data/pdf/AOI0000/AOI0000C51.pdf)

image

The automated check is failing on the "centre of footprint" check by 0.125mm, because I've placed the anchor at the centre of the rotating part, as per the datasheet diagram, and the pads are not symmetrical in size or placement.

As always, comments and suggestions welcomed.


poeschlr:
This part is marked as **obsolete**. I fear we can not accept it in the official lib for this reason.

As a learning opportunity i reviewed it anyways.

The silk outline should be much closer to the body. (It is of no particular help that far out from the part)
The fab reference can be a lot larger and still fit. There is no particular reason to reduce the size of the value field.

My suggested silk outline is drawn in pink
screenshot from 2018-12-13 23-33-45

---

There are also some mistakes in the land pattern. The pad number 2 is not correctly placed. And the courtyard clearance should be 0.25mm.

screenshot from 2018-12-13 23-51-17

MatthewStephenson:
Thank you for taking the time to correct my submission - I haven't done one for a while and I'm clearly rusty.

It's a fair point about it being marked as obsolete in the datasheet. It's a shame as it's the only one in a 3mm form factor available from RS for a low price in small quantities (but maybe that's because it is end of line).

Could you let me know if it's definitely not going to be accepted into the library? If so, I'll search for a replacement. But if it can go in, I'll make the required corrections.

MatthewStephenson:
On reflection, I will resubmit a different part.

PR #1146, opened 2018-12-06
Add TI DSBGA (YZR) footprint

RockyTV:
Required by https://github.com/pointhi/kicad-footprint-generator/pull/218 and https://github.com/KiCad/kicad-symbols/pull/824

Datasheets: http://www.ti.com/lit/ml/mxbg270/mxbg270.pdf, http://www.ti.com/lit/ds/symlink/lmv7272.pdf


evanshultz:
Thanks!

PR #1150, opened 2018-12-07
Add Mini-Circuits land-pattern PL-225

aewallin:
land-pattern: https://ww2.minicircuits.com/pcb/98-pl225.pdf
component: https://ww2.minicircuits.com/pdfs/LAT-3+.pdf
package: https://ww2.minicircuits.com/case_style/MMM168.pdf

pl-225
for review the spec-drawings:
pl-225pads
pl-225spec

symbol: https://github.com/KiCad/kicad-symbols/pull/1226

------------


aewallin:
I don't know how to fix this error in the footprint-editor (there is no grid-setting that low)

>  Footprint center calculated as (-0.00635,0.0)mm

poeschlr:
Similar namechange request to your other mini circuit part. (could you go through all your mini circuit parts and check their names?)

For parts like this you can ignore the center suggestion by travis. (The calculation only takes pads into account, Not really the right thing for this sort of part)

Regarding grid: That is not really a restriction. We normally do not really use the grid for footprints any more. Kicad can snap tracks to a pad centered on any position not only to the grid. (The grid stuff is simply a thing of the past.)

---

The silk outline can be a lot closer in this case. The clearance requirement is only relative to exposed copper.

The "vias" are not really inside the larger pads. Is this on purpose? Especially noticable on the right side. (Might be one of the reasons why travis complains about the center) This will influence the courtyard outline. (The reason why i added dimensions relative to the large pads as well)

body dimension in y direction is 0.12 inch right now. This is the maximum dimension, should be nominal. (nominal is 1.125)

The dimensioned drawing has even more info. (I feel dirty now that i used this strange unit system. But mini circuit sadly does not include metric dimensions in some of their drawings.)
mini svg

aewallin:
Description:

> Footprint for Mini-Circuits case MMM168, Land pattern PL-225, vias included, (case drawing: 
> https://ww2.minicircuits.com/case_style/MMM168.pdf, land pattern drawing: 
> https://ww2.minicircuits.com/pcb/98-pl225.pdf)

aewallin:
after changes:
mmm

poeschlr:
the large left pad looks a bit lower down (y direction) than one would expect. (I would expect it to be centered on the same y coordinate than the two vias. At least it kind of looks like that in the datasheet and i think it would make sense.)

poeschlr:
Body size in y direction is not quite right yet.
screenshot from 2018-12-16 20-14-08

aewallin:
y-size, in my footprint editor
- bottom line at y= 0.05675 inch
- top line at y = -0.05575 inch
- y-size = 0.1125 inch
Maybe FreeCAD truncates (not rounds?) to 3 decimal places?

poeschlr:
You have a one too much in there. It should be 1.125 not 1.1125

aewallin:
> You have a one too much in there. It should be 1.125 not 1.1125

I don't understand. The case drawing is here: https://ww2.minicircuits.com/case_style/MMM168.pdf
- Max Y-size 0.120 inch or 3.05 mm
- Min Y-size 0.105 inch or 2.67 mm
If we assume the nominal size is the average of these, and that is what we want for the F.fab outline then size should be
- Avg Y-size 0.1125 inch or 2.8575 mm
Now looking at F.fab of the footprint the Y-size is 0.1125 inch as far as I can tell both from f.fab line coordinates and from the measuring-tool:

poeschlr:
You are right. I seem to have copied the wrong numbers.

PR #1151, opened 2018-12-07
add Mini-Circuits land pattern PL-012

aewallin:
land-pattern: https://ww2.minicircuits.com/pcb/98-pl012.pdf
component: https://ww2.minicircuits.com/pdfs/RBPF-246+.pdf
package: https://ww2.minicircuits.com/case_style/CK605.pdf

pl-012

symbol e.g. https://github.com/KiCad/kicad-symbols/pull/1223

------------


aewallin:
re-named to include case name. improved description with link to case and land pattern.

myfreescalewebpage:
Hello @aewallin 

Had a look to the footprint, the dimension and position/size of the pads are not exactly the same as given in the datasheet. This is always some few mils, but is it really wanted ?

Cheers,
Joel

PS : see also travis errors

aewallin:
This is a strange error. I think it is caused by rotation of the footprint. It was first drawn in some orientation, with pad-coordinates exact in units of mils, but then after rotation so that pin1 is top left the coordinates have rounding error introduced!?

myfreescalewebpage:
I don't know, maybe this is the issue... Check also your pending PLxxx footprint, they have the same problem.

aewallin:
- all pads and f.fab outline are on integer coordinates/dimensions in units of mils
- the need for explicing GND-zone-pads can still be discussed

poeschlr:
I would suggest to add a footprint with the zone pre defined as with your other footprints.

I would refrain from adding a footprint with vias but without zones. (looks extremely strange having loads of tht pads floating in an otherwise empty area.) It would be possible to again add one footprint with both a zone and vias and one without both of them.

aewallin:
I added pads that define the GND-zone:
pl-012

myfreescalewebpage:
@aewallin this one looks great too. @poeschlr any comment ? Else we can merge.

PR #1152, opened 2018-12-07
add JAE DX07S024WJ3R400 vertical USB C

dlharmon:
Based loosely on the existing USB_C_Receptacle_Amphenol_12401610E4-2A.

Footprint drawing:
http://www.jae.com/z-en/pdf_download_exec.cfm?param=SJ117928.pdf

The 3d model does not yet exist.

If desired, I can add a version with round holes like some of the other USB connectors with slot holes. JAE calls for the slot holes in their drawing.

screenshot from 2018-12-07 11-27-22

------------


poeschlr:
Looks very nice. I have only two small points:
- Could you add a pin 1 (A1) marker on the fab layer? (Suggestion for new fab outline in green)
- could you move the silk outline a bit towards the outside? It should be completely outside the part dimension. (Suggestion for new silk screen in pink)

screenshot from 2018-12-13 21-35-19



---

For reference the dimensioned drawing created from this footprint.
screenshot from 2018-12-13 21-30-56

dlharmon:
Thanks for looking it over. I've made the suggested changes.

What tool did you use to create the dimensioned drawing? That seems useful.

screenshot from 2018-12-13 13-57-43

poeschlr:
A tutorial can be found here: https://forum.kicad.info/t/how-to-check-footprint-correctness/9279

PR #1156, opened 2018-12-08
Add CHIP and CHIP-Pro footprints

myfreescalewebpage:
Add missing CHIP and CHIP-Pro modules. After the end of NextThingCo, some documentation are available on the web, I have recorded a part of the hw doc files at https://github.com/myfreescalewebpage/chip-hardware.

Screenshot of the footprints:

chip

chip-pro


myfreescalewebpage:
Closing/opening to refresh the CLA.

evanshultz:
@poeschlr 
Do you want to add these missing footprints for a company that is insolvent? What value is there is making and merging these footprints into our library? Should we archive the symbols and be done with it?

myfreescalewebpage:
To get more information, this was broken when going from v4 to v5, as discussed here : https://github.com/KiCad/kicad-symbols/issues/975 so I pushed again the footprints.

In my point of view, this is not because something is out of date that it should not be accepted in the library. Many people still use these boards. Else we should remove many parts of the library :)

However, I understand if it is not a priority or not wanted at all, in this case it's good to status and I will add them in a personal GitHub repository, next to the Linux/uBoot/Yocto repositories I have for these boards.

Cheers,
Joel

PR #1157, opened 2018-12-11
Connector_USB: Add Molex 105133-0031

antoniovazquezblanco:
http://www.molex.com/pdm_docs/sd/1051330031_sd.pdf

screenshot from 2018-12-11 12-53-16

------------


poeschlr:
What is the difference between this part and the one that is already in the lib?
If they are identical then we might want to wildcard the footprint name instead of adding a new duplicate.

I would further suggest to change the fab pin 1 marker and also the silk outline. Something like this (green would be the new fab outline, pink the silk outline)

screenshot from 2018-12-13 20-38-04

antoniovazquezblanco:
Connectors differ in height and other 3D model details. We can go two ways:

1. Add the footprint and each footprint will have a corresponding 3D model.
2. Add a second 3D model to the existing footprint + footprint rename for the user to be able to choose the model.

Which way do you want me to go?

Thanks!

poeschlr:
With the current kicad feature set we need to go with option 1

antoniovazquezblanco:
I've updated the footprint to take into account a piece of plastic in the base of the connector that I previously missed. I will be updating the courtyard soon but meanwhile please tell me if the fab and silk layers are fine with you or if you want me to do any other thing.

Thanks!

screenshot from 2018-12-18 10-28-19
screenshot from 2018-12-18 10-28-30
screenshot from 2018-12-18 10-28-43

poeschlr:
the screenshots look good to me

antoniovazquezblanco:
Done!

antoniovazquezblanco:
Sorry about the last two commits. Just a mistake!

poeschlr:
Looks good, thanks

PR #1159, opened 2018-12-11
added ONSemi_SO-8FL_488 package

Shackmeister:
Datasheet: https://www.onsemi.com/pub/Collateral/488AA.PDF
Screenshot:
488aa

------------


aewallin:
@Shackmeister are you using a script or other tool to generate the courtyard that follows the f.fab and pads with a given clearance?
Or is it just skill and manual work? ;)

Shackmeister:
@aewallin Skill and manual work :) planning on adding rounding to some of the IPC scripts :)

poeschlr:
To be honest i am not so sure if rounded courtyards are such a good idea. They don't really give that much reduction in false positives while possibly increasing the calculation time needed by a lot.

It also seems that IPC does not really define anything other than straight lines. We would therefore stretch the definition of IPC compliant quite a bit.

Shackmeister:
Hmm that might be true. 
The reason I started adding roundings was to cut down on errors of the designs made before the courtyard check.
In squeezed and/or non rectangular designs it has helped me quite a bit.
Not sure how much effect it will have on DRC calculation time, but it might also be something to consider.

poeschlr:
I asked the devs. I am however doubtful that it is a good idea to round of every parts courtyard. A single rectangle has 4 lines to take care of. A rounded rectangle has 4 lines plus 4 arcs. (double the number of elements)

Even if lines and arcs are equally expensive for the algorithm that will still double the time needed for checking DRC violations. (And that assumes linear growth.)

evanshultz:
I recall John Beard did some testing and, if I understood, there was little hit to having arcs. (I can't find the thread in the mailing list archive right now.) So does that mean this is OK to review as it is? Or, Rene, do you prefer using square courtyards?

poeschlr:
I am still torn to be honest. Not because of a single component. I mainly am torn if we should go that way for the full library. (If we go that way with one footprint then the same reasoning that suggested it to be good for this one holds for all footprints in the lib.)

My question here is mainly: Is it worth the extra effort for both creating footprints and also for checking contributions. (Right now i lean towards the side of it not being worth it. Maybe kicad will add better tools to the footprint editor or we get better travis scripts that might make it more viable in the future. A travis check for closed courtyard outlines would be the least i would like.)

Shackmeister:
Hi guys

Sorry for the lack of response in the discussion.
For me it always depends on the design if it's nice to have, or doesnt make any difference.
In circular PCB's, with enough components to make angling nescessary, it's a must to have rounded corners, unless the courtyard DRC is disabled/ignored)
I started adding it to my own footprints since it typically doesnt take many seconds to add the roundings.
(after the courtyard has been made, make a 0,25/0,5mm 90 degree arc, place it in every corner, snap the old lines to the arc).
It could probably be added to the footprint scripts as well without too much troubles.

Reviewing is another story, we now got the script for checking if the courtyard is closed.
That saves a bit of trouble, but checking the arc might be annoying.


I don't mind removing the arcs on this footprint, perhaps we could also accept footprints with and without roundings?

poeschlr:
I can accept the odd footprint with arcs but would advice against making it a requirement.

poeschlr:
A few dimensions are wrong. The arcs on the courtyard are correct (you need to add 0.025mm to the measurement as stepup does not give me the center line.)

Screenshot from 2019-06-02 19-20-27

Shackmeister:
@poeschlr 
Those should be fixed now

poeschlr:
The edge cuts outline is not closed (massive hole at top left corner)

The dimensions for the "ears" of the large pad are still not correct.
Screenshot from 2019-09-04 22-30-52

Shackmeister:
After some Stepup, text editor and some more tricks I think I finally got the big pad size right.
And also fixed the arc issues

poeschlr:
looks good now
Screenshot from 2020-02-06 19-25-35

PR #1161, opened 2018-12-11
Button_Switch_SMD: Add a 6x6mm SMD push button

antoniovazquezblanco:
I was not sure if the component was generic enough to be named as I have currently submitted or if should be specific for the series of the following datasheet. Let me know if you prefer a more "atomic" naming. Thanks!

https://www.ckswitches.com/media/1471/pts645.pdf

screenshot from 2018-12-11 17-59-22

------------


aewallin:
are there cases where polarity matters? if so, pin1 should be indicated?
if they are atomic parts and you follow the datasheet it numbers pins as 1,2,3,4 (but ofcourse the generic symbol has pins 1,2...)

poeschlr:
The only reason i could see for a pin 1 mark on non polarized parts is that it could in some cases help with debugging. (This is true for every part.)

This is clearly a generic footprint intended to work with the generic symbol. (you can tell from the footprint name.) Footprints specialized for a single component include the full part number plus manufacturer name. (It does not really make sense for the official lib to go that far for parts like standard buttons. The most we can do is have generic footprints with fully specialized symbols or more likely aliases of a fully specified symbol. For buttons even that is too far in my opinion.)

antoniovazquezblanco:
Thanks @poeschlr,

Should we include the height of the device in the naming? (`SW_Push_1P1T_NO_6x6mm_H9.5mm`)

poeschlr:
I am torn to be honest. On the one had having the height would allow us to offer different generic footprints fitting different parts with a correct 3d model already build in. But on the other this would mean that we will need quite a few such generic footprints.

antoniovazquezblanco:
I will go whatever direction you prefer. Just let me know.

Thanks!

poeschlr:
Lets include the height in the name but don't go overboard with the number of generic footprints.

antoniovazquezblanco:
It should be ready for review.

Thanks!

poeschlr:
Pad positions in x direction are not quite right. Courtyard clearance is wrong (Will not be fixed by moving the pads to the correct position).

Screenshot from 2019-03-14 07-02-39

antoniovazquezblanco:
Messed up the PR. 

Re push at: https://github.com/KiCad/kicad-footprints/pull/1480

Sorry

Issue #1163, opened 2018-12-13
CRLF vs. LF: MountingHole_2mm and MountingHole_2.1mm


aewallin:
1438595d3987a6df81d4de376d6bed3544c83672 seems to cause errors for me:

$ git diff
warning: CRLF will be replaced by LF in MountingHole.pretty/MountingHole_2.1mm.kicad_mod.
The file will have its original line endings in your working directory.
warning: CRLF will be replaced by LF in MountingHole.pretty/MountingHole_2mm.kicad_mod.
The file will have its original line endings in your working directory.

I'm not sure what is going on. I'm on Ubuntu 16.04LTS.

fauxpark:
Looks like Travis picked up on the CRLFs: https://travis-ci.org/KiCad/kicad-footprints/builds/426012316
But it only produces a warning, not an error, so you get a green tick mark on the commit. Considering the use of the word "must" as opposed to "should" both in the check scripts and [the KLC itself](http://kicad-pcb.org/libraries/klc/G1.7/) maybe it should be an error.

poeschlr:
the autocrlf thing sadly does not help if users add stuff directly via the browser instead of going through git. When using a proper git client it should work as intended.

aewallin:
Is this something I can/should fix in my local repo?
It now prevents me from "checkout my_old_branch" before committting or stashing the CRLF modifications??

aewallin:
Thanks, resolved now.
I had to delete the offending files before I could do a "pull upstream master"

PR #922, opened 2018-09-08
Add 2 mm and 2.1 mm mounting holes

diegoherranz:
Modified from other existing mounting holes.
Thanks.

--------

Thanks for creating a pull request to contribute to the KiCad libraries! To speed up integration of your PR, please check the following items:

- [ ] Provide a URL to a datasheet for the footprint(s) you are contributing
- [ ] An example screenshot image is very helpful 
- [ ] If there are matching symbol or 3D model pull requests, provide link(s) as appropriate
- [x] Check the output of the Travis automated check scripts - fix any errors as required

PR #1164, opened 2018-12-13
MountingHole: fix 2 mm and 2.1 mm line endings

diegoherranz:
They were CRLF (Windows) and it was causing problems.

I submitted the symbols on https://github.com/KiCad/kicad-footprints/pull/922.
I created them on Windows but submtted them from Linux and because of that, they were committed with CRLF endings (Windows). Now git is showing them as modified everytime I checkout a recent commit.

I'll be more careful next time :)
Thanks!


aewallin:
this hopefully fixes #1163, which can be closed after this is merged.

SchrodingersGat:
Thanks :)

diegoherranz:
Actually, now that I read https://github.com/KiCad/kicad-footprints/issues/1163, I think that is what happened. I created the footprints on Windows and then submitted them via the Github website. Lesson learned.
Thanks!

PR #1167, opened 2018-12-14
add Xilinx Zynq 7 BGAs

dlharmon:
## CLG225 image
Other parts are similar.
screenshot from 2018-12-14 09-29-40

# Concerns

## Duplication of existing parts
Some of these are close to existing parts in the library and should probably be removed from the branch before merging.
 - CLG225/BGA-225_13x13mm_Layout15x15_P0.8mm_Ball0.4mm_Pad0.4mm_NSMD is unique
 - CLG400/BGA-400_17x17mm_Layout20x20_P0.8mm_Ball0.5mm_Pad0.4mm_NSMD is unique
 - CLG484/BGA-484_19x19mm_Layout22x22_P0.8mm_Ball0.5mm_Pad0.4mm_NSMD is unique
 - FBG484/BGA-484_23x23mm_Layout22x22_P1.0mm_Ball0.6mm_Pad0.53mm_NSMD is similar to the existing BGA-484_23.0x23.0mm_Layout22x22_P1.0mm and FB-BGA-484_23.0x23.0mm_Layout22x22_P1.0mm
 - FBG676/BGA-676_27.0x27.0mm_Layout26x26_P1.0mm_Ball0.6mm_Pad0.5mm_NSMD is similar to the existing BGA-676_27.0x27.0mm_Layout26x26_P1.0mm_Ball0.6mm_Pad0.5mm_NSMD
- FFG900/BGA-900_31x31mm_Layout30x30_P1.0mm_Ball0.6mm_Pad0.53mm_NSMD is unique
- FFG1156/BGA-1156_35x35mm_Layout34x34_P1.0mm_Ball0.6mm_Pad0.53mm_NSMD is similrar to BGA-1156_35.0x35.0mm_Layout34x34_P1.0mm

## Naming
 - should these include the Xilinx designator (CLG225, CLG400, etc) in the name?
 - FBG676, FFG676 and RF676 all have the same dimensions on the bottom side. They differ on the top side with a heat spreader vs exposed die. Should these be separate footprints for the sake of attaching 3D models later. The same is true of CLG484 and SBG485. 
 - 17x17mm vs 17.0x17.0mm - which is correct? The library has a mix and I wasn't even consistent within this PR. I'll fix this before the merge.
## Pad size
 - UG865 gives a maximum pad size only (0.4 mm for 0.8 mm pitch and 0.53 mm for 1.0 mm pitch) and that's what I have for now.
 - UG1099 gives a pad size of 0.4 mm for FB and FT devices (1.0 mm pitch), 0.51 mm for FF, FG, other 1.0 mm pitch devices. On 0.8 mm pitch, it is consistent with UG865, calling for an 0.4 mm pad. There is no difference in ball size for these packages (all have 0.6 mm balls).
 - UG1099 also calls for a solder mask expansion of 0.015 mm which is outside of most design rules. I've used the UG865 value of 0.05 mm which is possible at most fabs.

## should there be a more explicit pin 1 marker like others in the Kicad libs?
Example:
screenshot from 2018-12-14 09-52-17
------------


dlharmon:
#### Duplication
I don't find any occurrences of the older footprint names in the master branch of kicad-symbols, so removal of a few might be an option if there is a desire to do so. The various Xilinx FPGAs in the library don't specify any footprint.

#### Naming
I think these are fairly generic with other vendors having parts that would match these footprints. I'll leave out the Xilinx designation unless I am told otherwise. However, some designation for lidded/lidless will be needed for 3D models. The lidded parts tend to be thousand dollar chips, so the lidless may be a sensible default. We could use the thickness to designate that as it differs.

I'm planning to do the whole Lattice ECP5 series as well and some of those have unique patterns of missing balls, so will probably have the Lattice package designator in the name.

I will remove the NSMD from the names given the comment of it being the default.

#### Pad size
Since Xilinx only gives a max in the primary document, I prefer to use the pad sizes I've used in the past which are within the IPC range. I'll do another push to the branch later today with this.
0.4 mm balls will get 0.32 mm pads, 0.5 mm balls will get 0.4 mm pads and 0.6 mm balls will get 0.45 mm pads unless there are objections.

#### Pin 1
I wasn't aware of that in IPC. That's an excellent reason to go with it.

evanshultz:
**Duplication**
Then figuring out what footprints the Xilinx parts use should be done. They may be older parts before we had more rigorous requirements for the library and they do use the footprints there. Renaming footprints and then assigning them to symbols could be done in a couple PRs to take care of it. I'll volunteer to look at this unless you want to.

**Naming**
We don't have a good method to indicate missing balls, but I've discussed it in the past. Perhaps now is the right time to do that. http://ohm.bu.edu/~pbohn/__Engineering_Reference/pcb_layout/pcbmatrix/IPC-7x51%20&%20PCBM%20Land%20Pattern%20Naming%20Convention.pdf includes the precision information I mentioned earlier but the missing and hidden pads section doesn't really apply for BGAs. We need something more robust (even if it's just showing the total ball locations and the used ones (we probably won't run into many conflicts that way). https://github.com/KiCad/kicad-footprints/issues/436 is a subset of this issue, and going from that perhaps `BGA-256-232N...`?

I might then leave off any annotation about the lid for the common, lidless ones, and then add `_Lid` or something similar to the lidded versions (the exact naming TBD by committee at the appropriate time). And, since v5.1 should support multiple 3D models for one footprints, we may not ever need unique footprints for them as the generic footprint could have both lidded and lidless 3D models available. (This is working now so I expect it will show up in v5.1, but it hasn't been discussed before so we'll have to figure out how to use this feature in the library.)

**Pad size**
IPC-7095A (I don't have the newer versions) says:
- 0.6mm ball gets 0.45mm pad
- 0.5mm ball gets 0.4mm pad
- 0.4mm ball gets 0.3mm pad

All are +/-0.05mm.

So that sounds great to me and matches my reference doc. If you have a newer version of IPC 7095 and it suggests 0.32mm nominal for 0.4mm balls then all the better.

**Pin 1**
See, for example, http://www.ocipcdc.org/archive/What_is_New_in_IPC-7351C_03_11_2015.pdf  to get an idea of pin 1 silk marks.

dlharmon:
I've pushed the pad size and _NSMD removal from names change. I went with your numbers. My 0.32 mm was just a calculation.

Nothing is removed, so there is duplication. The following are existing symbols similar to ones in this PR:
 - FB-BGA-484_23.0x23.0mm_Layout22x22_P1.0mm.kicad_mod
 - BGA-484_23.0x23.0mm_Layout22x22_P1.0mm.kicad_mod
 - BGA-676_27.0x27.0mm_Layout26x26_P1.0mm_Ball0.6mm_Pad0.5mm_NSMD.kicad_mod
 - BGA-1156_35.0x35.0mm_Layout34x34_P1.0mm.kicad_mod

Once everyone is happy with this, I'll do the same for Artix-7, Spartan-7. I'm going to create an issue on kicad-symbols to discuss the FPGA symbols. At least on Artix-7 they need some fixes which I'm up for doing if others agree.

poeschlr:
Please make any changes to existing footprints in a separate pull request should they be required.

herostrat:
@evanshultz this can be closed, as the packages were merged in https://github.com/KiCad/kicad-footprints/pull/1596

@dlharmon Sorry for taking this over, I was already in the flow and wanted to complete the Xilinx packages (PS: I really enjoyed your presentation at KiCon!)

evanshultz:
@dlharmon 
As you can see at https://github.com/KiCad/kicad-footprints/issues/1560, there is still a little bit left to do with regards to tidying up older footprints, but all new Xilinx footprints are hopefully in the lib already. Would you mind reviewing this to see if everything looks good to you and close this PR if so? You put a lot of good though and effort into this and I don't want to close your PR without giving you a chance to take a look at what has been done first. Thank you!

dlharmon:
Looks good to me. Glad you enjoyed the KiCon presentation.

PR #1169, opened 2018-12-14
Add Bourns TC33X SMD potentiometer

MatthewStephenson:
Bourns TC33C SMD Potentiometer

Datasheet: [https://www.bourns.com/pdfs/TC33.pdf](https://www.bourns.com/pdfs/TC33.pdf)

image

The automated check is failing on the "centre of footprint" check by 0.1mm, because I've placed the anchor at the centre of the rotating part, as per the datasheet diagram, and the pads are not symmetrical in size or placement.

As always, comments and suggestions welcomed.


poeschlr:
Could you add the suffix `_Vertical` to the footprint name?

We might want to indicate the circular area required for the "wiper" (3.6mm diameter). Maybe add a circle to the comments layer. Optionally also some note on the comments layer telling the user that the wiper can be anywhere inside this circle. (The courtyard area is already bigger than this but it might still be valuable information to the user.)

For reference the dimensioned drawing:
screenshot from 2018-12-15 18-34-29

This is an alternative part to the obsolete one that would have been added in https://github.com/KiCad/kicad-footprints/pull/1143

MatthewStephenson:
Thanks for the review, I've made the requested changes. I kept the text about the wiper small so it fits inside the footprint which means you have to zoom in quite far to read it. Hope that was the right choice.

image

Out of interest, what tool did you use to make the dimensioned drawing?

poeschlr:
I would suggest to move the circle to the Dwgs.User layer.

I have a tutorial for creating these drawings: https://forum.kicad.info/t/how-to-check-footprint-correctness/9279

MatthewStephenson:
Ok, moved it to the Dwgs.User layer. You did say "add a circle to the comments layer"!

PR #1170, opened 2018-12-15
Xbga 121 10x10mm

Misca1234:
Added XBGA-121_10x10mm_Layout11x11_P0.8mm

Used by PIC24FJ256DA210-IBG in push
https://github.com/KiCad/kicad-symbols/pull/1266
http://ww1.microchip.com/downloads/en/DeviceDoc/39969b.pdf

Script push
https://github.com/pointhi/kicad-footprint-generator/pull/241

Foot print push
https://github.com/KiCad/kicad-footprints/pull/1170

------------


poeschlr:
How did you determine the pad size and mask clearance?

Misca1234:
copy paste from an existing one, are they to off ?

poeschlr:
Just wondering as i have no clue how to determine it for my self. (BGA is a bit of black magic to me at this point in time.)
We might want to double check with an application note.

Misca1234:
The closest I founf at microchip is this 
http://ww1.microchip.com/downloads/en/DeviceDoc/121L_TFBGA_10x10_BG_C04-0148F.pdf

So pad size should be 0.32, should I just keep the mask clearance or change it ?

Misca1234:
almost all paste_margin in the .yml file are 0.000001 that is 100 times more narrow

poeschlr:
Paste margins are set to that value to ensure that paste has basically the same size as copper.
The clearance above is for mask i would assume.

poeschlr:
Looks good, thanks

Misca1234:
They are moved now

bild

evanshultz:
https://blogs.mentor.com/tom-hausherr/blog/2011/01/08/pcb-design-perfection-starts-in-the-cad-library-part-9-bga-components/ has two charts for packages with balls. They are reproduced here for clarity, but read the whole article because there are lots of things you may not get just from looking at the charts (and I'll refrain from copying more of the article here).

Collapsible balls:
image

Non-collapsible balls:
image

I think we should be able to use this to determine footprint info for BGAs without it. Perhaps this, with a bit of supporting info, should be in KLC? @poeschlr ?

BTW, this agrees with IPC 7095 is anybody has that and uses it as their reference.

PR #1172, opened 2018-12-16
[Fix] fixes for remaining issues of sot-1123

poeschlr:
Fixes remaining issues of #977
- courtyard clearance on the right side
- increased silk clearance
- pad sizes

screenshot from 2018-12-16 14-51-08


Misca1234:
All looks ok

PR #1173, opened 2018-12-16
Connector_Audio: Add Jack_3.5mm_CUI_SJ-3523-SMT_Horizontal/Jack_3.5mm_CUI_SJ-3524-SMT_Horizontal

gmsotavio:
Add Jack_3.5mm_CUI_SJ-3523-SMT_Horizontal
Add Jack_3.5mm_CUI_SJ-3524-SMT_Horizontal

Datasheet: [https://www.cui.com/product/resource/sj-352x-smt-series.pdf](https://www.cui.com/product/resource/sj-352x-smt-series.pdf)

jack_3 5mm_cui_sj-3523-smt_horizontal
jack_3 5mm_cui_sj-3524-smt_horizontal

Regards
Otavio Augusto Gomes

------------


gmsotavio:
I have been contributing with footprints which I did. However I have not been using pick and place machine, so a precise anchor in the center of the footprint does not matter. In the footprints of the PR, the anchor was placed in the middle of F.Fab. I do not know if this is right. Datasheet is not clear about that. Some guidance to avoid this kind of mistake in the future would be a great help.

aewallin:
From the datasheet I did not understand if the model without a switch (your top screenshot) has the 4th pad also, but it is just not connected?
Just looking at the datasheet I would have guessed that both models have 4 pads??

poeschlr:
NP could mean that this pin is not included. (it could mean "No Pad")
Is there anyplace where we can get additional information about that? (example a distributor with a picture of the part. Also ok would be if @gmsotavio used the part in a design and can give us their word that that pin does not exist in the part.)

---

regarding pick and place center: The center is kind of given in the drawing on page 4. However there are many measurements missing to map this pick and place point to the part itself.
I would assume from the drawing that it is in the middle of the overall length which includes the back lead (but not the pad) and in the middle of the overall width (including the leads on both sides. but again not the pads.)

---

I did not yet check anything else. (only answering open questions right now.)

gmsotavio:
>From the datasheet I did not understand if the model without a switch (your top screenshot) has the 4th pad also, but it is just not connected?
Just looking at the datasheet I would have guessed that both models have 4 pads??

>NP could mean that this pin is not included. (it could mean "No Pad")
Is there anyplace where we can get additional information about that? (example a distributor with a picture of the part. Also ok would be if @gmsotavio used the part in a design and can give us their word that that pin does not exist in the part.)

@aewallin @poeschlr, the datasheet by itself is not much clear about the fourth pad, but I can confirm that It is not present (NP) in SJ-352**3**-SMT version. I used this part in a past project. If you need some kind of official documentation to verify this information, you can refer to 3D step model of this part provided by manufacturer at [https://www.cui.com/product/resource/3dmodel/sj-3523-smt-tr](https://www.cui.com/product/resource/3dmodel/sj-3523-smt-tr)

screenshot_2018-12-17_21-55-05
screenshot_2018-12-17_22-06-27

>regarding pick and place center: The center is kind of given in the drawing on page 4. However there are many measurements missing to map this pick and place point to the part itself.
I would assume from the drawing that it is in the middle of the overall length which includes the back lead (but not the pad) and in the middle of the overall width (including the leads on both sides. but again not the pads.)

I assumed the same thinking in the footprints I've submitted.
bitmap

I asked to CUI engineers about that. They often contact me back within 1 or 2 business days.


>I did not yet check anything else. (only answering open questions right now.)
No problem. Thank you!

gmsotavio:
The manufacturer contacted me back today. I can assume that the current pick and place position is correct, like depicted in my previous comment.

evanshultz:
Looks good to me. Thanks!
image

gmsotavio:
> Courtyard should be 0.5mm away from pads or body for connectors. It's too close.

I agree. Sorry, I have not realized that for connectors the Courtyard clearance is greater than other  type of footprints. I get fixed it in my local repo.

> I would like if you could please rotate all pads so they're at 0degrees and not 90degrees.

Sure. I drew the footprint in another direction, afterwards I rotated the entire body of the footprint, what explains the non zero degree. I get fixed it in my local repo for now.

>We haven't been doing named pads too long, but I should think that "T" is the equivalent of pin 1. (That's what I did in my similar PR). So then please put pad T in the top left corner.

**F4.2 - Pin 1 should be located at the top left: Footprints should be oriented such that Pin 1 is located in the upper left corner (IPC-7351).**

Is always TIP (T) the pin 1 in KLC convetion ? I followed the numeration from manufacturer datasheet.
screenshot_2018-12-18_22-09-08

After your answer I will commit the changes to PR.

gmsotavio:
> By the way, since this is your first PR I should say thank you for contributing to KiCad!

Thanks. It is a pleasure to contribute to this project!

evanshultz:
Thanks!

As mentioned, we haven't been doing named pads for long but that's the mapping convention I've used for audio connectors I submitted without any objections (pin T ~= pin 1). If no other librarian chimes in let's go with that.

gmsotavio:
I not fully agree with the idea to use named pads, because it is not standard across entire library. But, I think that in the particular case of TRRS connectors it is a good idea, because different manufactures give different numerations for terminals with same functionality, what allow us to use the same symbol for footprints regardless of manufacturer.

evanshultz:
The named pads is exactly to solve this issue, and used in only a few places like these audio connectors. I'm not sure if you're stating that your opinion (which matches the existing state of the library) or if you have a problem with something. Please clarify.

To me, the pin names you have are correct and all looks well with them.

gmsotavio:
> The named pads is exactly to solve this issue, and used in only a few places like these audio connectors. I'm not sure if you're stating that your opinion (which matches the existing state of the library) or if you have a problem with something. Please clarify.
> 
> To me, the pin names you have are correct and all looks well with them.

Sorry if I caused any misinterpretation. As you supposed, I have stated my opinion (which matches the existing state of the library), in response to your previous statement.

> **As mentioned, we haven't been doing named pads for long but that's the mapping convention** I've used for audio connectors I submitted without any objections (pin T ~= pin 1). If no other librarian chimes in let's go with that.

poeschlr:
The audio connector mapping stuff really needs to be documented in the KLC.
This is such that future librarians know what is going on and for users to know what to expect.

evanshultz:
@poeschlr 
Agreed that KLC needs to capture this.

@gmsotavio 
1. The NPTHs are rotated.
1. In the future would you please do incremental commits instead of force pushes? This makes it easier to track changes in the PR. I will assume the dimensioned drawings I made above are still valid.

I think everything else is ready to merge, right?

gmsotavio:
@evanshultz 

> 1. The NPTHs are rotated.

You are right. I will provide the fix.
> 2. In the future would you please do incremental commits instead of force pushes? This makes it easier to track changes in the PR. I will assume the dimensioned drawings I made above are still valid.

Ok. I will follow your recomendation in future Pull Request.

I have another PR reviwed by you, with pending changes. The delay is due to my heavy workload in last weaks. Give me some days to solve any pending changes in this and another PR.

evanshultz:
@gmsotavio 
Sure. Thanks for the update.

Note that 5.1 is expected to be released shortly, so if you would like to include your contributions in the official 5.1 library it would be best to get them merged in the next week or so. It could be longer but we can't know for sure. And of course they can be merged later, too.

gmsotavio:
@evanshultz 
I've made the proposed changes. Is there something more that do you think which need to be fixed in this PR ? Thank you for your careful review!

evanshultz:
Nope. This is good. Thank you!

PR #1177, opened 2018-12-17
Newly scripted footprints for LQFP 44 and 48

poeschlr:
This replaces the 44 and 48 pin versions of the LQFP footprints with
scripted ones. Script PR: pointhi/kicad-footprint-generator#245

# LQFP-44_10x10mm_P0.8mm
Original documents not found. Used one from the symbols linking this footprint: https://www.nxp.com/files-static/shared/doc/package_info/98ASS23225W.pdf?&fsrch=1

screenshot from 2018-12-18 00-31-20

```yaml
LQFP-44_10x10mm_P0.8mm:
  size_source: 'https://www.nxp.com/files-static/shared/doc/package_info/98ASS23225W.pdf?&fsrch=1'
  body_size_x:
    nominal: 10
  body_size_y:
    nominal: 10
  overall_size_x:
    nominal: 12
  overall_size_y:
    nominal: 12
  lead_width:
    minimum: 0.3
    maximum: 0.45
  lead_len:
    minimum: 0.45
    maximum: 0.75
  pitch: 0.8
  num_pins_x: 11
  num_pins_y: 11
```
# LQFP-48_7x7mm_P0.5mm
Original document not found. Used one of the symbol datasheets that link to this footprint: https://www.analog.com/media/en/technical-documentation/data-sheets/ltc2358-16.pdf
screenshot from 2018-12-18 00-37-13
```yaml
LQFP-48_7x7mm_P0.5mm:
  size_source: 'https://www.analog.com/media/en/technical-documentation/data-sheets/ltc2358-16.pdf'
  body_size_x:
    nominal: 7
  body_size_y:
    nominal: 7
  overall_size_x:
    nominal: 9
  overall_size_y:
    nominal: 9
  lead_width:
    minimum: 0.17
    maximum: 0.27
  lead_len:
    minimum: 0.45
    maximum: 0.75
  pitch: 0.5
  num_pins_x: 12
  num_pins_y: 12
```


evanshultz:
Only because I found the NXP LQFP-44 package drawing so ugly, I looked at the other parts using this footprint in our library. https://www.parallax.com/sites/default/files/downloads/P8X32A-Propeller-Datasheet-v1.4.0_0.pdf is much easier to read to me, but it has leads that are slightly tighter at 0.3-0.4mm wide.

I then searched at On Semi because I know they make this package:
- https://www.onsemi.com/pub/Collateral/561AX.PDF
- https://www.onsemi.com/pub/Collateral/561BK.PDF
- https://www.onsemi.com/pub/Collateral/561AA.PDF

Those show 0.3-0.45mm (as above) and 0.22-0.38mm in lead width.

So what should we do here? Would you like to open the YAML data up a bit to accommodate the wider lead width range? Leave it alone for now? Further specify footprint names to capture lead width?

poeschlr:
I am tempted to say "lets go with the newest" of these. which would either be nxp or on semi https://www.onsemi.com/pub/Collateral/561BK.PDF

These do not agree at all with each other. (They are the exact opposites)
I mainly used the nxp one as it was the one referenced in the original footprint.

In the end it looks like qfp might not be as standardized as one might hope. So any generic footprint will be a compromise solution. (I don't really care which part is the basis for this. The on semi one will definitely result in a larger pad to pad clearance which could be seen as a bonus for users of cheap fabs.)

PR #1181, opened 2018-12-18
Relay: Added Finder 36.11 footprints (copied from Sanyou relays).

mgottschlag:
The shape is identical, but the pins are numbered differently.

The footprint checker complains about some rules, but (except for F7.2) the original Sanyou relay footprint violates the same rules. I kept the footprint identical for consistency.

Due to the different pin names, the footprint checker also complains about pin A1 not being at the center. Should the symbol be moved? I kept the offset identical for now for better exchangeability with the Sanyou footprint.

------------

https://www.finder-relais.net/en/finder-relays-series-36.pdf

2018-12-18-090510_662x561_scrot
2018-12-18-090600_683x582_scrot

(no 3d models yet)


poeschlr:
Could you change the pad shape of A1 to be square or rounded rectangle to differentiate it from all others making debugging easier?
Plus add pin 1 markers pointing to that pin.

Some of the dimensions are also quite wrong.
releay svg

PR #1182, opened 2018-12-18
IPC conformal fiducials 

herostrat:
Find the previous discussion here:
https://github.com/KiCad/kicad-footprints/issues/1120

This PR:
 - moves old fiducials (non IPC conformal) to obsolte folder
 - adds multiple IPC conformal fiducials (only circular ones atm)

Some example images.
grafik

grafik

grafik

------------

Thanks for creating a pull request to contribute to the KiCad libraries! To speed up integration of your PR, please check the following items:

- [ ] Provide a URL to a datasheet for the footprint(s) you are contributing
- [ ] An example screenshot image is very helpful 
- [ ] If there are matching symbol or 3D model pull requests, provide link(s) as appropriate
- [ ] Check the output of the Travis automated check scripts - fix any errors as required

PR #1186, opened 2018-12-18
IPC conformal fiducials - update

herostrat:
Because I butchered my PR, here is a corrected one, where no wrong commits are included.
Previous one can be found here: https://github.com/KiCad/kicad-footprints/pull/1182
Corresponding issue can be found here: https://github.com/KiCad/kicad-footprints/issues/1120

Also changed the attributes and tags according to a previous PR, found here: https://github.com/KiCad/kicad-footprints/pull/1104
------------


evanshultz:
These are nice. Thanks!

- For `Fiducial_0.75mm_Mask2.25mm` we could scoot the courtyard radius to 1.38mm to put it on grid.
- Do you know why CI is complaining about the 1mm ones? They do have a circle on the fab layer.
- Please name the new folder with a capital "O" (like the "Sources" folder).

herostrat:
You're very welcome

 - done
 - I have no idea, I was confused when I checked the travis details. There is imho no difference as with the others - at least I can no spot one.
As you rightfully saw, there is a fab layer
 - done, I thought I use the same spelling as in the symbol library

evanshultz:
Weird. Now there are no issues about the fab layers.

One thought before merging: What about `Obsolete/Fiducial/` instead of just dumping them into a folder one level deep. If other footprints are obsoleted, this will maintain some organization. And, it's most similar to the symbols (where one file holds many symbols, but here one folder holds many footprints).

herostrat:
Yes, I like this idea. 
I do not know if more footprints get obsolte in the future (with v6 on the way?) but overview is always better.

Probably a bad day from travis.
But in all seriousness, I can not think of a way that triggered this bug.

herostrat:
Yes, aren't all folders without content (except subfolders) flat folders?
Is this a problem?

One workaround is to just add a .gitkeep file into Obsolete

evanshultz:
I wasn't sure but still expected so descend through two folders even in the web interface. But it's OK on my local drive so I guess it's just flattening on GitHub. Sorry for the confusion.

PR #1194, opened 2018-12-19
Adding Celduc Reed Relay footprint

omogenot:
New Celduc DIL Reed Relay footprint

------------


poeschlr:
please include a screenshot and a datasheet link in the pull request comment section

omogenot:
Datasheet link is a download link: F https://www.e-catalogue.celduc-relais.com/en/146-reed-relays-in-dip-enclosure (pick one).
Screenshot: 
capture d ecran 2018-12-19 a 21 36 23

poeschlr:
I doubt the footprint name is specific enough. Is there a series name that can be used? Otherwise we might resort to describing the parts similar to DIP packages. (We however don't really have a good naming convention for multiple missing pads)

The courtyard clearance does not seem to be correct. There should be 0.25mm between the pads and the center of the courtyard line.

Move the silk screen reference out from under the part (must be visible after assembly)
Silk screen lines should not be on top of pads. (There should be clearance between silk and copper. I suggest to aim for about 0.2mm)

We typically have the value field outside the part outline (towards the bottom)
A pin 1 marker is missing on the fab layer.
A pin 1 marker visible after assembly is missing from silk.

I did not yet check dimensions

omogenot:
I cannot agree more...
Especially because these Celduc relays have their pendants at Littelfuse/Hamlin:
https://www.littelfuse.com/~/media/electronics/datasheets/reed_relays/littelfuse_reed_relays_he700_datasheet.pdf.pdf
Ca we call them DIP Reed Relays?  By the way, looks like there are 2 types of footprints (see Littelfuse data sheet).
Can we call them DIP_8Pin_Reed_Relay and DIP_4Pin_Read_Relay ? just to agree about names before I redesign them accordingly?

myfreescalewebpage:
@omogenot why closing the PR ?

PR #1196, opened 2018-12-20
OptoDevice: Add APDS-9306-065

antoniovazquezblanco:
https://docs.broadcom.com/docs/AV02-4755EN

imagen

For symbol https://github.com/KiCad/kicad-symbols/pull/1294

No 3D PR.

------------


antoniovazquezblanco:
Scripts are checking all the elements inside OptoDevice although I have only added 1 element.

poeschlr:
Do you think this can be scripted with the DFN/QFN generator?

antoniovazquezblanco:
I probably could be done but as always I cannot find the time to setup the generator and play with it.

I will try to provide a scripted version for this and the other variant of this sensor. Before I do that I would like to discuss FP naming. See https://github.com/KiCad/kicad-symbols/pull/1294#issuecomment-449776511

evanshultz:
At https://www.broadcom.com/products/optical-sensors/ambient-light-photo-sensors/ I count 3 packages used in 13 parts, so combining them seems good. I don't know if other part categories or other vendors use the same package, but it's possible.

Let's keep Broadcom in the name unless we know this package is used by other vendors.

So how about `Broadcom_DFN-6_2.0x2.0mm`? The other would be `Broadcom_LGA-8_2.0x2.0mm`.

The scripts aren't too tough to use and ought to make this much faster than building footprints by hand. Let us know if you're stuck, but you've written a few footprints scripts so I think you'll be fine just running this one. :)

antoniovazquezblanco:
I've been playing with the DFN script a little bit.

https://github.com/pointhi/kicad-footprint-generator/pull/269

evanshultz:
New script PR at https://github.com/pointhi/kicad-footprint-generator/pull/293.

evanshultz:
Pads are longer than Broadcom says and not as wide, but should be OK. Review comments in the script PR.

evanshultz:
The DFN-6 looks OK to me, except for the weird double space in the description.

The LGA-8 seems to be missing one parameter which I commented in the script PR.

antoniovazquezblanco:
Updated

evanshultz:
Because they were missing or stale, here are current screenshots:
image
image

I'll go with it. Thanks!

antoniovazquezblanco:
Thanks!

PR #1199, opened 2018-12-20
Scripted replacements for qfn-20-1ep 3x3 ep1.8x1.8

poeschlr:
This footprint is wrongly used by some parts in the lib.
The pull request will in the end add all footprints that are required to replace it with the correct ones. (starting PR that early to be able to reference it in the commit messages for the symbol side)


evanshultz:
`QFN-16-1EP_3x3mm_P0.5mm_EP1.45x1.45mm`: lead_len tolerance should be 0.1mm.

`QFN-16-1EP_3x3mm_P0.5mm_EP1.7x1.7mm` shows 1.7mm on the reference footprint, but the package's thermal pad is 1.5mm min and 1.8mm max, which is 1.65mm nom. Before you said go with the datasheet size so I'll merge, but it's a bit odd to me to deviate in just this one place.

Also, the datasheet can be simplified to https://www.st.com/resource/en/datasheet/tsv521.pdf.

`TQFN-16-1EP_3x3mm_P0.5mm_EP1.23x1.23mm`: The nominal thermal pad size of the T1633-5 package is 1.1mm, unless I'm misreading it. Why is it 1.23mm in this footprint?

`VQFN-16-1EP_3x3mm_P0.5mm_EP1.45x1.45mm` is referencing a drawing with 20 leads so I can't check it.

`VQFN-16-1EP_3x3mm_P0.5mm_EP1.68x1.68mm`: The EP size tolerance should be 0.07mm.

poeschlr:
Correct datasheet for `VQFN-16-1EP_3x3mm_P0.5mm_EP1.45x1.45mm`: http://www.ti.com/lit/ds/sbos354a/sbos354a.pdf

`TQFN-16-1EP_3x3mm_P0.5mm_EP1.23x1.23mm`: There is a land pattern drawing linked in the datasheet: https://pdfserv.maximintegrated.com/land_patterns/90-0032.PDF (Will add that as a second reference file)

`QFN-16-1EP_3x3mm_P0.5mm_EP1.7x1.7mm` it was mainly your influence in past submissions of mine that got me to accept the suggested ep size from the datasheet over always using the nominal size. (For better functionality i even added a separate field for that to the generator. This will allow the yaml file to be used for both footprint and 3d model generation)

`QFN-16-1EP_3x3mm_P0.5mm_EP1.45x1.45mm` good catch

`VQFN-16-1EP_3x3mm_P0.5mm_EP1.68x1.68mm` good catch

evanshultz:
OK. I guess I can't make up my mind about EP sizing then. All looks good. Fixing up all the QFN stuff was a lot of work so thank you!

poeschlr:
By the way i documented our current way of doing things in the KLC.
If you want you can look over it. http://kicad-pcb.org/libraries/klc/F6.3/

evanshultz:
There is a typo of "pads" in the final 2.b. of the KLC update. And it could probably a written a bit more clearly. But regardless, nice update!

PR #1203, opened 2018-12-21
Removing impossible qfn footprint

poeschlr:
3x3mm body size with 2.7x2.7mm ep size would not really be possible.
This footprint is simply named wrong as it has a 1.8x1.8mm ep. (such a
footprint is already scripted in the lib)


evanshultz:
OK. I will assume you checked that no symbol is referencing this footprint.

PR #1204, opened 2018-12-21
Display_7Segment: Add DE122, a 6-digit 7-segment LCD

piotr-wiszowaty:
Datasheet: http://www.display-elektronik.de/filter/DE122-RS-20_635.pdf
Symbol: https://github.com/KiCad/kicad-symbols/pull/1309
de122-fp

------------


aewallin:
would it be useful to indicate the position of the actual digits on f.fab? The the PCB-drawing could be used for enclosure design where we want an opening in the enclosure for the display?

piotr-wiszowaty:
Footprint with digits outline:

de122-fp-2

myfreescalewebpage:
Hi @piotr-wiszowaty, thanks for contributing,

A single comment I have during my review:

- [ ] For consistency with other DE footprints in the library, can you change the size of the pin 1 with Y=1.524mm as shown below ?

capture

I'm thinking to the usage of the Dwgs.User rectangle and I don't know how it is possible to use it ? @aewallin maybe you can explain how you can use this layer to know where you should open the enclosure ? The rectangle isn't printed on the PCB and has no chance to be visible on 3D view...

Cheers,
Joel

myfreescalewebpage:
@piotr-wiszowaty thanks. Just want to wait feedback of @aewallin about the User layer.

aewallin:
For designing enclosures I think the display-area-outline is useful. I don't have any comments on what layer it should be on. The workflow will differ case-by-case, but at minimum you can take distance-measurements from mounting-holes and the PCB-outline to the display-area once you have a finished board.

myfreescalewebpage:
Okay, thanks @aewallin. I have just review the KLC about this kind of information on the footprint, and particularly I have read KLC F4.5 (http://kicad-pcb.org/libraries/klc/F4.5/) in which there is:

> Descriptive text may be included on the Cmts.User layer. This text is used to provide further information about keepout requirements to the user.

So I propose @piotr-wiszowaty to move the display area outline rectangle from the Dwgs.User to the Cmts.User layer. If you wish to add some digits, still on the Cmts.User layer, to look like the drawing page 5 of the datasheet, it's also possible but not mandatory.

Thanks,
Joel

PR #1206, opened 2018-12-22
Added DCC6C footprint for saw filters (like EPCOS)

omogenot:
DCC6C footprint for saw filters

------------
Datasheet : https://en.rf360jv.com/inf/40/ds/ae/B3717.pdf
3D package under https://github.com/KiCad/kicad-packages3D/pull/485


omogenot:
Travis-Ci says that anchor is not in the middle, but I guess the test fails the test because pad1 is offset by 0.3 mm, which is not a common situation.

poeschlr:
A screenshot of the footprint here in the pull request would help a lot.

omogenot:
Here you go ....

dcc6c footprint

omogenot:
Don't you have a cue card (just a one page quick guide summary of list items) of requirements? That would be very helpful to limit the try/fail cycle (and the frustration associated with it I must admit).
I've been using Kicad for many years and designed these symbols at the time no Fab nor CYrd layers where existing and never had to design or used them.

aewallin:
F1 to F9 of KLC. Not sure if there is a summary/shorter version..
http://kicad-pcb.org/libraries/klc/

omogenot:
New trial ;-)

dcc6c footprint

aewallin:
- I agree that the footprint-editor should use default widhts/settings.
- a "check KLC" button would also be very handy - it would report problems immediately
- I don't know if the KLC can be shortened to one page, or if someone already made an effort to do this.

For your new screenshot check F5.1.3 "For SMD footprints, silkscreen must be fully visible after boards assembly (no silkscreen allowed under component)"  
Maybe you can use this style to indicate pin1 on silk: https://user-images.githubusercontent.com/4781841/38120284-75f8bb9a-3413-11e8-98ce-87a61a77e9fb.png

sorry no review of geometry/dimensions by me yet..

aewallin:
The courtyard-tolerance is quite tight. It is probably best to use an automated tool for this.
maybe the "--fix" option from the library-utils: https://github.com/KiCad/kicad-library-utils

I also created my own courtyard-maker, because getting it right by hand is so frustrating:
https://gist.github.com/aewallin/a9b3d785c13584c0bac84e6099a8f5ed
(this may or may not be the same that "--fix" option does - I have not checked)

omogenot:
Well, definitely the silkscreen will still be visible after assembly. And the courtyard is probably larger enough looking at the resulting 3D view.

dcc6c 3d

Using the type of marker you suggested was resulting in getting questions from my factories. They don't like this type of marking and prefer dots or bevels.

poeschlr:
The fab outline should represent the part body size not some framing around the pads.

poeschlr:
The datasheet does not contain a suggested footprint. How did you determine the required pad sizes?

omogenot:
Just to illustrate what I was just saying (frustrating). Different grids for different layers. Difficult to remember them all (especially for an old man, close to retirement, like me)..

dcc6c footprint

omogenot:
> The datasheet does not contain a suggested footprint. How did you determine the required pad sizes?

Google: DCC6C landing pattern ==>
http://52ebad10ee97eea25d5e-d7d40819259e7d3022d9ad53e3694148.r84.cf3.rackcdn.com/UK_ACT_ACT9401-866MHz-DCC6C-F866BMB_DS.pdf

I've been using successfully this footprint for production for the last 8 years.

omogenot:
> The fab outline should represent the part body size not some framing around the pads.

So, we agree that the Fab layer would cross the pads?

omogenot:
dcc6c footprint

poeschlr:
Fab is not printed onto the pcb so there is no problem with crossing the pads.
The pin 1 chamfer should be 25% of the smallest body dimension or at most 0.25mm

You really should read the full KLC not just a summary it seems :grin:

omogenot:
> Regarding you being successful with this patter: Does not proof anything to be honest. I don't know which manufacturing methods you use. Others can not know anything about that. We therefore prefer the use of industry standards (IPC) or manufacturer suggestions (datasheet, application note).
> 
> Please link the actual document you used to design the footprint both in the pull request description and inside the footprint description.
> 

Point taken !
Last trial. After that, just cancel my PR.
I'll keep my symbols and foot prints for me in the future, not being adapted to a so strict list of requirements. I understand that things must be controlled, however sometimes, some flexibility isn't bad as well...
Have a nice Sunday, and thanks a lot to both of you for your contant help and support.
Best Regard and Happy new year.

poeschlr:
The strict requirements are there such that users can trust them

If you add a footprint from the official lib you would like it to be checked against the full set of rules. Not every user will care about all our rules. (Some might not care about courtyard at all, others might not really care about the exact dimensions on the fab layer, .... But after enough usages every rule will be important to somebody.)

This is just how it is when you want your stuff to be useful to others. There is a bit more work required than using it just for yourself. (You know what your requirements are. Sharing something with others means documenting every parameter. This is what the KLC is there for. If we then randomly decide some rules are not as important then how do users know which footprint follows the rules they care about?)

omogenot:
Peace !
I had in the past to throw away 100 PCB/A because at some time the official QFN28 with 1EP footprint was wrong and all the pads were too closed to the exposed pad and the device was in short circuit in the end. As fixing it would have required to unsolder the component, slightly cut all pads around and re-solder the component, this manual operation would have cost too much.
I bite the bullet, thrown all the boards to the trash, made my own footprint. I did not complain to you.
What I meant by this, is that even if you try to be as strict as you can, you can't embrasse all cases.
What was really frustrating was to argue about the length of the bevel needing to be 0.75mm instead of the 0.45mm as it was.
By the way the KLC F5.2 2.b says:

> Bevel should be 1mm or 25% of package size (whichever is smaller)

It says should, not must :-D (Just joking !)

Just cancel this PR and we move on. It does not hurt my feelings, I didn't do all this to be famous, but rather to kindly contribute and not just stay as a consumer of your hard work.
I tried, I failed. No problems. I just understand that I'm not good at making drawings (which by the way I knew for a long time ;-) ). Don't worry I'll get over it :-D
Peace again. And Happy new year.

poeschlr:
Now with the datasheet present i can do a proper review:

Add the Filter prefix to the footprint name (to fit other footprints in the same lib)

The chamfer line on silk is a bit close to the pad. (There is less than 0.05 mm clearance. I doubt any manufacturer can reliably add this silk line.)

The top fab line is not quite horizontal (left point is out by 0.01mm)

The pad positions are not quite right (At least they do not agree with the datasheet you linked above)
The courtyard clearance is not quite right. (But if you fix the pad position that will change.)

screenshot from 2018-12-23 11-06-11

poeschlr:
> It says should, not must :-D (Just joking !)

Yup we know that. The person writing it was simple too polite. There is an open issue regarding the proper use of a formal language. We could need help with that as well: 

---

Regarding your destroyed board: Do you remember which footprint it was? Is that one still in the lib?

We started an effort to create IPC compliant footprint scripts. We have not yet transferred over all footprints that can be generated with them. (again something where we could need help)
In this endeavor i discovered that the standards where a lot lower in the past (well i already kind of knew that. But i never thought it could have been that bad. Especially EP sizes where really wrong in a lot of cases.)

omogenot:
I probably mixed up between symbol and footprint requirements. You see too complicated for a simple old man as me.
Never mind, just cancel the whole thing.

poeschlr:
SX1276 is not in the lib. However SX1272 is. And that one is assigned to the wrong footprint.
Seems i am right in suspecting that a lot of old symbols using qfn footprints did not particularly care about assigning the correct ep size. (Well it seems i still have a bit of work to do.)

By the way: If you ever find such a problem again then please report it. We can not fix things we do not know about.

PR #1208, opened 2018-12-23
Re export the qfn-2x footprints

poeschlr:
This updates the qfn-2x footprints with the fixes added to the qfn
generator since they where last updated.
(Mainly fixes to chamfered pads. Automated pad to ep clearance.)


evanshultz:
Since these are scripted I'm just going to merge and go with it. Most seem to only change the generation date and I didn't individually check any of the others.

PR #1215, opened 2018-12-25
Sip package link fixes

Misca1234:
SIP4_Sharp-SSR_P7.62mm_Angled_NoHole
Changed the 3D model path to
${KISYS3DMOD}/Package_SIP.3dshapes/SIP4_Sharp-SSR_P7.62mm_Angled.wrl
So it is the same as SIP4_Sharp-SSR_P7.62mm_Angled

SLA704XM
Added data sheet link

PowerIntegrations_eSIP-7F
Re created from scratch, the original foot print does not match the component,
I guess the original creator misunderstood the drawings (it is the odd "rotate down into table" format) in combination with the belief that the metal pad should be against the PCB

The actual component look like this
bild

old foot print
bild

New

bild

bild

bild

------------


poeschlr:
The courtyard clearance is not quite right

screenshot from 2019-01-16 22-42-07

evanshultz:
The Power Integrations footprint doesn't have a pin 1 mark on fab or silk layers. I rather liked the leads being shown, but not a big deal.

Shouldn't `SLA704XM` be `Sumzi_SLA704xM`? There are symbols that would be affected. Perhaps the impact is small enough to do that rename now, though it's not terribly important?

Misca1234:
I was not sure about that pin 1 marker because it is not possible to mount this component in the wrong way

Misca1234:
pin 1 marker added
right f.crtyrd line moved 0.01

bild

Misca1234:
Not sure how to do this one, so this became the result

bild

poeschlr:
Looks good, thanks

PR #1217, opened 2018-12-26
add Fischer FK244_13_D2_PAK SMD heatsink

aewallin:
https://www.fischerelektronik.de/web_fischer/en_GB/heatsinks/C04/Heatsinks%20for%20D%20PAK%20and%20others/PR/FK244_13_D2_PAK_/$productCard/parameters/index.xhtml

Heatsink for e.g. TO-263 parts.
- I'm not sure how to indicate to the user that the "wings" of the part are actually 10mm off the board. Should the silk/f.fab outline just the parts touching the PCB or the whole part as viewed from above?
- There is no recommended land-pattern so the pads are positioned with about 1mm overlap/tolerance sideways and no overlap length-wise.

fischer_fp
fk244_13d2pak

------------


poeschlr:
You could add  a text on the comments layer to tell the user that the wings are high up. But leave the courtyard unchanged as one can not really place components there and expect pick and place machines to be able to place both the component below it and the heat sink without special treatment. (One would need to somehow communicate that the component below should be placed first. Meaning the DRC error plus the text on the comments layer should tell the user everything they need to know.)

---

Could you link directly to the pdf drawing instead of the product page? https://www.fischerelektronik.de/pim/upload/fischerData/cadpdf/base/fk_244_13_d2_pak.pdf

The solder fillets seem a bit large to me. (>=0.5mm both for heel and toe.) Did you find some suggestion by the manufacturer or was this done by eye?

screenshot from 2019-01-25 16-47-24

aewallin:
- link directly to PDF in description.
- I added a text "H10mm" and a box on the cmts.user layer.
- pad sizes and positioning is my own estimate - I did not find a suggested land-pattern.

fischer

poeschlr:
I entered the dimensions of your part into the gullwing generator and got the following dimensions for the pads. (I included a tolerance of 0.1. No tolerance information is given in the datasheet. I think this would be a reasonable assumption for such a part.)

Position x: 7.7
Size x: 3.25
Size y: 13.2

aewallin:
> Position x: 7.7
> Size x: 3.25
> Size y: 13.2

Done. Also new 0.25mm courtyard. I did not change f.fab or silk or comments-layer.

poeschlr:
Your pads are of different size now (y direction)

aewallin:
travis OK now for this footprint (but checks a lot of other ones - not sure why).

poeschlr:
That symbol could simply filter by library.

So add a filter to it in the form: `Heatsink:*`

poeschlr:
Looks good, thanks

PR #1219, opened 2018-12-27
Footprint for Quectel BC66

jneiva08:
Add footprint for Quectel BC66

Datasheet: https://www.quectel.com/UploadImage/Downlad/Quectel_BC66_Hardware_Design_V1.1.pdf
**Note:** to access datasheet needs to be register on Quectel website.

Screenshot:
bc66

------------


poeschlr:
I would suggest the use of rounded rectangle pads with 25% round radius ration (maximum 0.25mm)
This is in preference to the oval pad shapes shown in the datasheet. (As defined by IPC)

The datasheet contains following note:
> The module should be kept about 3mm away from other components on the host PCB.

We could implement this with the courtyard layer. (use a clearance of 3mm between the courtyard and the body.)

Some of the inner pads are positioned wrong
screenshot from 2019-01-23 22-15-21

poeschlr:
Why 29% roundrect radius? IPC would specify 25%

jneiva08:
the footprint have the following aspect:

bc66

Have fixed the inner pad, and moved back the roundrect radius to 25%, and added courtyard to 3mm instead of 0.25mm.

poeschlr:
You used the 3mm clearance relative to the pads. I would read the datasheet as if it should be relative to the body.

The rest of your fixes look good.

screenshot from 2019-01-25 16-33-00

jneiva08:
Removed extra 1mm from F.CrtYd

modem

poeschlr:
Looks good now. thanks

jneiva08:
thanks, for your time :+1:

PR #1222, opened 2018-12-28
Added Footrpints for Hanrun HR911105A 10/100 Ethernet Magjack & Traco…

jamesrlucas:
… TMR 2WI Range of DC-DC Converters

Datasheets:

Hanrun HR911105A - http://www.kosmodrom.com.ua/pdf/HR911105A.pdf
TMR 2WI Converter - https://www.tracopower.com/products/tmr2wi.pdf

------------


myfreescalewebpage:
Hi @jamesrlucas , thanks for contributing,

Converter_DCDC_TRACO_TMR-2xxxxWI_THT.kicad_mod:
- [ ] A datasheet link (PDF) should be provided in Document link field
- [ ] The 3D path is not correct, a wrong device is displayed in 3D viewer (see also travis output)
- [ ] Silkscreen error in travis output to be corrected
- [ ] I do not agree with the Fab dimension, the part should be 9.3x26mm
- [ ] Drill size of the pad is very large, I suggest 0.8mm, what do you think ?

RJ45_Hanrun_HR911105A:
- [ ] Why "horizontal" in the name ?
- [ ] A datasheet link (PDF) should be provided in Document link field
- [ ] I do not agree with the Fab dimension, the part should be 15.88x21.4mm
- [ ] Drill size is not enough, you should use a 0.2mm margin (http://kicad-pcb.org/libraries/klc/F7.6/)
- [ ] Globally, sizes/positions to be reworked according to the datasheet (only reviewed the fab layer at this moment, the other must follow)

Cheers,
Joel

jamesrlucas:
Hello Joel,

Fixed everything you've asked me to :) Looks like a screwed up on a few bits there, my apologies.

Travis seems to be failing on stuff I haven't touched however...

For the Hanrun Magjack, I've measured the pins on the real thing, and it looks like the datasheet dimensions are the recommended PCB hole size, rather than the actual pin dimensions.

James

myfreescalewebpage:
Thanks.
No more comment on the DCDC converter.
For the RJ45 conenctor, I have still some issues regarding to the positions of the pin/holes etc. You should not truncate the values, keep 0.01mm precision has indicated in the datasheet. Also, the CrtYd layer margin is too much important, 0.25mm is enough (http://kicad-pcb.org/libraries/klc/F5.3/)
Joel

myfreescalewebpage:
For example:
- distance from left side to pin1
- Y position of pin 9/10/11/12 not always the same
- distance between mechanical hole
...
that's always a difference of about 0.01mm but it troubles the review because the drawing is not very easy. Maybe I will suggest to have a precision of 0.005mm to simplify the review...

jamesrlucas:
The holes sizes on the datasheet appear to be the recommended hole sizes - Ive measured them from the real thing as follows:

LED Pins - 0.6-0.7mm across the Diagonal
Mounting holes 3.05, but need to be a tighter fit to 'snap' in
Shielding Pins - 1.25mm
RJ Pins 0.4-0.5mm Dia

James

myfreescalewebpage:
Well, from the attached image I understand this is "Comp side view", so it's the dimensions of the part itself, not the recommended layout...

capture

jamesrlucas:
I just think that the connector will be very loose in the holes if they are increased any further. I've already got some PCBs in the post using this connector, I can send a photo over of how they fit in the holes if you like?

I've also just emailed Hanrun to clarify the datasheet, see what they come back with!

James

myfreescalewebpage:
> I've also just emailed Hanrun to clarify the datasheet, see what they come back with!

Great solution, will see if they respond.

> I've already got some PCBs in the post using this connector

Online or made yourself ?

jamesrlucas:
> > I've already got some PCBs in the post using this connector
> 
> Online or made yourself ?

Made myself using this footprint (I know some of the holes won't line quite line up!)

myfreescalewebpage:
Great job James, thanks, accepting the PR now !
Don't think we have a place for the datasheet....
Joel

jamesrlucas:
Awesome, thanks Joel!

PR #1227, opened 2018-12-28
Rerun script for qfn-4x footprints

poeschlr:
This updates the qfn-4x footpritns to the current state of the script.

Main changes are to chamfered pads (paste layer of exposed pads with
thermal vias)
But also to the way we address pages of datasheets and the new ep to
pad clearance calculation.


evanshultz:
Changes don't look major so I'll just merge it.

PR #1228, opened 2018-12-29
Delete unneeded qfn-44 8x8 (6.6x6.6 ep)

poeschlr:
This one footprint is not needed for any symbol after
https://github.com/KiCad/kicad-symbols/pull/1345 is merged.

I could not find a datasheet that contains a package with the dimensions
of this footprint. (The assigned datasheet was from another package)

---

This footprint should have been deleted in my first round of scripting qfns. Not sure where i got the parameters from for it. (Neither of the datasheets that was ever assigned to it had the parameters used.)

The closest i found was http://ww1.microchip.com/downloads/en/DeviceDoc/44L_QFN_8x8_ML_C04-0103D.pdf but that has a better fitting footprint already in the lib.


evanshultz:
The other PR is merged so merging this.

Issue #1141, opened 2018-12-05
Manufacturer-specific connectors with double series name


evanshultz:
Pulled from https://github.com/pointhi/kicad-footprint-generator/pull/224#issuecomment-442506843, which was spawned from https://github.com/KiCad/kicad-footprints/pull/1107.

@SchrodingersGat @poeschlr 
Some footprints fall into the "Manufacturer Specific Connectors" section of http://kicad-pcb.org/libraries/klc/F3.6/, I believe, where `Series` comes before `MPN`. Neither field is optional. This can result in text for `Series` and the same text again in `MPN`; this looks weird, especially if the the two pieces of text follow each other, but follows convention. I wonder if we should do something different.

None of the connectors in the Harwin repo have a datasheet or appear to be relatively modern so they don't make a good example, but let's look at the JST repo where I know many of them are recent. They have doubled text to follow KLC:
![image](https://user-images.githubusercontent.com/1936989/49164836-d6557900-f327-11e8-802d-faf66d2b1a9b.png)

That does look a bit bizarre and does not add clarity to me.

This convention, however, works well for Molex (https://github.com/KiCad/kicad-footprints/tree/master/Connector_Molex.pretty) where you have Series names like "CLIK-Mate" and "Micro-Fit_3.0" which are quite distinct from the MPN.

The Samtec repo (https://github.com/KiCad/kicad-footprints/tree/master/Connector_Samtec.pretty) doesn't double text in the scripted LSTM footprint names and I think it's clear.

Should KLC state that the Series field is required only if it is not a substring of the MPN? This would require updating a number of footprint file names if we change existing ones, but we could start a new precedent (perhaps updating all for v6?) to avoid doubled text. I agree that removing one "M20" looks better and doesn't lose anything, but since we're noticing this inconsistency now let's sort it out.

evanshultz:
@poeschlr @SchrodingersGat (and anybody else @Misca1234 @antoniovazquezblanco @DanSGiesbrecht @myfreescalewebpage )
Thoughts on the above?

If we decide to do any renaming, should that wait until after 5.1? Wait for 6?

What about renaming if the MPN above is in error? For example, ACH headers should start with "BM" but our library shows "BBM". See http://www.jst-mfg.com/product/pdf/eng/eACH.pdf and the screenshot above. Can that be done now or just after 5.1 since it's a bug?

Misca1234:
I don't mind it is double, better to stick to convention than create exceptions

[...]

PR #1238, opened 2018-12-30
Correct JST MPNs

### LINKED ISSUE #911: JST connectors have wrong PN
The EH series MPNs are shown in [the datasheet](http://www.jst-mfg.com/product/pdf/eng/eEH.pdf) here:
image

Here is our current library:
image

Issue 1: Our PNs have a leading zero when the pin count is <10 (`B03B-EH-A` instead of `B3B-EH-A`).

Issue 2: Keywords for the [vertical 3 pin header](https://github.com/KiCad/kicad-footprints/blob/master/Connector_JST.pretty/JST_EH_B03B-EH-A_1x03_P2.50mm_Vertical.kicad_mod) are `connector JST EH side entry`. The term "side entry" is no longer used in our library and also it's wrong (the vertical and horizontal footprint descriptions should be swapped).

These might apply to other series of JST connectors as well.

evanshultz:
Fixes #911

This all started from https://github.com/KiCad/kicad-footprints/issues/1141 and I documented the first couple issues at https://github.com/KiCad/kicad-footprints/issues/911. I've now looked at all JST connector footprints and fixed only the errors in MPN and make other non-breaking updates (fixed keywords, etc.).

General single-row connectors:
- Changed "side entry" to "horizontal" and "top entry" to "vertical" in keywords

ACH (http://www.jst-mfg.com/product/pdf/eng/eACH.pdf):
- MPN started with "BBM" and should be "BM"

EH (http://www.jst-mfg.com/product/pdf/eng/eEH.pdf):
- When the pin count is <10 there is no leading zero in the MPN
- There is no 16 pin version

J2100 (http://www.jst-mfg.com/product/pdf/eng/eJFA-J2000.pdf, page 7):
- Keying was not in the footprint name; this is unclear (and affects 3D models) so could somebody check me on this?
- Changed pin numbers a1 through b5 to 1 though 10 (the "Line No" in the datasheet is not related to the pin number)

PH (http://www.jst-mfg.com/product/pdf/eng/ePH.pdf):
- Abstracts away the exact PN for normal vs low insertion force THT headers, but I left it alone

SFH (http://www.jst-mfg.com/product/pdf/eng/eSFH.pdf):
- Corrected MPN

XH (http://www.jst-mfg.com/product/pdf/eng/eXH.pdf):
- When the pin count is <10 there is no leading zero in the MPN
- Removed JST_XH_B11B-XH-AM_1x11_P2.50mm_Vertical (doesn't exist)

ZE (http://www.jst-mfg.com/product/pdf/eng/eZE.pdf):
- For SMT, the top entry was massing the final "T" and the side entry had a final "T" but shouldn't have

The following appear to be OK to be in footprint naming, so just keyword updates: AUH, GH, JWPF, LEA, NV, PH, SHL, SH, SUR, VH, XAG

I used Python 3.7.0.

The number of additions and deletions seems odd to me, but I'm quite sure I was careful with only updating the changed files so I think this is correct.

@poeschlr
Would you be willing to correct the EH and XH 3D models? Those are the only ones at https://github.com/KiCad/kicad-packages3D/tree/master/Connector_JST.3dshapes that are affected and it looks like you submitted the existing files. Since you're not busy with anything else I hope that's OK. :)

------------


evanshultz:
Script update is at https://github.com/pointhi/kicad-footprint-generator/pull/256.

Travis shows a red 'x' above but the Details page is all green that I can see. Am I doing something wrong? Maybe the quantity of file changes is too great for Travis to handle and it bails?

poeschlr:
The travis error comes from the library table check. Looks like i missed the problem in the pull request i merged.

poeschlr:
At least JST EH,PH and XH also have 3d models in our lib. Meaning you would need to also update these scripts and the 3d models in our lib.

evanshultz:
@poeschlr 
Is there a fix that's needed for the table check?

Would you be willing to do the 3D models? Or @Shackmeister or @Misca1234 ?

poeschlr:
The table stuff is already fixed in a pull request of mine.

evanshultz:
Thanks, I looked but didn't catch it right away. All is green now with Travis.

poeschlr:
Thanks, with the 3d models now ready this can be merged.

PR #1246, opened 2018-12-31
Added 7 segment display SA15-11

kajusK:
7seg display for symbol from https://github.com/KiCad/kicad-symbols/pull/1130

2018-12-31-180814_1920x1080_scrot

Datasheet: http://www.kingbrightusa.com/images/catalog/SPEC/SA15-11SRWA.pdf


myfreescalewebpage:
Hi @kajusK, thanks for contributing,

- [ ] The full part name is SA15-11SRWA, you should rename the part.
- [ ] The global dimensions of the part are not consistent with the datasheet, you should not truncate the values.
- [ ] The external diameter of the pad seems very large too. You can reduce to 1.5mm I think.
- [ ] The F.Fab should line be interrupted near pin 1 to get a single closed form.
- [ ] Pin 1 should be in the top left corner (KLC F4.2), please rotate the part.

CA56-12CGKWA already available in the same library is a good example. 

Cheers,
Joel

kajusK:
Thanks for a review, according to the name, the SA15-11 is manufactured in multiple variants, e.g. SA15-11EVA, SA15-11GWA with name based on the display color. Therefore I tried to make the name of the footprint generic.

myfreescalewebpage:
Thanks to point me out on this. I just made some research and found:
- SA15-11GWA which is green ("G")
- SA15-11EWA which is red ("E")
- SA15-11SRWA which is super bright red ("SR")

"W" seems to mean "White Diffused" and "A" if for "Common Anode".

To conclude, I would rename the footprint "SA15-11xxx" (first x for the color, second x for the diffused, third x for the common pin) and on the symbol side, I would like you to create one symbol "SA15-11EWA" (for example) with two alias "SA15-11GWA" and "SA15-11SRWA", with description properly indicating the right color/diffused/common pin.

What do you think of that ? Do you know if the part is also available with other colors/diffused/common pin ?

Thanks,
Joel

kajusK:
@myfreescalewebpage Pong. I'm sorry, I'm a bit overloaded by some work so I completely forgot to check the pull requests. Anyway, here's the proposed change to the footprint, the symbol update is in the original pull request.

I'm not aware of any other variants, I guess it's not necessary to add all possible variants in this pull request, I guess if anyone needs other variants, more pull requests will follow...

Jakub

myfreescalewebpage:
Hello @kajusK 

Thanks for renaming the part, no worries I was just ping to notify you but I understand you are busy :)

During the change of part name:
- [ ] Datasheet link is incorrect, should be `http://www.kingbrightusa.com/images/catalog/SPEC/SA15-11SRWA.pdf`

When you will have time to fix the previous comments of the review I will check them with pleasure

Cheers,
Joel

kajusK:
Hello @myfreescalewebpage 
I've fixed the remaining issues, or at least I hope I fixed them :smile:. If you see any further issues, just let me know.

myfreescalewebpage:
Thanks, clearly better, great job here!

I have some few comments about the global dimensions of Fab layer:
- [ ] Part is 33.15x47mm, your footprint is 33.2x47mm => To be fixed
- [ ] Left border should be at -3.18mm from the left pads, you have -3.2mm
etc.

About the other layers:
- [ ] CrtYd clearance with the Fab layer should be 0.25mm, you have 0.3mm actually => to be reduced
- [ ] Footprint laye also a bit large, 0.15mm clearance is great

kajusK:
@myfreescalewebpage 
thank you for the comments, fixed.

myfreescalewebpage:
Hello @kajusK 

Sorry but this is worse than the previous version. Following are two screenshot in top left and near pin 1.

top-left

pin-1

Cheers,
Joel

kajusK:
What the.... Well, the footprint was in this state long time before I've made the commits, it seems the kicad failed to save the latest changes correctly, I do remember clicking save and confirming the footprint name... Never mind, here's the fix, at least I hope

myfreescalewebpage:
Hello @kajusK thanks for quick modification, the footprint is great for me. Merging.

PR #1247, opened 2018-12-31
Add ALPS SKRK series button footprint (finish #456)

poeschlr:
This finishes #456

Product page: http://www.alps.com/prod/info/E/HTML/Tact/SurfaceMount/SKRK/SKRKAHE020.html

screenshot from 2018-12-31 22-27-45

screenshot from 2018-12-31 22-31-07


Misca1234:
I see that your spacing between F.Fab and F.Silk is 0.12, I usually use 0.13, according to 
http://kicad-pcb.org/libraries/klc/
In chapter F5 the distance is not mentioned.
What is the rule here, 
"as long as it is outside F.Fab but no closer to exposed copper than 0.2" ?

poeschlr:
Silk is really not critical. As long as it is fully outside the body dimensions and looks nice it should be good to go. No reason to overly regulate this. (Silk in my opinion is the least important part of footprints.)

Issue #1245, opened 2018-12-31
MountingHole library should be renamed Mechanical


myfreescalewebpage:
Following the addition of #1041 I would suggest that the MountingHole library should be renamed Mechanical. This is more consistent with the symbol library which is already named Mechanical (in which enclosure, heatsink, holes and din rail adapter are available).

Any objections ?

Cheers,
Joel

poeschlr:
It is to be expected that we will have more footprints for these parts compared to symbols. It makes therefore a lot of sense to have a mechanical symbol lib that collects symbols for many very different mechanical parts. On the footprint side i would however prefer these to be split up.

myfreescalewebpage:
Thanks @poeschlr 
So moving DIN rail in a new library, alone ?
Joel

poeschlr:
Maybe something like `MountingEquipment` could be generic enough to include different specialized parts intended to mount the pcb to something.

evanshultz:
Agree that `MountingHole` stays but this footprint doesn't fit.

What footprints are we likely to add? `MountingEquipment` is pretty broad so we could start there and see what other contributions come up. (I'm not in love with the name but don't have a betters suggestion right now.) If we get a lot more footprints and split up a library that's a lot of churn later on but we also don't want to start out with overly-specific libraries. `MountingRails` would be an option if we were likely to add more rail adapters. `MountingAdapters` would be even more generic to include rails and other stuff. Personally, I wouldn't predict a huge amount of contributions here and support Rene's suggestion unless we can come up with a more elegant name.

myfreescalewebpage:
@evanshultz totally agree with you, we cannot say what will be proposed in the future, so let's go with `MountingEquipment`.

I will make the modification for this DIN rail part soon.

Joel

myfreescalewebpage:
Done.

myfreescalewebpage:
@poeschlr you have merged the footprint but not the 3D package. See above referenced PR. Probably we need also to close this issue once it is done ;)

Cheers,
Joel

PR #1252, opened 2019-01-02
MountingEquipment: moving DIN Rail adapter

myfreescalewebpage:
Only moving the DIN Rail footprint and 3D following discussion at KiCad/kicad-footprints#1245. No other modification of the part.

------------


poeschlr:
I seem to remember that this one is connected to a symbol. Does this need fixing as well?

ferdymercury:
Yep, here it was done with MountingHole: https://github.com/KiCad/kicad-symbols/pull/1095

myfreescalewebpage:
@poeschlr you are righ, but the symbol has no default footprint :)

poeschlr:
In this case it is ok. I however feel the symbol has a way too specific footprint filter. I would not really suggest to list the screw size. Might be something that should be fixed.

myfreescalewebpage:
@poeschlr probably yes, will see in the future depending of the footprints proposed by the community. Actually the available footprint is M3 only.

PR #1259, opened 2019-01-04
Added 5400BL15B050E.

matthuszagh:
https://www.johansontechnology.com/datasheets/5400BL15B050/5400BL15B050.pdf

image

I've placed this in RF_Module, but I'm not convinced that's the right footprint library for it. Also, the schematic for this doesn't exist yet. I'll create it momentarily, but I wanted to create this first so I could reference the footprint in the schematic.

------------


aewallin:
do you have 0.2mm clearance on the non-GND pads?
seems the datasheet suggests at least 0.2mm to the GND-zone.

poeschlr:
I am not convinced RF_Module is the correct library for this.

We might either want to create a new library for rf signal converters.
Maybe `RF_Converter` would be a good name with description "Specialized footprints for RF signal converters (Like Attenuators, Baluns, Mixers, etc.)"

---

The pin 1 marker on the fab layer should be a chamfer with 45 degree. chamfer size (x and y) should be 25% of shortest device dimension or at most 1mm

Move the fab reference inside the outline and rotate it by 90 degree. Right now the thickness of is a bit high. The thickness should be 15% of the text height.

The courtyard clearance must be 0.25mm all around (rounded to nearest 0.01mm grid point)

screenshot from 2019-01-06 10-51-14

matthuszagh:
@poeschlr, thanks for the feedback. I've made the proposed changes. Let me know if any issues. I also added "Couplers" to the library description, but let me know if you'd like me to change it to the description you proposed.

image

PR #1260, opened 2019-01-05
SiT9121 Differential Oscillator Footprint

awygle:
This PR adds the footprint for the SiT9121 differential oscillator, 3.2mmx2.5mm variant.

I tried using scripting for this part but the existing SMD oscillator script doesn't seem to handle pads of different sizes, which the datasheet recommends for this part.

Datasheet can be found at https://www.sitime.com/datasheet/SiT9121.

Screenshot: 
image

------------


aewallin:
- rotate so pin1 is top left
- not sure about f.adhes layer - nothing in KLC on that? is there something in the datasheet? or IPC-guidelines?
- can't see from screenshot if f.fab outline has a bevel to indiacete pin1. (bevel should be <=1mm or 25% of width)

aewallin:
F4.2 - "Pin 1 should be located at the top left" is quite universally followed I think.
Not sure if dimensions always have to be in order e.g. X mm x Y mm.
Bevels usually leave out the 'cut' corner-part of the outline - F5.2 figure with examples.

poeschlr:
I documented the wrong zero orientation of the oscillator lib in this issue: https://github.com/KiCad/kicad-footprints/issues/1264

awygle:
Fixed the rotation and the bevel:

image

I made sure the pads were at 0 rotation too, having had that asked of me before.

poeschlr:
The courtyard clearance should be 0.25 all around (0.5 is only for connectors and strangely for crystals.)

screenshot from 2019-01-19 19-41-16

myfreescalewebpage:
@awygle any news on this subject ? Thanks!

PR #1253, opened 2019-01-02
Added footprint for Silicon Labs Si512/513 dual frequency oscillator

fhw72:
Added footprint for Silicon Labs Si512/513 dual frequency oscillator
with 6 pin (3.2x2.5mm²) package.

------------

Thanks for creating a pull request to contribute to the KiCad libraries! To speed up integration of your PR, please check the following items:

- [x] https://www.silabs.com/documents/public/data-sheets/si512-13.pdf

screenshot 2019-01-02 22 21 49

PR #1261, opened 2019-01-05
Scripted fp for Silicon Labs Si512/513 oscillator

poeschlr:
This is a replacement for #1253

Script PR: https://github.com/pointhi/kicad-footprint-generator/pull/258

https://www.silabs.com/documents/public/data-sheets/si512-13.pdf

screenshot from 2019-01-05 21-10-11

```yaml
Oscillator_SMD_Silicon_Labs_LGA-6_2.5x3.2mm_P1.25mm:
  device_type: 'LGA'
  library: Oscillator
  fp_name_prefix: Oscillator_SMD
  manufacturer: 'Silicon_Labs'
  size_source: 'https://www.silabs.com/documents/public/data-sheets/si512-13.pdf'
  ipc_class: 'qfn_pull_back'

  body_size_x:
    nominal: 2.5
    tolerance: 0
  body_size_y:
    nominal: 3.2
    tolerance: 0

  lead_width:
    minimum: 0.45
    nominal: 0.5
    maximum: 0.55
  lead_len:
    minimum: 0.65
    nominal: 0.7
    maximum: 0.75
  lead_center_to_center_x:
    nominal: 1.65
    tolerance: 0

  pitch: 1.25
  num_pins_x: 0
  num_pins_y: 3
```


fhw72:
Hi René,
please let me know when you've pushed your changes.
Thanks,
Frank

poeschlr:
I am not allowed to push anything myself to the repo. Some other librarian has to look over it.

myfreescalewebpage:
Hi @poeschlr 

Few comments:
- [ ] The pitch is 1.27mm, not 1.25mm
- [ ] X position of the pins should be -1.3mm/+1.3mm
- [ ] Size of the pins should be 0.8x1.7mm

I have looked at datasheet https://www.silabs.com/documents/public/data-sheets/si512-13.pdf page 17 for reference, let me known if I'm wrong ?

Cheers
Joel

fhw72:
Hi Joel!
On page 17 is the land pattern for 3.2x5.0mm² device. However the scripted land pattern is for the 3.2x2.5mm² device which is specified on page 20. So this one is correct. Please Check!

Cheers,
Frank

fhw72:
I rechecked: Pitch is correct with 1.25mm but I'm unsure about width and len of lead.
Cheers,
Frank

myfreescalewebpage:
ooooooo sooorry !!! redoing the review right now :)

poeschlr:
As with the other footprint: This is not derived from a suggested footprint but directly via industry standards from the dimensioned drawing of the package

myfreescalewebpage:
> In my opinion the name of this footprint suggests it's for a "Silicon Labs" part; therefore
it makes sense to also adapt their recommendation!

Agree with @fhw72 !

Cheers,
Joel

poeschlr:
If the package has different dimensions than it is clear that the footprint will look different. As this one is based on the dimensions of silicon labs it should be good for this one part. 

silicon labs simply uses an older version of the same standard. We use version IPC-7351B while they used IPC-7351. Our version is from 2010 theirs from 2005 so 5 years older. 
Version C is already actively worked on. Some details from that later standard are already known and therefore incorporated in this footprint. One of these new things is the use of rounded rectangle pads which increase solder reliability. 

Another one is the use of closely contoured courtyards. (Not applicable to this particular footprint)

fhw72:
Hi René,

In order to solve this I'd suggest the following:

1. Renaming the footprint you've created to reference the underlying
standard (e.g. "Oscillator_SMD_IPC-7351B_LGA-6_2.5x3.2mm_P1.25mm" instead
of "Oscillator_SMD_Silicon_Labs_LGA-6_2.5x3.2mm_P1.25mm").

2. Creating a second NEW footprint that follows the **manufacturer recommendation**
and naming it either "xxx_Silicon_Labs_xxx" or giving it a name referring to the (older) standard.

Obviously we have some kind of review-deadlock!? What do you think about this pragmatic (but still workable) solution in order to get this thing solved finally? Any other suggestion?

myfreescalewebpage:
Where is it possible to read IPC7351B content ?

fhw72:
If I'm right IPC7351B is not available without paying for it. Maybe @poeschlr can help.

myfreescalewebpage:
Thanks @evanshultz for your comment on this PR. Merging the footprint.

PR #1262, opened 2019-01-05
SJ1-3535NG stereo 3.5mm audio jack with switch

ppelleti:
[Datasheet](https://www.cui.com/product/resource/digikeypdf/sj1-353xng.pdf).

I have created one footprint with the land pattern specified in the datasheet, and I've created another footprint with circular holes.

These footprints are meant to be used with the `AudioJack3_SwitchTR` library symbol.

This jack is identical to the SJ1-3533NG jack already in the footprint library, except that the SJ1-3535NG contains two switches (and thus has 5 pins instead of 3).

sj1-3535

sj1-3535-circular

------------


poeschlr:
Sorry it took us so long.

Looks good, thanks
Screenshot from 2019-06-01 10-47-53

Issue #238, opened 2018-01-10
Missing fuse holders


u382903:
Looks like good footprints for fuse holders were deleted. They were described as "Fuse_Holders_and_Fuses:Fuseholder5x20_horiz_open_universal_Type-III", type II and I (also other footprints with 5x20 and 5x25 in name were good). Please restore them.

poeschlr:
If you can link at least one product to each of them, then we can include them.

u382903:
Here you have list of many products that were compatible with these footprints - first 8 for example from link below:
https://www.digikey.com/products/en/circuit-protection/fuseholders/140
or here: https://www.tme.eu/gb/katalog/#id_category=100088&page=1&s_field=niski_prog&s_order=ASC
or this: https://www.tme.eu/gb/details/zhl77/pcb-fuseholders/stelvio-kontek/ptf77/
Type III was universal, it was made from type I and II.
So these footprints were good for 'bare' fuse clips, which are used widely.

stephf0716:
Please link to a specific product for each footprint. The first two links are not specific, and the suggested land pattern for the third link you provided does not match any of the three footprints you mentioned. 

Here are links to fuse clip datasheets from three manufacturers: [littelfuse](http://www.littelfuse.com/data/en/data_sheets/littelfuse_fuse-clips.pdf), [Keystone](http://keyelco.com/userAssets/file/M65-fuse-mountinglayout.pdf), and [Cooper](http://www.cooperindustries.com/content/dam/public/bussmann/Electrical/Resources/product-datasheets-a/Bus_Ele_DS_2131_1AXXXX_Series.pdf)

The recommended footprints for 5x20mm fuse clips do vary from part to part.

u382903:
For Type I this would fit: https://www.digikey.com/product-detail/en/littelfuse-inc/01000056Z/F3784-ND/2498892 (in catalog it's described as series 100)
For Type II: https://www.digikey.com/product-detail/en/littelfuse-inc/01110501Z/F4189-ND/2518598 (in catalog it's described as series 111)
And Type III would fit both of those fuse clips series.
Recommended footprints indeed vary from part to part, but I think there should be footprints for 'bare' fuse clips. It will be laborious process to add most of them, but these footprints will be good for beginning.

poeschlr:
[...]

PR #1268, opened 2019-01-06
Add ESKA 120.700 fuse holder.

mgottschlag:
This pull request adds a footprint for the ESKA 120.700 fuse holder.

The component is actually a single clip, and this pull request adds a footprint for a *pair* of clips (as the fuse has a fixed size, the distance between the clips is fixed as well). This will obviously mess up order lists generated from kicad schematics, but seemed better to me as it is a lot more error-proof.

The manufacturer does not provide any drill pattern and the drawing is somewhat incomplete, so I guessed some dimensions and added a generous courtyard. As I have not had the part in my hands yet, I cannot say whether this is 100% correct - if you want, we can leave this pull request open for ~1-2 months until I have a PCB with the part.

F7.3 is violated (pin 1 shape), but the fuse holder is non-polarized, so the pad has to be round.

------------

http://www.produktinfo.conrad.com/datenblaetter/500000-524999/524314-da-01-ml-HALTEFEDER_1207_f_6_3x32mm_de_en_fr.pdf
2019-01-06-163456_710x180_scrot
(no 3d model, no new symbol)


herostrat:
Hallo mgottschlag,

thanks for your contribution.
If you have time have a look at https://github.com/KiCad/kicad-footprints/issues/238

There is a discussion for "generic" fuseholders and how we could realize them.
Because you are already familiar with the topic I hoped you can look at the models there and if it fits your needs or if there are more models out there that get used often.

poeschlr:
How did you determine the x distance between the two clips?

PR #1270, opened 2019-01-06
Crystal: Add Abracon ABS25 4Pin 8.0x3.8mm

piotr-wiszowaty:
Datasheet: https://abracon.com/Resonators/abs25.pdf

abs25

------------


poeschlr:
Rotate the footprint such that pad 1 results nearest to the top left corner (rotate by 90 degree clockwise).

The pads are too short.
The courtyard clearance would be too large for the current pads, but it will be correct for the correctly sized pads.

screenshot from 2019-01-19 16-50-55

poeschlr:
Your previous pin 1 marker on silk was much clearer than what you have now. Either recreate the old version or use something like this:
screenshot from 2019-01-20 19-13-46

Your adhesive circles should either be replaced with a single adhesive only pad or removed.

poeschlr:
Thanks, looks good now

PR #1271, opened 2019-01-06
Smaller  Polarity Center Symbols

Hitmare:
Added smaller  Polarity Center Symbols than 6mm

------------


pointhi:
This is the original source used for the other polarity markers:

* https://commons.wikimedia.org/wiki/File:Polarity_marking_center_positive.svg
* https://commons.wikimedia.org/wiki/File:Polarity_marking_center_negative.svg

Hitmare:
I will update the files so that the Travis CI build runs without errors

poeschlr:
I think you can ignore nearly all travis errors except the one regarding the line endings. (Symbol footprints do not need a visible reference or value field. They also do not need a courtyard outline)

PR #1272, opened 2019-01-06
add texas SOP-8 package

balcer:
http://www.ti.com/lit/ds/symlink/iso1050.pdf

footprint

https://github.com/KiCad/kicad-symbols/pull/1342

I found 3D model on Texas Instrument website:
https://webench.ti.com/cad/dlbxl.cgi/newstep/DUB0008A.stp

------------

Thanks for creating a pull request to contribute to the KiCad libraries! To speed up integration of your PR, please check the following items:

- [x] Provide a URL to a datasheet for the footprint(s) you are contributing
- [x] An example screenshot image is very helpful 
- [x] If there are matching symbol or 3D model pull requests, provide link(s) as appropriate
- [x] Check the output of the Travis automated check scripts - fix any errors as required

PR #1273, opened 2019-01-06
Scripted SOP-8 for ISO-1050 (replaces #1272)

poeschlr:
replaces #1272 

http://www.ti.com/lit/ds/symlink/iso1050.pdf
script https://github.com/pointhi/kicad-footprint-generator/pull/259

screenshot from 2019-01-06 22-54-39

```yaml
SOP-8_6.62x9.15mm_P2.54mm:
  size_source: 'http://www.ti.com/lit/ds/symlink/iso1050.pdf'
  body_size_x:
    minimum: 6.37
    maximum: 6.87
  body_size_y:
    minimum: 8.75
    maximum: 9.55
  body_height:
    maximum: 4.85

  overall_size_x:
    minimum: 10.1
    maximum: 10.7
  lead_width:
    minimum: 0.355
    maximum: 0.555
  lead_len:
    minimum: 1.15
    maximum: 1.45

  pitch: 2.54
  num_pins_x: 0
  num_pins_y: 4
```


myfreescalewebpage:
Hi @poeschlr 

I do not agree with X position of the pins. On left: -4.525mm ; on the right: 4.525mm. This gives a distance between pins of 9.05mm, but the datasheet says 9.1mm.

capture

Also, seen a HV version is designed in the datasheet, we never take care of it ? For this part the difference is very small... but that's a question.

Chers,
Joel

poeschlr:
the script uses ipc standards (version B so two versions newer than the drawing you show) and derives stuff directly from the package dimensions. The suggested footprint (should it exist) is completely ignored.

myfreescalewebpage:
Where do you find version B ? Well in any case this is a difference of 0.05mm, so it's very small, if you indicate this is acceptable for this generic (SOP) footprint, I think we can accept the PR like that :)

poeschlr:
We found version B on some file sharing server some time back. I sadly did not save the link but i have a copy of it on my disk (Another sad thing is that it is a series of pictures instead of a nice pdf)

myfreescalewebpage:
Okay @poeschlr so I understand it is possible to be merged like that.

PR #1279, opened 2019-01-07
NKK GW12xxH SPDT right-angle rocker switch

ppelleti:
[Datasheet](http://www.nkkswitches.com/pdf/GW.pdf).

I was not sure how to handle the pin numbering, so let me know what I should do.

The datasheet specifies the pins as 4, 5, and 6, but I numbered them 1, 2, and 3, to allow the footprint to be used with the existing library symbol for an SPDT switch.  The alternative would be to create a new library symbol where the pins are numbered 4, 5, and 6.

I left the mounting pins unnumbered.  The alternative would be to number them MP.  This would then require a new library symbol, too, since there does not seem to be a symbol for a switch with a mounting pin.

gw12xxh

------------


poeschlr:
We have a more detailed naming convention for switches that should be applied for new contributions. https://github.com/KiCad/kicad-symbols/issues/580

So i might suggest: `SW_Rocker_1P2T_NKK_GW12xxH`

ppelleti:
I have increased the width from 7.0 to 7.5mm, and I've added a simple single-line representation of the paddle on the fab layer.

gw12-with-paddle

poeschlr:
I took another look at the naming for this. We might want to include the letter code for the rocker version in the part number. meaning use `SW_Rocker_1P2T_NKK_GW12RxH` (Sorry i missed that the last time around.)


The size fixes seem ok.
screenshot from 2019-01-25 20-46-29

I would also favor a bit more detail for the rocker to be honest. The current graphic looks a bit strange to my eye. (But i would accept it if you are not prepared to invest more time.)

ppelleti:
> We might want to include the letter code for the rocker version in the part number. meaning use `SW_Rocker_1P2T_NKK_GW12RxH` (Sorry i missed that the last time around.)

Wouldn't it be `SW_Rocker_1P2T_NKK_GW12LxH`, since I have drawn the fab layer and courtyard based on the "paddle" actuator?  (The copper and silk should be good for either actuator style.)

gw12-part-number

> I would also favor a bit more detail for the rocker to be honest. The current graphic looks a bit strange to my eye.

I agree; I'll try to improve it.

poeschlr:
You are right this is the paddle and should use letter code L

poeschlr:
We might also want to change the prefix from Rocker to Lever in this case.

poeschlr:
Thanks

Issue #144, opened 2017-11-30
Button_Switch_SMD.pretty naming convention not continuous


pointhi:
We have some mixture of rotary encoder without a prefix, switches without a ```SW_SPST``` prefix and switches with/without manufactur name

Ratfink:
Also: a lot of pushbutton switches are currently prefixed with `SW_SPST`.  Should they be renamed to use an `SW_Push` prefix?

Shackmeister:
SW_Push doesn't give the electrical configuration of the switch. A pushbutton could also be SPDT, SPDT etc

Ratfink:
Good point.  I think it would be best to put both in the name, which would allow for better footprint filters.  Really, our switch naming needs a lot of improvement and standardization, and it should extend to the symbols as well.

poeschlr:
We can think about a scheme that is used for new contributions. But existing symbols and footprints should not be changed in a manner that would break existing projects that use them. (Otherwise we are unable to supply users with bugfix updates of the library.)

poeschlr:
We have documented a naming convention for new contributions here: https://github.com/KiCad/kicad-symbols/issues/580

PR #1282, opened 2019-01-09
Added footprint for Miyama MS-500A-VT switch.

mgottschlag:
This commit adds a footprint for the Miyama MS-500A-VT switch.

------------

http://www.miyama.co.jp/pdf/MS-500X-X-X.pdf
2019-01-09-082715_1189x338_scrot
(no related pull requests)


mgottschlag:
I guess this depends on #144?

mgottschlag:
I guess SW_DIP_SPDT_Miyama_MS-500A-VT would be the best name?

Also, what about the 500B-E versions which have the same footprint, should that be indicated in the name? Note that 500F and following have a different footprint, so ...Miyama_MS-500X-VT or something with a placeholder would not be correct.

poeschlr:
The naming convention has reached a temporary conclusion with this: https://github.com/KiCad/kicad-symbols/issues/580

`SW_Lever_1P2T_Miyama_MS-500A-x-VT`

poeschlr:
The land pattern looks good in general. But the courtyard clearance should be 0.25mm all around. Note specifically that you have the mounting pads protruding. (This clearance is marked pink in the drawing i made to make it stand out.)

screenshot from 2019-01-16 20-58-17

PR #909, opened 2018-09-02
add Oscillator_SMD_Abracon_ASCO-4Pin_1.6x1.2mm

penoud:
add Oscillator_SMD_Abracon_ASCO-4Pin_1.6x1.2mm footprint. Based on https://abracon.com/Oscillators/ASCO.pdf
------------

Thanks for creating a pull request to contribute to the KiCad libraries! To speed up integration of your PR, please check the following items:

- [X] Provide a URL to a datasheet for the footprint(s) you are contributing
https://abracon.com/Oscillators/ASCO.pdf
- [X] An example screenshot image is very helpful 
image
image

- [X] If there are matching symbol or 3D model pull requests, provide link(s) as appropriate
I will create the PR soon and update this with the link.
- [X] Check the output of the Travis automated check scripts - fix any errors as required

PR #1284, opened 2019-01-09
Added Kyocera 6 pin tcxo.

matthuszagh:
https://global.kyocera.com/prdct/electro/product/pdf/kt2520_e.pdf

image

------------


matthuszagh:
The reason I did that is because the [KLC](http://kicad-pcb.org/libraries/klc/F5.3/) says crystals should have a 0.5mm courtyard clearance. Still want me to reduce it?

myfreescalewebpage:
Good point, I was basing my comment on https://github.com/KiCad/kicad-footprints/pull/909 recently merged where a 0.3mm margin is kept. I wonder if "crystal" include SMD oscillator...

@poeschlr can you help on this point, you have merged https://github.com/KiCad/kicad-footprints/pull/909 but we wonder about the respect of KLC F6.3... Thanks

Joel

poeschlr:
The device you link to had 0.25mm courtyard clearance not 0.3. (0.3mm is never mentioned in the KLC. And also not in IPC standards.)

To be honest i am not really sure where the 0.5mm recommendation for crystals comes from. (It was there before i worked on the library. It is not part of IPC recommendations.)
Which makes it harder for me to determine what devices should be covered by it. If i would have a guess then only crystals should use the 0.5mm clearance if this is even true for them.

myfreescalewebpage:
@poeschlr thanks for your response, so this device need only 0.25mm clearance. @matthuszagh I have updated my previous comment, please can you make the modification ?
Joel

matthuszagh:
Clearance changed.

myfreescalewebpage:
Thanks @matthuszagh no more comment, merging.

PR #1287, opened 2019-01-10
KiCad Logos: Master Update to Build

codemonkeyrawks:
I have fixed several issues: If you would like to merge in:

Updates
5mm Logo - Added Silkmask @4x1md : thanks
6mm Logo - Fixed Bounds & Labels: @SchrodingersGat
8mm Logo - Fixed Bounds & Labels @SchrodingersGat

Standardized Logo Set into four categories: Front Silk, Front Copper, Front Mask,  Back Copper.

Files attached below.
------------


aewallin:
is there a script that generates all these symbols from the design-file? If so, it would be nice to link to it in the footprint description, or in this PR thread, or at least somewhere...

codemonkeyrawks:
Hi @aewallin : there is not a script that generates these files - I did them all by hand with pulling in from @4x1md files and @SchrodingersGat files and spent time correcting bounds on each. I had basically taken the logo from the prior files then copied this to every file then only took the text from the other file and added this and changed options. I did however write a script to do checking on multiple files using the diff command when placing this together by hand which I deleted yesterday.

I am not sure if you wanted to see but the check had done basically:

Check #1:
diff <5mmItem1> <5mmItem2> - diff <5mmItem1> <5mmItem3> - diff <5mmItem1> <5mmItem4>

Check #2:
diff <folder0/5mmItem1> <folder1/5mmItem1>

The main point is I followed this: Standardized the format on each file
- Added Modules and File Name
- Added Yard
- Added Logo
- Added Text

poeschlr:
I am not so sure logos need a courtyard. The courtyard is there to allow space for pick and place machines, soldering equipment, ...
In other words for things where you need access to.

codemonkeyrawks:
@poeschlr : I did not know in either case if it was OK breaking the check build or not on certain items. I have updated the branch to remove the courtyard. If this accepted there may need to be a change on what is allowed vs not allowed for courtyards for the build to pass. 

Thank you.

herostrat:
Hey, very nice work!

Because one can just change the layer for the footprint/silk from the top to the bottom, wouldn't one version (the top one) be sufficient?
And if it is decided that both sides are needed, is there a reason the ones on the bottom layer ones are not mirrored?

codemonkeyrawks:
Hello @herostrat,

1. I created four different layers because of the following below and each have been named so its easy to add/remove (following: KiCad-LogoX_Xmm_Layer):

a. I did not want to have to go though the process each time and change from the copper layer logo's to the silk screen layer. I myself actually use the silkscreen for printing vs using a copper layer. There is no bottom silk but I typically don't use this but would add one other option if you see fit?

b. I also noted that on one push there was a mask logo : personally I don't use this however was trying to keep what already pushed. It almost would make more common to add a silk bottom and remove this but am unsure if people would be upset. I just was trying to come with a solid base for the logos such that if this gets pushed I may look at updating others the same way (in figuring out what layers are required vs optional) and work around this.

2. Mirroring: I am still a little new to KiCad however the way I see the logo it matches both sides without being mirrored: just means it flows the opposite instead of lining up. I am happy to change this if needed.

Let me know if there is something to change.

pointhi:
please note that logos on the backside are mirrored ^^

poeschlr:
Do not add any bottom side logo. All footprints are designed on the top and can be flipped to the bottom by the user once placed in pcb_new.

codemonkeyrawks:
@poeschlr : Done - There is only: Top Copper - Top Silk - Top Mask.

poeschlr:
Sorry i missed something else. Could you hide the reference and value fields? I doubt that the user will want them shown in a typical usecase for these. (And yes i am aware that this will be something travis will complain about. We simply have no way of defining rules on a per library basis so travis will always have falls positives for some rare edgecases.)

codemonkeyrawks:
@poeschlr : done

poeschlr:
Thanks

codemonkeyrawks:
@poeschlr : Just for future: Is there a common naming scheme to use as I noted my branch may be slightly different than what is in there. Are you OK if I look at updating a few other logos to match a similar format with: Top Copper - Top Silk - Top Mask. I am just trying to find the best way I can support in knowing what standards are for this library.

Thank you!

poeschlr:
I did not really look at that to be honest. Now that you ask i would not have included the specification if the layer is top or bottom. Only if it is copper, silk or mask.

codemonkeyrawks:
@poeschlr : I will work on renaming a few of the files to follow the same format in the branch (copper and silk). Looking further it looks like both Copper and Silk are common so will remove the mask files also. I will create a new pull request when I am ready.

Once I complete this I will look at updating a few others to match and/or add top copper/silks that are missing. The main idea is : top copper - mask and named the same.

codemonkeyrawks:
@poeschlr : https://github.com/KiCad/kicad-footprints/pull/1312 should fix everything mentioned.

Thanks!

PR #1288, opened 2019-01-10
Add PAT1220.

matthuszagh:
land pattern datasheet: http://www.susumu-usa.com/pdf/Foot_Print_38.pdf

regular datasheet (contains package dimensions): https://www.susumu.co.jp/common/pdf/n_catalog_partition16_en.pdf

image

------------


aewallin:
- there are two footprints here, I'm guessing you meant to PR only the one with the longer name?
- I put a similar part #1150 into the "Package_SO" lib, along with other RF footprints, not sure if these deserve a lib of their own?
- fwiw I created this symbol for the mini-circ attenuator https://github.com/KiCad/kicad-symbols/pull/1226 maybe it can be used (with modifications) here also?

matthuszagh:
Ah sorry about that @aewallin, meant to take the other one out.

I placed it in the same lib as another PR (see [this comment](https://github.com/KiCad/kicad-footprints/pull/1259#issuecomment-451729203)). If that library is created, I do think it probably makes sense to put this there as well. But happy either way.

Yeah I'd be happy to use the existing symbol. Like that better?

matthuszagh:
The only thing is I don't really like pins in and out here since they really are passive. I also do like that the graphic exactly represents what the circuit is. My only complaint is how it covers the "GND" text.

poeschlr:
@aewallin this one is definitely not a fit for the SO lib. The SO lib should really only hold gullwing parts. (And yes the SO lib will be one of the libs that need to be split and properly organized with the v6 release.)

---

@matthuszagh thanks for your contribution. All looks well.

screenshot from 2019-01-19 13-03-34

PR #1289, opened 2019-01-10
Add skyworks dfn 6.

matthuszagh:
pages 7 and 8: http://www.skyworksinc.com/uploads/documents/SKY65404_31_201512K.pdf

image

------------


myfreescalewebpage:
@matthuszagh already indicated in the symbol footprint, according to the datasheet, the package is "QFN", not "DFN".

matthuszagh:
I've changed the name to QFN in accordance with the datasheet, despite the fact that there are pins on only two of the 4 sides...

poeschlr:
_Note for the future: This is not a true qfn or even dfn package. The pads look very different (they have very elaborate shapes)_

I don't think DFN/QFN is the right library for this part. It is too specialized.
Lets make a new lib simply called RF for holding RF parts that do not really fit any other library. This one would then be the first to be put in that library.

If skyworks does not rally use this package for anything else then simply name the footprint `Skyworks_SKY65404-31`. (The datasheet does use QFN at only two places. It does not rally use it to describe the dimensioned drawing.)

---

The courtyard must take pads into account.

screenshot from 2019-01-19 12-41-28


I would suggest the use of rounded rectangle pads for the 6 outer pads.

matthuszagh:
Changes made. Let me know if any further issues.

poeschlr:
Same suggestion regarding the pin 1 marker as in your other pull request. Lets make it similar to other dfn packages:
screenshot from 2019-01-20 19-26-42

---

Your fixes look good.
screenshot from 2019-01-20 19-51-16

matthuszagh:
Also fixed.

image

poeschlr:
Looks good, thanks

PR #1294, opened 2019-01-12
add QFN-56-1EP_8x8mm_P0.5mm_EP6.1x6.1mm

penoud:
http://intantech.com/files/Intan_RHD2000_series_datasheet.pdf
image

Was generated using the footprint generator. Please refer to PR:


myfreescalewebpage:
Hello @poeschlr I have reviewed several symbol waiting for some footprints, like this one, but I'm locked because of the review in conformance with IPC-7351B. I do not found the right information to say if these footprints are okay or not. I have looked at several AN, guidelines on Internet and compared with existing footprints in the lib. It does not match. Have you a good reference to review standard footprints (DFN, QFN, etc...) ? If can't be shared publicly, maybe you can share by private message ?
Thanks for your help,
Joel

poeschlr:
@myfreescalewebpage  github has no private messaging system. You do not have an e-mail listed in your account. (I have no way to send you something in private)

A note: IPC does not give specific dimensions for specific parts but rules how to get from part dimensions to a footprint. These formulas are implemented as given in the IPC documents inside my scripts. They have been checked multiple times. 
There where bugs in the past. The current version of the script is trusted for now. If anybody does another check and truly finds another error then we can update the script, regenerate all footprints and merge them into the repo. This is one of the major benefits of the script.

A similar benefit exists for the day when version C of the same standard is available.

poeschlr:
@penoud is it on purpose that you used the old generator where you specify the pad sizes and positions instead of using the new generator that uses IPC rules to calculate these measurements from the package dimensions?

myfreescalewebpage:
@poeschlr thanks for your answer. I have just sent you a mail. I don't know if checking with the formulas is difficult yet, but anyway this is interesting to look.

penoud:
@poeschlr No it isn't on purpose. Do you mean a new version is available? I need to update the tool and correct the script right? Could you also send me the IPC-7351B on this email pernecker@sevjnet.ch, thanks in advance.

poeschlr:
There should be a folder called Package_Gullwing__QFP_SOIC_SO under the Packages sub directory. If there is not then you are definitely behind with your personal fork.

poeschlr:
I checked your pull request and you did indeed base it on an old version of the repo. It does have the new generator but i think i added some fixes to it so please base your new contribution from the current state of the master branch.

assuming you have pointhis repo added under the name upstream then you can ensure that you create your branch at the correct point with the following commands:

```bash
git fetch upstream           # get the current state of the upstream repo
git checkout upstream/master         # switch to the top of the master branch
git checkout -b new_branch_name       # create your branch
git push --set-upstream origin new_branch_name       # setup your local branch to sync to your fork on github 
```

penoud:
I update it. But I will do the modification another day. It's late ;-)

penoud:
@poeschlr I tried with the new version of the script. I used the Package_DFN_QFN instead of  Package_Gullwing__QFP_SOIC_SO since QFN wasn't define in the last one. I get a nice footprint but the IPC rules doesn't seem to give the requested footprint from the datasheet. For exemple. the datasheet request a pad length of 0.9 but the generated footprint is 0.8. Then I have concern on how to solve it. We have to options 1) Make the hypothesis than the IntanTech doesn't respect IPC rules and we used the one generated 2) Stick with manufacturer request. What do you think.

penoud:
updated screenshoots:
image
image

poeschlr:
Note to self: the datasheet has a very bad drawing of this package. We might want to check if there is another source for the same package that we can use. One such option is JEDEC.

antoniovazquezblanco:
Removing the scripting required label as the fp is currently scripted.

myfreescalewebpage:
No news of the author on the footprint generator PR, indicate this is Abandoned

Issue #1267, opened 2019-01-06
Schurter 0031 8002 fuse holder is not 5x20mm (footprint name)


mgottschlag:
While drawing a new footprint I noticed that the "Fuseholder_Cylinder-5x200mm_Schurter_0031-8002_Horizontal_Open" footprint is much bigger than my new 5x20mm fuse holder footprint. Indeed, it seems to be a 6.3mmx32mm fuse holder, the name should be changed accordingly.

Misca1234:
This have been addressed in 
https://github.com/KiCad/kicad-footprints/pull/1295
closing

PR #1295, opened 2019-01-12
Added data sheet links to a number of fuse foot prints and renamed one fuse foot print

Misca1234:
Added a the data sheet link to a number of fuse foot prints.

Also, renamed 
Fuseholder_Cylinder-5x20mm_Schurter_0031-8002_Horizontal_Open.kicad_mod
into
Fuseholder_Cylinder-6.3x32mm_Schurter_0031-8002_Horizontal_Open.kicad_mod

The original size of the foot print was way to big for a 5x20 fuse, it must be for a 6.3x32 fuse.
The foot print was renamed and the layout was simplified

bild

bild

------------


mgottschlag:
This fixes #1267.

Misca1234:
Oh, jeee, double work, I noticed it when I made the 3D models

If you want you can add the data sheet links to your push and I close this

mgottschlag:
No double work at all, I opened the bug because I did not include that in my PR.

Misca1234:
ah, ok, i thought you pushed a foot print

evanshultz:
`Fuseholder_Cylinder-6.3x32mm_Schurter_0031-8002_Horizontal_Open`:
1. Rather than "way too big for a 5x20mm fuse" we should be referencing the datasheet. At https://us.schurter.com/bundles/snceschurter/epim/_ProdPool_/newDS/en/typ_OG__Holder__6.3x32.pdf we can see this MPN is for a 6.35x32mm fuse.
1. The NPTH is 3.25mm but the drawing shows a hole only 3.2mm. If this is to fit a screw, only a 3.2mm screw can go in the hole so why is the NPTH oversized?
1. Please remove the German word in the keywords.
1. The courtyard is a bit too wrong all the way around. It should be X=-3.25mm/+40.75mm and Y=+/-7.75mm.
1. The pads are could be rotated to 0 degrees but that's my preference.

It looks like the 3D model you submitted at https://github.com/KiCad/kicad-packages3D/pull/511 matches the footprint name so that's great.

Misca1234:
The NPTH hole is 3.2 mm
Only English words in keywords
Pads rotation is 0 degree
Courtyard changed

bild

evanshultz:
Here is the file changed in the web interface:
image

I didn't see a commit for the courtyard you mentioned above.

Misca1234:
It seems that I forgot to push the changes, but now so

evanshultz:
Thanks.

None of these have a 3d model (yet?) so that won't be affected. And since it's a bug fix to correct the fuse size in the renamed footprint that's OK too.

Misca1234:
I will do another 3D race when most of my current pushes are merged, they are so many so I cant sort them out

PR #1298, opened 2019-01-13
LED_SMD: Added a footprint for 4mm*3.6mm Osram SIDELEDs.

mgottschlag:
Also usable for some Dialight LEDs.

Some datasheets have a separate wave soldering footprint with larger pads. This is the reflow soldering footprint found in all datasheets I have seen so far.

------------

https://dammedia.osram.info/media/resource/hires/osram-dam-6455804/LW%20A67C%20binning%20FK0PN0_EN.pdf

2019-01-13-175047_710x215_scrot

Without fabrication layer:

2019-01-13-175056_145x170_scrot

(no symbols/3d models)


mgottschlag:
Oops, footprint name fixed (LED_ prefix).

poeschlr:
SMD parts are not allowed to have silkscreen below them. (For manufacturing reasons. Silk below a part increases the probability that something goes wrong during reflow soldering.)

You can add the led direction to the fab layer.

screenshot from 2019-01-19 10-38-41

Your body outline is too large in y direction
screenshot from 2019-01-19 11-08-00

It might be a good idea to add the version with the enlarged soldermask covered pads in addition to the small one. As a suffix i would suggest something like ThermalPads

PR #1300, opened 2019-01-14
RJ45: Add a new Amphenol RJHSE5380-08 footprint

manio:
Small parts based on Martin Kröll work (RJHSE5380)


manio:
Hi, is the 3D model a strict requirement for every footprint?

myfreescalewebpage:
Hi @manio 

> Hi, is the 3D model a strict requirement for every footprint?

No it isn't, however, please indicate path in the footprint (name of 3D = name of footprint, look at other devices) so it is possible to create it later without to come back to the definition of the footprint.

Cheers,
Joel

manio:
@myfreescalewebpage 
Thank you Joel, I've updated this PR as requested

myfreescalewebpage:
Thanks.

poeschlr:
For reference the screenshot of the part:

screenshot from 2019-01-19 09-17-47

The datasheet (dimensioned drawing) https://www.amphenolcanada.com/ProductSearch/drawings/AC/RJHSE538X08.pdf

@manio please provide these two in the future.

---

**Review**

The courtyard must take all pads into account. (All shield pads in your case)

You have the same pin numbers added multiple times. This will mean you can not connect the connectors independently but only in parallel. I doubt this would be on purpose.

Any dimension must agree to all significant digits given in the datasheet. (No rounding, No tolerance) If a datasheet gives metric and imperial dimensions then the metric ones are preferred.
rj svg

manio:
@poeschlr
Thank you very much for your time reviewing the footprint :)
I think I've fixed all issues you've found.
Please only keep in mind the following:

1. You've told me to provide the screenshot while when creating a PR the info is saying "An example screenshot image is very helpful". I just assumed it is not a strict requirement. Thank you for providing the screenshot yourself (maybe it will be better to change this info for PR creators?).

2. Regarding the main pin adjustment (marked 2.04 on your schema): I was just using the Amphenol RJHSE5380 as a base for my footprint. So the dimensions are cloned from there. But you're absolutely right that it has to be 2.03 as in the drawing, so I've fixed this. We only need to note that eventually the RJHSE5380 also has to be fixed the same way.

3. Regarding the pink "15.74" mark - the problem here is IMHO it was done correctly. I was placing these two bottom shield pads relative to center big "hole" - as in the drawing. But there is another problem regarding this: "[...] the metric ones are preferred". As you can see in the drawing it's main dimensions are inches. But if I took the milimeters in the drawing for this placement you can see that 7.87+7.87 is not equal 15.75. Rounding the inches to mm was not with enough precision, so to fix the error you marked I've just converted the inches myself and I've put those two pads in with better metric precision.

4. The last thing is the most bottom dimension you've marked (from outer left-right shield pads). On the drawing it is 126.49 and this is exactly as in my footprint, while you have it 126.76. I just believe it is correct in the footprint.

I hope it is now OK for you - so please review again. Thank you in advance! :)
regards!

poeschlr:
Please to not use force pushes! I now can not tell what you changed and need to review the whole thing again. (Well will take a while as i am really not motivated for that right now.)

manio:
@poeschlr
Oh sorry - i just squashed it because I thought that it would be eventually quicker for you to merge it upstream as a single commit. I have the discrete changes in my local branch, so I now recovered it to you to single commits :)

poeschlr:
We use the squash option to merge pull requests (if there are many commits by the contributor). Github offers an easy interface for that.

We also use that to clean up commit messages by contributors. (Most contributors don't really have any experience with git. Meaning they do not really know how to write proper messages.)

poeschlr:
Thanks

manio:
@poeschlr 
I only want to confirm that using this footprint the element fits just perfectly fine! (as well as single-port RJHSE5380). PCB was manufactured at jlcpcb. Thank you!
image

myfreescalewebpage:
@manio thanks for the photo ! That's nice to have a return using the footprint :)

PR #1303, opened 2019-01-15
BatteryHolder: Add CH273-2450 (1 x CR2450)

piotr-wiszowaty:
Description: http://www.comf.com.tw/ProductDetail.asp?no=148

ch273-2450

------------


poeschlr:
Do you know of a more detailed drawing for this part? The one on the website is under constrained. (There is no dimension given that relates the pads to the body. The lead diameter is not given.)

Is there a particular reason for having the silkscreen so far out?
I would suggest to put it nearer to the fab outline. (There is no strict requirement. But it should be near enough to be of use for manual part placement.)

---

The courtyard clearance should be 0.25mm it is 0.5mm now
The circular features are not quite precise. (I know kicad is not really good at handling arcs but i feel you can do a better job than that. You might also be able to use freecad and stepup to make your live easier.)

screenshot from 2019-01-19 01-05-11

piotr-wiszowaty:
This is what the component looks like:
ch273-2450

poeschlr:
So does this mean the part does not at all agree with what is given in the datasheet?
I am not really happy about having such parts in the official library to be honest.

piotr-wiszowaty:
That's right - the spec drawing is significantly different from the real 
thing.

I've just sent the manufacturer an e-mail asking to provide a correct 
drawing.
Let's see what their reply will be.

poeschlr:
That would be perfect. If they can produce a drawing that aligns with what you see then we can include this part.

Maybe the better drawing has even the one missing dimension.

piotr-wiszowaty:
The manufacturer sent back this datasheet:
[CH 273-2450LF.pdf](https://github.com/KiCad/kicad-footprints/files/2873363/CH.273-2450LF.pdf)

They also said that they will correct the drawing on the webpage.

poeschlr:
Thanks for taking care of this. I will check your footprint against that new drawing shortly.

The question now is how to link to this datasheet such that users of the footprints can verify that it fits their part?

piotr-wiszowaty:
They've actually updated the drawing on their webpage, but rounded the dimensions.
I propose putting the datasheet on github in some (maybe newly created) repository.

poeschlr:
I would go with their rounded version of the datasheet to be honest. They are after all free to define the part as they wish (rounding probably gives them more wiggle room than the precise drawing you got.)

poeschlr:
Please remove the version you based on the unofficial datasheet. (I know it seems crazy but we really can not base stuff in the official lib on unofficial information that nobody can later verify. Especially as that unofficial information is not binding anyways.)

The dimensions of the version complying with the drawing on their website fits with these dimensions.

General remark: It is kind of annoying that they are still missing the dimension that fixes where the pins are in relation to the body. (the 1.22mm) It is also kind of annoying that they do not simply use the detailed drawings from the pdf or at least link the pdf on their site (I suspect they want to be able to easily switch supplier if they find something cheaper)

poeschlr:
Thanks.

PR #1309, opened 2019-01-17
Added SO-8 for the PCF8523 RTC Chip

johnfwhitmore:
https://www.nxp.com/docs/en/data-sheet/PCF8523.pdf

Signed-off-by: John Whitmore <johnfwhitmore@gmail.com>

This footprint is marked in the PCF8523 Datasheet as an SO-8 but it appears to be different to the SO-8 already in the footprint lib. 

https://www.nxp.com/docs/en/data-sheet/PCF8523.pdf

screenshot from 2019-01-17 20-02-36


poeschlr:
The body size must be given as the nominal size. You used the maximum size for the x dimension. I have no clue where the 5.5mm size for y direction would come from as the datasheet gives 4.8...5.0 as the tolerance range.

I would suggest that this is scripted with the gullwing generator: https://github.com/pointhi/kicad-footprint-generator/tree/master/scripts/Packages/Package_Gullwing__QFP_SOIC_SO

johnfwhitmore:
Sorry long weekend, I'll have a look at the generator and try that. On the sizes I looked at the sizes on the SO-8 already in the library to try and see where the sizes in the name came from. I got it wrong.

johnfwhitmore:
I got the 5.5mm size from figure 47. That shows the footprint for the SO-8 and has a measurement of 5.5mm so I used that.

Looking at this I can see why we have a problem. You're looking at Figure 39 which I didn't use because I was interested in the footprint rather then the actual dimensions of the physical device. I can see how you have a question as to my 5.5mm measurement. 

So now I have a problem too as the two figures which both seem to represent the SO8 are different.

johnfwhitmore:
I guess the footprint figure just gives a bit (0.5mm) of wriggle room around the device. So should I use the footprint measurements or the device measurements?

johnfwhitmore:
OK The script is brilliant, but having gone through the documentation all I got was Python, not what you described. I'll re-read it again, and hope for a better understanding this time of where I put those device measurements.

poeschlr:
The repo holds many things. One part is KiCadModTree which is a python library for creating footprints. And then there is the script directory holding our scripts. The script you are looking for is found here: `kicad-footprint-generator/scripts/Packages/Package_Gullwing__QFP_SOIC_SO/` (Direct weblink: https://github.com/pointhi/kicad-footprint-generator/tree/master/scripts/Packages/Package_Gullwing__QFP_SOIC_SO)

This directory holds the generator script (ipc_gullwing_generator.py) and a subdirectory where the size definition files are found (called size_definition)
Inside that size definition folder there are a number of yaml files. These files hold the information about the package dimensions which are used by the python script to generate IPC and KLC complaint footprints.

The script requires python3. (I did not take special steps to have it compatible with python2. This means it could work with pyhton 2 but i am quite sure that there will be problems. Meaning you should really use pyhton 3.)

To run it open a console inside the `Package_Gullwing__QFP_SOIC_SO` directory.
Call the script with `python3 ./ipc_gullwing_generator.py size_definition/soic.yaml` (as an example)
That should result in a new Package_SO.pretty directory holding a number of newly created SOIC footprints.

You would need to add a new size definition file and call it simply `SO.yaml`. (copy one of the other yaml files for a starting point.)

The yaml files have entries similar to this one. The Header tells the script how the library should be called and what prefix to use for the footprint name.

```yaml
FileHeader:
  library_Suffix: 'SO'
  device_type: 'SOIC'
    
SOIC-4_4.55x2.6mm_P1.27mm:
  size_source: 'https://toshiba.semicon-storage.com/info/docget.jsp?did=12884&prodName=TLP291'
  body_size_x:
    minimum: 4.4
    nominal: 4.55
    maximum: 4.8
    # alternatively one can also specify the tolerance instead of minimum and maximum values. 
    # If minimum and maximum values are given then the nominal value is optional. 
    # It should only be included if it is given in the datasheet. 
    # (If it is in the datasheet then it must be included.)
  body_size_y:
    minimum: 2.45
    nominal: 2.6
    maximum: 2.85
  overall_size_x:
    minimum: 6.6
    nominal: 7.0
    maximum: 7.4
  lead_width:
    minimum: 0.37
    maximum: 0.39
  lead_len:
    nominal: 0.5
    tolerance: 0
  pitch: 1.27
  num_pins_x: 0
  num_pins_y: 2
```

poeschlr:
The indentation is crucial in the yaml file. So make sure you copy not only the text but also whitespace.
(A tip would be to first start with what i gave above and then slowly modify it to fit your footprint.)

PR #1311, opened 2019-01-18
add Onion Omega2+ footprint

balcer:
https://raw.githubusercontent.com/OnionIoT/technical-drawings/master/Mechanical/OM-O2.PDF

omega footprint

------------


myfreescalewebpage:
@balcer, you have several travis errors to be corrected. For the 3D model, you should provide the URL even if it does not exist (have a look to some Arduino module in the lib for an example). The travis error F9.3 will remain anyway but it is not a problem.

Thanks,
Joel

balcer:
@myfreescalewebpage Sorry I was on holidays. Footprint ready for the first review. All fix from now on I will make without --force option.

balcer:
Hi, @myfreescalewebpage I made changes according to your suggestions. I additionally added some mechanical features according to the datasheet. I think that I should also add keepout area for sd card but Dwgs.User and Cmts.User layers are not available during footprint design.

myfreescalewebpage:
@balcer, thanks for the modifications. Regarding to the CrtYd layer, you have not read correctly my comment, the 0.25mm clearance is between the CrtYd and Fab layers. You have set 0.25mm between Silkcreen and CrtYd layers... Agreed with you regarding to the SD card, but maybe in the future due to impossibility to add it actually. Your design with internal drawing of the module is great !! :)

myfreescalewebpage:
@balcer thanks, looks great now ! Merging.

Issue #1308, opened 2019-01-17
add Mini-Circuits footprints without GND-zone/vias


aewallin:
As discussed in #1133 it was agreed to add "plain" versions of the Mini-Circuits footprints that do not contain GND-zones or GND-vias.

Here is an attempt to list the footprints that are merged, but do not have corresponding "plain" (non GND-zone, non GND-via) footprint in the library.

- [x] BK377 PL-005 #1126 UPDATE: plain footprint PR is #1314
- [x] GP731 PL-176 #1136, plain footprint PR is #1315
- [x] GP1212 PL-176 #1258 (same as GP731 but different height), plain PR is in #1315 
- [x] HF1139 PL-230 #1134, plain footprint PR is #1317
- [x] MMM168 PL-225 #1150, plain footprint PR is #1318
- [x] QQQ130 PL-236 #1135, plain footprint PR is #1323
- [x] TT1224 PL-258 #1128, plain footprint PR is #1324
- [x] YY161 PL-049 #1137, plain footprint PR is #1327
- [x] CK605 PL-012 in #1151, plain footprint PR is #1328
- [x] TTT167 PL-079 in #1133, plain footprint PR is #1329
- [x] HZ1198 PL-247 in #1132, plain footprint PR is #1330

Creating these footprints is straightforward: copy the existing footprint, re-name without "LandPattern" and code, then remove all THT/SMD pads that are GND-zone, i.e. not part of the package pads proper.

herostrat:
Just to understand, do you plan to do the work yourself or is this a "need help" type of issue?

aewallin:
These are not at the top of my to-do list - I may tick them off slowly one by one but don't hold your breath ;)
Please do feel free to work on this!

aewallin:
I think these suggested new "plain" footprints create a bit of an issue with the footprint-filters in the merged symbols.
The plan above is to have two symbols
- mini-circuits_<Case>_LandPattern<PL-Code> with gnd-zones/vias
- mini-circuits_<Case>  the "plain" footprint, no gnd-zones/vias

But in work that is now merged the fp-filter is set to "mini-circuits*<PL-Code>*", so it will *not* match the plain footprint!?

poeschlr:
the footprint filter for the symbol might best mention the case code.
The default footprint should of course use the full footprint name including the land pattern.

aewallin:
these now merged. closing this.

Issue #1249, opened 2019-01-01
split package_so by lead style


poeschlr:
Right now we have nearly every lead style in existence inside the package_so lib.

I would suggest to split it into separate libs per lead style.
Meaning there would be a lib for gullwing, j-lead and even no lead (the later is the package_son lib right now. but some mini circuits parts that are in the package_so lib would also fit into it.)

PR #1314, opened 2019-01-19
Mini-Circuits BK377 plain footprint

aewallin:
https://ww2.minicircuits.com/case_style/BK276.pdf

Discussion in #1308
- remove GND-vias, GND-zones, 
- remove clearance-values
- all pad-connections to "from parent footprint" (not "solid")
- new 0.25mm courtyard.
- parent footprint was reviewed in #1126

bk377


poeschlr:
I would suggest to move the min circuits packages into their own library. I think only a few have been part of the library before you added many of them recently. So they have never been part of a release and we should be free to move them.

My reasoning is that these are really not fitting to the SO lib that i would see as a library for gullwing lead style parts.

My suggestion would be to create a `Package_MiniCircuits` library specifically for these specialized footprints.

We can either decide to do this after this one has merged or you already place these into the new lib and we move the rest afterwards.
I would however suggest that you align all your still open symbol pull requests to fit that or at least make a note about the future move such that we do not create a problem for us later.

More details see https://github.com/KiCad/kicad-footprints/issues/1249

aewallin:
OK, I guess creating a new library is just:
- move the files with "git mv" into a new directory
- update fp-lib-table to include the new library
- somehow (how?) figure out which symbols use the moved footprints - so the symbols can be flagged for updating the library

poeschlr:
> somehow (how?) figure out which symbols use the moved footprints - so the symbols can be flagged for updating the library

a good text editor should do the trick. (one that can search a full directory. Example atom https://atom.io/)
Meaning it is quite easy to find out for stuff that is already in the library. But open pull requests are harder. Here i assume that you are the only one currently relying on the footprints you added very recently. I would further assume that you only have one footprint open right now that would be affected: https://github.com/KiCad/kicad-symbols/pull/1223

---

For the parts that where already in the lib by tag 5.0.2 (So the ones before you added many more)
- Mini-Circuits_CD541_H2.08mm (Not used)
- Mini-Circuits_CD542_H2.84mm -> used in the transformer lib
- Mini-Circuits_CD636_H4.11mm -> used in the transformer lib
- Mini-Circuits_CD637_H5.23mm -> used in the transformer lib

---

Another way of searching is by using `Package_SO:Mini-Circuits`. This results in the following:
- 1 symbol in the oszillator library
- 13 symbols in the transformer library.
- 14 symbols in the rf library

---

An alternative to having a manufacturer specific lib (which would not really fit the rest of the kicad library organization) would be to move the mini circuits stuff into the oszilator lib, the rf lib (new library but also already needed for another contribution) and transformer lib (also a new one).

aewallin:
- I don't have any strong opinion on the organization - someone with a better overview can decide on the libraries/directories and I'll try to adapt.
- Yes the ones that are open now are this one (BK377) and symbol https://github.com/KiCad/kicad-symbols/pull/1223
- if there's a schedule for the next release then it's good to know about it asap.

poeschlr:
The footprint looks good. Lets leave the reorganization for after you finished #1308

PR #1321, opened 2019-01-22
TO-SOT SMD: Add footprint for On Semiconductor ECH8 package

mrnuke:
ECH8 is the On Semiconductor's flat lead version of SOT23-8. It's fatter, shorter, and cheaper (less metal).

------------

- [x] Provide a URL to a [datasheet](https://www.onsemi.com/pub/Collateral/TIG058E8-D.PDF) for the footprint(s) you are contributing


mrnuke:
The footprint, as seen by the footprint editor
Comparison with DPAK and 0805

mrnuke:
[3D model PR](https://github.com/KiCad/kicad-packages3D/pull/516)

poeschlr:
The courtyard clearance in y direction is not quite right.

The roundrect ratio should be 0.25 (25%) according to IPC. (With a maximum radius of 0.25mm.)

screenshot from 2019-01-23 21-10-46

poeschlr:
The purpose of courtyard is to allow enough space for pick and place machines and possibly for rework equipment (The later is the reason for the much enlarged clearance for BGAs.).

Courtyard is to be designed relative to pads and the nominal part body to allow for this. The nominal part body is also used for the fab layer according to KLC. This means the fab layer outline can be used to determine the courtyard area.

Silk does not play a role here.

---

You use rounded rectangles. If we use them then we should follow current industry standards. In this case we follow what is already known about the future IPC-7351C (It has already been announced that rounded rectangle pads will be added to that version. https://ocipcdc.org/archive/What_is_New_in_IPC-7351C_03_11_2015.pdf#page=9 )

mrnuke:
Thank you for taking the time to elaborate. Issues that were pointed out should be fixed in latest push.

onsemi-ech8-improved

poeschlr:
Thanks.

A remark for the future: please do not use force pushes (commit --amend) it makes reviews harder than they need be. We use squash merging anyways so there really is no need for removing the history from your personal branch.

PR #1333, opened 2019-01-23
Footprint for SO-8_3.9x4.9mm_P1.27mm

johnfwhitmore:
This is for the PCF8523 RTC device

https://www.nxp.com/docs/en/data-sheet/PCF8523.pdf

Signed-off-by: John Whitmore <johnfwhitmore@gmail.com>

This footprint was produced by the kicad-footprint-generator. The yaml file for which is in the pull request:

https://github.com/pointhi/kicad-footprint-generator/pull/273

screenshot from 2019-01-23 21-39-34


poeschlr:
Review added to the script pull request.

johnfwhitmore:
screenshot from 2019-01-29 18-31-05

poeschlr:
Thanks for this contribution and a special thanks for looking into the scripting stuff.

johnfwhitmore:
@poeschlr Thank you for your help! Was struggling with that a bit but it's easy when you know how. I'll try do another soon just to keep this in my head.

evanshultz:
@poeschlr 
Did you notice this has a 'SO' prefix instead of 'SOIC'? This footprint is not identical to, but I believe compatible with, https://github.com/KiCad/kicad-footprints/blob/master/Package_SO.pretty/SOIC-8_3.9x4.9mm_P1.27mm.kicad_mod. The NXP doc references JEDEC MS-012 so even if NXP dimensions slightly differently (tighter) the footprint can still be compatible. Why do we have essentially the same footprint with two different names?

PR #576, opened 2018-05-08
creation

Misca1234:
Overhaul of the DHT11
(recreated push from https://github.com/KiCad/kicad-footprints/pull/407)

Foot print push

Symbol push

3D model push

3D model source code push

bild

bild

bild

bild

bild

------------

Thanks for creating a pull request to contribute to the KiCad libraries! To speed up integration of your PR, please check the following items:

- [ ] Provide a URL to a datasheet for the footprint(s) you are contributing
- [ ] An example screenshot image is very helpful 
- [ ] If there are matching symbol or 3D model pull requests, provide link(s) as appropriate
- [ ] Check the output of the Travis automated check scripts - fix any errors as required

PR #1336, opened 2019-01-24
Added the foot print for sensor DHT11

Misca1234:
Push of the sensor DHT11

Datasheet
http://akizukidenshi.com/download/ds/aosong/DHT11.pdf

Symbol push
https://github.com/KiCad/kicad-symbols/pull/1456

Foot print push
https://github.com/KiCad/kicad-footprints/pull/1336

3D model push
https://github.com/KiCad/kicad-packages3D/pull/520

3D model script push
https://github.com/KiCad/kicad-packages3D-source/pull/205

bild

bild

bild

------------


poeschlr:
I would assume this is replacing https://github.com/KiCad/kicad-footprints/pull/576

myfreescalewebpage:
@Misca1234 

- [ ] We agreed previously (chinese datasheet) that the distance between the pin and the back of the package was 2.55mm, you have moved the package to center the pin, which is wrong.
- [ ] Pin 1 should be a square, why using a rectangle ?

Joel

Misca1234:
> I would assume this is replacing #576
Yes, dont need the back pack from all comments


> we agreed previously (chinese datasheet) that the distance between the pin and the back of the package was 2.55mm

There was no agreement around 2.55, you asked where I got it front and I pointed to the Chinese data sheet. 
However we agreed on that in the case where the English data sheet conflicted with the Chinese then the English should rule over.
I ignored the 2.55 in the Chinese data sheet because in the English the thickness of the component is 5.5 but in the Chinese it is 5.8 so the English and Chinese version of the data sheet do not even agree on thickness of the component so i just placed the pins in the middle as the image in the English data sheet strongly suggest.

Misca1234:
> Pin 1 should be a square, why using a rectangle ?
Pin 1 have been modified to be square

myfreescalewebpage:
@Misca1234 you previously indicated it's an alias, so adding the alias to DHT11 will be easy :)

Misca1234:
I hope so, lets see how many variants of the data sheet there is and if someone actually measure it in real life

PR #1338, opened 2019-01-24
Add LTST-S326 series right-angle LEDs

EvilMav:
This is a series of dual right-angle LEDs of different colors. The manufacturer has a separate datasheet for each, but seems like the packages are the same.

[Example datasheet](http://optoelectronics.liteon.com/upload/download/DS22-2000-287/LTST-S326KGJRKT.PDF)
kicad_ltst-s326

I've used a circle silkscreen mark as the package has a dot at the same spot.

KLC warning F6.2 is due to anchor position: I didn't find a proper recommendation in the datasheet but it seems like positioning the anchor in the middle of the non-transparent part of the packages makes the most sense.

------------


poeschlr:
Sorry it took us so long. Looks good, thanks. (First time right is very refreshing)
Screenshot from 2019-06-01 11-01-57

PR #1339, opened 2019-01-26
Moving mini circuits footprints out of Package_SO lib

poeschlr:
The mini circuits footprints in the Package_SO lib have been placed there by mistake.
(Non of them are gullwing parts.)

This is a partial fix for #1249

Most of these footprints have been added recently by @aewallin (They have never been part of a kicad release)
Meaning we should be free to move them. But we really need to get this done till 5.1.0 is released! (We are stuck with whatever library they are in at that point in time till we start v6 development.)

Symbol pull request: https://github.com/KiCad/kicad-symbols/pull/1459

---

Why a special RF_Mini-Circuits library?

Well mini circuits uses the same package for many different parts. Even one of the packages already in the lib is used for both a mixer and a transformer:
screenshot from 2019-01-26 13-39-28


aewallin:
There are mini-circuits parts also in Filter.pretty  and in DFN_QFN - not sure if you want to move them also?
- Filter_Mini-Circuits_FV1206-X.kicad_mod X=1,4,5,6,7
- DFN QFN Mini-Circuits_DL805.kicad_mod
- DFN QFN Mini-Circuits_FG873

if these are not moved - then for future contributions some nice simple rule of where new footprints should go is needed. 
a simple rule would be "mini-circuits parts go into RF_Mini-Circuits" - but then this means current footprints in Filter.pretty and DFN_QFN should also be moved?

poeschlr:
The ones in the filter lib look like they are truly specialized to filters. (At least all but one of them)

Maybe some of the ones i moved now are also specialized to some other lib but it is imperative to me that they vanish from the SO lib as they really definitely do not fit there. We can discuss splitting the RF_mini-circuits lib at a later time. (This can be something that can wait for v6 to be honest. If we have time befor v5.1 then we can do it before that.)

myfreescalewebpage:
@poeschlr The modification currently violates F9.3, should be corrected.

poeschlr:
@myfreescalewebpage a tip for the future: write out the problem. Nobody knows which rule is for what. ;)

Yes i will fix the 3d path as it is a problem created by this pull request

myfreescalewebpage:
Sorry, that's usually what I do, but I know you are experienced and in this case it's in the travis log :)

Thanks,
Joel

poeschlr:
@myfreescalewebpage 3d settings have been fixed

myfreescalewebpage:
@poeschlr checked, fine for me, merging.

myfreescalewebpage:
Same time with @evanshultz ;-)

evanshultz:
Sorry. No offense. Just got to my PC and saw Rene's request so I merged it to get it done.

myfreescalewebpage:
No worries @evanshultz I was left last 2 days for business trip :)

PR #1345, opened 2019-01-27
Update of LED_PLCC_2835

blackvladimir:
Changes:
 -  added reference to fabrication layer
 - changed fabrication layer, original has a plus mark on pin 2 which is confusing. Fabrication layer changed to include bevel on pin 1 (same as the physical package)
 - moved the anchor to the middle of the part
 - moved courtyard so it is 0.25mm from part and footprint (was short and wide)
 - updated datasheet URL (original was moved)
 - more distinct pad 1 indication in silk
 - switched silk and fabrication texts

Datasheet:
http://www.everlight.com/file/ProductFile/201407012101031690.pdf
Other part using the same package:
http://datasheet.octopart.com/LL-R2835PGC-G5H-M60-Lucky-Light-datasheet-41210093.pdf

Before
before
After
after

---


poeschlr:
There is a handsolder version of this. Could you align it to the same style as well?
Also check the pad rotations. (currently at 180 degree, there would be no change to the resulting footprint if you set it to 0)

The pad positions and sizes do not quite agree with the datasheet.

screenshot from 2019-02-12 22-08-07

blackvladimir:
> There is a handsolder version of this. Could you align it to the same style as well?

Done. It is the same, only pads are 0.3mm longer on both sides

> Also check the pad rotations. (currently at 180 degree, there would be no change to the resulting footprint if you set it to 0)

You are right, I rotated it to 0.

> 
> The pad positions and sizes do not quite agree with the datasheet.

I used the second datasheet http://datasheet.octopart.com/LL-R2835PGC-G5H-M60-Lucky-Light-datasheet-41210093.pdf I think better to use larger one.

poeschlr:
Updated review against that datasheet (smaller pad is wrong in x direction, both pads too small in y direction)

screenshot from 2019-02-13 19-47-40

blackvladimir:
Fixed:
plcc

poeschlr:
Sorry for the long wait. I have been quite busy at work.

You still have the first datasheet in the description of the footprint, please update that as you say you used the second one to design the footprint.

The pad sizes and positions look good now.

blackvladimir:
Link updated. I used address directly to manufacturer instead of octoparts. The datasheet is little bit different but footprint drawing on page 2 is the same.

https://www.luckylight.cn/media/component/data-sheet/R2835BC-B2M-M10.pdf

poeschlr:
Thanks for your contribution.

PR #1347, opened 2019-01-28
RF_Module: Fix issues of HOPERF_RFM9XW_THT (#1346)

### LINKED ISSUE #1346: RF_Module: Issues in HOPERF_RFM9XW_THT
During my inspection I've found the following problems:

- Pin 1 is circular (Should be squared instead);
- Holes drill size is 0.762mm (Should be 0.8 mm, as per datasheet);
- F.SilkS thickness is 0.1mm (The current recommended value is 0.12 mm, as per KLC);
- F.Fab thickness is 0.12mm (The current recommended value is 0.1 mm, as per KLC);
- F.CrtYd is 0.33 mm away from body edge of the footprint (The current recommended value is 0.25 mm, as per KLC);
- Missing pin 1 indication at F.SilkS layer
- Outdated manufacturer datasheet link (broken)
Current: http://www.hoperf.com/upload/rf/RFM95_96_97_98W.pdf
Updated: https://www.hoperf.com/data/upload/portal/20181127/5bfcbea20e9ef.pdf
- Holes groups are 12.8mm away from each other (The correct value is 13.6mm).
image823

The correct distance information is missing in the RFM9X series datasheet. I found the correct distance in the datasheet from a similar product of the same manufacturer; RFM69HCW (https://www.hoperf.com/data/upload/portal/20181127/5bfcb8284d838.pdf).
rfm_modules_package_dimension

which looks like correct

I've provided the due corrections and I intend to open a PR.

Otavio Augusto

gmsotavio:
Fix #1346 

Before
hoperf_rfm9xw_tht_old

After
hoperf_rfm9xw_tht_new

Datasheet link: https://www.hoperf.com/data/upload/portal/20181127/5bfcbea20e9ef.pdf

Otavio Augusto

---


gmsotavio:
@poeschlr Why "Pending changes" label ?

evanshultz:
I'll chip in.

Please provide a summary list of the changes for each PR. It's much easier to review and understand your objective than seeing only images and combing through the file diff.

- The silk pin 1 mark is blocked after assembly. Please move the chevron outside the body or, my preference, remove silk in the pin 1 corner.
- Shouldn't the two opposite corners of the fab rectangle be at (-1,-1) and (15,15)?
- The description mentions it's an SMD footprint but that's clearly not the case.
- The link mentions only SMD modules. I can't find a document on their site for a THT version but I suppose this footprint is meant for taking the existing module, which has drill holes and also allows fillets on SMD pads, and adding pins to it. In that case, how did you come up with the drill and pad sizes? You changed them so I assume you have a reason for not keeping the existing padstacks.

poeschlr:
I seem to have forgotten to post my review. (or github screwed up)

If i remeber correctly then i would have requested the pin 1 marker to be moved to the outside such that it is visible even when the module is placed. (so the same as evan already requested.)

The review by evan is much more precise as i remembered mine.

evanshultz:
@gmsotavio 
Will you be coming back to this PR soon?

gmsotavio:
@evanshultz . Sorry by delayed response. I will come back to this PR, but due my heavy workload I'm not having any time in the last few weeks. I intend to return to contribute here in a few weeks. Sorry.

gmsotavio:
@evanshultz and @poeschlr, first of all, thanks for the careful reviews. Some comments about @evanshultz reviews

> The silk pin 1 mark is blocked after assembly. Please move the chevron outside the body or, my preference, remove silk in the pin 1 corner.

I agree. Like you, I prefer remove Pin-1 mark. But I ask myself if it is allowed to do this. KLC (http://kicad-pcb.org/libraries/klc/F5.1/) states "_Pin-1 designator is provided on the F.SilkS layer_". For me KLC is not clear enough if pin-1 mark is either mandatory or not.

> Shouldn't the two opposite corners of the fab rectangle be at (-1,-1) and (15,15)?

I've not understand why. Before I changed the footprint the drill holes groups was 12.8mm away from each other. The correct distance information is missing in the [manufacturer documentation](https://www.hoperf.com/data/upload/portal/20181127/5bfcbea20e9ef.pdf). I found the correct distance in the datasheet from a similar product of the same manufacturer; RFM69HCW (https://www.hoperf.com/data/upload/portal/20181127/5bfcb8284d838.pdf). It looks like that 13.6mm is the correct distance, like depicted bellow
51804716-832a7480-224b-11e9-828e-6a30dd35eff3

(16 - 13.6)/2 = 1.2. Then I think the two opposite corners of the F.Fab rectangle must be at (-1.2,-1.2) and (14.8,14.8)? 14.8 - (-1.2) = 16. What do you think ?

>The description mentions it's an SMD footprint but that's clearly not the case.

I agree. Actually should be "THT". I'll fix it.

>The link mentions only SMD modules. I can't find a document on their site for a THT version but I suppose this footprint is meant for taking the existing module, which has drill holes and also allows fillets on SMD pads, and adding pins to it.

This is the thought. According with manufacturer [documentation](https://www.hoperf.com/data/upload/portal/20181127/5bfcbea20e9ef.pdf) (p. 121) it are de facto an SMD package, however, it can be used as a THT module due the drill holes. Remember that I haven't created this footprint. I have been fixing it only.

>In that case, how did you come up with the drill and pad sizes? You changed them so I assume you have a reason for not keeping the existing padstacks.
Good question! But, I also ask myself how the original author come up with this drill and annular ring sizes. The manufacturer specify the drill radius as being 0.4 mm, so the diameter should be 0.8 mm. I've changed from 0.762 mm (original) to 0.8 mm. The annular ring is not specified, so I used the same values found in the PinHeader P2.0mm from the kicad library, that satifies KLC (http://kicad-pcb.org/libraries/klc/F7.5/). Am I wrong ? Is there alternative suggestions?

Screenshot_2019-04-05_22-30-25

Regards,
Otavio Augusto

evanshultz:
Thanks for making these updates!
- I wasn't saying remove the pin 1 mark entirely. Just remove it in the corner. Around the rest of the part he silk can remain. Removing the silk just in the pin 1 corner will work as well as adding some silk by pin 1. You can see many of our BGA footprints do this.
- I think we're both wrong. I caught 1mm down and missed the 13.6mm row spacing. It should be (-1.2,-1) and (14.8,15). Is that right? (Note the minus signs aren't correct for KiCad but you get the idea.)
- HopeRF just tells us their module drill size, not the size of the pins on the header. So we will have to guess. I don't know the idea of the original footprint, but I think your suggestion is fine. I just wanted to how you came up with the updated padstacks. Thanks for answering.

Could you please post a new screenshot after making updates?

gmsotavio:
> I wasn't saying remove the pin 1 mark entirely. Just remove it in the corner. Around the rest of the part he silk can remain. Removing the silk just in the pin 1 corner will work as well as adding some silk by pin 1. You can see many of our BGA footprints do this.

Done! I've followed your suggestion.

> I think we're both wrong. I caught 1mm down and missed the 13.6mm row spacing. It should be (-1.2,-1) and (14.8,15). Is that right? (Note the minus signs aren't correct for KiCad but you get the idea.)

Done. I think that I got it. Additionally, I have changed the F.Fab pin-1 indicator style to follow the style of the SMD version.

> Could you please post a new screenshot after making updates?

Here it is


Screenshot_2019-04-13_23-27-09

evanshultz:
This isn't the pin 1 mark I had in mind on the silk layer, but I'll merge anyway. Here is a footprint showing the removal of silk to indicate pin 1:
image

PR #1354, opened 2019-02-01
DRC0010J via to solid

jneiva08:
Changed vias to be solid, forgot to change in the initial commit.

---


poeschlr:
Why? We typically do not do this. The only pads that we set to solid are polygon pads as they do not offer thermals as an option.

jneiva08:
Used this footprint for a dc-dc, if i put solid when have ground plane in the bottom, it have the best dissipation even the other components have thermal reliefs for soldering

poeschlr:
Yes but that should be made in the zone settings not at the footprint level. At least not for generic footprints in the library.

What is however missing is a rectangular pad on the bottom layer covering all vias. (See any qfn thermal vias variant.)
That pad should have the same pad number as the vias and only copper activated (no mask and no paste active) See: http://kicad-pcb.org/libraries/klc/F4.4/

jneiva08:
I will have attention for next time.
@poeschlr  can I close this PR?

poeschlr:
I created an issue such that we remember that this one is missing the bottom pad: https://github.com/KiCad/kicad-footprints/issues/1400

PR #1355, opened 2019-02-01
Added SIM28ML footprint

kajusK:
Adding footprint for SIM28ML gps module
https://simcom.ee/documents/SIM28ML/SIM28ML_Hardware%20Design_V1.01.pdf

2019-02-01-122841_1920x1080_scrot


kajusK:
Ahh, I was using old version of fp-lib-table and didn't notice the RF_GPS library, fixed. Thanks for pointing this out.

poeschlr:
Looks good, thanks.

For reference:
screenshot from 2019-02-10 20-17-48

Issue #759, opened 2018-07-18
Inconsisten footprint naming in C&K switches


antoniovazquezblanco:
Examples:
`SW_SPST_CK_KSC6xxJLFx`
`SW_CuK_JS202011AQN_DPDT_Angled`

What should we use for https://www.ckswitches.com/?

antoniovazquezblanco:
I know this is low-priority and targets v6 but if we define a naming convention, then we will be able to merge new additions without having to rename them in the future.

Alternatives:
- CK (Used in their URL)
- C&K (Used in their legal texts https://www.ckswitches.com/legal/)
- CuK (Only seen in KiCad)

robkam:
It should be more generic than that. Leave out the manufacturer name and model. There are multiple brands and part names sharing the same footprints.

antoniovazquezblanco:
Thank you for your input @robkam but unfortunatelly some of the footprints in the repository are manufacturer specific so we should anyways reach a consensus about how we name this manufacturer.

poeschlr:
Can the ampersand sign be used in footprint names?

poeschlr:
I just tested it. & is ok for kicad. It is however maybe a different story for our footprint download page. (& would need to be escaped in urls)

antoniovazquezblanco:
I would avoid using it. Footprint include that characters in filenames and could cause problems in some FS types. Given that the ampersand character is not accepted in G1.1 I would not use it. Just in case.

PR #1356, opened 2019-02-01
add a few C&K toggle switches

robkam:
add vertical THT C&K toggle switches, SPST, SPDT and DPDT, miniature & sub-miniature

Mini https://www.ckswitches.com/media/1394/7000toggle.pdf

Sub-mini https://www.ckswitches.com/media/1391/ttoggle.pdf

capture


robkam:
Please review.

poeschlr:
Just from the screenshot alone i can already tell you that these are missing pin 1 markers on both silk and fab layers.

I assume the bottom ones are circular hole versions of the top ones. I am not so sure the ones on the bottom right have enough pad to pad clearance to be of any use. It might simply be that we can not offer such an option for every part.

diegoherranz:
Thanks, I'll comment on that PR.

robkam:
I left out pin 1 markers because these're non polarised. There's a few footprints already in Button_Switch_THT without pin 1 markers on silk/fab layers, e.g. the similar SW_NKK_G1xJP Could I have same clarification please.

Yes the bottom ones are the circular hole versions. Included because the holes are cheaper to have produced. The clearance is 0.45 mm between the sub-mini circular pads, at their closest.

poeschlr:
To make separate pull requests you need to have one branch per contribution.

I am assuming you have the official library repo added as a second remote and called it upstream:
``` bash
git fetch upstream  # get the state of the official repo
git checkout upstream/master  # change to the current head of the official repo
git checkout -b new_sensible_branch_name # create a new branch from that state
git push --set-upstream origin new_sensible_branch_name # push that new branch to github
# after you made your changes work normally with
git add -u # or git add list_of_files
git commit # enter sensible commit message as usual
git push # push your changes online
```

if you then are requested to do some changes you would need to checkout the correct branch and then do your changes.

``` bash
git checkout branch_connected_to_pull_request
# do your changes
git add -u # or again git add file_list
git commit # detail what you changed in the commit message
git push # push your fixes to github (will automatically appear in the pull request)
```

---

To add the official repo as a second remote called upstream (one time setup):
``` bash
git remote add upstream https://github.com/KiCad/kicad-footprints.git
```

robkam:
Thanks that's useful :-) ... except I'm doing it all through a browser via GitHub.

poeschlr:
The browser workflow has its limitations. So it might be time to start looking into having a local git clone.

poeschlr:
Regarding workflow: I would guess that even the browser workflow offers some way to make a new branch for every pull request you open. But i have never used the browser for anything as it is simply way more work that way.

robkam:
I'll look at the command line as I seem to have got myself stuck with the browser.

poeschlr:
There are git tools with graphical interfaces if you would feel more at ease that way. But the only thing that is truly universal is the command line.

evanshultz:
I started to look at this but didn't have time to get anything written down before something else came up. However, I will note that these will need the manufacturer name on them or some real definition about what type of switch fits in this footprint. Even if there are multiple switches that could be so you don't want to be too specific, the current naming is way too generic.

robkam:
Something like `SW_CK_Toggle_DPDT_Vertical_Miniature.kicad_mod` or `SW_CK_DPDT_T_Series.kicad_mod`?

evanshultz:
I'm not sure that the switch configuration matters for these switches. One footprint could support a variety of switch configurations.

For https://www.ckswitches.com/media/1394/7000toggle.pdf how about:
- `SW_CK_T71xx`
- `SW_CK_T72xx`
- `SW_CK_T73xx`
- `SW_CK_T74xx`

And for https://www.ckswitches.com/media/1391/ttoggle.pdf how about:
- `SW_CK_T10x`
- `SW_CK_T2xx`

I recall a discussion about switch names which I may not have found just now, so there might be some tweaks. But that ought to at least be in the ballpark. https://github.com/KiCad/kicad-footprints/issues/759 and https://github.com/KiCad/kicad-symbols/issues/580 (even though it's about symbols there are lots of linked issues and PRs) are good places to look.

poeschlr:
The switch configuration matters if we want to have better footprint filters on the symbol side. (I assume with switch config you mean things like normally open, number of possible positions, ...)

robkam:
Then:
SW_CK_Lever_DPDT_72xx.kicad_mod
SW_CK_Lever_DPDT_72xx_CircularHoles.kicad_mod
SW_CK_Lever_DPDT_T2xx.kicad_mod
SW_CK_Lever_DPDT_T2xx_CircularHoles.kicad_mod
SW_CK_Lever_SPDT_71xx.kicad_mod
SW_CK_Lever_SPDT_71xx_CircularHoles.kicad_mod
SW_CK_Lever_SPxT_T10x.kicad_mod
SW_CK_Lever_SPxT_T10x_CircularHoles.kicad_mod
?

evanshultz:
Ahh, right. Thanks, Rene.

If there are no options besides `Lever` that seems superfluous, but otherwise good.

poeschlr:
The idea behind including Lever (and other such options) is such that similar switches are grouped together.

To be honest it is kind of an experiment. I would suggest that we keep the current naming scheme for a while and see how it develops. If it is a hindrance to what we want to achieve then we can look at it again and see if we can find something better for v6.
(Right now we only use the naming convention for new stuff. That is such that we don not necessarily break old projects.)

PR #1365, opened 2019-02-03
Added USB_C_Plug_JAE_DX07P024AJ1.kicad_mod

Misca1234:
This is a repush of an assumed abandoned push
https://github.com/KiCad/kicad-footprints/pull/507

DX07P024AJ1 (Slim Plug)
http://www.jae.com/z-en/pdf/MB-0301-2E_DX07_PLUG.pdf

The easiest way to review this is to move the bottom F.Fab line to x, y=0,0 via "move exactly".

bild

bild

bild

bild

---


poeschlr:
Is the line that is currently on the fab layer intended to be the edge cut? If so then move it to the edge cuts layer please.

My assumption would be that the part is then outside of that outline. So that side might be best included as well to show the user where the board ede can not return. (Would suggest the full outline on the fab layer to be honest. But no silk as it is fully outside the board.)

---

The A side side pads should not have paste (See note 3)

I am not so sure the large shield pins are on both the A and B side. (They are not in the drawing that i assume to show the A side.)

---

A few dimensions are not quite right. And the drawing has at least two degrees of freedom (For the large S1 pads. Y direction is not fixed at all. The orange dimensions.)

screenshot from 2019-02-10 19-59-44

Misca1234:
> Is the line that is currently on the fab layer intended to be the edge cut? If so then move it to the 
> edge cuts layer please.

Is it not possible to "draw line" in Edge.Cuts, nor change a line to that type, do you want me to add it manually via an text editor ?

poeschlr:
Yes the text editor might be required. (Wayne was against adding this to v5 for some strange reason. Even if it would be a hidden feature. https://bugs.launchpad.net/kicad/+bug/1251393)

Misca1234:
It looks like this now

bild

poeschlr:
Check the edgcuts outline. There are a few points where the endpoints of the next elements do not quite meet. (marked pink)
There is also a short line in there that should not be there. (marked red, reported by DRC when trying to use the footprint.)

Screenshot from 2019-03-21 20-56-07

I did not check the measurements again as i would need to check them again after you fixed the endpoint alignment.

Misca1234:
Now so, it should have been fixed

poeschlr:
It seems like the JAE datasheet is not available today. Will check it again later but we might need to search an alternative.

Misca1234:
They seemsed to have rearranged their sire, I find it again and have updated the foot print
https://www.jae.com/en/searchfilter/?topics_keyword=DX07P024AJ1&mainItemSelect=1

poeschlr:
The large shield pads are only in the bottom side of the drawing. You have them on both sides.
There is also still the question on how you came up with their dimension and positioning in y direction.

There are still some dimension errors.

usb svg

Misca1234:
Made some modification after review

bild

poeschlr:
And the question about the y direction (orange dimensions in my drawing)?

Misca1234:
I can not find any width for the pads, more than looking how they "cover" the pads below, 

I went to their site and asked for their 3D model so we can "attach" it to the foot print and see how good it looks, I have not yet got the download link for the model, I guess it is manually accepted

poeschlr:
My guess would be that the pads should have the same distance from the edge cuts on both sides so possibly move both to the left. (If you did not have any reference for them. Maybe also ask for clarification as it is kind of strange that they leave 3 degrees of freedom in their documentation.)

Misca1234:
I got hold of the 3D model, when I apply it to the foot print and mess around a little it is clear that the "wide" shield pad is to large

bild

I could not find hoe to change the thickness of the PCB in the 3D viewer so I moved it around a little manually

bild

bild

Misca1234:
Going to away for 5 days, will not be able to fix anything until then

Misca1234:
@poeschlr 
Should I change the footprint so it fits the 3D model?

poeschlr:
Only things that are not dimensioned in the datasheet.

Misca1234:
As you can see in the last image the left pad does not really fit, but the foot print should be according to data sheet, otherwise I can not see any changes

poeschlr:
The edge cuts measurement i marked in my last review is still not exact.
(The leftmost edge cuts line is on -4.3025 in x direction but should be 4.3000)

Misca1234:
Done 

Changed the leftmost edge cut x coordinate to -4.3 and aligned the end point in y direction to match next edge cut line

poeschlr:
thanks

PR #1366, opened 2019-02-03
Scripted SOIC-8-1EP footprint for HIP2100

poeschlr:
https://www.renesas.com/eu/en/www/doc/datasheet/hip2100.pdf#page=13

symbol PR: https://github.com/KiCad/kicad-symbols/pull/1481
script PR: https://github.com/pointhi/kicad-footprint-generator/pull/287

screenshot from 2019-02-03 15-34-32

```yaml
SOIC-8-1EP_3.9x4.9mm_P1.27mm_EP2.514x3.0mm:
  size_source: 'https://www.renesas.com/eu/en/www/doc/datasheet/hip2100.pdf#page=13'
  body_size_x:
    minimum: 3.8 # rounded from 3.811
    maximum: 4.0 # rounded from 3.99
  body_size_y:
    minimum: 4.8
    maximum: 5.0 # rounded from 4.98
  body_height:
    minimum: 1.43
    maximum: 1.68

  overall_size_x:
    minimum: 5.84
    maximum: 6.2
  lead_len:
    minimum: 0.41
    maximum: 0.89
  lead_width:
    minimum: 0.35
    maximum: 0.49

  pitch: 1.27
  num_pins_x: 0
  num_pins_y: 4

  EP_size_x:
    nominal: 2.514 # only maximum given but that is not supported by the script
    tolerance: 0

  EP_size_y:
    nominal: 3.0 # only maximum given but that is not supported by the script
    tolerance: 0

  EP_num_paste_pads: [2, 2]

  thermal_vias:
    count: [2, 3]
    drill: 0.2
    paste_via_clearance: 0.1
    EP_num_paste_pads: [2, 2]
    EP_paste_coverage: 0.7
    grid: [1.4, 1.0]
    paste_avoid_via: False
```

PR #1369, opened 2019-02-03
Remove old SOIC-8 ep footprint

poeschlr:
This footprint did not fit any of the symbols that currently use it. New
scripted footprints have been added in other commits to replace it for
the symbols.
The datasheet assigned to the footprint does not really contain any
drawing that would fit this footprint.

Sister pull requests:
- #1364
- #1366
- #1367
- #1368


evanshultz:
All referenced PRs are now merged.

PR #1371, opened 2019-02-03
Scripted version of non EP SOIC footprints

poeschlr:
Screenshots are a bit unnecessary here.

The additions to the script are in this commit: https://github.com/pointhi/kicad-footprint-generator/pull/287/commits/29f2e243f1b4bedfd5db12ed739bbd9c48502a30 (part of pull request https://github.com/pointhi/kicad-footprint-generator/pull/287)
(No need to copy them here.)


evanshultz:
We could link to the (free with registration) JEDEC spec, such as https://www.jedec.org/system/files/docs/MS-012G.pdf (3.9mm wide SOIC) and https://www.jedec.org/system/files/docs/MS-013F.pdf (7.50mm wide SOIC). This would be a little more official than an individual company's drawings and probably more broadly applicable for this extremely common package.

For example, 3.9mm wide SOICs have 0.31-0.51mm wide leads by JEDEC, but www.st.com/resource/en/datasheet/l272.pdf#page=9 (used on SOIC-16_3.9x9.9mm_P1.27mm) is 0.35-0.46mm. While ST shows this for some reason, I bet if we looked at a larger number of datasheets we would find more of them that align with JEDEC than deviate.

That being said, the JEDEC docs show more max only and nom only dimensions instead of min/max or min/nom/max like the datasheets or package drawings you liked.

I just looked at the referenced datasheets but I can break out the info in the JEDEC docs if you like. If not, I checked the YAML file and everything looked good to me.

poeschlr:
The "free with registration" part is what worries me. I tried to register with jedec a month or so ago and did not yet receive a valid account. So i would really rather link something that is more easily available.

I can however look a bit more for a datasheet that agrees with jedec. (Sadly atmel did not provide such nice drawings for every one of our packages. So i need to look a bit longer for it.)

evanshultz:
Hmm. I signed up long ago and recall it came immediately. Could it have perhaps gotten marked as spam?

poeschlr:
I checked all my mail accounts and could not find anything in spam for any of them.
Did you give them your work address plus state the company you work for by any chance? (I suspect their automatic system does not like public mail servers (gmail) with company name set to KiCad)

evanshultz:
Indeed I did use my work email and address. That's disappointing if it's the reason.

Has anybody else registered for JEDEC? Already having an account, making one for another email address of mine might not be the typical registration process. @Misca1234 @antoniovazquezblanco @Ratfink @DanSGiesbrecht @diegoherranz @myfreescalewebpage 

I would still prefer linking to the JEDEC docs. We are using IPC info in the scripts which is only available by paying (or from "unofficial" sources). Even if there can be trouble signing up, the JEDEC docs are intended to be freely available and are accessible online.

@poeschlr 
There's no milestone above but I assume you would like this in 5.1?

poeschlr:
I would like any symbol side stuff marked as "bug" to be merged soon. (preferably before the 5.1 tag)  The "footprint side only" changes don't really change that much for users. It is mainly better for us as the scripted versions are much more easily maintained. (And yes the footprints are arguably better and more uniform so there is a small improvement for users as well. But that can be said about nearly any currently open pull request. Meaning there is no special priority to be given to these changes.)

I would however not block it just because we are not fully happy with the documentation links.

myfreescalewebpage:
@evanshultz I'm not registered for JEDEC.

evanshultz:
Hmmm... we use non-free IPC docs as source material in quite a few scripts. Those documents may be found online with some searching.

Here, we could specify the "source" package drawings from JEDEC, which are ostensibly free but also available online with some searching.

I'm not understanding why we don't get as close to the source of information in both cases, since the barrier to the information seems no greater here.

poeschlr:
I am prepared to add the JEDEC stuff as a second link. I would however keep the really easily available datasheets as the main ones as users need easy access to these to find if the footprint fits their part. (They don't need the same level of access to the IPC definitions as long as they do not intent to check the scripts for correctness.)

evanshultz:
OK. If you update the footprints with the JEDEC link then we can probably get this in before RC1.

poeschlr:
Well there is a complication. I don't have access to jedec stuff so somebody with access would need to supply the links and tell me which one goes with which footprint.

evanshultz:
See https://github.com/KiCad/kicad-footprints/pull/1371#issuecomment-460461545 above. I posted the JEDEC docs and their landing pages for the body widths.

MS-013 doesn't cover two footprints here:
- 16 pin, 12.8mm length (only 10.3mm for 16 pin)
- 28 pin, 18.7mm length (only 17.9mm for 28 pin)

There might be other JEDEC docs which have these body sizes. But for example, https://www.analog.com/media/en/package-pcb-resources/package/pkg_pdf/ri_soic_ic/ri_16_1.pdf mentions MS-013AC but it doesn't actually match those dimensions.

evanshultz:
@poeschlr 
Ping! Shall we get this merged and into 5.1?

Here are the JEDEC doc numbers and variation codes for the footprints included here:
- SOIC-14W_7.5x9mm_P1.27mm.kicad_mod: JEDEC MS-013AF
- SOIC-14_3.9x8.7mm_P1.27mm.kicad_mod: JEDEC MS-012AB (actually the spec shows 3.9x8.65mm)
- SOIC-16W_7.5x10.3mm_P1.27mm.kicad_mod: JEDEC MS-013AA
- SOIC-16W_7.5x12.8mm_P1.27mm.kicad_mod: no JEDEC spec
- SOIC-16_3.9x9.9mm_P1.27mm.kicad_mod: JEDEC MS-012AC
- SOIC-18W_7.5x11.6mm_P1.27mm.kicad_mod: JEDEC MS-013AB (spec is 7.5x11.55mm)
- SOIC-20W_7.5x12.8mm_P1.27mm.kicad_mod: JEDEC MS-013AC
- SOIC-24W_7.5x15.4mm_P1.27mm.kicad_mod: JEDEC MS-013AD
- SOIC-28W_7.5x17.9mm_P1.27mm.kicad_mod: JEDEC MS-013AE
- SOIC-28W_7.5x18.7mm_P1.27mm.kicad_mod: no JEDEC spec
- SOIC-8_3.9x4.9mm_P1.27mm.kicad_mod: JEDEC MS-012AA

poeschlr:
I am a bit busy right now. So i am not sure if i get all this done in time for 5.1. (I advice against rushing anything just to get it into an arbitrary release.)

evanshultz:
Straight from the JEDEC docs...

Body sizes can have their dimensions adjusted:
- 7.5mm wide:
-- Body X is 7.5 +/-0.1mm (7.4-7.5mm).
-- Overall X is 10.3 +/- 0.33mm (9.97-10.63mm).
-- Tolerance on body Y is +/-0.1mm.
- 3.9mm wide:
-- Body X is 3.9 +/-0.1mm (3.8-4.0mm).
-- Overall X is 6.00 with no tolerance given in rev G of the doc, at least that I can see, but the same tolerance (`bbb`) from the above doc is +/-0.2mm which would give us 5.8-6.2mm. And I found the rev F doc online which is 6 +/-0.2mm so I think that's it.
-- Tolerance on body Y is +/-0.1mm.

Pin widths for all are 0.31-0.51mm. These also have an additional tolerance shown but I think that's bogus and this is the correct info to input to the script.

poeschlr:
A note: the footprint name does not always agree with how the size definition defines the body size. I rounded in the name to avoid needing to change a fuckton of symbols. (Can be done but i would suggest to do this separately one package at a time.)

evanshultz:
Sure. Just posting it here so you can adjust tolerances or other non-breaking items, if desired.

poeschlr:
Footprints updated as suggested.

New pull request for script changes as the original pull request got merged too early: https://github.com/pointhi/kicad-footprint-generator/pull/321

evanshultz:
Thanks!

PR #1375, opened 2019-02-04
Add MSOP-8 EP (mask defined) for MIC5355-S4YMME

poeschlr:
Note in datasheet states "as much copper as possible". Therefore very
large copper pad with mask at same size as nominal EP.

datasheet: http://ww1.microchip.com/downloads/en/DeviceDoc/mic5355_6.pdf#page=15
script: https://github.com/pointhi/kicad-footprint-generator/pull/289
symbol: https://github.com/KiCad/kicad-symbols/pull/1491

screenshot from 2019-02-04 23-08-46

```yaml
MSOP-8-1EP_3x3mm_P0.65mm_Mask1.73x2.36mm:
  size_source: 'http://ww1.microchip.com/downloads/en/DeviceDoc/mic5355_6.pdf#page=15'
  force_small_pitch_ipc_definition: True
  body_size_x:
    nominal: 3
    tolerance: 0.1
  body_size_y:
    nominal: 3.0
    tolerance: 0.1
  body_height:
    nominal: 0.94
    tolerance: 0.1

  overall_size_x:
    nominal: 4.9
    tolerance: 0
  lead_width:
    nominal: 0.33
    tolerance: [-0.1, 0.03]
  lead_len:
    nominal: 0.53
    tolerance: 0.1

  pitch: 0.65
  num_pins_x: 0
  num_pins_y: 4

  EP_size_x:
    nominal: 1.73
    tolerance: 0.15
  EP_size_x_overwrite: 2.5
  EP_mask_x: 1.73
  EP_size_y:
    nominal: 2.36
    tolerance: 0.15
  EP_size_y_overwrite: 3
  EP_mask_y: 2.36
  # EP_paste_coverage: 0.65
  EP_num_paste_pads: [2, 2]

  thermal_vias:
    count: [2, 2]
    drill: 0.2
    # min_annular_ring: 0.15
    paste_via_clearance: 0.1
    EP_num_paste_pads: [2, 2]
    # paste_between_vias: 1
    # paste_rings_outside: 1
    EP_paste_coverage: 0.7
    grid: [1, 1]
    # bottom_pad_size:
    paste_avoid_via: False
```


evanshultz:
1. The tolerance of the body size is closer to 0.12mm than 0.1.
1. Where did you get the 0.15mm tolerance on the EP? I'm not seeing that.
1. The EP width could be a bit more than 2.5mm since that leaves 0.25mm gap between the pads (assuming you still target 0..2mm gap). But this is OK too.

poeschlr:
The metric dimensions of the datasheet state 3mm +/-0.10 (or am i reading that bad resolution wrong?)

 ep tolerance is indeed wrong. (will not influence the footprint so no change needed here)

I wanted to keep the ep size on a nice number. (and keep a reasonable clearance while still kind of maximizing the size.)

evanshultz:
Alright then. It appears I was the one misreading the numbers.

PR #1382, opened 2019-02-06
Added footprint for RECOM RPM-series of DC/DC converters (1A, 2A, 3A and 6A).

fhw72:
Added LGA-25 footprint for Recom RPM-family of DC/DC converters.

- [x] Please see page 8
https://www.recom-power.com/pdf/Innoline/RPM-3.0.pdf

screenshot 2019-02-07 09 35 05


poeschlr:
The pin 1 marker on the fab layer is a bit small. should be 25% of the largest size or at most 1mm (I think the later would limit it here) The size of the chamfer is measured in x and y direction.

The body outline is too small (should be 12.19x12.19mm)
The current courtyard clearance is wrong. But the fab layer does not represent the body size right now so the measurement is not correct either. (Calculate it against the correct body size)

This is similar enough to BGA so i would guess we might want to use the same courtyard clearance. Meaning 1mm.

screenshot from 2019-02-10 19-10-42

fhw72:
I just pushed the changes.

poeschlr:
Is there a reason you used 0.5mm courtyard clearance?

screenshot from 2019-02-12 18-21-12

fhw72:
No. This was just me misinterpreting the additional margin to be 1mm in total and not on all edges. I increased it by 0.5mm on each side now.

fhw72:
Courtyard distance is now exactly 1mm

poeschlr:
It is now 1mm clearance but you need to round it to the 0.01mm grid (that is why there is a bit of tolerance for the courtyard clearance. In your case you would need 1.05mm clearance to fulfill both requirements.)

fhw72:
Rounded to 1.05mm.

poeschlr:
Thanks

PR #1383, opened 2019-02-06
Add Molex 200528 series (scripted via kicad-footprint-generator)

chapatt:
PR for script: https://github.com/pointhi/kicad-footprint-generator/pull/291

Datasheet: https://www.molex.com/pdm_docs/sd/2005280040_sd.pdf

footprint
3d_axonometric
3d_top


chapatt:
Regarding the TravisCI pipeline results:

- I wasn't aware 3D models were required for footprints to be submitted to the library (indeed, some of those in the existing library don't have 3D counterparts).
- I put the origin at the center of the "pick up area" designated on the datasheet.

Shackmeister:
only the 3D model reference is required. This way we won't have to edit the footprint if we are gonna add 3D models.

Pickup area beats Travis :)

poeschlr:
I checked the 13 pin version (random choice)
Some measurements of the body are not quite right.
The courtyard clearances in y direction are not correct.

screenshot from 2019-02-10 21-26-05

chapatt:
Thanks for the review!
 - The 3.96 was a typo and corrected to 3.95.
 - I was using a bad reference to calculate the x-coordinates of the actuator--fixed.
 - I was mistakenly offsetting the courtyards from the silkscreen (rather than the actual pads and device).
 - I believe the bits of the body that bracket the pins are indeed actually 0.35mm. If you zoom in the drawing, the contact patch of the lead does indeed extend 0.4mm, but the features in question are a bit short of that. The 3D model provided follows this line of thought and the features do measure to be 0.35mm for the plastic bits and 0.4mm for the leads. I believe I also got the 0.55mm width of these protrusions from the model.

poeschlr:
Looks good now. Thanks

screenshot from 2019-02-12 18-11-19

PR #1387, opened 2019-02-07
Add spark gap lib with one initial footprint

evanshultz:
This is to kick off a new footprint library for spark gaps. I created one to get started but I will make more if there is a reference or somebody requests the design.

Would it be possible to get this squeezed in for 5.1 so the lib is at least started? Then we can add more between this being merged and v6 instead of delaying this longer?

Here it is:
image

Notes:
- Suggestions on the footprint naming convention?
- I didn't add any fab drawings because I'm not sure it applies.
- No 3D model since our antennas don't have them and they're also just copper.
- I set the mask opening to exactly match the pads without any mask pad/polygon between the pads.

I didn't search for spark gap designs by any governing body. If somebody has a good reference that would be nice. Thanks!

---


poeschlr:
Do you have a document that shows why the footprint looks as it looks? Or is this one done by experience?

evanshultz:
Actually, I copied this from a random board so I don't know the design philosophy for sure. The application is a spark gap across the winding of a common mode choke with no parallel resistor and that's why the gap is so small. I've always avoided corners on parallel edges (corners pointing towards each other are OK) and I assume that's why the pads are oval. It all made sense to me so I figured it would be fine to kick off this library and get feedback from other folks on what spark gaps they want.

poeschlr:
I mean yes we can simply specify gabs by their dimensions. But i have seen very complex spark gabs out there. So it might be preferable to find some more generic way to describe them. (I don't really like rushing things in shortly before a release just such that they are in. This can easily result in us being stuck in a corner long term.)

Example this: https://forum.kicad.info/t/how-to-build-a-customized-pad-pcb-spark-gap/12872

evanshultz:
That's certainly fair. Removed milestone.

I think specifying the gap is something that belongs in the footprint name because it related to the breakdown voltage and that will be one (of maybe the) main point of the spark gap.

The `Oval` is for the pad shape, and it could be `Triangle` for use in the example above. Or, even better, `Teeth` to separate a single triangle from multiple triangles.

So the naming convention proposal is `SparkGap_Pad<pad x>x<pad y>mm_Gap<gap distance>mm_<pad shape such as Oval, Triangle, Teeth, Needle (the kind in the post above)>`.

Are there other points you're aware of to consider? Anybody else please chime in too.

poeschlr:
I think we might want to move the pad shape forward to be honest. Different shapes seem to be suggested for different applications.

evanshultz:
So  `SparkGap_<pad shape>_Pad<pad x>x<pad y>mm_Gap<gap distance>mm`?

poeschlr:
The pad size might not really be applicable for all shapes. zig zag types for example might not really be defined by it. Would the length of the "active edge" be a better specifier plus maybe a descriptor for used area? (But that would not work for types where only a sharp tip is used.)

Another option would be that we have a naming convention specific to every shape. The zig zag type would then have the shape and gab in common with your type. But might include "number of triangles" and maybe eight/width of triangle instead of a single size.

PR #1073, opened 2018-11-07
Creating a few parts popularly used with Eurorack synth modules

robkam:
Jack_3.5mm_QingPu_WQP-PJ398SM_Vertical.kicad_mod
was #820
Create footprint for mono vertical 3.5mm jack with oval pads and holes QingPu_WQP-PJ398SM
http://www.qingpu-electronics.com/en/products/WQP-PJ398SM-362.html (can't find a pdf)
mono oval holes qingpu_wqp-pj398sm

Jack_3.5mm_QingPu_WQP-PJ366ST_Vertical.kicad_mod
was #822
Create footprint for stereo vertical 3.5mm jack with oval pads and holes QingPu_WQP-PJ366ST 
http://www.qingpu-electronics.com/en/products/WQP-PJ366ST-364.html (no pdf)
stereo oval holes qingpu_wqp-pj366st

Jack_3.5mm_QingPu_WQP-PJ366ST_Vertical_CircularHoles.kicad_mod
was #823
Create footprint for stereo  vertical 3.5mm jack  with circular pads and holes QingPu_WQP-PJ366ST 
http://www.qingpu-electronics.com/en/products/WQP-PJ366ST-364.html (no pdf)
stereo circular holes qingpu wqp-pj366st

PR #1388, opened 2019-02-07
add 9 mm Alpha vertical pots

robkam:
add 9 mm Alpha single and dual vertical pots

was in "Creating a few parts popularly used with Eurorack synth modules" #1073 

http://www.taiwanalpha.com.tw/english2016/p_e_062.htm

annotation 2019-02-07 094655


robkam:
@[diegoherranz](/diegoherranz)

diegoherranz:
Thanks, @robkam, for splitting that PR on separate ones. It makes everything easier.
I'll review this one.

diegoherranz:
Things which look OK:
- Main pins 1mm diameter, 2.5mm pitch - OK
- Main dimensions: 9.5mm x 11.35mm - OK
- Courtyard: 0.25mm clearance - OK
- 7 mm diameter circle on F.Fab on centre of the shaft - OK

Things which need some changes:
- [ ] The website link doesn't seem to work. Maybe http://www.taiwanalpha.com.tw/downloads?target=products&id=113 ?
- [ ] I think that they should be called `Potentiometer_Alpha_RD901F-40-00D_Single_Vertical`. The `-00D` indicates that it's with bracket, which these ones are. That's the name that Alpha uses on the datasheet. Similar for the Dual one. Actually, the full part number is much longer and it includes more information about the shaft, resistance, etc. and optional modifiers for the bushing. If we were to have separate 3D models for different shafts, for instance, we would need a different footprint for each one. But I think that we can have a single footprint at the moment and if we eventually had multiple 3D models, we could split/replicate the footprints.
- [ ] The mechanical holes should be slots. If you also want a version with round holes, we could have additional footprints with this suffix `_CircularHoles.kicad_mod`. Example: [BarrelJack_CUI_PJ-063AH_Horizontal.kicad_mod](https://github.com/KiCad/kicad-footprints/blob/master/Connector_BarrelJack.pretty/BarrelJack_CUI_PJ-063AH_Horizontal.kicad_mod) and [BarrelJack_CUI_PJ-063AH_Horizontal_CircularHoles.kicad_mod](https://github.com/KiCad/kicad-footprints/blob/master/Connector_BarrelJack.pretty/BarrelJack_CUI_PJ-063AH_Horizontal_CircularHoles.kicad_mod) 
- [ ] Clearance between silkscreen and copper features is too small. See [KLC F5.1](http://kicad-pcb.org/libraries/klc/F5.1/).
- [ ] Pin 1 should be rectangular. See [KLC F7.3](http://kicad-pcb.org/libraries/klc/F7.3/). There seem to be many potentiometers in the library that don't meet this, but I can't see a reason not to.
- [ ] Please remove the manufacturer and part number from the descriptions and keywords. See [KLC S6.3](http://kicad-pcb.org/libraries/klc/S6.3/).
- [ ] I think that the keywords should all be lower case (not capitalised), although I can't find it in the KLC.

Many thanks!

robkam:
Changed to use data sheet at http://www.taiwanalpha.com.tw/downloads?target=products&id=113 and other suggested fixes done.

Is there any point having the 7 mm diameter F.Fab circle?

Original two foot prints now have _CircularHoles in name.

capture

robkam:
Oops, not sure what I've done with Alps pots above. Do I need to fix something?

diegoherranz:
> Oops, not sure what I've done with Alps pots above. Do I need to fix something?

It seems like you have removed them. Can you add them back (You can copy them from this online repo) and commit them in again? I think that should fix it.

I'm reviewing your latest changes.

Thanks!

robkam:
thanks, that seems to have fixed the deletions

diegoherranz:
Thanks for the changes. It's looking very good. I've updated the tickboxes on the changes requested before.
A few extra comments:

New checks which look OK:

- Oval slot matches footprint: 1.1mm x 1.8mm
- `_CircularHoles` variants are identical to the slot ones except for the slots vs round holes and the name. The 3D model is the same for both - OK

Things which need some changes:

- [ ] Courtyard: I did check the clearance left and right but I forgot to check it north and south (near the big mechanical pins). That needs adjustment (it's too big at the moment). It will be slightly different fot the slot version and the CircularHoles version.

- [ ] The description should start capitalised ("Potentiometer"). My comment before was just for the keywords. Sorry if I wasn't clear enough.

Thanks!

robkam:
Thanks for the review.

Courtyards fixed and capitalised Potentiometer in descriptions.

capture

Potentiometer_THT.pretty/Potentiometer_Alpha_RD901F-40-00D_Single_Vertical.kicad_mod
```diff
@@ -0,0 +1,41 @@
+(module "Potentiometer_Alpha_RD901F-40-00D_Single_Vertical" (layer F.Cu) (tedit 5C69D274)
```
diegoherranz: @robkam, when you've modified and then fixed the names, some quotation marks `" "` have been added. The `module` line doesn't use quotations (you can check around in the library). Can you fix that? Thanks!

robkam:
Is there any point having the 7 mm diameter F.Fab circle?

poeschlr:
It would allow users to see where the shaft is. Could be useful for housing design.

Potentiometer_THT.pretty/Potentiometer_Alpha_RD901F-40-00D_Single_Vertical.kicad_mod
diegoherranz: This has been fixed now. Thanks.

diegoherranz:
Cool, thanks for the changes again.
It seems that most of the content of the files had been reordered so it's been difficult to check the changes but I think we're good to go except for one small thing:
- [ ] On both the dual footprints, the location of the north mechanical pin is off by 0.1mm. I hadn't noticed before. Could you fix that one? I think that the courtyard doesn't need updating.

Thanks again!

robkam:
Thanks, I've tweaked n&s mech holes positions and the courtyard. I might have used the single footprints 11.4 mm dimension previously, now using 11.3 mm.

diegoherranz:
Cool, the mech holes and courtyard are looking OK now.
It seems that with your latest changes, you've moved the `REF**` on the Fab layer of `Potentiometer_Alpha_RD902F-40-00D_Dual_Vertical.kicad_mod`, 
unwittingly I guess. We're not very strict with the placement of that but that footprint one looks a bit weird now compared to the others.
pots

Can you fix that one?

Thanks!

robkam:
I hope that fixed it.

diegoherranz:
Hi, @robkam 

I think that you changed the placement of the slots on the `Potentiometer_Alpha_RD902F-40-00D_Dual_Vertical` again. I think that they were correct. Was that a mistake?

How are you making all these changes? Is it through the github website? It is hard to follow the changes.

Thanks!

robkam:
> I think that you changed the placement of the slots on the `Potentiometer_Alpha_RD902F-40-00D_Dual_Vertical` again. I think that they were correct. Was that a mistake?

Must be a mistake. Oops, sorry about that ...

> How are you making all these changes? Is it through the github website? It is hard to follow the changes.

On Windows with the footprint editor and then Git Bash. I don't know why it's mangling the files.

diegoherranz:
> Must be a mistake. Oops, sorry about that ...

No problem. We're nearly there :)

> On Windows with the footprint editor and then Git Bash. I don't know why it's mangling the files.

Don't know, then. If you're using git bash, you can check the commits before pushing them. And you can use amend to fix them if you see something wrong (see https://help.github.com/articles/changing-a-commit-message/), to reduce the number of commits (note: you shouldn't amend after having pushed, since that changes the history).

Thanks!

robkam:
Probably messed up because I moved 0,0 to the centre of the part, did the edits and then moved it back to pin 1.

robkam:
By the way in the [data sheet](http://www.taiwanalpha.com.tw/downloads?target=products&id=113) for the dual pot there are two pin 1s. I've changed the second pin 1 to 3 because they're electrically separate.

diegoherranz:
> By the way in the [data sheet](http://www.taiwanalpha.com.tw/downloads?target=products&id=113) for the dual pot there are two pin 1s. I've changed the second pin 1 to 4 because they're electrically separate.

Sounds like the most sensible option possible.

Thanks for the latest change. The holes are correct now, but the courtyard for `Potentiometer_Alpha_RD902F-40-00D_Dual_Vertical` is wrong now. Everything else is correct as far as I can see.

Thanks!

robkam:
Scrambled the file again, all I changed was the courtyard coordinates, in the fp editor.

diegoherranz:
> Scrambled the file again, all I changed was the courtyard coordinates, in the fp editor.

That is KiCad, then. Don't worry, I've checked the contents of the file to make sure nothing else has changed.

It all looks OK now. Many thanks for your contribution!

robkam:
Hooray. Many thanks for the review :-)

PR #1389, opened 2019-02-07
Add LED-APA102-2020

cp-aquila:
Add LED-APA102-2020

---

Datasheet:
http://www.led-color.com/upload/201604/APA102-2020%20SMD%20LED.pdf
Better(sic) quality picture of the land pattern:

I am not that happy with the Pin1-Marker.
But I can not fit it within the courtyard without violating F5.3. In the example-picture of the KLC, the CP_ELEC behaves similar, the "+" sign is partially outside the courtyard.
The Datasheet is pretty sparse, so I am not 100% sure on how to handle the center pads. I choose to epxlicitly connect them, that seems most plausible for me right now.

Symbol PR: https://github.com/KiCad/kicad-symbols/pull/1501


poeschlr:
A not to answer an implicitly made question:

Silk is completely ignored by the courtyard. The courtyard is responsible to ensure place for tooling required to solder the part. (Pick and place machines need that space)
But also tooling for rework (Space for the solder iron tip, the specialized heat gun tools, ....)
The later is the reason why bgas need such a large courtyard area.

IPC specifies courtyard simply against pads and nominal part body. The later is also what we use to define the fab outline meaning courtyard in kicad is relative to pads and fab outline.

cp-aquila:
Thanks for the clarification @poeschlr 

Upon closer inspection the courtyard in the first commit is wrong. It is based on the 6 'regular' pads.
It does not have the 0.25mm spacing related to the inner (exposed) pads (1 & 6).

What does the audience think, should I add 'ears' to the courtyard to have room for the inner pads, or just increase the overall square courtyard?

poeschlr:
Both options are valid. One will get fewer false positive DRC errors if you contour the courtyard. (A bit more work during fp design with a bit of payoff for every user of the footprint)

cp-aquila:
grafik

poeschlr:
Looks good, thanks

screenshot from 2019-02-13 22-44-36

cp-aquila:
So I used this footprint for some prototypes and it was fine.
However, in the first small production run the factory managed to rotate the parts the wrong way.
There was confusion about the orientation.
* This FP has correct Pin1 Marking in F.Fab and F.Silk
* The device itself has a white dot on pin3
* Page1 of the apa102-2020 Datasheet shows a PLCC-6 package
* Page2 of the Datasheet has the proper pinout, but is barely readable

My proposal is to add the Pin3 marking to the footprint.
grafik

myfreescalewebpage:
@poeschlr what do you think of the last comment of @cp-aquila ?

evanshultz:
The same, or at least similar, thing has come up before. See https://github.com/KiCad/kicad-footprints/pull/418 and the linked case. Perhaps the same solution would apply here?

Two other notes:
1. The fab ref des should be shrunk to fit inside the fab outline.
1. It is most common to put the ref des on top and fab layer value below.

cp-aquila:
What does IPC say about that? Does the marking always go to Pin1?
Anyway, that is something that we should clarify in the KLC (F5.1, F5.2). Right now KLC does only specify Pin1 or polarization, but not the device marking.

Related question: is there a public source for the IPC documents?

cp-aquila:
grafik

evanshultz:
IPC 7351 contains info on packages broken down by package type. Note that IPC uses the term "polarity" frequently to indicate placing the part in the proper orientation which may be a higher-level term to capture packages which don't have a pin 1 (even a simple disagreement like a BGA where is marked as pin A1). Here are a few snippets so you can get an idea of how IPC says packages should be marked (the latter two might be the closest to this package):
SOIC
image
QFP
image
QFN
image
SON
image

So having no marking is possible but I don't see them indicate marking pin 3 is OK. That being said, IPC doesn't define packages (that would be JEDEC, JEITA, etc.) so the package designers may follow another standard. And even if IPC does recommend or require a marking, it doesn't mean we'll actually see that in practice. And maybe LEDs are specifically non-conformant for some reason?

With regards to footprint markings, you have this:
image

I don't see much from IPC that helps here, but looking at a land pattern generator or more specifically at some LED footprint info may be illuminating? (Pun intended.)

IPC docs are not free, but take a look online and surely you can find them.

To avoid being too narrow, adding an allowance in KLC for alternative markings and making the above an example sounds quite reasonable to me.

A new PR will need to be created since this was already merged. Referencing this one is perfect.

cpresser:
In the meantime I found IEC-61188 which is similar to IPC-7531. Only in german, sorry about that.
2711807-1 svg

Their wording is weird, basically its saying
> the pin1 mark of the footprint should if applicable be identical to the polarity-indicator of the part

So still no help about what to do if the part has a mark that differs from pin1.

My gut is telling me to keep the pin1 in silk and fab as it was, but add another mark in both layers.

PR #1390, opened 2019-02-07
Add shield pads

cp-aquila:
The current FP has an error. Two of the pins are actually connected to the metal enclosure.
This PR makes them 'shield' pins.

---

https://katalog.we-online.de/pbs/datasheet/7499010121A.pdf
grafik

A symbol PR will follow.


cp-aquila:
It might also be an option to name the pins 'MP' according to KLC.
S3.7: "For connectors it must connect to the shield of the cable." that depends on the cable; For most RJ45 Cables, it will not connect.
But on the other hand, those two pins to connect to the metal enclosure.

evanshultz:
Good catch. Thanks!

PR #1391, opened 2019-02-07
add C&K spxt dpdt mini and sub-mini switches

robkam:
was "add a few C&K toggle switches" #1356, now has pin 1 markers on fab & silk

add vertical THT C&K toggle switches, SPST, SPDT and DPDT, miniature & sub-miniature

Mini https://www.ckswitches.com/media/1394/7000toggle.pdf

Sub-mini https://www.ckswitches.com/media/1391/ttoggle.pdf

annotation 2019-02-07 095205


poeschlr:
Sorry for the long delay.

---

**Naming**

Regarding naming convention take a look at https://github.com/KiCad/kicad-symbols/issues/580#issuecomment-450632899 (Nearly fleshed out now so we can start thinking of adding it to the website soon.)

I think yours should therefore be named `SW_Lever_2P2T_Vertical_CK_72xx` (meaning DPDT -> 2P2T [SPDT -> 1P2T] and move the manufacturer name plus mpn towards the back of the name. Plus add Vertical to show the orientation in clear text.)

Add SPDT or DPDT to the keyword field such that searching for this alternative nomenclature works.

We might also need to include much more info into the part number. (Right now you only have the series prefix. But i think you will also need to include the lead specification.)
Meaning 72xx-xxx-xx-Z (I assume Z type because this is the only one that would require such long oval holes. The only alternative without extra mounting holes would be C type.)


----

**Hole Sizes**

The oval holes seem a bit short to me. (The straight part before the circular part starts has the same lengths as the lead. I would add 0.2mm to it like you did with the width.) Especially as the lugs are quite short. Might need to get your solder iron in there and solder from the "inside" if you use a standard 1.6mm pcb and are not willing to cut away the isolator between the rows of the switch.

The circular hole is ok as is. (diagonal is 2.17 you used 2.37, and 1.6 diagonal vs 1.8 hole for the T series which is also ok)

---

**Dimensions**

generally ok except oval hole size see above.

Screenshot from 2019-06-01 12-14-02

robkam:
The delays are extremely discouraging.

chschlue:
I guess nothing will happen until you make those changes or at least comment on them.

robkam:
Fixed the names. Z type was in error, it should have been C type termination for PCB mounting. also fixed. C has circular pins so oval holes aren't required anyhow,

Capture2

robkam:
Pin 1 markers on fab & silk to follow.

robkam:
Please review.

C&K SPxT and DPDT mini and sub-mini toggle switches, vertical PCB through hole termination.

Mini https://www.ckswitches.com/media/1394/7000toggle.pdf

Sub-mini https://www.ckswitches.com/media/1391/ttoggle.pdf

Capture

poeschlr:
I am a bit confused now. Why are there suddenly no oval hole footprints in here and why did the circular ones loose the suffix? Did you change what this pull request represents?

PR #1393, opened 2019-02-07
Add TRAPC5M 5-Pin Mini-XLR Connector

cp-aquila:
Add TRAPC5M 5-Pin Mini-XLR Connector

---

grafik
* http://www.switchcraft.com/Drawings/TRAPC_X-TRASM_X_SERIES_CD.PDF
* No Symbol PR, there is already an XLR5 symbol
* The coordinates look ugly, thats because I drew the FP completely in Imperial. The datasheet does not provide values in metric.
* I tried to follow the KLC for "Connectors with Standardised Shape", but thats pretty hard and lots of options remain. I am open for suggestions for a better name
* I put it into Connectors_Audio since XLR is mostly used for audio stuff.


poeschlr:
Rotate the part such that pad 1 is nearest to the top left corner.
Add a pad 1 marker on the fab layer.
We do not typically include things like thread sizes in the footprints. (This information is not needed for pcb design.)
Move the value field outside the outline. (The final result should be that the value field is below and the silk reference above the part in the footprint editor.)

The drill sizes are not quite right.
The courtyard offset is not as it should be (0.5mm rounded to nearest 0.01mm grid -> 0.495mm to 0.505mm)

screenshot from 2019-02-10 07-56-11

cp-aquila:
Thanks for the fast review!
grafik

I would like to have some kind of graphical indication that the 'front' part of the device has an outside thread. Modelling the thread with graphical lines is not an option, it will look ugly. The second option that comes to mind is to follow the 'drawing standart' for thread lines. 
Like so:

But I suspect most people won't understand that, and mixing different line styles on the Fab-Layer would also violate KLC.
So In the end, I choose to keep the text, but moved it to another layer (Dwgs.User).

poeschlr:
Indicating the thread really is overkill for a pcb design. Especially on the fab layer. (I doubt your mechanical engineer will be happy if you simply print the fab layer for them to indicate what parts have threads as it includes way too much information they are not interested in. For example they will not really care about the few hundred small components like resistors, caps, esd diodes, ...)

In reality the best option to communicate with the mechanical engineer is via a well made 3d model.

poeschlr:
The non plated through holes must have the same pad size as hole size.

The courtyard clearances are not quite right.

screenshot from 2019-02-12 18-02-03

cp-aquila:
The errors spotted in your review should be fixed in the latest commit.
I am still looking for a nice workflow to check the courtyard while editing it in pcbnew. So far everything is just a stupid workaround. Right now I am dragging a circle with radius r=0.5 along the fab-lines, drills and pads.

I did remove the thread information. I guess my view is biased, usually I am both the electrical and mechanical engineer. IMHO every bit of information helps.
Instead I choose to add the text "PCB Edge", similar to some of the USB-Micro-B plugs already found in the lib.

grafik

poeschlr:
Thanks looks good now.

PR #1395, opened 2019-02-07
Add Texas RGE0024C footprint

chemicstry:
Add Texas RGE0024C footprint

Datasheet: http://www.ti.com/lit/ds/symlink/pca9548a.pdf#page=31

Same as existing RGE0024H footprint, but smaller EP. Generated by a script: https://github.com/pointhi/kicad-footprint-generator/pull/290

However there are a few problems with the currently existing RGE0024H footprint as well as this one (generated by the same script):
- Manufacturer specs have much shorter pads, that barely extend past the package size (making it way harder to hand solder). Most of the kicad QFN packages have longer pads. Should I fix both of the packages to match manufacturer specs?
- Script generates 9 thermal vias (3x3) instead of 5 (cross) in the specs. Should I remove the extra?

Symbol PR: https://github.com/KiCad/kicad-symbols/pull/1495

image

image


poeschlr:
I requested some changes over at the script repo

chemicstry:
@evanshultz
I'm not sure how to fix the F7.4 violation on thermal vias. Is it something with the way I used the script:
```
python ipc_noLead_generator.py size_definitions/qfn/qfn_texas.yaml
```

evanshultz:
Our check script isn't smart enough. Those errors are fine.

evanshultz:
BTW, the questions you raised above are good. Thanks for taking a look at thinking about this!

Our scripts make generic footprints using the IPC standard which is prevalent in the industry. We accept them as they are because it is a universal foundation. Footprints built exactly as a manufacturer specifies and are unique get a special prefix to note that; for this footprint the prefix would be `Texas_`. But the IPC footprints are fine here so we can accept the deviations from the datasheet. (There are also lots of older footprints which haven't been update to scripts, or we haven't made scripts yet, or to which scripts don't apply, so you will find a lot of variation.)

PR #1403, opened 2019-02-09
Add Stocko_MKS_16XX Connectors

balcer:
This is reopened https://github.com/KiCad/kicad-footprints/pull/912. Fixed all reviewer suggestions. 

Datasheet: https://www.stocko-contact.com/downloads/steckverbindersystem-raster-2,5-mm.pdf
stocko

---


poeschlr:
For reference this is the script pull request: https://github.com/pointhi/kicad-footprint-generator/pull/169

Also this part is on page 15 of the linked document. Please modify the datasheet link to include this by adding the suffix `#page=15`

I would suggest to use oval pads that are thinner in x direction than the circle is now. You could even decrease the width by 0.5mm and still have 0.25mm copper left. (And result in 1mm clearance which would really be more than enough.)
I typically have the minimum annular and pad to pad clearance as script parameters. [Check my JST XH connector generators for an example.](https://github.com/pointhi/kicad-footprint-generator/blob/30841f4e2b95f3df421f754d05d2287c490f2695/scripts/Connector/Connector_JST/conn_jst_xh_tht_top.py#L118-L120)


The courtyard clearance is still not right. (Remember it is relative to the body size -> The fab layer represents the body size so the courtyard is relative to the Fab layer. Silk is ignored by it.)

There are a few dimensions not given in the datahseet. (Not a big deal) but do you have a source for them?) I marked them orange in the drawing.

screenshot from 2019-02-10 22-36-59

balcer:
Hi, @poeschlr I added information about page (it is  13, not 15).  Pins now are oval. Courtyard clearance corrected. Sorry, I forgot to change this in my script. I measured orange dimensions on the real component now they are correct. I will also check this after my board will be manufactured and make the correction if needed. Next time I will try to use more helpers from your script. This was my first attempt so I started with example script from [generator](https://github.com/pointhi/kicad-footprint-generator). More exaples would make it easer. No I move to 3D scripting.

poeschlr:
The pin 1 would also need the same y dimension as the oval pads (elongated pad instead of square)

If you use the array node instead of generating all pads manually (in a for loop) yourself then this should happen automatically (The array replaces pad 1 with rectangle or maybe even rounded rectangle depending on settings.): https://github.com/pointhi/kicad-footprint-generator/blob/30841f4e2b95f3df421f754d05d2287c490f2695/scripts/Connector/Connector_JST/conn_jst_xh_tht_top.py#L128-L138
(The kicad4 compatibility stuff is not absolutely necessary. Meaning you could simply always supply it with rounded rectangle.)

---

Regarding examples: Well i would have assumed people copy the generator of a similar accepted part in preference to starting from simplified examples. (We could maybe add a note in the example directory that the script directory holds better examples if one plans to implement something that is in any way similar to existing generators.)

---

The dimension fixes look good in general:
screenshot from 2019-02-12 17-33-00

balcer:
@poeschlr Hope, now it is ready for merge. :)

poeschlr:
Looks good, thanks

balcer:
Thanks, @poeschlr for help, see you soon with next pull requests.

PR #1404, opened 2019-02-11
Add M.2 TE 2199230 series footprints

samt:
Add M.2 TE 2199230 series footprints

---

## Datasheets and Screenshots:

**[2199230-8 M.2 A-KEY](https://www.te.com/commerce/DocumentDelivery/DDEController?Action=showdoc&DocId=Customer+Drawing%7F2199230%7FB2%7Fpdf%7FEnglish%7FENG_CD_2199230_B2.pdf%7F2199230-8)**
screen shot 2019-02-11 at 11 10 15

**[2199230-3 M.2 B-KEY](https://www.te.com/commerce/DocumentDelivery/DDEController?Action=showdoc&DocId=Customer+Drawing%7F2199230%7FB2%7Fpdf%7FEnglish%7FENG_CD_2199230_B2.pdf%7F2199230-3)**
screen shot 2019-02-11 at 11 10 21

**[2199230-4 M.2 E-KEY](https://www.te.com/commerce/DocumentDelivery/DDEController?Action=showdoc&DocId=Customer+Drawing%7F2199230%7FB2%7Fpdf%7FEnglish%7FENG_CD_2199230_B2.pdf%7F2199230-4)**
screen shot 2019-02-11 at 11 10 26

**[1-2199230-5 M.2 M-KEY](https://www.te.com/commerce/DocumentDelivery/DDEController?Action=showdoc&DocId=Customer+Drawing%7F2199230%7FB2%7Fpdf%7FEnglish%7FENG_CD_2199230_B2.pdf%7F1-2199230-5)**
screen shot 2019-02-11 at 11 10 31

## Other Notes

* I've included placeholder 3d models to be added at a later date
* Cross-checked keying with wikipedia article on [M.2](https://en.m.wikipedia.org/wiki/M.2#Form_factors_and_keying)
* Printed (on paper) and roughly verified everything is correctly positioned with the parts I have on my desk

---


samt:
I made some minor tweaks since the screenshots to make the grid align with 0.01mm, but it should be good to go.

I have no models yet and symbols are also to come.

myfreescalewebpage:
Hello @samt thanks for contributing !
For this PR we will have to consider also https://github.com/KiCad/kicad-footprints/pull/879.
We are also currently reviewing https://github.com/KiCad/kicad-footprints/pull/1124 and it's good if we can do something consistent.
I will do a review according to choices on the two above PR too.
Cheers,
Joel

samt:
Hi, @myfreescalewebpage - I wanted to check on the status of this and if there is any action needed on my part. Thanks!

myfreescalewebpage:
Hi @samt 
We are quite busy at this moment because of the 5.1 release, and we try to integrate important fixes.
I have not made a complete review of the footprint right now, they look good, but I'm waiting some strategic decisions in the PR attached in my previous comment to indicate you if some changes are required.
I will keep you in touch asap anyway, do not worry, but I think this will take a while.
Thanks for your understanding,
Joel

evanshultz:
I would like to name these by the keying and connector height, so the 3D model can match, since they are a standard. This would be something like `M.2_<key>_<height>` where `<key>` is `A`, `B`, `E`, or `M` and `<height>` is `H2.25mm`, `H3.2mm`, `4.2mm`, or `<MidPlane>`.

And then put them into a different repo like what was discussed starting at https://github.com/KiCad/kicad-footprints/pull/1124#issuecomment-461936355. Let's try not to spread the conceptual part of the discussion around, if we can.

A few things I can see above:
1. It looks like silk is on top of the fab lines, which should be at the component edges, but silk should be offset. Can you move it, say, 0.11mm outside of the fab lines?
1. Fab lines should be a rectangle at the component extents.
1. There can be silk lines across the key (gap in the pads).

Is all that OK?

myfreescalewebpage:
@evanshultz for consistency with other footprints I propose `Bus_M.2_<key>_<height>` format

samt:
Sorry this is months late, I believe I was traveling and missed the notifications here.

@evanshultz I'll make the requested changes, but was confused here:

> And then put them into a different repo like what was discussed starting at #1124 (comment). Let's try not to spread the conceptual part of the discussion around, if we can.

Do you mean to put this in a separate tree within this repo? I know I have a fab drawing to indicate where the card edge goes to, but thats not a cut line at all.

In addition, should I create new footprints or modify these to indicate where the mounting screws go? IIRC, there are a few standard lengths but I also put the card edge so a user can measure it for his or herself.

poeschlr:
Does this contain the same parts as https://github.com/KiCad/kicad-footprints/pull/1929

samt:
It appears so. I have no preference, personally. I just use the M.2 footprint and this is the one I've personally been using.

samt:
It appears so. I have no preference, personally. I just use the M.2 footprint and this is the one I've been using.

poeschlr:
I disagree with the assessment by evan that such connectors would not fit into the footprint repo. What does not fit is something that defines a full outline but the partial outline of a connector alone is fully ok. And most importantly this contribution does not even define any edge cut as it is for the connector that receives such board edge connector.

---

Review:

The body outline is missing.
Move the remarks to the cmts.user (text) and dwgs.user (graphics) layers.
The pads for the mounting pads would need the pad number MP.
The fab reference should be inside the body outline (preferably at the center)

---

From a KLC standpoint i would say the duplication of this might be a better fit. (So it might be less work for @ki5libs to get their contribution ready for merging than it would be for you.)

samt:
I noticed a small issue when using the 2-row, 75 pin connected that was merged back in May (https://github.com/KiCad/kicad-library-utils/pull/272). The pin counts were not matching up because M.2 always has a key, which is effectively a chunk of "missing" pins. For example, in an M.2 E-Key setup, pins 24-31 are missing, but the numbering continues at pin 32.

What should happen here, exactly? Do I need to make each one with no-op pads? Do we need different symbols per key instead?

samt:
## Updated footprints

### A Key
M.2 Connector footprint A-Key

### B Key
M.2 Connector footprint B-Key


### E Key
M.2 Connector footprint E-Key


### M Key
M.2 Connector footprint M-Key

---
## Changes

* [x] Added body outline
* [x] Remarks on correct layers
* [x] Mounting pads are named 'MP'
* [x] Fab ref is inside fab lines
* [x] It looks like silk is on top of the fab lines, which should be at the component edges but silk should be offset. Can you move it, say, 0.11mm outside of the fab lines?
* [x] Fab lines should be a rectangle at the component extents.
* [x] There can be silk lines across the key (gap in the pads).

---
For what it's worth, I made a script to do these changes. The template sits in its own `.pretty` folder which I load into the footprint editor for editing and I have a script that modifies/populates the resulting template and copies it to a new location. I have it doing 4 variations, one for each key.

https://gist.github.com/samt/db077785ddf99e2797c9d65ee49022d4

PR #1406, opened 2019-02-11
Footprint for display NHD-C0220Biz

aracosta1:
New footprint for the corresponding display
The footprint center is the same as the datasheet which is slightly off-center 

image


myfreescalewebpage:
@aracosta1 I have not performed a full review, but I see on the screenshot that the silkscreen is not correct. When the device is soldered, the silkscreen should remain totally visible, so for this part it should follow all the corners and screw holes to achieve this.

aracosta1:
Thanks for the advice, I'll make the correction and resubmit the footprint

poeschlr:
The description and keyword fields of the footprint need to be filled (look at similar parts for inspiration)

Please add a link to the datasheet you used to design this somewhere in the pull request description and in the footprints description field. (We need it to double check your footprints dimensions. No footprint is allowed to enter the lib without being checked against its datasheet.)

Rotate the footprint such that pin 1 is nearest to the top left corner (180 degree rotation should do the trick)

The courtyard layer is there to define where no other component can be placed. You used it wrong unless this part has a massive hole in the middle where other components can be placed.

You do not have anything on the fab layer (It must contain the nominal part outline. It is allowed to simplify it.) It must have the value field visible and a copy of the reference. And a pin 1 mark.

Silk is not allowed to overlap pads.
There must be a pin 1 marker on silk. (visible after assembly)

aracosta1:
I made some of the changes requested.
Footprint rotation is not correct from my point of view since this is a display and it's oriented so the text is with the correct orientation.
The center of the component is the display area center as indicated in the datasheets.
The components doc are available at : [newhavendisplay](http://www.newhavendisplay.com/specs/NHD-C0220BiZ-FSW-FBW-3V3M.pdf)

aracosta1:
@poeschlr , is there something else I should do on the footprint for conforming to the guidelines ?

poeschlr:
Sorry for the delay i have been busy at work (and uni).

---

If the simplified outline on fab is correct then the silk outline is definitely wrong. (Right now the fab layer looks like the component is larger then both the silk outline and the courtyard outline. My guess is you did not really care about the fab layer.)
My suggestion is to either have all layers detailed (silk, fab and courtyard) or none of them. 

Of these 3 layers the silk layer is possibly the least important (only needs to look good.) The fab layer is arguably the most important layer and really needs to fit the nominal dimension of the component (to all significant digits given in the datasheet.) Courtyard is a close second as it defines where no other compoent can be placed. (It makes sense to have a contoured courtyard for this part but that would require the fab layer to be contoured as well. This is because we really need the fab layer here to check if you have the correct courtyard offset all around.)

I am also unable to properly check the relative pad positions without at least a fab outline for the main body as they are referenced relative to that in the datasheet.

---

The text note is best moved to the comments layer (user cmts)
There still is no pin 1 mark on silk and you did not rotate the footprint as requested.

For reference a current screenshot. For future changes please add updated screenshot.

Screenshot from 2019-03-21 06-42-51

---

Regarding mounting holes (the 4 non plated through holes) Which screw dimension is intended to be used here? If it is M2 then there is no reason to go with a 2.7mm hole as the tolerances are already mostly taken into account by the 2.5mm holes in the component itself.

---

The main pad dimensions look suspiciosly like the kicad default values. You might want to double check if they fit your part as suggested in http://kicad-pcb.org/libraries/klc/F7.6/

For more details about calculating this for non round leads see: https://forum.kicad.info/t/tutorial-how-to-make-a-footprint-from-scratch/11092/4 (TlDr: use Pythagoras to get the diagonal dimension of the lead. This dimension is then used as the basis for the calculation of the required hole size as described in the KLC.)

It might be best to use oval drill shape for the LED pads. (or add _CircularHoles as a suffix to communicate the fact that this uses circular holes where slots would make more sense.)

aracosta1:
First thank you for all the suggestion and the time taken to explain them clearly.

Regarding the fab layer I thougth it was just a placeholder reference for placement print. I will update it to a detailed contour like the coutyard and silk. I will also review the fab and courtyard dimensions.

> 
> The text note is best moved to the comments layer (user cmts)
> There still is no pin 1 mark on silk and you did not rotate the footprint as requested.

The pin 1 mark is probably to small and you missed it, I will make it more visible. 
Regarding the rotation I disagree with this because it will end with a display with inverted characters if placed with pin 1 up. I think this could generate confusion and lead to boards with display mounted upside down.

> Regarding mounting holes (the 4 non plated through holes) Which screw dimension is intended to be used here? If it is M2 then there is no reason to go with a 2.7mm hole as the tolerances are already mostly taken into account by the 2.5mm holes in the component itself.

Datasheet states the holes are 2.5 mm, I can change the hole to 2.5 from 2.7.

> It might be best to use oval drill shape for the LED pads. (or add _CircularHoles as a suffix to communicate the fact that this uses circular holes where slots would make more sense.)

I preferred a circular hole because it is more easier to fab versus a slotted hole, I'll add the suffix and maybe later make also the oval version.
If I change the footprint name should I close this pull request and open a new one ?

myfreescalewebpage:
@aracosta1 I have not read all the thread, just seen your last question: keep working on this pull request even if you change the name, this will be automatically updated by git when you push to your branch :)

aracosta1:
I made several changes as suggested.
the final footprint this:
Immagine

As stated before I think is not a good idea to rotate this part to have the pin 1 on the top as it will reverse the display and can induce error in users.

For the dimensions check, note the footprint center correspond to the origin indicated on the datasheet drawing.

poeschlr:
There are some wrong measurements.

I am not convinced that the drill sizes of the oval pads is optimal (too short but too wide). See http://kicad-pcb.org/libraries/klc/F7.7/

The y alignment of these pads is also not really dimensioned in the datasheet. Makes me a bit worried.

Some dimensions of your footprint do not agree with the datasheet.

d svg

Move the display area text to the user comments layer and the outline of the display to either fab or maybe better user drawings layer.

myfreescalewebpage:
@aracosta1 why closing this ? We can just keep it open until you or somebody else want to continue working on this device.

poeschlr:
It might be a reaction to me wrongly removing the ready for review tag.

aracosta1:
No, as I wrote on the other pull request, I just wanted to move the changes from the main brach to a separate branch. It was my iitial fault to create a pull request from the master branch and I needed to change it.

poeschlr:
Clarifying this when closing might be best for everyone involved.

PR #1407, opened 2019-02-12
Added footprints for Intel - Altera EN53xx series regulators

luca-della-vedova:
Hi all! First of all this is my first footprint contribution so I'm more than grateful for any feedback to make future contributions smoother!

Added footprint for Intel / Altera EN53xx (i.e. EN5322QI) regulators. Datasheet here https://www.intel.sg/content/dam/www/programmable/us/en/pdfs/literature/ds/en5322qi_03454.pdf

The footprint is a QFN24 in a 4x6mm package with a quite particular layout of thermal pads below, I created two variants, one with thermal pads (following the guidelines in datasheet page 14) and one without.
image

image

This is the current 3d view:

image

Now there is a few issues that I get in the CI script that I'd like to check with you:

* Violating F6.3. The two pads are supposed to be connected by a layer of copper that, following the datasheet, should not be exposed and as such I removed the solder mask / paste, it looks fine in the 3d view but I wonder if that is the right way to do it?

* I had to superimpose an oval and a rectangular pad to create the particular shape and now the pad numbers are hard to read, not sure how to fix that.

* Thermal vias variant is violating F7.5 with the annular ring being too small. I could increase the size, however that would make it impossible to get the layout specified in the datasheet, page 14. I chose the drill size recommended there (0.33mm) and making the annular ring 0.66 would not allow the suggested via layout.

* I have no 3d model for this (and I'm not sure how to generate one?), will start to work on the symbol and create a pull request for that in a few days, however symbols are much easier than footprints and I have contributed a few in the past so I guess that merge should be quite straightforward.

Look forward to your feedback and suggestions!

Cheers,
/Luca

---


poeschlr:
We do not typically use the one sided oval pads. The industry standard IPC suggests to use rounded rectangular pads with 25% round rect radius instead. (They acknowledge that many manufacturer datasheets will likely suggest something different for a long time. Mainly because updating datasheets has a very low priority.)

luca-della-vedova:
Hi Rene,

Thanks for your feedback! I edited the footprints to remove duplicated pads and changed all the pad shapes to rounded 25% rectangles, example below.

image

Also, I created the matching symbol PR for this footprint, available here https://github.com/KiCad/kicad-symbols/pull/1520

/Luca

poeschlr:
I just noticed you put this into the dfn/qfn lib. This is not quite the right place for it. I would suggest the Converter_DCDC library

luca-della-vedova:
Done, new name for components are
Converter_DCDC_Intel_EN_53xx_QFN

and

Converter_DCDC_Intel_EN_53xx_QFN_ThermalVias

/Luca

evanshultz:
I haven't quite got time to do any dimensioning just yet, but here are some things I can see from the screenshot:
- Move the silk closer to the fab lines. Offsetting the silk by 0.11mm will work.
- Move the silk lines to within 0.12mm of the pads.
- Change the pin 1 silk indicator to keep the horizontal line near pin 24 and then remove the vertical section near pin 1.
- The big copper pad should be a single pad with a custom shape, with mask and paste layers disabled. Set the zone connection of this shape to "Solid".
- Then add two pads with no pin number and only the mask layer turned on where indicated on top of the custom copper pad.
- In this case the via annular ring doesn't matter much because they're in a bigger copper shape, so you can enlarge it to 0.66mm if the via annular ring doesn't stick out past the copper shape. It should be fine even if the annular rings overlap.
- Page 14 says to use as many thermal vias as possible, but I guess you tried to copy that layout image and guessed the via locations?
- Is there any recommendation about paste placement for this part? I don't see it.
- Is this package used by other parts too? I guess that's why you called it "53xx"? If so, you can simplify the footprint name to `Converter_DCDC_Intel_EN53xx_`.
- I think of this as an IC package, and not a DCDC converter. Yeah, it has integrated magnetics and mostly of all the other parts. Maybe that's just me. And Rene is the boss.

luca-della-vedova:
Hi Evan!

I tried incorporating your suggestion in one footprint, if it looks OK I'll copy the changes over the second one.

image

> * Move the silk closer to the fab lines. Offsetting the silk by 0.11mm will work.

Done
> * Move the silk lines to within 0.12mm of the pads.

Didn't quite understand this part? I guess they are closer to the pads now that they are closer to the fab lines.
> * Change the pin 1 silk indicator to keep the horizontal line near pin 24 and then remove the vertical section near pin 1.

Also didn't quite understand this part? I rotated the marker from vertical to horizontal but not sure this is what you meant.
> * The big copper pad should be a single pad with a custom shape, with mask and paste layers disabled. Set the zone connection of this shape to "Solid".

Done, I am a bit worried about the fact that there is no solder paste on the pad and it looks a bit wrong in 3d view now.
> * Then add two pads with no pin number and only the mask layer turned on where indicated on top of the custom copper pad.

Also done, it's actually one pad only.
However, integrating these changes made the 3d layout look a bit wrong, the copper on the pad doesn't seem exposed?
image

> * In this case the via annular ring doesn't matter much because they're in a bigger copper shape, so you can enlarge it to 0.66mm if the via annular ring doesn't stick out past the copper shape. It should be fine even if the annular rings overlap.
> * Page 14 says to use as many thermal vias as possible, but I guess you tried to copy that layout image and guessed the via locations?

Will do these two once the design without thermal pads seems reasonable, so I will bring its changes over to the second footprint.
> * Is there any recommendation about paste placement for this part? I don't see it.

Didn't find any, so I just went with the defaults.
> * Is this package used by other parts too? I guess that's why you called it "53xx"? If so, you can simplify the footprint name to `Converter_DCDC_Intel_EN53xx_`.

Done
> * I think of this as an IC package, and not a DCDC converter. Yeah, it has integrated magnetics and mostly of all the other parts. Maybe that's just me. And Rene is the boss.

Yea but I guess this is such a particular footprint that it will never be shared among any other IC so might as well put it in the DCDC library.

/Luca

poeschlr:
@evanshultz i am not quite sure what the benefit would be to move the silk closer to pads for this very large part. There is enough space to keep a larger distance and allow this footprint to be used by users of cheaper fabs (or cheaper options of the same fab)

I know IPC gives the minimum of the silk line width. But that is a minimum suggestion. We can use larger clearances where it makes sense.

poeschlr:
@luca-della-vedova your pin 1 mark looks strange. Look at other similar parts (qfn footprints) and see how it is done there. (There is no need to move the line away from pin 24. The pin 1 mark is already defined by the missing vertical line.)

luca-della-vedova:
My bad, should have looked at the QFN packages silk screen more carefully before creating the part.

Latest version with the longer silk screen and proper pin 1 marking:

image

evanshultz:
@poeschlr 
I asked to make the silk closer because it was far from pin 24 and pin 1. Now it looks better and is only missing near pin 1.

@luca-della-vedova 
Looking better now. Thanks!

It's hard to explain the layers thing with words so take a look at this:
[Converter_DCDC_Intel_EN_53xx.kicad_mod.txt](https://github.com/KiCad/kicad-footprints/files/2866188/Converter_DCDC_Intel_EN_53xx.kicad_mod.txt)

The first pad is only copper and I reconfigured it a bit so the pin number is bigger. I then added mask pads where the copper should be exposed. Finally I added the paste pads as shown in the newer datasheet you mentioned. And, since we want to capture that info with the footprint, I changed the link.

Let me know if this is making sense. And also please check everything because it visually looks OK but I didn't dimension the entire footprint to verify.

I would like to put a via keepout (maybe even a copper keepout) in the gray areas shown on the datasheet, but KiCad doesn't support this feature and those areas aren't dimensioned.

evanshultz:
Also note that you need to reference a 3D model even if one isn't created already. If the 3D model is added later, the footprint will link to it automatically since the reference is there in the footprint.

luca-della-vedova:
> @poeschlr
> I asked to make the silk closer because it was far from pin 24 and pin 1. Now it looks better and is only missing near pin 1.
> 

So is my current part OK? I checked a QFN footprint (i.e. QFN 16) and it seems it looks the same? Where there is no silkscreen near pin 1 in order to mark it.
image

> @luca-della-vedova
> Looking better now. Thanks!
> 
> It's hard to explain the layers thing with words so take a look at this:
> [Converter_DCDC_Intel_EN_53xx.kicad_mod.txt](https://github.com/KiCad/kicad-footprints/files/2866188/Converter_DCDC_Intel_EN_53xx.kicad_mod.txt)
> 
> The first pad is only copper and I reconfigured it a bit so the pin number is bigger. I then added mask pads where the copper should be exposed. Finally I added the paste pads as shown in the newer datasheet you mentioned. And, since we want to capture that info with the footprint, I changed the link.
> 
> Let me know if this is making sense. And also please check everything because it visually looks OK but I didn't dimension the entire footprint to verify.
> 

I see! So the idea is that instead of  defining  the whole pad as a custom polygon we can create one rectangular pad for the one area then add a custom polygon for the next one, sounds good didn't think about that!
I checked the solder paste you added and there was just a minor issue of misalignment, everything else looks good! I am doing the thermal viases version right now.

I have a few queries as I finalize the footprint:

I'm actually wondering whether this approach of keeping two separate footprints, one with thermal vias and one without, makes sense, since it seems the typical application always recommends putting thermal viases, any input on that?

The naming of the footprint itself is a bit tricky. You can see a list of Intel's ICs in https://www.intel.sg/content/www/xa/en/power/programmable/devices.html and the devices that share the same footprint are all the QFN24-4x6 mm (that is also the reason why previously I went with QFN footprint instead of IC name, so we could match by package type).
However, the ICs are:
EN5322QI (with its own symbol, slightly different pinout) and
EN5319QI, EN5329QI and EN5339QI, pin compatible with each other different current rating.

There is only another device with 24 pins, in a 4x4mm package, called ER6230QI, but there is plenty of other EN5xxxQI with a different pin number and package.

That being said, how do you suggest calling the package? Technically a name EN5xxxQI would capture all the ICs, but would also capture a lot of "false positives" and that could be tricky if new parts were to be added in the future.

luca-della-vedova:
More details on the thermal vias version, below is the final one:

image

I followed the directives of the datasheet to place the thermal vias.
Since the directive of the datasheet is to have a via hole of 0.33mm and a pad size of 0.6 mm, we are failing annular ring size (that is 0.135 instead of 0.15). Now the choice is whether to follow the datasheet and keep the failing check or adapt the vias position by a bit and increase the annular ring width (at the current state increasing it would cause the ring to fall outside the copper region), or even decrease the hole size to 0.3mm and keep everything unchanged.

Finally, the last issue about the name also carries to the 3d model assignment, since if we just assign a wildcard name it will be impossible in the future to add other ICs that follow the wildcard, and if we assign a specific name we would be missing some ICs.

After these issues I believe this should be almost done.

Thanks again!
/Luca

poeschlr:
The screenshot looks good now. We still need to check the dimensions of your footprint against the datasheet and do some sanity checks within kicad. I hope i get around to that sometime this weekend.

evanshultz:
Page 24 of https://www.intel.com/content/dam/www/programmable/us/en/pdfs/literature/ds/en5319qi_08325.pdf shows 0.2mm vias with 0.6mm pads. Since we often use 0.3mm holes in the library I would accept 0.3mm vias with 0.6mm annular rings, which will address that CI failure. 0.2mm/0.5mm would also work.

You also need to do https://github.com/KiCad/kicad-footprints/pull/1407#issuecomment-463722110. That should give you a green check by CI.

Mask pads don't need the zone connection set to solid. You can change them to "None".

Regarding footprint name, that's a good webpage you sent because it makes it clear there is an issue. We can still keep whatyou have but add in the specific package info. Perhaps `Converter_DCDC_Intel_EN53xx_QFN-24_4x6mm_P0.65mm`? The "EN53xx" could potentially be removed.

luca-della-vedova:
Hi,

I have done all the changes you mentioned, the name that I implemented in the end is Converter_DCDC_Intel_QFN-24_4x6mm_P0.65mm , which I also used for the 3d model.
I went for 0.3mm holes in the end, the datasheet was a bit ambiguous mentioning 0.33mm in one page and 0.2mm in another one so I went for something inbetween.

/Luca

luca-della-vedova:
Hi all,

Any update on this, did anyone have a chance to look at the footprint?

Cheers!
/Luca

myfreescalewebpage:
@evanshultz ping :)

luca-della-vedova:
If it can make any difference we fabricated and assembled our boards with this footprint and it works so there shouldn't be any "major" mistakes.

/Luca

evanshultz:
1. Can you list PNs, wildcarded or not, that can use this footprint in the description? That will help users to search for this footprint in the future, I believe.
1. Pads 5, 12, 17, and 18 should be at 2.275mm in the Y direction.
1. I'd appreciate if you could unrotate the pads.
1. The paste pads don't need their zone connection set to Solid.
1. The pads aren't the shape specified by Intel. We can let this slide, but I'll mention it.
1. The center pad, mask pads, and paste pads are missing the fillets shown in the datasheet, but we can probably accept that too. Especially since I don't see the fillet radius given.

Dimensioned drawing:
image

luca-della-vedova:
Hi and thanks for taking time to dig this up!
> 1. Can you list PNs, wildcarded or not, that can use this footprint in the description? That will help users to search for this footprint in the future, I believe.
Done, it fits the EN5322QI and all the EN53x9QI devices (19,29 and 39), added in description.
> 2. Pads 5, 12, 17, and 18 should be at 2.275mm in the Y direction.
I guess you meant 5,12,17,24, fixed them, good spotting!
> 3. I'd appreciate if you could unrotate the pads.
Done
> 4. The paste pads don't need their zone connection set to Solid.
Done
> 5. The pads aren't the shape specified by Intel. We can let this slide, but I'll mention it.
Yea this was part of the early PR review process, I used the datasheet shape but was told to just use rounded rectangle instead.
> 6. The center pad, mask pads, and paste pads are missing the fillets shown in the datasheet, but we can probably accept that too. Especially since I don't see the fillet radius given.
Yea unsure what to do there.
> 
> Dimensioned drawing:
> image

Cheers!
/Luca

myfreescalewebpage:
@evanshultz  I think this is ready for review following previous changes

PR #1412, opened 2019-02-15
Vishay OLED-128O064D-BPP3N00000

jeanthom:
Hi,

Here's my footprint for Vishay's OLED-128O064D-BPP3N00000 OLED display.

There's is no specification for the component footprint inside the manufacturer's datasheet. Each pad on the flex PCB is 2x0.5mm, so I did 2.5x0.5mm pads on the PCB to make soldering easier.

F.Fab layer's frame corresponds to the area of the flex PCB that should be bent. 

oled-128o064d-bpp3n00000

[Associated symbol](https://github.com/KiCad/kicad-symbols/pull/1531) - [Datasheet](https://www.vishay.com/docs/37902/oled128o064dbpp3n00000.pdf)

---


myfreescalewebpage:
@jeanthom is it a standard connector footprint ?

jeanthom:
There's no connector, you're supposed to solder the OLED screen directly to the PCB:

ssd1306-10a-back

[Source](https://golemparts.wordpress.com/2017/07/24/ssd1306-1-3-oled-spi-breakout-board/)

poeschlr:
Well if the designers of the display would have taken a sane pitch then there would be connectors available. (0.5mm or 1mm pitch fpc connectors exist. Digikey even has 0.8mm pitch options. So close and yet so far.)
With the current solution you are right there really only is the option of a cheap hack.


It might be a good idea to include more of the cable in the footprint. To me it really looks like the pcb edge should be somewhere near where it is in your picture (At the least) So it might be hard to get any component fitet there.

---

Include the manufacturer in the footprint name.

---

In general the footprint looks good. It might however be a good idea to decrease the pad width slightly to increase the pad to pad clearances. (This footprint will never be used for automatic assembly so i would guess we should make it as easy as possible for users to solder this by hand.)

Screenshot from 2019-04-02 22-23-04

evanshultz:
Two notes:
1. There is no datasheet link for this footprint.
1. Can anything be wildcarded? Or do we not want to do that because of the 3D model? Since the placement of the display seems like it would be quite variable (like different orientations of the flex cable, for one) perhaps there's not one single 3D model that will be suitable and thus a wildcarded footprint works well here? There are multiple options at https://www.vishay.com/displays/oled-graphic/.

PR #1414, opened 2019-02-15
Create missing Infineon MLPQ-48 footprints

evanshultz:
Split off from https://github.com/KiCad/kicad-footprints/pull/981 to support merging of https://github.com/KiCad/kicad-symbols/pull/1395.

image

Infineon gave me non-public documents showing the footprint and stencil for this package. :(

---


evanshultz:
Hang on...

evanshultz:
These parts can use this footprint:
- https://www.infineon.com/dgdl/irs2093mpbf.pdf?fileId=5546d462533600a401535675fb892793
- https://www.infineon.com/dgdl/irs2052mpbf.pdf?fileId=5546d462533600a401535675d3b32788

Just rechecked my docs and this footprint matches what Infineon gave me:
image

@easyw
When I import into FreeCAD, you can see above that the main pad and one custom polygon are shown. But the second custom polygon is not shown.

evanshultz:
I'm asking Infineon a question so please hold off on merging this.

easyw:
@evanshultz 
I noticed that you made a pad as a square plus a triangle as polygon and a rectangle as an other polygon...
I would suggest to create the EP as a single polygon coming from the external perimeter.
That should simplify both the kicad fp and freecad imported.

easyw:
@evanshultz 
here my sample
[Infineon_MLPQ-48-1EP_7x7mm_P0.5mm_EP5.55x5.55mm-v2.zip](https://github.com/KiCad/kicad-footprints/files/2870844/Infineon_MLPQ-48-1EP_7x7mm_P0.5mm_EP5.55x5.55mm-v2.zip)

evanshultz:
@easyw 
Updated as mentioned. That is better. Thanks!

@myfreescalewebpage 
If you're open to merging this, it allows https://github.com/KiCad/kicad-symbols/pull/1395 to be merged which gets back symbols that were in the lib earlier and lost at v5.

myfreescalewebpage:
@evanshultz looks great for me, however, I just realized there is also a footprint "Infineon_MLPQ-48-1EP_7x7mm_P0.5mm_Pad5.55x5.55mm" (difference is EP vs Pad in the name. Yours is totally the same except pin 49. I would like to suggest the old one should be removed.... (maybe for milestone v6 ?)

Tell me your position before I merge this one. Thanks,
Joel

myfreescalewebpage:
@evanshultz just looked at the usage of "Infineon_MLPQ-48-1EP_7x7mm_P0.5mm_Pad5.55x5.55mm" and also "Infineon_MLPQ-48-1EP_7x7mm_P0.5mm_Pad5.15x5.15mm", both are not used at all in the symbol library. I suggest they should be removed right now.

evanshultz:
@myfreescalewebpage 
As mentioned above, I split this off from https://github.com/KiCad/kicad-footprints/pull/981 where those footprints will be deleted. To avoid branch conflicts I made minimum changes here. And they aren't used because the IR symbols from the old ir.lib weren't copied to the v5 libs. So if this and the symbol PR are merged all will be well with IR symbols, and the above-referenced PR will continue to handle the cleanup activities it originally included. Is that OK for you?

myfreescalewebpage:
Thanks for the answer, I have not read all the content of #981, it's OK for me. Merging this PR and Infineon symbols now ! Great job !!

PR #1415, opened 2019-02-15
Add 10 pin DFN, 4x4mm, 0.65mm pitch, with 2.65x3.05mm EP footprint.

stambaughw:
Datasheet used to create this footprint can be found at:
https://www.ichaus.de/upload/pdf/MZ_datasheet_B2en.pdf

Datasheet URL:
https://www.ichaus.de/upload/pdf/MZ_datasheet_B2en.pdf

3D model pull request:
https://github.com/KiCad/kicad-packages3D/pull/530

Obligatory screenshot:
image

---


myfreescalewebpage:
Hello @stambaughw 
I do not agree with the EP size, should be 2.8x3.2mm (datasheet page 3).
Cheers,
Joel

stambaughw:
Joel, the KLC does not make it clear whether to use the actual component
EP size or the footprint EP size.  I opted for the component EP size.  I
can change it if that is indeed the policy but the KLC should be
clarified that the EP size is the footprint size not the component EP size.

poeschlr:
@stambaughw the footprint as it is now is a generic dfn footprint. We try to get all of these generated with our IPC scripts. (We kind of require new ones to be generated that way.)

The scripts are found under https://github.com/pointhi/kicad-footprint-generator/tree/master/scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON
You would need to add the package size definition (including tolerances) into the size definition file. The script then generates a footprint from that using IPC formulars and constants.

---

There is however a slight problem with that. The datasheet does not give us tolerancing information for the package. Meaning with we sadly can not go this route for that part without requesting additional information from the manufacturer (or finding a better datasheet or at least a more detailed package drawing for it.)

Another option would be to make this into a manufacturer or even part specific footprint. (meaning prefix the manufacturer name to the footprint name.) That way we can duplicate the suggested footprint instead of going through the script.
We could even move the footprint into a more specific library (One of the sensor libs might be a good fit.)

---

Assuming you choose option 2, we should still align the pin 1 marking style to what the script would generate.
IPC defines that pin 1 is marked by a missing line for packages that have their pads below them (DFN, QFN, ...)

As inspiration see screenshot of `DFN-6-1EP_3x3mm_P1mm_EP1.5x2.4mm`:
screenshot from 2019-02-16 14-02-57


---

Regarding EP size definition: The footprint pad is to be used. See: http://kicad-pcb.org/libraries/klc/F2.2/ We could add the word footprint in front of pad to make it even more clear. (I assumed that pad always refers to the land pattern as the word lead or slug is used for the package features.)

poeschlr:
Is there a reason why you use roundrect radius set to 0.5? Would it be better to use oval pads in that case.

If you want to use roundrect then i suggest to go with IPC suggestion of 25% radius ration (maximum of 0.25mm radius but that would only apply to the exposed pad.)

stambaughw:
Are you sure the radius is not .175mm?  That is what is recommended by
the manufacturer on page 3 of the datasheet
https://www.ichaus.de/upload/pdf/MZ_datasheet_B2en.pdf.  It's hard to
believe that I would have made a typo from .175 to .5.

poeschlr:
I did look directly at the footprint file (Not within kicad as i do not have access to kicad right now).
The footprint does not specify a radius in mm but only a radius ration (KiCad does not support defining the radius directly in footprints. One can enter the radius in new nightlies but all that does is calculate the required ratio for the footprint)

Right now they are on 0.5 (equals 50%) -> This results in a pad equal to the oval pads. I prefer the use of oval pads in such cases (backwards compatibility)

---

The 25% suggestion comes from IPC-7351C (Draft for the new standard replacing IPC-7351B expected to come out not too far into the future.)
IPC acknowledges in that standard that manufacturers will likely suggest oval pads for some time to come (Main reason is that there is a concern for tool support. KiCad is not the only tool that was missing rounded rectangle support until recently. IPC seems to have found that most if not all tools have some way of working with rounded rectangles now this is why they updated their suggested pad shape recently.)

stambaughw:
@Rene, it looks like option 2 is the correct solution.  I cannot find a
single mention of a DFN-10 4x4mm, 0.65 pitch component used by any other
manufacturere so it looks like this should be a manufacturer specific
footprint.  I'm fine with making the changes but rather than massive
back and forth of me getting it wrong, I have a few questions.

Would the correct footprint file name be
iC-Haus_DFN-10-1EP_4x4mm_P0.65mm_EP2.8x3.2mm.kicad_mod?

Would the 3D model file name be the same sans extension.

What is the correct library (path) for the footprint?

What is the correct library (path) for the 3D models?

Once I have this information, I will update everything accordingly.

stambaughw:
@poeschlr Did you get a chance to look at my comments about getting this footprint fixed and committed?  There is no hurry.  I would just like to get this merged so I can resync my repos.

poeschlr:
The footprint name you suggested is ok. I would put it into the Sensor library. It might be easiest to have the 3d model named the same and put into the same library. 

In theory we could also point to the generic 3d model (Model named after generic package, in dfn library.) 
But it might be confusing as we do not have a generic footprint that would go with that model. (And we can not really make one from the bad drawing provided in the datasheet.)

stambaughw:
@poeschlr, Just so I'm clear, if I can get the package tolerances from the manufacturer than I can just add this footprint to the generic DFN script?  This would be my preference.

poeschlr:
TlDr: In general the component at hand can use a generic footprint. To make such a footprint we need the tolerance information such that we can use the script to generate a footprint with current industry standard rules. (And place it as a generic footprint into the generic library.)

---

Long and detailed answer with my reasoning behind why i would handle it this way:

Without tolerances we sadly really only have the option to blindly duplicate the suggested footprint. 

We typically only duplicate footprints for such generic packages exactly if the component needs a specialized footprint for a reason. Such a footprint then needs to include the manufacturer name (possibly even the part number or part series designation). In a lot of cases it even makes sense to be placed in more specialized libraries. (Yes this is only slowly developing. Until recently we had no tools to generate true generic footprints so there was less of a need to take this all into account. Lately i try to get the generic libs to use more generic footprints. -> Slow and tedious work so it will take a while till this is done.)

---

Having a specialized footprint for similar packages mostly makes sense for MEMS components that really need special treatment. Such footprints typically include a keepout marking or need special pad shapes or sizes to function properly. Even if their packages looks very similar to a typical generic one. (In other words: For such parts it would not even matter if we have tolerancing information. Other factors would force us to use/create a specialized footprint.)

As far as i can tell this does not really apply to your component. (Hall sensors are not really sensors that need such specialized footprints. Only thing i could think of is that one can not have high current traces near or below the part. There is however not even a remark in the datasheet about something like that so i guess not even this is a real concern for this component.)

stambaughw:
@poeschlr I got a response from the vendor and the dimensions for the component are per JEDEC MO-229F.  I have a copy of this standard but it's watermarked with my company's name so I don't think I should be posting it.  The JEDEC MO standards are "freely" available if you register on the website.  If this is the information that you need to make this a standard footprint, let me know and I will see if I can figure out how to update the script to generate the appropriate footprint.

poeschlr:
@stambaughw i tried to register with JEDEC a month or so ago. Never got my access granted. Either i misstyped my mail address or they where unhappy with me registering using a gmail address plus giving KiCad as the company. (I could try to use my university address but i will loose access to it as soon as i finish my theisis. The same is true with using the address of the industry partner who supports my theisis.)

poeschlr:
Oh and yes the jedec standard likely has all the information one needs to make a proper ipc footprint for this part. If you are able to get the information from jedec then we have that option. I thing @evanshultz also has access to jedec so he can verify the result.

evanshultz:
@stambaughw 
I'll be happy to review if you reference a JEDEC standard. https://github.com/KiCad/kicad-footprints/pull/1371 is another PR where we're planning to do that.

poeschlr:
@evanshultz i assigned you as you have access to jedec standards and already offered to make the review.

evanshultz:
@stambaughw 
Long, long delay. Trying to clean things up before the final migration. As Rene mentioned above, it appears we can use the generator script to create this footprint. I don't see a PR from you at https://github.com/pointhi/kicad-footprint-generator/pulls. Would you mind to submit the script-generated footprint?

PR #1392, opened 2019-02-07
add QingPu 3.5mm jacks, mono and stereo

robkam:
was "Creating a few parts popularly used with Eurorack synth modules" #1073 

QingPu_WQP-PJ398SM (mono)
oval pads and holes 
circular pads and holes (already in lib but lacked pin 1 marker on F.Fab)
http://www.qingpu-electronics.com/en/products/WQP-PJ398SM-362.html (no pdf)

QingPu_WQP-PJ366ST (stereo)
oval pads and holes 
circular pads and holes 
http://www.qingpu-electronics.com/en/products/WQP-PJ366ST-364.html (no pdf)

annotation 2019-02-07 101017

PR #1416, opened 2019-02-17
add pin 1 indication on F.Fab

robkam:

accidentally split from "add QingPu 3.5mm jacks, mono and stereo" #1392

capture


evanshultz:
The picture above does not show the pin 1 mark on the F.Fab layer. But anyway, it's a duplicate of the silk pin 1 marker and that's not necessarily what may be best since avoiding pads isn't required on fab layers.

We first need to decide what pin to mark. This affects the silk pin 1 mark, the pad shapes, and the footprint orientation. I have opened https://github.com/KiCad/kicad-footprints/issues/1420 to capture this topic.

In addition:
- The silk lines are on top of fab lines in some places and offset in other places. Please pick one method and stick with it everywhere.
- Move the value outside of the part body, either above or below.
- Even if I only make the pad clearance 0.12mm the silk lines intrude. Give at least as much clearance as the silk line width. The silk pin 1 mark is almost totally inside this clearance which means the line will be dramatically thinned or totally removed on an actual board (unless the PCB fabricator intervenes).
- The lower courtyard line can be moved up 0.01mm to give to required 0.5mm offset (+/-0.005mm).
- The fab layer ref des should not overlap with other fab layer elements (right now it hits a circle).
- It seems odd that that the "S" pad is different in X and Y dimensions by 0.1mm. Why such a small difference? How did you arrive at this pad design.
- I would prefer pads are not rotated, but this isn't required.

Can you make the updates I enumerated above? The community can build consensus on the pin 1 marking (or lack thereof) and then this PR can be merged. Then the changes here can be pushed to the other PR you mentioned and they can all be done.

robkam:
> We first need to decide what pin to mark. This affects the silk pin 1 mark, the pad shapes, and the footprint orientation. I have opened #1420 to capture this topic.

On the data sheet the pins are named 1, 2 and 3. So here pin 1 is marked as such. However I've then renamed them to S, TN and T as per Audio jack fix #1014. The symbol uses e.g. S, TN, T and when the footprint had 1, 2, 3 CvPcb would complain.

robkam:
> * It seems odd that that the "S" pad is different in X and Y dimensions by 0.1mm. Why such a small difference? How did you arrive at this pad design.

I can't recall nor work it out, also not for the hole.
http://www.qingpu-electronics.com/en/products/WQP-PJ398SM-362.html gives rectangular hole 
dimensions 1.3 x 0.6 mm. I assume this already includes the 0.2 mm clearance on l and w. 

BTW using digital calipers to measure the pins and an unsoldered PCB with round holes already available to suit these jacks:
pin 1 - 0.82 x 0.28 mm and hole 1.31 mm
pin 2 - 1.02 x 0.28 mm and hole 1.58 mm
pin 3 - 1.03 x 0.17 mm and hole 1.55 mm

robkam:
Data sheet actually doesn't include pin dimensions, except for 0.8 mm for one dimension of pin 1 / tip (I'm waiting to hear back from Qingpu).

The hole sizes here (based on measured pin sizes above) are not very different to the holes in a (small) commercial produced PCB https://docs.google.com/document/d/1rFD1qOnU61r5iWUj4tl1WYmum1hYpuAtiI1hz0R8bA0/edit
or to those in an EAGLE .lbr also used in (small) commercial product runs without problems https://github.com/TomWhitwell/MTM-Parts-Library 

capture

robkam:
Qingpu replies that ground / pin 1 / S is 0.8 mm x 0.3 mm, the other two are 1.0 mm x 0.3 mm, but allow for tolerances.

capture

robkam:
capture

robkam:
Should be merged as soon as possible to fix errors with one currently in the footprint library.

robkam:
bump

evanshultz:
1. Silk clearance to pads is still too little. See my comments above.
1. It would be nice to unrotate the pads.
1. Were the old holes too large? Because of the uncontrolled tolerance I'm inclined to use bigger hole sizes than what you have now. If the old drill sizes weren't causing problems let's stay with it.

What error is in the library now? This PR is only talking about adding a pin 1 indicator. Is it that, or is there more wrong?

robkam:
I don't understand "unrotate the pads"?

evanshultz:
This is a rotated pad:
image

Note that, depending on the pad rotation and size, the pad's X and Y dimensions may need to swap.

robkam:
Fixed silk clearances. Pads unrotated. I can't work out whatever it was I thought in error five months ago. I can only find a few tenths of a millimeter difference between hole sizes before and after the correction. 

Capture

robkam:
Silk clearances still need fixing.

robkam:
Capture

evanshultz:
Silk spacing looks good now. Thank you.

The drill sizes are all smaller and some pads are bigger and some smaller. See my comment at https://github.com/KiCad/kicad-footprints/pull/1416#issuecomment-518868815. If there's no reason, bigger holes sounds better to me.

Also, set the solder mask clearance to default (zero).

PR #1417, opened 2019-02-17
add footprint for dual isolated BNC jacks

lorem-ipsum:
This adds a footprint designed to be used with the generic `Conn_Coaxial_x2_Isolated`. It is based of the drawings for an connector from [Amphenol](https://www.amphenolrf.com/downloads/dl/file/id/2980/product/644/031_6575_customer_drawing.pdf) and [Bomar Interconnect](https://www.mouser.de/datasheet/2/53/DWG_364A2595Rev_A-1158937.pdf) (which seems to be the same as Winchester Interconnect). Both have the same positions and diameters for the mounting holes, only the outer dimensions of the connectors are different. As those there the only ones I could find at all I made it a generic footprint (even though I am not sure whether the name is correct). I used the existing `BNC_Amphenol_B6252HB-NPP3G-50_Horizontal` as starting point and adjusted the dimensions, but kept the drawing.

I used the larger dimensions which are those from Bomar Interconnect. The pin labeling is not stated in the Amphenol, but looking at the projections this seems to be correct. Bomar labels them, but they label the outer and inner pin of the coax connector the other way round than the symbol library (they use 1 for out, 2 for in). Because of this 1 and 2, and 3 and 4 must be flipped, so the two match up.

screenshot_20190217_154032


myfreescalewebpage:
Hello @lorem-ipsum 

> the outer dimensions of the connectors are different

Thus, this is not a generic footprint. You should propose an atomic footprint and rename according to the footprint chosen.

Thanks,
Joel

lorem-ipsum:
I had it first as `BNC_Amphenol_031-6575_Horizontal` with the smaller Amphenol dimensions. But there are multiple variants of this connector from Amphenol with exactly the same dimensions. Same for the Bomar connector. The relevant dimensions are the positions of the pins and the connectors towards the front panel, and these are identically for both components.

I can make it a specialised connector if that makes more sense to you, but I think it will just result in plenty of identically footprints in the library.

myfreescalewebpage:
I agree with you that position of pins and outside connector is very important, but silkscreen and CrtYd layer are very important too during a PCB design. That's why the new parts will need a new footprint.

lorem-ipsum:
Ok, I will add the two variants of this connector which I need. Is the naming scheme I proposed in the second comment fine?

Should I also make ~20 copies for the identical components with other names or is there another way to handle this?

myfreescalewebpage:
Yes `BNC_Amphenol_031-6575_Horizontal` is great format.

> Should I also make ~20 copies for the identical components with other names or is there another way to handle this?

Submit only the first you need at the moment, but if you can give a list of the equivalent connectors that's great!

lorem-ipsum:
I made the split into the connectors and documented the reasoning behind the naming in the last two commit messages. The Bomar/Win connector now matches all existing variants, the Amphenol would require the creating of 4 identical footprints. Then there are many more which are not isolated, but for the isolated ones this seems to be it.

poeschlr:
Do we have symbols that fit these parts? (the one listed in the description does not exist)

I fear a few measurements of your footprints are wrong.
Screenshot from 2019-06-02 19-51-58

lorem-ipsum:
thanks for the review. I've fixed the wrong dimensions. The symbol is part of the Connector library under the mentioned name. It was added in [this PR](https://github.com/KiCad/kicad-symbols/pull/1103).

poeschlr:
One minor problem remains. (threaded length of amphenol)
Screenshot from 2019-06-03 20-04-19

lorem-ipsum:
This is fixed by now, just forgot a ping afterwards.

poeschlr:
Is it on purpose that the WIN connector now has its grounds connected? I thought these are isolated connectors (stated in your original pull request title)

lorem-ipsum:
Sorry for not posting this, yes this is on purpose. I've got some of these connectors here, and after some debugging I realised that the WIN connectors are isolated from the front (plastic shielding), but not from each other and the mounting pins. I've also checked the Amphenol connectors, they are isolated as posted in the original version of this PR.

Sorry for the late changes, got hit by surprise as this was not clear to me based on the (little) available documentation.

poeschlr:
Well thanks for the update.

Could you add a bit more description to the description field? After that this is good to go.

lorem-ipsum:
ok, added some description.

poeschlr:
thank

PR #1419, opened 2019-02-19
Added the footprint of a shielded TO-46-4 can, which is used by the L…

PatrickBaus:
This is a TO-46-4 footprint with an additional can around it to thermally shield it. It is used by the LM399 Zener reference (https://www.analog.com/media/en/technical-documentation/data-sheets/199399fc.pdf).

The courtyard is increased on purpose, because the thermal shield is not fixed to the TO-46 can. It is loosely mounted around it.

screenshot from 2019-02-19 10-10-14

I will also add a 3D model in a seperate pull request.
---


PatrickBaus:
Needs to be added to the generator scripts

PR #1421, opened 2019-02-19
Add reverse mount 18.4x8mm TSOP-I-32

fauxpark:
A TSOP-I-32 footprint with pins mirrored around the Y axis, for mounting on the other side of the board. Modified from the original symbol.

fp

Symbol: https://github.com/KiCad/kicad-symbols/pull/1441

Datasheet: http://www.futurlec.com/Datasheet/Memory/628128.pdf#page=18

Note that pin 1 is at the *top right* of the footprint. This violates the "pin 1 at top left" rule, but it is meant to match the normal pinout footprint so I hope this is a reasonable exception. When placed on the back of the board the pins line up:

fp

---


poeschlr:
What do you mean with mounting on the other side of the board? Typically if you place a part on the "bottom" side you simply flip the existing footprint. This is because the real part does not care on which side it is mounted.

Is this meant for embedding the part in the pcb? If so then there would also need to be an edge cut defined.

fauxpark:
See the second image above for what I mean. The top symbol is the existing normal footprint, and the bottom one is this new one.

The use case appears to be ease of routing double sided PCBs, when you are placing this part on both sides. If you only used the existing footprint, you would get this:
bad1
or this:
bad2

evanshultz:
@poeschlr 
It's a normal TSOP-I with the leads bent the other way. At least it appears to be so.

@fauxpark 
We would normally rotate the footprint so that pin 1 is still in the top left corner. I don't see why that wouldn't work here.

I'd be remiss not to mention that this is only to support adding a symbol for an old, obsolete part that didn't even exist before you took on the initiative to update the symbol. I checked out a pile of datasheets for compatible parts on Digi-Key and all of them in TSOP had pin 1 as A11 so this reverse pinout may be quite rare. I tried to check out current memory ICs from Hitachi but gave up; they make a lot of crap and neither their site search nor Google helped me zoom into their current memory offerings.

fauxpark:
@evanshultz
Well, this reverse TSOP and the normal TSOP footprint seem to be a matching set, so I figured the orientation should reflect that.

Me fixing up the 628128 symbol was intended to help tick off the Memory_RAM item in https://github.com/KiCad/kicad-symbols/issues/520, and now I can see why it hasn't been yet 😁

evanshultz:
Yeah, no good deed goes unpunished, as they saying goes. Would you mind rotating to stick with the existing convention? Then I'll merge.

fauxpark:
Done, and updated the OP image.

evanshultz:
Can you make the text right-side up? No need to rotate that.

Also, this is my preference, but what about not rotating the pads? That will require flipping their X and Y values. It just makes the file cleaner and a bit smaller but has no effect on the actual footprint so it's not required. Though it is quick and easy with search and replace.

fauxpark:
Whoops, forgot about that. Done.

evanshultz:
Beautiful!

PR #1422, opened 2019-02-20
Add Texas_VQFN-HR-12_2.0mmx2.5mm

Puchinger:
Add Texas_VQFN-HR-12_2.0mmx2.5mm

[link](http://www.ti.com/lit/ml/mpqf508/mpqf508.pdf)

footprint


poeschlr:
Are you ok with your contribution being published under the kicad library license? If so please sign the license agreement.

This one should be scriptable with https://github.com/pointhi/kicad-footprint-generator/tree/master/scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON
Have a look at the size definition file "special_qfn.yaml" as it generates a very similar footprint.

(It might even be a lot faster to learn how to use the script then to fix the many errors found by our test script for you contribution.)

The only thing our script can not do is add the 4 vias suggested by TI. Meaning a true duplication of the suggested footprint would require manual work.

Puchinger:
@poeschlr 
Thank you for your advice.
I hope that everything fits now.
footprint

Puchinger:
@poeschlr 
Is there any other mistake I should fix?

poeschlr:
I assume you scripted this now. Can you add a link to your pull request that adds the size definition to the script?

Puchinger:
I could only generate the skeleton with the script. 
The pads 1,2,7 and 8 have different lead widths, therefore I have to place it per hand.

poeschlr:
I somehow missed that when i first looked at it. Well another thing i need to add to the script in the near future. (Should be reasonably easy to implement. I however have another pull request open right now that would interfere with implementing support for creating such a footprint so i will hold off for now.)

evanshultz:
I was going back to old PRs trying to clean them up and found this. A few issues remain:
1. Beef up the description so it's more than just a URL.
1. You're missing the vias. Since this is a TI-specific footprint we should built it to the doc.
1. Silk lines are just touching the pads. Give 0.2mm clearance from silk to pads.
1. Set the mask clearance to get 0.05mm over the pads as the doc shows.

@Puchinger Are you able to wrap this up?

myfreescalewebpage:
@Puchinger any news here following previous comment ?

Puchinger:
@myfreescalewebpage unfortunately, I am not able to fix this right now, I can do it in one month at the earliest.

PR #1423, opened 2019-02-20
QFN-20, 0.4mm pitch, 1.7x1.7mm EP

jeanthom:
Hi,

This is a QFN-20, 0.4mm pitch, 1.7x1.7mm EP required by [SGTL5000XLNA3](https://github.com/KiCad/kicad-symbols/pull/1553).

qfn20

[Datasheet of the SGTL5000, see page 61 for package description](https://www.nxp.com/docs/en/data-sheet/SGTL5000.pdf) - [SGTL5000XLNA3 pull request](https://github.com/KiCad/kicad-symbols/pull/1553)

---


poeschlr:
Footprints for standardized packages (like this one) must be scripted using our IPC generator scripts found here: https://github.com/pointhi/kicad-footprint-generator/tree/master/scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON

jeanthom:
Hi! Sorry for the delay, I am quite busy this month.

I'm wondering if that rule should still apply since KiCad 5.0 brought an integrated footprint generator (which I used to generate this footprint).

evanshultz:
Yes. The integrated footprint generator is an embarrassment.

jeanthom:
Uh ok ^^ Could you elaborate on that ?

Is there documentation for the python kicad-footprint-generator ?

evanshultz:
In every way. But most importantly, the footprint generator Rene linked above uses the IPC industry-standard rules to generate pad sizes and is much easier to maintain in the future.

The documentation is lacking, but I'll help you get started.
1. Go to https://github.com/pointhi/kicad-footprint-generator/blob/master/scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON/size_definitions/qfn/qfn-20.yaml which is the file that contains the package info from the datasheet or package drawing.
1. You can find `QFN-20-1EP_3x3mm_P0.4mm_EP1.65x1.65mm` which is quite similar, so we will copy that to make the new footprint. I will reference the lines of this entry in the changes for the new entry.
1. Update the datasheet link on line 72.
1. There is no tolerance given for the outside dimensions, so remove lines 78 and 81.
1. Lines 83-84 change to using the minimum and maximum dimensions of 0.5mm and 0.6mm from page 62 of the datasheet.
1. Change lines 87-88 to use min/max numbers instead of nom and tol. Same for lines 90-91, 95-96, and 98-99.
1. The rest should be OK so, assuming you have Python 3 installed and in the PATH env var, at the command line (I use Windows) type `python ipc_noLead_generator.py ./size_definitions/qfn/qfn-20.yaml`. A new folder with the QFN-20 footprints should be created.
1. Pick out the new footprints (with and without thermal vias) and submit them in this PR.
1. Create a new PR at the script repo to submit the update to the YAML file.

How to configure the generator or what it can do and what it cannot do isn't well written down. I think in this case all should be OK but there are plenty of edge cases. Let me know if that helps and it you get stuck anywhere.

PR #1426, opened 2019-02-21
add AD cp-32-2 footprint for LFCSP-32 package

lorem-ipsum:
This adds the footprint for a package from Analog Devices they call CP-32-2. It is a LFCSP-32-1EP footprint which is used by symbols in [this PR](https://github.com/KiCad/kicad-symbols/pull/1558). The mechanical specification can be found [here](https://www.analog.com/media/en/package-pcb-resources/package/414143737956480539664569cp_32_2.pdf). The footprint was generated using the footprint generator with the definition added in [this PR](https://github.com/pointhi/kicad-footprint-generator/pull/297). The package has chamfers which are not added, but they also only have a maximum specification.
screenshot_20190221_222839


poeschlr:
Looks good, thanks.

PR #1428, opened 2019-02-23
Remove copper layer from paste (stencil) pads

SchrodingersGat:
This PR fixes a bug I found in the Vishay PowerPak Single footprint.

The stencil pads do not have a number (which is correct) but they incorrectly have the front copper layer set. This prevents the footprint from being routed as the copper pads is un-numbered.

For this PR I have simply removed the copper layer.

I think this should be something that the KLC should be able to detect, probably have to implement a collision detection routine for copper shapes.


myfreescalewebpage:
@SchrodingersGat thanks for this fix ! No comment on the PR, merging.
Yes probably travis scripts should be enhanced (more generally !)
Joel

PR #1429, opened 2019-02-23
Mentor 4 way lightpipes 1276.1004 and 1276.2004

lorem-ipsum:
This adds two very similar lightpipes from Mentor. They can be found in their [product overview](http://www.mentor-bauelemente.de/katalog/ll/MENTOR-LL.pdf) at page 28. The two different variants have a spherical and a planar end, but otherwise they are identical.
I was first thinking of not extending the silk screen all the way to the front of the lightpipe, but then the silk screen is completely symmetric. Do you maybe have a suggestion on how to solve this? The front part of the lightpipe will most likely not be on the PCB anymore.

screen


lorem-ipsum:
note to self before this is merged: The silk screen needs to be adjusted as it collides with the pin 1 marker of PLCC-2 LEDs placed withitin.

lorem-ipsum:
I reduced the silk screen to look like this:
screenshot_20190224_173513

This is how it looks with PLCC-2 diodes placed inside (Fab and courtyard layers disabled):
screenshot_20190224_173427

poeschlr:
Looks good in general. I however found a small break in the fab outline. (Minor problem but would still be nice to be fixed. Marked with red arrow)

Screenshot from 2019-06-01 18-36-29
Screenshot from 2019-06-01 18-36-24

lorem-ipsum:
thanks for the review. Should be fixed now.

poeschlr:
thanks

PR #1430, opened 2019-02-23
Stocko fix

balcer:
During 3D model generation I noted that I broke footprints. 

stocko

https://www.stocko-contact.com/downloads/steckverbindersystem-raster-2,5-mm.pdf#page=15

---


myfreescalewebpage:
@balcer many files changed here, is it really wanted ? Please have a look at the git diff, thanks!

evanshultz:
The CONNFLY USB connector is probably an accident. All other changes files are Stocko MKS parts. Note that I didn't check each one, just quickly looked at changed files.

https://github.com/pointhi/kicad-footprint-generator/pull/298 is the PR for the script fix that drove this, so these are generated by script, which is why some footprints only had the time stamp changed.

myfreescalewebpage:
> The CONNFLY USB connector is probably an accident

Yes I think too

balcer:
Now should be ok. sorry for problems

evanshultz:
I just quickly checked the script fix and it did add the -X offset at 3.6mm from pin 1, which is the correct dimension in the datasheet. I added more notes at the script PR that don't affect the footprint.

Looking at the code paths, and even though they were submitted and checked at https://github.com/KiCad/kicad-footprints/pull/1403, I dimensioned both 2-pin, the 3-pin, and the highest pin count footprints:
image

balcer:
thx

PR #1433, opened 2019-02-25
update TQFP-100-EP5x5mm and add ThermalVias

lorem-ipsum:
new version is scripted using the footprint generator, matching definition can be found [here](https://github.com/pointhi/kicad-footprint-generator/pull/301).
screenshot_20190225_083038


poeschlr:
The parameters needed for the footprint are ok. Small details added to the script pull request to be fixed before the script addition can be merged.

PR #1435, opened 2019-02-25
Oscillator: Add Abracon ASDMB

antoniovazquezblanco:
https://abracon.com/Oscillators/ASDMB.pdf

Sym PR: https://github.com/KiCad/kicad-symbols/pull/1574

imagen

---


evanshultz:
1. Could you put the link at the end of the description?
2. Are their rules for the adhesive layer or should it be removed? I didn't recall seeing this layer used before but I see several SMD oscillator footprints are using it. I would expect this to be in KLC if it's to be used in the official library.
3. I'm wondering if there's a better silk pin 1 mark. This seems a bit ambiguous to me. Could you take a look at other, similar footprints and see if there's anything more clear. If not I can merge like this, but at first glance it wasn't clear to me.

evanshultz:
@antoniovazquezblanco Ping!

antoniovazquezblanco:
1. Done
2. My design was based on a very similar device in the lib. Datasheet does not specify an adhesive requeriment. Removed.
3. Added more silk to make pin 1 marker clearer. Not standard at all but could not find anything better among other oscillator devices in the lib.

Screenshot from 2019-03-28 16-21-36

evanshultz:
Thanks.

Now looks like this (with 0.12mm mask margin):
image

A couple of things I think of:
1. Could you remove the `^2` in the description? That is like our typical descriptions and understandable.
1. The ref des text pokes out of the fab outline just a bit with the 5 characters. 0.6mm/0.09mm works.
1. I'm not sure why the oscillators have a different pin 1 fab mark than elsewhere, but probably because scripting or the first one was made that way and nobody else touched it?
1. Maybe just put silk on the pin 1 corner and nowhere else?
image

FYI, since most of them don't have pin 1 at the top left we'll allow it.

antoniovazquezblanco:
Done,

Screenshot from 2019-04-01 10-18-51

Thanks!

evanshultz:
Nice. Thank you!

antoniovazquezblanco:
Thanks!

PR #1436, opened 2019-02-25
Add Myrra EF20 footprint

johnbeard:
Used for 74087, 74088, 74089 flyback transformers.

http://myrra.com/wp-content/uploads/2017/09/Datasheet-74087-74088-74089-rev-A.pdf

2019-02-25_172844_718x634_screenshot
2019-02-25_173751_884x551_screenshot
2019-02-25_174829_606x643_screenshot

---


evanshultz:
Hi John!

1. Do we need to include the `_no_pin3` in the footprint name? Since the part is always missing this pin, and the footprint name is explicit for only this transformer, I think we can leave it off.
1. 2mm pads is a bit small for the relatively large leads and drills. Maybe 2.5-3mm?
1. Pin 1 should be at the top left. Can you rotate the footprint. If you don't mind, rotate each pad again so it's at 0 degree rotation.

If you will avoid force pushing the changes it will be a bit more quick to review. I will squash merge this branch to keep the commit history tidy. Thanks!

johnbeard:
@evanshultz hi, thanks for the review!

1. You are right, looks like all the Myrra EF20's have a missing pin 3. It's now `Transformer_Myrra_EF20`.
2. Made them 2.5mm. Is there an IPC-type standard for such big holes? IPC Level B indicates 1.2 + 0.6, which is a very skinny annular ring!
3. Done.
2019-02-26_103052_511x524_screenshot

johnbeard:
So the datasheet drawing is apparently from the pin side, but it's not on the datasheet on the Myrra website, it's in the catalogue.

So the pins, viewed from the top, start on the right and go anti-clockwise:

2019-02-26_150030_579x611_screenshot
2019-02-26_150224_846x524_screenshot

evanshultz:
1. Looking at the datasheet again, there is a 5-digit PN for these parts. I think it's best to include that since EF20 is the core size and numerous transformers could be made with that core. What about `Transformer_Myrra_EF20_7408x`? That is probably specific enough and then we can merge.
2. IPC-2221 covers pad sizes, but only for a minimum annular ring. As you found, Level B is mainstream (these days it's actually conservative) and requires 0.55mm over the drill size. My personal experience is that ~1mm is good for moderate-sized and -weight parts like this. Thanks!

Awfully kind of them to leave off this critical info from the datasheet! Good catch on orientation.

Lastly, the 3D  model. As it is now, any Myrra EF20-core transformer without pin 3 would be valid for this model (understood it doesn't exist now). We don't have a 3D model yet and transformers (both footprints and models) aren't sorted out by the librarian team. So we can go with it, at least for now.

johnbeard:
Name is now `Transformer_Myrra_EF20_7408x`.

3D model is still `${KISYS3DMOD}/Transformer_THT.3dshapes/Transformer_Myrra_EF20_no_pin3.wrl`.

evanshultz:
Thanks John!

PR #1438, opened 2019-02-25
add footprint for E-Switch TL1250

lorem-ipsum:
push button with included LED. The pin numbering is changed from the
[datasheet](https://www.e-switch.com/system/asset/product_line/data_sheet/148/TL1250.pdf) to match the generic SW_Push_LED symbol. I did not mark the number 1 pin as there is no clear number 1. For the switch it also doesn't matter.
screenshot_20190225_221549


poeschlr:
Sorry for the delay.

The naming for this footprint should be: `SW_Push_1P1T_[Latching*]_NO_LED_E-Switch_TL1250` (*: Latching if this button is mechanically latching. I can not really tell from the datasheet.)

---

Pin numbers look good (fit the symbol available)
Dimensions look good
Screenshot from 2019-06-01 19-06-00

lorem-ipsum:
Hi @poeschlr,
thanks for the review, I've renamed it to follow the correct pattern.
Cheers,
Hendrik

poeschlr:
thanks

PR #1440, opened 2019-02-26
Use parent FP zone connection on rect pads

evanshultz:
Jumper pads that are not custom should have their zone connection set to "From Parent Footprint".

The open triangular jumper footprints, and only those, have mask pads between the jumper pads. Should mask be removed between all pads of our jumper footprints? I think so and I'm happy to do that here if y'all agree.

---


poeschlr:
thanks

PR #1441, opened 2019-02-26
Package_SO: Add a new HTSSOP-20-1EP_4.4x6.5mm_P0.65mm_EP3.4x6.5mm_Mas…

antoniovazquezblanco:
…k2.4x3.7mm

http://www.ti.com/lit/ds/symlink/bq24006.pdf

imagen

Sym PR: https://github.com/KiCad/kicad-symbols/pull/1576
Script PR: https://github.com/pointhi/kicad-footprint-generator/pull/302

---


evanshultz:
The mask size shouldn't be included in the 3D model. There is already an existing 3D model without a mask size that will work.

However, the script may not be smart enough right now to know that leaving off the mask for 3D models is OK. And we have a 3D model identical but with different mask size already, so it's too late to catch this from the start.

If you just point to the 3D model with no mask manually that will be OK for now and script updates can happen outside of this PR.

antoniovazquezblanco:
Opened a issue at the corresponding repo and fixed the path. From what I've seen, multiple packages should be updated to reflect the latest changes.

Thank you!

antoniovazquezblanco:
Thanks!

PR #1442, opened 2019-02-26
Update Linear_DE14MA.kicad_mod

evanshultz:
Split off from https://github.com/KiCad/kicad-footprints/pull/981.

Currently an unused footprint, I believe, but will be used by https://github.com/KiCad/kicad-symbols/pull/1133.

http://www.analog.com/media/en/package-pcb-resources/package/pkg_pdf/ltc-legacy-dfn/05081731_C_DE14MA.pdf

image

The dimensioned drawing isn't even worth doing since what I changed are pads 4 and 9 and KSU doesn't import the custom pad shapes (yet?).

---


evanshultz:
Here is the dimensioned drawing before fixing the courtyard offset. Because the custom pad shapes aren't imported completely the drawing looks... incomplete.
image

myfreescalewebpage:
@evanshultz looks ok for me on this footprint, merging.

PR #1443, opened 2019-02-27
Add QFN 14 for SKY13575-639LF.

matthuszagh:
http://www.skyworksinc.com/uploads/documents/SKY13575_639LF_203270D.pdf

image

image

Script PR: https://github.com/pointhi/kicad-footprint-generator/pull/304

Thermal vias look a bit weird on this...

---


poeschlr:
you might want to use the parameter chamfer_edge_pins to increase the clearance between for example pin 1 and pin 14.

I also doubt that 4 thermal vias are a good idea for this small center pad. I would not even include a thermal vias version here.

I additionally am unsure if this package is really a generic qfn. Might be better to put it into the respective RF library and give it a more specialized name.

One might even consider to duplicate the suggested footprint in this case as the pad shapes might be suggested that way to guarantee matched impedance. (I am a bit unsure about that. So i could be persuaded to accept a generic footprint.) If you want to go that way then freecad with stepup might really be the best way to go. For a demo of the stepup workflow see: https://forum.kicad.info/t/kicad-stepup-the-sketcher-for-footprint-generation/9582

matthuszagh:
Made changes

image

poeschlr:
If you want to keep the original script generated version then i suggest to use the parameters `chamfer_edge_pins` and `edge_heel_reduction` to increase the clearance at the corners. (It might however be required to extend the script to be able to limit parts of these to one direction only to get a proper result.)

---

For the manual made one:
Split the mask layer of the central pad.
The complex pads are set to pad connection none please select pad connection solid.
The courtyard outline is skewed on the right side

Screenshot from 2019-05-05 11-47-51

matthuszagh:
I've made the changes and here's a new screenshot. I hope the mask layer is correct, I wasn't entirely sure there.

image

poeschlr:
Sorry i kind of screwed up my previous review. The paste layer should be split not the mask layer. (Meaning change the layer of the 4 small central pads to paste and add mask to the large pad instead.)

You might also want to consider using normal rectangle pads for the paste pads that are just rectangles.

matthuszagh:
I changed the layers. I haven't changed the paste layers to rectangles because I had implemented them as graphic polygons not pads. The EP itself is not split, its just a single, unbroken custom shape. I'd be happy to change the EP to be the same shapes as the paste layer (3 rects and 1 custom shape) if you want. Just let me know.

poeschlr:
We kind of require paste to be determined with pads so please change that.

matthuszagh:
Let me know if this is better. It looks like a newer versions of kicad quote names, refs etc., which have made it into the last commit. Let me know if that's an issue and I'll take them out.

poeschlr:
Please remove all polygons (that are not part of pads)
Also make sure there is a copper pad for the central pad. (Copper and mask should be the same pad. Only paste should be separate. These paste pads should only have the paste layer and no pin number)

matthuszagh:
Better? The central paste layers are polygons and are not part of pads, but I'm not sure how to separate them without doing this. Let me know.

poeschlr:
Well you already have some other polygon pads. Make the central past only polygon pads the same way and then simply select "no copper" and deselect all other layers except paste.

You might want to look at https://forum.kicad.info/t/tutorial-how-to-make-a-footprint-from-scratch/11092/3 as i show how to make the paste layout for an exposed pad in this tutorial (normal pads but the same applies to complex pads that have a polygon as part of their definition)

matthuszagh:
I looked at the tutorial which was helpful, thank you. Hopefully this is now correct. Let me know.

poeschlr:
looks good now, thanks

Issue #1434, opened 2019-02-25
Rename Packages_SO:SOIJ-8_5.3x5.3mm_P1.27mm to Packages_SO:SOIC-8W_5.3x5.3mm_P1.27mm


Icenowy:
The SOIJ-8 name is not a common name for this package. By searching on google with this keyword, only materials from Microchip can be found.

However, this package is popular for 25 SPI NOR Flash, and many manufacturers do not use the SOIJ name, they just declare it as a wide version of SOIC/SOP or mention the 208mil size.

See the following:
https://www.winbond.com/resource-files/w25q128fv%20rev.l%2008242015.pdf
http://www.gigadevice.com/datasheet/gd25q127c/
https://www.micron.com/~/media/documents/products/data-sheet/nor-flash/serial-nor/m25p/m25p80.pdf
http://www.xinyahong.com/upLoad/product/month_1411/201411201249025495.pdf (This is EN25Q64)

So I suggest to rename it to follow the scheme of other wide SOICs, rather than using a vendor-specific name.

poeschlr:
SOIJ is typically a package with J lead instead of gull wing. (leads bend inwards instead of outwards) Are you sure about the renaming? (Disclaimer: I did not check the current footprint as i do not have kicad access right now. A screenshot of the current footprint would have made this a lot easier.)

evanshultz:
Here is the footprint mentioned above:
![image](https://user-images.githubusercontent.com/1936989/53355321-9af22880-38dd-11e9-9ecb-8970faf2e9e7.png)

The doc referenced by this footprints is an older version of 
http://ww1.microchip.com/downloads/en/PackagingSpec/00000049BQ.pdf. However, I don't find this package type called "SOIJ" in that doc.

This footprint is used by the following symbols (just searching in Eeschema):
![image](https://user-images.githubusercontent.com/1936989/53355550-148a1680-38de-11e9-93b5-48000dc100eb.png)

I agree that it would be best to change the footprint name and it's not a substantial amount of work so it could be done right away if the other librarians agree.

Icenowy:
Datasheet with the SOIJ name remaining can be easily acquired via Google.

e.g. http://ww1.microchip.com/downloads/en/PackagingSpec/00049AU.pdf

This document says "8-Lead Plastic Small Outline (SM) – Medium, 5.28 mm Body [SOIJ]", and it has no J lead mentioned.

[...]

PR #1446, opened 2019-02-28
Package_SO: Rename SOIJ-8_5.3x5.3mm_P1.27mm to SOIC-8W_5.3x5.3mm_P1.27mm

Icenowy:
The SOIJ-8 package name is only used by the package specification
documents by Microchip. Other manufacturers usually just call it a
variant of SOIC/SOP with 208mil (5.28mm) body. This package is currently
modern among 25 SPI Flash chips from many manufacturers, including
Winbond, GigaDevice, Micron, ST, etc.

As the SOIJ-8 name is vendor specific, and will mislead people with the
"J" as J lead, rename it to SOIC-8W, meaning a wider version of SOIC-8.

Signed-off-by: Icenowy Zheng <icenowy@aosc.io>

### DIFF
--- a/Package_SO.pretty/SOIC-8W_5.3x5.3mm_P1.27mm.kicad_mod
+++ b/Package_SO.pretty/SOIC-8W_5.3x5.3mm_P1.27mm.kicad_mod
@@ -1,11 +1,11 @@
-(module SOIJ-8_5.3x5.3mm_P1.27mm (layer F.Cu) (tedit 5A02F2D3)
-  (descr "8-Lead Plastic Small Outline (SM) - Medium, 5.28 mm Body [SOIC] (see Microchip Packaging Specification 00000049BS.pdf)")
+(module SOIC-8W_5.3x5.3mm_P1.27mm (layer F.Cu) (tedit 5A02F2D3)
+  (descr "8-Lead Plastic Small Outline (SM) - 5.28 mm Body [SOIC] (http://ww1.microchip.com/downloads/en/PackagingSpec/00000049BQ.pdf)")
   (tags "SOIC 1.27")
   (attr smd)
   (fp_text reference REF** (at 0 -3.68) (layer F.SilkS)
     (effects (font (size 1 1) (thickness 0.15)))
   )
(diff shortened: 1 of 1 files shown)


myfreescalewebpage:
@poeschlr I'm not sure this is wanted, my opinion is that it will lead to confusion then when using  the parts, and when adding new parts too. What do you think ?

Icenowy:
@myfreescalewebpage I think the size can already prevent confusion.

Where I buy components usually (LCSC Chinese ver) just call it SOIC-8_208mil, and narrow SOIC-8 as SOIC-8_150mil.

myfreescalewebpage:
@Icenowy do you have example of devices using the SOIJ footprint ? What is indicated in the datasheet ? If SOIJ is used in the datasheet, then it's better to use SOIJ too here to avoid confusion.

I haven't seen any symbol PJ from you attached to this footprint PR, the footprint is probably used today, symbols should be changed too if we decide to accept this modification.

Icenowy:
@myfreescalewebpage For discussion, see #1434.

Symbol changes see https://github.com/Icenowy/kicad-symbols/tree/soij8-fix.

myfreescalewebpage:
:D great link, you should have started here, I haven't seen the content of the issue before ! So let's go with the modification.

Can you make the pull request for the symbols ?

Thanks

evanshultz:
Could you please do two things:
1. Go back to the general Microchip document, but the `00000049BQ.pdf` revision. This package is on page 174 of http://ww1.microchip.com/downloads/en/PackagingSpec/00000049BQ.pdf.
1. Change the 3D model file name to the new footprint name.

antoniovazquezblanco:
@Icenowy ping :)

evanshultz:
@Icenowy 
Can you please make the two changes I requested above?

Icenowy:
@evanshultz how should I change 3D model file name to match this commit?

evanshultz:
This is now correct. The 3D model name (even if it doesn't exist), should match the footprint name. This is good now. Thank you!

PR #1451, opened 2019-03-01
Add Fujitsu FTR-LY Form C relay

johnbeard:
The slightly protruding pad A1 is accounted for in the courtyard
such that the relays can "tesselate" end-to-end if needed.

11, 12, 14 pins sized at 2.4 to avoid extending the courtyard.

DIN EN 50005 pin naming used for compatibility with generic symbols.

---

https://www.fujitsu.com/downloads/MICRO/fcai/relays/ftr-ly.pdf

2019-03-01_174411_714x325_screenshot
2019-03-01_174957_812x359_screenshot
2019-03-01_175222_583x188_screenshot


poeschlr:
@KiCad/librarians i need your input.

---

Looks good. Only trouble is we don't have a generic relay symbol right now.

I wonder if we should start a new relay footprint lib specifically for holding parts compatible with the generic pin numbering allowing us to easily add a filter to a newly added generic symbol.

Or we make a specialized symbol for now and switch to generic once we start with development for v6.

I am kind of torn between these two. (Both have their ups and downs.)

---

We might further also use a similar naming approach as we do with switches which would mean exchanging SPST -> 1P1T.

Misca1234:
@poeschlr 
I vote for waiting for v6

@johnbeard 
You have stuff in the F.Silk layer that is beneath the component when the component is mounted. They are kinda hard to see, perhaps they should be removed, they are kinda useless?

poeschlr:
Ok in this case we would then need a fully specified symbol for the time being.

PR #1453, opened 2019-03-02
add 3 stackable headers used in Arduino Uno

forrestbao:
Stackable headers are frequently used in Arduino Uno. But KiCad does not have a footprint nor 3D model for it. So I created the versions for 6-pin, 8-pin, and 10-pin to use by myself. I now wanna share them with the community. For a centralized location of footprints, 3D models, and 3D model source file, see this: https://github.com/forrestbao/pebble/tree/master/KiCad/Stackable

---

* [6-pin, 4UCon 20516](http://www.4uconnector.com/online/object/4udrawing/20516.pdf)
* [8-pin, 4UCon 20517](http://www.4uconnector.com/online/object/4udrawing/20517.pdf) 
* [10-pin, 4UCon 20518](http://www.4uconnector.com/online/object/4udrawing/20518.pdf)
3D top
3D model PR https://github.com/KiCad/kicad-packages3D/pull/534
No symbol PRs as it can use any male or female connector's symbol. 
1. I created a new library called Connector_Stackable for all stackable connectors as they are gaining popularity because of their wide use in shield boards. However, this makes my PR cannot pass fp-lib-table test. Let me know whether I must put my new parts in an existing library. 
2. 3D models are in STEP format because [KiCad 5.0 is in favor of STEP (a better CAD exchange format) over WRL (dimensionless)](https://forum.kicad.info/t/scaling-issues-exporting-vrml/13941). But testing rules still require WRL.


poeschlr:
These are just standard pin sockets for which we already have footprints in the Connector_PinSocket library. I know the 3d model is different here but if we create a footprint/3d model pair for every such connector then we end up with way too many footprints to handle. This is related to https://github.com/KiCad/kicad-footprints/issues/130

A better way forward might be to create generic generator scripts for both footprints and 3d models and ship these in addition to the standard footprints. That way users can create the footprint/model pair they need on demand but can also use the generic version if they do not care about that level of accuracy.

myfreescalewebpage:
@poeschlr I suggest this pull request should be tagged Invalid and closed, what do you think ?

poeschlr:
Yes i think we sadly really can not include specialized footprints for every such component. That would make the library unwieldy. I am truly sorry.

PR #1456, opened 2019-03-03
add SMA jack Amphenol 132203-12

lorem-ipsum:
Based of the existing `SMA_Molex_73521-2200_Horizontal`.
[Datasheet](https://www.amphenolrf.com/media/downloads/1769/132203-12.pdf)
screenshot_20190303_213822


aewallin:
I have reviewed the relevant dimensions from the datasheet - looks OK.
KiCad 5.1 seems to have a nice new snap-to-endpoint feature for the dimension tool.
sma_amp_review

Possibly remaining questions for librarian before merge:

- the length of the non-threaded part of the stick-out is I guess just estimated from the datasheet?
- with 2.25mm pads the annular rings are a bit different width, since the center hole is 1.5mm and the GND-holes are 1.7mm - but the datasheet doesn't give a recommended pad-size, so I guess this is ok.

lorem-ipsum:
Yes, the non-threaded stick out is estimated. It is sadly not specified, so I chose something which looks about right. I wasn't sure about the annular rings, if you have any recommendations for the size I gladly change it.

poeschlr:
Looks quite good, the threaded section is not 1/4" however.

Screenshot from 2019-06-02 21-28-00

lorem-ipsum:
I changed the dimension to 1/4" and also tried to match the 36 thread increase.

poeschlr:
Thanks

PR #1457, opened 2019-03-03
Create Linear MSOP-12-16 for LT3748

myfreescalewebpage:
Creation of MSOP-12-16 3x4mm for LT3748, copied from Linear_MSOP-12-16-1EP_3x4mm_P0.5mm and removed the expose pad.

File comparison between "Linear_MSOP-12-16-1EP_3x4mm_P0.5mm.kicad_mod" and "Linear_MSOP-12-16_3x4mm_P0.5mm.kicad_mod":

capture

---

https://www.analog.com/media/en/technical-documentation/data-sheets/3748fb.pdf

capture

https://github.com/KiCad/kicad-symbols/pull/1600

No errors at this time


evanshultz:
If you compare the footprint in this datasheet with the one in the existing footprint (https://www.analog.com/media/en/technical-documentation/data-sheets/3630fd.pdf), you will see that the outer edges of the pads is 5.1mm min whereas the datasheet above says 5.23mm min. The footprint is right at 5.1mm.

However, all the package dimensions are the same so this footprint is likely to be fine. So I'll merge after I hear back from you if you want to make any changes or go with it.

myfreescalewebpage:
@evanshultz thanks for the review, I haven't seen this details before, I have only compare the packages dimensions themselves, which are exactly the same, and I agree the difference is acceptable here.

evanshultz:
On the assumption that this will be scripted in the future, which requires updating the script to allow "missing" pins like on this package, any issues should go away anyway.

myfreescalewebpage:
An optional list with missing pins in the config file of the script and it's can be easily achieved I think :)

evanshultz:
Yeah. I believe some script already has this but it's straightforward to imagine how to do it if not.

PR #1459, opened 2019-03-04
Fix up footprint for TUSB7340

evanshultz:
Split off from https://github.com/KiCad/kicad-footprints/pull/981. Leaving old footprints to be removed there so when it's merged they'll get cleaned up.

As requested at https://github.com/KiCad/kicad-symbols/pull/1394.

Rename of `Package_DFN_QFN/Texas_WQFN-MR-100_3x3-DapStencil` and `Package_DFN_QFN/Texas_WQFN-MR-100_ThermalVias_3x3-DapStencil` with updated silk pin 1 mark and merged multiple adjacent pads into a single one.

image

---


myfreescalewebpage:
@evanshultz  thanks for separating the PR. Looks great, but some comments at this moment:
- [ ] That's a lot of paste on pin A53, maybe 4 small squares like done on other footprint is better ?
- [ ] The CrtYd rectangle is very large, isn't wanted and why ?
- [ ] The Axx pins are 0.25x0.55mm in the datasheet, 0.25x0.6mm in your design
Cheers,
Joel

evanshultz:
I only renamed the symbols and merged pads, and specified in #981 and above. There is no change from the footprint as it exists in the library currently. I didn't plan to do anything more but let me take a look since these seem relatively easy (the last one might be TI changing pad sizes over time...).

myfreescalewebpage:
> I only renamed the symbols and merged pads

What was the old names ? The git diff do not show it...

evanshultz:
See my comment in the first post. All answers are there, I believe.

I found that the TUSB7340 datasheet includes the same package drawing and footprint design as the two separate docs used already, so I changed the documentation a bit.

I believe 0.25x0.6mm pads are correct. See this:
image

Updated courtyard to 4.95x4.95mm and split up the paste pad into 9 separate 1.5x1.5mm pads (67% paste coverage and just touches the vias).

Anything else?

myfreescalewebpage:
@evanshultz thanks, I think it's ok. Only one remaining question, I have seen your comment about the original footprint renamed, but I was pointing that the git diff should show that you have removed the original ones. Or maybe you want to keep the old footprints ?

evanshultz:
I remember writing this somewhere but don't see it above. Maybe it is buried in a commit message?

Anyway, #981 will delete the old footprints and so I'm leaving the removal to stay in the old PR.

myfreescalewebpage:
@evanshultz the two old footprints are not currently removed by #981 ... I invite you to achieve this, because a conflict will occur as soon as I will merge this PR.

evanshultz:
Git knows the old footprints were really renamed, or effectively removed once that PR is merged, and I resolved the branch conflicts so it should work out.

Thanks for reviewing and merging this.

PR #1460, opened 2019-03-04
Package_DFN_QFN: Add Texas R-PVQFN-N24

antoniovazquezblanco:
http://www.ti.com/lit/ds/symlink/bq24133.pdf

imagen

imagen

---


antoniovazquezblanco:
I will create a sym PR soon.

evanshultz:
The generator script was used to make this footprint. Are you submitting a PR for the script? Or did you use the script for the basic design and then modify it? If we're calling this `Texas_` then I think sticking exactly to the TI footprint is the best idea. And this one isn't. What do you think?

antoniovazquezblanco:
> The generator script was used to make this footprint. Are you submitting a PR for the script? Or did you use the script for the basic design and then modify it?

I've tryed to use the scripts but it would have required extensive modifications so I continued manually.

> If we're calling this Texas_ then I think sticking exactly to the TI footprint is the best idea. And this one isn't. What do you think?

I will look at what is different.

Thanks!

antoniovazquezblanco:
Fix pad sizes. Is there anything else?

Thanks!

antoniovazquezblanco:
Ping @evanshultz

evanshultz:
Sorry for the long delay.
1. Let's call the footprint `Texas_RGY` since we know that's the most clear package name from TI.
1. To reiterate, since this is a TI-specific package it should match the datasheet. Pad size, pad location, pad shape, thermal vias design, all stencil pads, and maybe other things don't match the datasheet.
1. The footprint doesn't have swelled mask as shown in the datasheet.

antoniovazquezblanco:
Done!

imagen

imagen

poeschlr:
The pin count must still be included in the footprint name. I therefore suggest to use the full string as used by TI.

evanshultz:
@poeschlr 
Why is that? For CvPcb? Doesn't it match up symbol pin count by footprint pin count? We have quite a large number of footprints without pin count in the footprint name so I hope this isn't a major problem or a new requirement. I'm sure there's a reason but I'm missing it right now.

If by "full string" you mean "R-PVQFN-N24", it has been extensively documented that this does not reliably differentiate TI packages. Even if you want to include that text, "RGY" has to be included as well as it is the proper package designation by TI.

poeschlr:
No user will know intuitively from these 3 letters what to expect. So do me a favour and include this info. The 3 letters can be there in addition to that.

I do not particularly care if you add the 3 letters in front or the back (TI has them in front so that would be kind of natural)

antoniovazquezblanco:
Should we go `Texas_RGY_R-PVQFN-N24` and `Texas_RGY_R-PVQFN-N24_ThermalVias`?

poeschlr:
add the size definition of the thermal pad and i am happy.

antoniovazquezblanco:
Then:
`Texas_RGY_R-PVQFN-N24_EP2.05x3.1mm`
`Texas_RGY_R-PVQFN-N24_EP2.05x3.1mm_ThermalVias`

Could be?

poeschlr:
Yes that would be ok to me.

antoniovazquezblanco:
Done!

evanshultz:
Review notes:
1. The 3D model needs to be updated to the new footprint name.
1. The via annular rings can be increased to meet KLC without any detriment.
1. There is no bottom copper. Perhaps copy the top copper as our script does?
1. Custom pad shapes need the zone connection set to 'Solid'. With these partial-custom pads I'm not sure what can happen in all permutations of zone connection so perhaps it's worth a quick check?
1. I think it would result in a smaller file to set the overall paste and mask clearance instead of on each outer pad. On the center pad the paste later is turned off and the mask clearance should still be OK.

evanshultz:
image

evanshultz:
Thanks!

The B.Cu copper pad on the via one shouldn't have mask turned on. Otherwise I think it's good.

antoniovazquezblanco:
Done!

evanshultz:
Thanks!

antoniovazquezblanco:
Thank you!

PR #1464, opened 2019-03-05
Added Omega2S Footprint

fandres77:
Added Foorptint Onion2S

---

Datasheet: 
https://github.com/OnionIoT/Omega2/raw/master/Documents/Omega2S%20Datasheet.pdf
screenshot_20190305_140239

Library: https://github.com/KiCad/kicad-symbols/pull/1599


myfreescalewebpage:
Hi @fandres , thanks for contributing,

- [ ] The symbol should be centered on pin 1

I have other comments, but the above one will move every thing so we should first start here.
Question : do you know what is the square below the module ?

Cheers,
Joel

fandres:
it seems that it is for refrigeration

fandres:
Oh this you mean: The symbol should be centered on pin 1
Screenshot_20190308_220804
because it breaks the KLC rule F6.2.

poeschlr:
Why do you think it should be centered on pin1? This would only be true for through hole components. Is it possible you set this footprints attribute to trough hole?

See: http://kicad-pcb.org/libraries/klc/F6.1/

fandres:
Hi @myfreescalewebpage , I think I'm not understanding what you really want
The type of footprint placement is established in surface mounting for SMD footprints
Screenshot_20190309_105748

Could you tell me what do you mean by: The symbol should be centered on pin 1 ??

myfreescalewebpage:
oooops you are right, that's only for through hole devices ! (KLC F7.2)

myfreescalewebpage:
My comments from now:
- [ ] The anchor is not vertically centered on the device body
- [ ] The silkscreen clearance is too large, only 0.12mm from the device body is enough
- [ ] The CrtYd layer clearance is too large between pin 1 and pin 63, only 0.25mm is enough (it's okay on the three other sides)
- [ ] The Dwgs.User square is not useful
- [ ] Seems the center pad is an expose pad for dissipation as suggested above with your photo, but I'm not sure it is to be connected to the end board. Do you think it is ? If yes, we should set this pin number to 64, add it to the symbol stacked with GND
- [ ] If yes to above question, the center pad is too large to have so much Paste. You can draw a single Pad of 6x6mm with no paste and 4 pad with paste only. If you do not see what I mean, please ask or have a look at "QFN-64-1EP_8x8mm_P0.4mm_EP6.5x6.5mm" package for example.

fandres:
Hello Joel, in fact the central pad has continuity with GND
I have little free time these days ... as soon as I can fix it

fandres:
Hey @myfreescalewebpage the [KLC rule F5.1](http://kicad-pcb.org/libraries/klc/F5.1/) indicates that:
"Clearance between silkscreen and exposed copper elements must be at least 0.2mm" not 0.12mm. or I'm not understanding you.

Screenshot_20190313_163632

myfreescalewebpage:
I have troubles about the clearances, it was 0.12mm and 0.25mm in the past, don"t understand if the KLC has been changed maybe... @poeschlr does KLC F5.1 has been changed recently ?

@fandres position of anchor not fixed.

myfreescalewebpage:
@fandres finally retrieved ! @evanshultz previously written (https://github.com/KiCad/kicad-footprints/pull/1050#issuecomment-434022340):

> http://kicad-pcb.org/libraries/klc/ are the official library rules and http://kicad-pcb.org/libraries/klc/F5.1/, clause 2, specifically says 0.2mm. However, this hasn't been updated and we've been allowing IPC 7351C draft recommendation of the line width (so just 0.12mm). Either one is an acceptable silk-pad clearance.

0.12mm is IPC-7351C recommendation that is currently followed even if 0.2mm is ask in the KLC. I will submit the issue in the KLC repo right now.

Joel

fandres:
ok, 0.12 mm and 0.2 mm are accepted. I already set it to 0.12mm

fandres:
Joel the anchor I have not moved him because of this:
_Checking footprint 'Onion_Omega2S':
  Violating F6.2
    For surface-mount devices, footprint anchor is placed in the middle of the footprint (IPC-7351).
    Footprint anchor is not located at center of footprint
     - Footprint center calculated as (0.0,0.78)mm_

fandres:
@myfreescalewebpage update with:  

> Footprint center calculated as (0.0,0.78)mm >

myfreescalewebpage:
@fandres the clearance is to be taken between the center of the line, as shown on the following schema (red is incorrect measurement, the right one is the blue) Thus, your clearance is too much important. You simply have to take X/Y coordinate of the Fab line and add/remove 0.12mm to get X/Y coordinate of the Silkscreen line.

Capture

Let see after that about the center, but it is strange because X=-18.04 for the top line and X=15.96 for the bottom line...

fandres:
@myfreescalewebpage  if it is from center to center is as follows, Figure. Are you sure? basically there is no separation between F.SilkS and F.Fab
Screenshot_20190316_131904
If it's strange, I initially designed it centered but then the script made me move it. Is it possible that the script is wrong?

myfreescalewebpage:
I'm sure.
And the clearance of the CrtYd layer is 0.25mm center to center.
Joel

myfreescalewebpage:
@fandres thanks for the fixes. Two adjustments to do on CrtYd layer:
- bottom line is not correct, Y coordinates should be 17.74 (0.25mm from the bottom pads)
- top line is not correct, Y coordinates should be -17.51mm (0.25mm from the Fab layer)

fandres:
Ok, Joel, I still have to learn.

myfreescalewebpage:
No worries, we do it step by step :)

Last errors:
- [ ] The %R text is under pin 64 and it is not visible without hidding the F.Cut/F.Paste/F.Mask. Can you move it somewhere else int he symbol ? Screenshot below.

Capture

- [ ] The footprint is not vertically centered, as shown below.



(this should be both 17mm)

You should center the footprint using the anchor tool in Kicad:

a

myfreescalewebpage:
Not yet valid for centering the symbol, but ok for the %R text.

fandres:
@myfreescalewebpage, is centered between the CrtYd layers
Screenshot_20190325_105735

myfreescalewebpage:
@fandres I have a doubt reading your previous comment, because you are right, you have centered regarding to the CrtYd layer. I have checked other devices in the lib, the footprint should be centered looking at the **Fab** layer.
Due to the specific positions of the pins on the footprint, travis will still complain about the center but it is normal.
Joel

evanshultz:
`%R` can stay under pads since the assembly print will only show the fab layer. It's OK where it is, too, just FYI.

myfreescalewebpage:
ok @evanshultz but it was not visible at all in the footprint designer, that's why I ha requested moving it.

myfreescalewebpage:
@fandres77 great job ! Travis complain bu it is normal here.
No more comment to do, merging :)

fandres77:
Ok, Great, @myfreescalewebpage , thank you for your advice

PR #1465, opened 2019-03-05
Create WE 750343373

myfreescalewebpage:
WE Flyback Transformer 750343373

---

https://katalog.we-online.com/ctm/datasheet/750343373.pdf

foot

https://github.com/KiCad/kicad-symbols/pull/1608

No error at this moment


evanshultz:
1. Can you please beef up the description?
1. The annular rings are a bit small for such large pins. Maybe at least 1mm over the drill size?
1. Missing pin 1 mark on fab and silk layers.
1. I think not mentioning transformers at http://kicad-pcb.org/libraries/klc/ is a mistake. 0.25mm is too small for a part so big in all three dimensions.

myfreescalewebpage:
Hi @evanshultz thanks for the review!
About your points:
1- fixed, added description already used on symbol
2- right, don't known if we have common rules, increased to 1mm over drill size
3- added
4- I agree too, maybe some rules should be added depending of the size of the part
Joel

evanshultz:
1. For a footprint I think the description should be physical characteristics like size. The info there now doesn't necessarily affect anything about the footprint.
2. See https://github.com/KiCad/kicad-footprints/issues/595 for some info. Not cannonized.
3. Looks like this now:
image

I just found one thing and that's all to do before merging: Replace `WE` with `Wuerth`. That's the manufacturer name we used.

myfreescalewebpage:
Renamed :)

evanshultz:
Thanks! Could you adjust the description so it refers to footprint stuff and not electrical info which isn't related to the physical component?

myfreescalewebpage:
Well I do not know what I can write, have you got suggestion ? Is there a common format we have for transformer ?

evanshultz:
Ha! You're full of jokes today! Just teasing. There are no guidelines for transformer footprint descriptions AFAIK. Really, all footprint descriptions are pretty wild and uncontrolled.

Well... how about `Transformer, horizontal core with bobbin, 10 pin, 3.81mm pitch, 15.24mm row spacing, 22x23x17.53mm (<url>)`? Is that descriptive of the footprint (and joined 3D model)? I'm thinking of some transformer footprints I will be submitting and what terms and needed to describe them.

myfreescalewebpage:
Not pad ! I'm confused when this is footprints. Every thing is simple with symbol but not so much clear for me with the footprints :) Modification done.

evanshultz:
Agreed. This is an area that could probably use some rough guidelines to help steer things in a useful direction without getting too detailed and overwrought.

Thanks!

PR #1466, opened 2019-03-05
Add Texas RNN0018A for TPS568215

myfreescalewebpage:
Add Texas RNN0018A for TPS568215

---

http://www.ti.com/lit/ds/symlink/tps568215.pdf

footprint

https://github.com/KiCad/kicad-symbols/pull/1610

Not sure the 0.05mm margin required in the datasheet is properly set....


evanshultz:
1. Put the fab ref des under the part body.
1. Why is the clearance set per pad? If it needs to be done, do it for the whole footprint. But I don't see a reason why this needs to be set at all.
1. Pin 1 fab layer bevel is too small. http://kicad-pcb.org/libraries/klc/F5.2/. Use 0.875mm (though 0.5mm would probably look OK and I might not have noticed it).
1. Pull the silk in closer to the fab lines.
1. Mask should be positive, not negative, as the preferred NSMD footprint design has the copper smaller than the mask. See page 34 of the datasheet. Then make sure it's really 0.05mm bigger than the copper.
1. Paste clearance should not be defined for the whole footprint. Set it to 0 and then turn off the paste layer for pads 2, 6, 7, and 11. Make custom paste pads for those pads. See page 35 of the datasheet.
1. Can you rotate any pads that need it so they all have a rotation of 0?

Dimensioned drawing (it looks messy but everything is correct):
image

evanshultz:
8. Perhaps remove the silk in the corner closest to pin 1. Right now the removed silk is closest to pin 18.

myfreescalewebpage:
Hi @evanshultz thanks for the review,
About the list of points:
1- ok, done
2- because of the following detail page 34 of the datasheet, but maybe I'm wrong ?
Capture
3- 25% is large and look strange, but why not, done
4- done
5- this question seems to be linked to point 2 but I'm not sure of the right settings to do. I have modified only pin 1, can you indicate me if this one is correct, else, can you detail ? If it is correct I will fix the other pins of course.
6- custom pads made, but not sure I have understood every thing, again if you can guide me to the remaining modification to do, that's very nice :)
7- done, all pads are now 0 or 180°
8- done
Thanks,
Joel

PS : how do you draw the 2D dimensions of the footprint like the one you have done in your previous comment ?

evanshultz:
1. Thanks. Shrink the size a bit to fit inside the part outline.
2. I'm not sure what happened with my words. You have now set mask oversize for the entire footprint, which is good, but use a positive number instead of a negative one to make the mask bigger than the pads. And remove the individual setting from pad 1 since this applies to all pads. No individual pads should need any paste or mask adjustments, nor should the footprint as a whole need any paste adjustment.
3. Yeah. It doesn't always look pretty to me either.
5. I think this is covered above. Hopefully in an understandable way this time.
6. Yeah, you got the general idea. Paste pads for 2 and 11 should be 0.36mm tall. Turn off the paste layer for copper pads which has a unique paste layer. And make sure the pads pads only have the paste layer turned on. And then please rotate paste pads so they're not at 90 degrees. Lastly, check all of them to be sure paste pads have 0.05mm corner radius (a few copper pads need this adjusted too).

Rene wrote up how to dimension footprints at https://forum.kicad.info/t/how-to-check-footprint-correctness/9279. I just followed his instructions.

myfreescalewebpage:
1- done
2- I have made some modifications, but I'm not sure.
3- ok
4- above
5- Paste pads for pin 2 and 11 are clearly 0.35mm in the datasheet. Fixed other points but not sure it's totally correct...
Sorry i'm not used to create/review this kind of symbol so I'm learning on this...
Joel

evanshultz:
2. Super close! Set mask clearance to `0.05` for the entire footprint. Adjust the corner radius of copper pads 2-5 and 8-11. The paste-only pads need their corner radius corrected (all 8 of them). Turn off the `F.Cu` layer for the paste pads (set the `Copper` pulldown to `None`):
image
5. They should be 0.36mm tall. The datasheet is very dense. See this from page 35:
image

Between workarounds in KiCad and the craziness of some footprints we're all learning. :)

myfreescalewebpage:
> Between workarounds in KiCad and the craziness of some footprints we're all learning. :)

Yeahhh!!!

I think I have address point 2. Let me known. I have looked a lot between global and pin settings and understood several thinks now :)

About point 3, I do not agree, the datasheet indicate 0.35mm as shown below for the red lines, which are the solder paste dimensions. Blue lines (0.36mm) are the dimensions of the expose metal according to my understanding of the datasheet (also clearly indicated for pins 6 and 7).

Capture

Joel

evanshultz:
0.35mm is the distance from the drawing center line to the center of the pad. 0.36mm is the paste pad height.

myfreescalewebpage:
Ouuuuups, read this several times, and each time I have made the error. Agree with you. Fixed now.

evanshultz:
Great!

Paste pads for pads 2 and 11 can be adjusted a bit to get closer to 0.05mm corner radius. Can you please fix that? Then I think it's ready.

myfreescalewebpage:
Sure, that's a percentage, I have forgotten to adjust it when resizing the pad !

evanshultz:
Looks great. This was a tricky footprint. Thank you!

myfreescalewebpage:
Thanks for your help @evanshultz !!!

PR #1467, opened 2019-03-06
Fix Tensility 54-00023

gcormier:
This part is not a Micro USB port, but a mini port rather

1. Change filenames to Mini
2. Change module to Mini
3. Change 3d footprint names (there is no 3D model in any case) to Mini


evanshultz:
@gcormier 
You are correct. Thanks for noticing this and submitting a fix.

Please see the CI notes. There is still a bit more renaming to be done in these files.

Since Pcbnew sucks in footprints, and the user can update later if they like through CvPcb, I don't see a problem to merge this right away. But since 5.1 is just a week away, barring a major bug, let's hold off a few more days. This footprint has been in the lib a while so it doesn't seem like it must be done now.

gcormier:
Sounds good! Fixes are in.

evanshultz:
@gcormier 
Thanks for this. I appreciate your contributions recently and look forward to more!

PR #1468, opened 2019-03-06
Added USB_A SS-52100-001 connector

Duckle29:
I was working on a project and noticed that there wasn't any female USB A connectors other than one stacked double connector.

I added one based on the Stewart SS-52100-001 connector.

---

[Datasheet](https://belfuse.com/resources/drawings/stewartconnector/dr-stw-ss-52100-001.pdf)

image

image

image


diegoherranz:
Sorry for the delay looking at this. I'll get at it soon. Thanks for the patience.

Duckle29:
No problem :) The footprint has been successfully used here: https://github.com/Duckle29/ATX-Power-breakout

(well it turns out it's blocked by the rpi it plugs onto, but that's just a pcb design issue :P)

diegoherranz:
Thanks for your contribution. It all basically looks pretty good except for this tiny picky thing:
- [ ] The clearance on the Courtyard is 0.5 mm (correct) all around because it is a connector, but it is a bit too much near the shield pins
clearance

- [ ] Also, taking advantage of the update, could you add "receptacle" to the keywords?

Things which I've checked and look OK:
- Holes sizes and locations match datasheet
- Personally I think I would have made the annular rings a bit more generous, but it's OK as is.
- Lines width OK: Fab 0.1 mm, courtyard 0.05 mm, silkscreen 0.12 mm
- Pin 5 used for shield which matches the symbol in the library
- Fab matches datasheet dimensions. Flange thickness not defined in datasheet but thickness used looks reasonable
- Silkscreen matches Fab (OK for THT) but avoiding pads - OK

If you fix those tiny things, I'll merge it. Thanks again!

diegoherranz:
Thanks for the changes.
The courtyard is better now. It is still a bit more than 0.5mm clearance around the shield pins but I think it is close enough now.
I've seen you've increased the annular rings size for the main 4 pins. It wasn't really necessary but it's OK like this too.

I'll merge it. Thanks for your contribution!

PR #1469, opened 2019-03-07
Battery: Add the 6.8mm Keystone 2998 coin cell holder

alistair23:
The datasheet is available here: www.keyelco.com/product-pdf.cfm?p=763

Signed-off-by: Alistair Francis <alistair@alistair23.me>

---
www.keyelco.com/product-pdf.cfm?p=763

keystone-2998-6 8mm-footprint

There is no matching symbol

There are no violations


alistair23:
Here are some pictures of what it looks like on a produced board.

WP_20190528_09_04_52_Rich_LI
WP_20190528_09_05_19_Rich_LI
WP_20190528_09_05_28_Rich_LI

poeschlr:
Outline on the fab layer is missing (can not really fully check the contribution as the fab outline is a major part of it.)

Not a single dimension of the land pattern agrees with the datasheet you provided.
Screenshot from 2019-06-02 22-35-25

alistair23:
Thanks for reviewing, I have pushed a v2.
KiCAD-Uploade

alistair23:
Ping!

poeschlr:
Courtyard clearance should be 0.25mm all around (relative to nominal body size and pads.)

And please double check your measurements before pinging us for a review. I checked the pads and they are still not agreeing with the datasheet (even the central pad does not and that one is really easy to check even from within kicad).

alistair23:
Thanks for looking at this. Sorry it took so long for me to reply and update the PR.

I have increased the courtyard clearance to 0.25mm.

As for the pad sizes, I double checked them and they look fine to me. I have rounded up (by a little bit) from the minimum specified in the datasheet. As the datasheet has a error margin of +/- 0.15mm I think this seems like an ok thing to do. Especially as it makes the numbers nice and round :)

If you want the pad size to be to be exactly the minimum in the datasheet I can update that.

The latest picture is below:
Picture

poeschlr:
Rounding of dimensions is not allowed. If the datasheet gives you 20 significant digits i expect you to use them. We simply guarantee to our users that every supplied footprint is exactly as specified by either the datasheet or an industry standard.

So either use exactly what the datasheet specifies or use an industry standard and calculate the pad dimensions from there (The easy route is the datasheet as the other one will require quite a bit of math)

alistair23:
Ok. I have updated the footprint. This is the new picture:

Pic

alistair23:
Another small update

Pic

alistair23:
Sorry for the noise, updated to pass the checks.

Pic

poeschlr:
Thanks

PR #1472, opened 2019-03-09
Add Murata 1200RS inductor footprint

a-wai:
Murata 1200RS series inductors
Datasheet can be found at https://www.murata-ps.com/data/magnetics/kmp_1200r.pdf

Note: as it is a non-polarized device, only round pads are used.

murata_1200rs

---


gmsotavio:
Hi,

I noticed that "MuRATA" is bad capitalized. The correct capitalization would be "muRata", but most of footprints from current KiCad libraries uses the "Murata" capitalization.

gmsotavio:
I'm think the "MuRATA" capitalization" is wrong, according with http://kicad-pcb.org/libraries/klc/G1.6/. Similar library footprints are using this wrong capitalization. It should be "muRata" instead, I believe. Maybe an librarian can help us. @poeschlr @evanshultz @myfreescalewebpage do you agree?

myfreescalewebpage:
@gmsotavio you have totaly reason. @a-wai the name of this manufacturer in the library should be spelled `Murata`.

a-wai:
Fixed, thanks for the review.
I can make another PR to fix existing components in the library too, if you want.

evanshultz:
@gmsotavio 
Good catch! Thanks for pointing this out.

@a-wai 
Let's not touch other footprints. This will break existing footprint associations. Doing this as v6 nears release will be good, but that's well in the future.

1. Since the inductor winding pins are 4x5mm, let's change the pitch to 6.4mm.
1. Please update the description to `Inductor, Radial, Pitch=6.4mm, Diameter=10.5mm, Murata 1200RS, http://www.murata-ps.com/data/magnetics/kmp_1200rs.pdf`.
1. Remove the double space in the keywords. They can be simplified too, if you wish, to reduce duplication with the description. Keywords can be very brief.

Lastly, please don't force push so it's easier to review. I'll squash merge.

evanshultz:
@a-wai 
So let's use the XxY pitch scheme suggested above. Are you able to make these changes?

a-wai:
@evanshultz Sure, I'm on it

evanshultz:
Thanks! This looks ready to merge.

FYI, we don't have defined guidelines for accuracy so `x.00` is OK. (We could copy this from IPC but it's undefined so far.)

PR #1477, opened 2019-03-12
added FP for https://www.johansontechnology.com/datasheets/2450AT43B1…

cyber-murmel:
Footprint for 2.4 GHz WLAN, Home RF, Bluetooth, 802.11 b/g, Antenna
---

https://www.johansontechnology.com/datasheets/2450AT43B100/2450AT43B100.pdf


poeschlr:
Your pull request title is less than ideal. For the future do not include the datasheet in the title but a clear meaningful description of what you add. (I updated the title to be more meaningful.)

cyber-murmel:
> Your pull request title is less than ideal. For the future do not include the datasheet in the title but a clear meaningful description of what you add. (I updated the title to be more meaningful.)

Ah, sorry. I put the datasheet link into the commit message and didn't change it when I opened the PR.

poeschlr:
We have detailed rules on how the keepout area should be documented please align your contribution to it: http://kicad-pcb.org/libraries/klc/F4.5/

The courtyard layer must contain a single closed polygon that tells kicad where no other component is allowed to be placed. This is checked by DRC since version 5. The DRC check relies on it being drawn correctly.

Silk is definetly too close to the pads at places. The pin 1 mark is also hard to identify. I have two suggestions in the screenshot provided. Green might be better suited for manual assembly as it guides the exact placement of the antenna. Pink would however fit the current industry standards and is also what we use for other polarized components.
antenna

Rotate the footprint such that pin 1 is on the left. (Zero orientation type A according to IPC)

The endpads are too far inwards. (pink dimension should be 1.6) The orange dimension would be partially fixed by fixing that but you will still need to move the courtyard/keepout area.

I am unsure about the dimension marked in blue. It is given differently in the datasheet but i am not sure if the datasheet suggestion would make any sense at all. (The component is symmetric why would they suggest an asymmetric footprint.)
Screenshot from 2019-03-29 22-03-38

PR #1483, opened 2019-03-16
Add monolithic power QFN-20 3x5mm

HubertD:
This adds a footprint for what MonolithicPower simply calls "QFN-20", used for their integrated switching regulators such as e.g. the MPM3620(A) and the MPM3630
 
https://www.monolithicpower.com/pub/media/document/MPM3620_r1.1.pdf#page=24

MPS_QFN-20_3x5mm

planning to do a 3D model, but not there yet

No violation intended, but:
As clearly visible in the screenshot, Pins 4-5-6 and 7-8-9 form a common pad.
Since these triples are used everywhere in the datasheet, I decided to create on larger pad (pins 5 and 8) and adjoining pads (4/6, 7/9).
Not sure if that is the best way to do it.


HubertD:
Travis whines about the footprint not being centered, violating KLC F6.2.
In fact, it is perfectly centered (package center is at 0/0, see F.Fab and F.CrtYd layers).
Warning is triggered because of the non-symmetric pad layout, e.g. there are no pads at the bottom despite MPS calling it Q-FN :-)

HubertD:
added a 3d model PR: https://github.com/KiCad/kicad-packages3D-source/pull/214

poeschlr:
The courtyard clearance is not quite right (Unless i missed a note in the datasheet requesting it that way).

I would assume there is a way to have the middle pad of the connected pads as a rounded rectangle (Easiest way would be to use overlapping pads instead of a polygon pad. This would also allow the use of thermal spoke connection instead of solid.)

I am not so sure the benefit of having the first pad chamfered outways the downsides of not being able to use thermal connections or an easy way to include rounded corners. Might i suggest the use of a normal rounded rectangle here? (There is no technical reason to have this chamfer in the footprint. I think it is suggested for visual identification of pin 1 if no silk is included.)

Screenshot from 2019-03-25 21-02-58

poeschlr:
Could you also add an explanation on the Cmts.user layer explaining that the keepout area is for top layer copper? Something like "KEEPOUT no copper directly below part"

PR #1490, opened 2019-03-19
Add footprint for E73-2G4M08S1C BLE5 module (with ceramic antenna)

GitWellBack:
Datasheet: ([http://www.ebyte.com/en/pdf-down.aspx?id=627](http://www.ebyte.com/en/pdf-down.aspx?id=627)) (page 10 for C variant)

Symbol PR: #1654

Screenshot: (internal pads are numbered but don't show at this zoom level)
E73-2G4M08S1C

Credit: This new component is one I am using in several projects and needed the footprint - this footprint is inspired by the work of @isundaylee for the E73-2G4M04S module.

KLC violations:

**F6.2** is violated to keep the module horizontally centred on edge pads (5 and 35), which are the closest pads to the centreline but it aids in track routing on a standard pitch.
**F9.3** is violated as there is currently no 3dshape for this module. The plan is to add one soon - just getting to grips with FreeCAD 0.18!

---


fvollmer:
Hi,
just a quick comment regarding **[F9.3](http://kicad-pcb.org/libraries/klc/F9.3/)**: 

> All non-virtual footprints **must have** 3D model references, **even if the 3D model is missing** (does not yet exist). This allows the 3D model to be added later without requiring the footprint to be edited again. (It is not required to supply a 3d model file.)

GitWellBack:
> Hi,
> just a quick comment regarding **[F9.3](http://kicad-pcb.org/libraries/klc/F9.3/)**:
> 
> > All non-virtual footprints **must have** 3D model references, **even if the 3D model is missing** (does not yet exist).

Fair point - just wasn't sure how to go about it. I'll just commit a change in the next hour or teo and then make the model later today (I was considering doing the components on the board under the shield but that seems like mostly a vanity trip, so the rest should take a couple of hours). Thanks for the rule reminder.

GitWellBack:
https://github.com/KiCad/kicad-packages3D/pull/537 is present for 3dshape.
Footprint has been moved to RF_Bluetooth (as has 3dshape), by request in Symbol PR (https://github.com/KiCad/kicad-symbols/pull/1654#issuecomment-480622856)

GitWellBack:
Is there anything in particular needed to have this merged? The symbol is waiting for this to be merged before it can also be added. Been a while.

poeschlr:
How did you determine the land pattern for this? The datasheet does not provide a suggestion. If there is no suggestion by the manufacturer we will need to use IPC standards. The datasheet is however missing information required even for doing this (we would need to know the width of the castelated contacts at the very least. Including tolerance information meaning simply measuring a component will not do.)

Misca1234:
The model is placed in 
RF_Bluetooth.3dshapes

Is that ok  (to create a new folder) or should it go into RF_Module.3dshapes ?

G-Pereira:
The reason for choosing that library is discussed in the symbol PR https://github.com/KiCad/kicad-symbols/pull/1654#issuecomment-480622856

G-Pereira:
In my opinion this module is complete and this PR is blocking everything else. I am able to help in whatever it needs to merge this. I couldn't get the footprint anchor centered, but is that the reason this hasn't been merged yet?

poeschlr:
I requested a datasheet that we can use to check this contribution. It has not been provided yet so how do you expect us to review this contribution?

G-Pereira:
Here they are
[E73 Module Datasheet](https://github.com/KiCad/kicad-footprints/files/4162474/E73-2G4M08S1C_Usermanual_v1.8.pdf)
[nRF52840 Datasheet](https://github.com/KiCad/kicad-footprints/files/4162476/nRF52840_PS_v1.1.pdf)

poeschlr:
The question i asked above still stands as there is no suggested footprint in the datasheet you provided (same as with the original datasheet): How were the pad sizes determined? I need a satisfactory answer to this question. (It looks strange that the pads are centred on the board outline. I would be really surprised if this is anything close how it should be.)

---

By the way nothing says you can not use the asset as provided in your project. Your own standards can be lower than the ones of the official lib.

Or if you want you can provide me with a drawing showing the resulting fillet sizes with this footprint and i can double check how bad they are when compared to what IPC would suggest for castelated terminals (i much rather invest the time needed into a pull request where the contributor is prepared to invest the time needed.) My guess is that i will then request changed pad sizes and somebody will need to implement that (as the original contributor never reacted i would guess that this will fall onto you)

G-Pereira:
I don't get where your anger comes from but I am sure I have nothing to do with it. I would like to ask for cooperation in this process.
I am actually using this footprint with success so I would like to help on everything I can to make it adhere to KiCAD requirements to be included in the official library so it is easier for other to use this component.
You can find the pad sizes on page 5 after the module overall drawing and right before the pin table. What it seems to me that the manufacturer doesn't provide is the spacing between the inside pads in respect to the outside pads. About that I can't find no other drawing on the internet with that specification all I can say is that I have measured it myself and the dimensions on this PR match 10 units that I currently have in stock. I don't know if this module can ever be on the KiCAD libraries since there is no drawing with the required measurements so you or any other librarian can confirm it.

poeschlr:
I am not angry. I just need to ensure that i use my time efficiently. I am working on >100 footprints so if i ask for information then it is to offload work to the contributor who is most likely working on <10. (In the past i had more time so i took on even the task to search for information like that. Now i work full time and am still finishing my education so i really need to be a lot more strict with where i spend my time)

G-Pereira:
> I am not angry. I just need to ensure that i use my time efficiently. I am working on >100 footprints so if i ask for information then it is to offload work to the contributor who is most likely working on <10. (In the past i had more time so i took on even the task to search for information like that. Now i work full time and am still finishing my education so i really need to be a lot more strict with where i spend my time)
>
> Edit: this by the way includes the fact that i do not search for more than 5 mins in a datasheet for a drawing. So if you link one give the page where the relevant info is to be found. (especially as you linked a >600 page document)

I agree with you, I should have said that the NRF52840 was for reference only and not useful to read and should also guided you to page 5 of the module's datasheet. I value the work librarians don't get me wrong.

> Edit2: the drawing is not defining the pads on the receiving PCB but the one on the component which is comparable to how the terminals of a castellated LED are given.

I see, but what can I do to extrapolate the right pads? IPC for that?

G-Pereira:
@poeschlr I contacted the manufacturer and they can't provide me more information other than the one that it's described in the datasheet above, which means we have no more information about the drawing. What can we do?

poeschlr:
It's sad but expected that they do not give us more info. (Not sure why as every PCB manufacturer can give their general tolerances and this part is basically just a PCB. I fear they just don't want to be held to a standard that might lock them into using a subset of possible PCB manufacturers.)

I think we can still use IPC if we use reasonable assumptions for the tolerances. This should give us a general idea how the pads might look like which can be further refined on the final footprint (I suspect that the inner row of pads might get in the way if we choose tolerances that are too large)
Be aware that the rules for castellated pads are kind of turned around (heel and toe are exchanged compared to other components)

G-Pereira:
I would like to say that I will no longer work for this PR for lack of knowledge of footprint design because I don't have a clear idea how I can instruct myself. For the time being I am not ready to help.

poeschlr:
I do not expect you to take this on. But maybe this will be of help in your endeavour anyway https://forum.kicad.info/t/tutorial-how-to-make-a-footprint-from-scratch/11092/5

G-Pereira:
Meanwhile the manufacturer has sent me a message saying "manufacturing tolerances are 0.0254mm"

poeschlr:
Thanks for your great effort. Now we have the information needed to properly calculate how a proper footprint should look like. I will however mark this pull request as abandoned as we did not hear back from the original contributor in more than half a year.

poeschlr:
For the outer pads i calculated the following suggestion as per IPC-7351B concave chip array suggestions (Fillets T=0.45, H=S=-0.07, rounded to nearest 0.05) Padsize = 1.55x0.75mm, position in x direction 6.25mm from center of board meaning 0.25mm inward of the board outline. This might need adaptation depending on what the values are for the inner pads. (The heel fillet might need reducing to achieve a reasonable clearance between the pads)

There is however another hitch. The inside pads are not documented in any of the datasheets we found so far (The nominal size is at least documented in the datasheet found by @G-Pereira but their positions are not documented. It does not look like they fall under the centreline of the outside pads so at least the distance between edge and inner pads is missing in both datasheets.)

G-Pereira:
I noticed those dimensions were missing and the manufacturer said it would come back to me with an answer. As soon he answers me I will post here.

G-Pereira:
Meanwhile, the manufacturer has sent me what I believe to be [Altium footprint library](https://github.com/KiCad/kicad-footprints/files/4210344/E73.2G4M08S1C.zip). Let me know if this can be of use.
http://ebyte.com/pdf-down.aspx?id=1227

poeschlr:
Maybe the importer by @pointhi can already handle it but otherwise we will need to wait till this is either ready or somebody with Altium access can decode the missing info for us.

G-Pereira:
The importer seems to be nearly ready now: https://gitlab.com/kicad/code/kicad/-/merge_requests/60
Quoting the developer (two days ago post at https://forum.kicad.info/t/work-in-progress-native-altium-importer/20712/90)

> I think my importer is quite good now

PR #1491, opened 2019-03-19
add Ninigi RJ45GE connector

balcer:
https://www.tme.eu/en/details/rj45ge/rj-connectors/ninigi/
https://en.ninigi.com/product/ninigi/socket/rj45ge

rj45

---


poeschlr:
I can not find a dimensioned drawing of this. Could you link to the drawing you used to create the footprint found in this pull request?

balcer:
tme

poeschlr:
For future reference here the source: https://www.tme.eu/html/gfx/image_17594.gif

poeschlr:
I think the dimensions are ok. (The drawing you found might be wrong as it seems to give the dimension i marked orange as 8.89.)

I would however suggest to either generally reduce the rest ring or at least reduce it in x direction. (right now the rest ring is 0.445 could easily be reduced down to 0.3 which would increase the clearance to ~1mm. See green dimensions.)

Screenshot from 2019-06-03 19-49-36

balcer:
@poeschlr You are wright about wrong dimension given in drawing. This is a standard ethernet connector. Pad size reduced. Thanks for rereview.

poeschlr:
thanks

PR #1498, opened 2019-03-25
DFN_QFN: Add 10-pin DNF, 3x3mm with 1.7x2.5mm EP

zonque:
This is very similar to `DFN-10-1EP_3x3mm_P0.5mm_EP1.55x2.48mm` except for a slightly different middle pad and smaller side pads.
    
This package is used by MPS components such as this one:

  https://www.monolithicpower.com/pub/media/document/MPQ2483_r1.05.pdf

A symbol definition which uses this footprint will follow.

image


poeschlr:
Footprints for standardized packages like this should really be scripted using our IPC generator scripts found here: https://github.com/pointhi/kicad-footprint-generator/tree/master/scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON

zonque:
I don't quite understand I must admit. I didn't create that footprint from scratch but used `DFN-10-1EP_3x3mm_P0.5mm_EP1.55x2.48mm` as a template, and changed the sizes of the pads. How is that the wrong approach? IOW - what would be the difference to generating it from scratch?

poeschlr:
The script makes it easier for us to maintain the library in the future. This is why all new additions that can be scripted must be scripted. (I am slowly going through the old existing footprints and transferring them over to be scripted. I just did not get to the one you used as a basis for your contribution.)

Another reason is that checking a scripted footprint is much easier. I only need to compare the size definition you entered against the datasheet. When it is not scripted i would need to create a dimensioned drawing of your copper, fab and courtyard layers and compare them against the datasheet, industry standard and library convention which takes up about half an hour per footprint. (Checking a script contribution takes 5 minutes if everything is correct and 10 minutes if there are mistakes to be documented.)

---

Oh and IPC is the industry standard comity that determines how to get from a package dimension to a footprint.

---

The script is a standalone python3 program. Run it from a console using `python3 ipc_noLead_generator.py size_definition/dfn.yaml` (assuming you started the console in the directory i pointed to.)

Enter your size definition into the appropriate size definition file (in this case dfn.yaml) Use a similar part as the basis of your size definition.
Every dimension can be entered as nominal+tolerance, minimum+maximum or minimum+nominal+maximum. Choose the same style as used in the datasheet.

If you need additional help feel free to ask here.

evanshultz:
Our scripts generate footprints according to IPC rules using, where possible, an industry-standard package drawing (like from JEDEC). This means the footprints are based on known parameters instead of being whatever the IC company decides to publish.

In addition, maintaining the footprints is dramatically easier because scripted footprints can be updated en masse instead of any manual work (even search and replace).

Here, with a scripted footprint the pads would be rounded and this is one obvious sign it wasn't generated by a script.

We have been working through the footprints and updating them in groups but we haven't gotten to all gullwing footprints, like this one, yet.

zonque:
Alright, thanks for the explanation! Will dig into that generator then.

zonque:
Okay, very nice tool. A few question though, sorry to bother you.

* When I enter the values in the datasheet, the footprint that is generated differs from the recommended landing pattern the vendor. I can tweak the input parameters to match the recommendation, but I guess that's not the point? Should we just fly with what the generator comes up with?

* How is a new footprint end up in the library eventually? Would I submit a PR for the generator and things are then regularly synced over? Or would I copy the generated one and then resubmit here?

* What about the 3D shape? My current version reuses the one of another DFN-10 3x3mm package, and that worked for me. What't the correct way of doing that in the generator?

Thanks!

evanshultz:
1. Yes, trust the generator.
1. Submit a PR for the script at https://github.com/pointhi/kicad-footprint-generator, and then update this PR with the new footprint.
1. You can reuse an existing 3D model if it's compatible. If the thermal pad is unique in the part, it should be unique in the model in our library. The DFN 3D model generator is also scripted, I believe.

poeschlr:
regarding 3: do not change the 3d model selected by the generator. Or more precisely do not edit the output of the generator as we will lose these changes the next time we regenerate all footprints from the generator for any reason (for example if kicad adds new features to the footprint file format allowing some things to be represented better than now.)

zonque:
Got it. I submitted a [PR for the generator](https://github.com/pointhi/kicad-footprint-generator/pull/325) and updated the content of this one here. The 3D part is not there yet, but IIUC, this can be done independently as the 3d package is now referenced but non-existing?

poeschlr:
We do not require every 3d model to exist. The long term idea is that we use the same size definition file that generates the footprint to also be used by the 3d generator. But that is something that we can start once we have the footprints converted or if we get additional manpower. (This is why we started to include the part height in the footprint size definition even thought it is not used for the footprint generation itself.)

poeschlr:
I assume the changes i requested over on the script do not change the generated footprint. Meaning i will merge this as it is now. Thanks for your contribution and a special thanks for looking at the script.

PR #1501, opened 2019-03-25
Fix #1497 mounting hole positions in Keystone 2468

### LINKED ISSUE #1497: Keystone 2468 screw holes are backwards
The footprint has two mounting holes, currently they are in the wrong position.  They are oriented as for mounting the battery pack on the back side of the board, where the rest of the footprint puts the battery pack on the front side.  See http://www.keyelco.com/product-pdf.cfm?p=1033

ianrrees:
Fix #1497 mounting hole positions in Keystone 2468

The issue has a photo of the part in question, and a PCB made with the original footprint.

http://www.keyelco.com/product-pdf.cfm?p=1033

---

### DIFF
--- a/Battery.pretty/BatteryHolder_Keystone_2468_2xAAA.kicad_mod
+++ b/Battery.pretty/BatteryHolder_Keystone_2468_2xAAA.kicad_mod
@@ -86,8 +86,8 @@
   )
   (pad 1 thru_hole rect (at 0 0) (size 2 2) (drill 1.02) (layers *.Cu *.Mask))
   (pad 2 thru_hole circle (at 0 12.7) (size 2 2) (drill 1.02) (layers *.Cu *.Mask))
-  (pad "" np_thru_hole circle (at 8.636 8.6995) (size 3.5 3.5) (drill 3.5) (layers *.Cu *.Mask))
-  (pad "" np_thru_hole circle (at 38.608 3.9624) (size 3.5 3.5) (drill 3.5) (layers *.Cu *.Mask))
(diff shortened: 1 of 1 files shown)


evanshultz:
The dimensions are roughly correct, and I'll merge as-is if you request it. I read your message about having trouble before and I'd rather you feel welcome and happy to contribute than put up barriers.

That being said, this may have been converted from an Imperial footprint before. All parts in the library should be in millimeter and there are a number of corrections to make this footprint match the datasheet. Please update this PR if you're willing to really correct this, or let me know and I'll take over.

ianrrees:
So I think there are two separate issues here:

  1. The footprint has mounting holes which are mechanically incompatible with the part.
  2. The footprint does not conform to some aesthetic guidelines.

Issue 1 is the bug that this pull request fixes.  Issue 2 is far below my personal "worth the time to fix" threshold.

Basically what I care about, is being able to use KiCad to make PCBs that work, with minimal irritation along the way.  Functionally the fixed footprint is plenty good for my purposes, but the current one has contributed to a batch of discarded PCBs and lost time.

evanshultz:
Which I why I asked if you prefer me to merge as-is. Thanks for this fix!

ianrrees:
Thanks @evanshultz .  I am not trying to be difficult, but am trying to explain that this footprint having been converted from imperial, for instance, is outside my area of interest.  So, from my perspective, the second paragraph in https://github.com/KiCad/kicad-footprints/pull/1501#issuecomment-476360251 is a distraction that adds friction to this process.

I think it's great that there is so much care being taken with library management, and don't want to dissuade you or anyone else from working in that direction.  However, I think it would be great if small contributions in the right direction were simply accepted as such.

evanshultz:
I understand. Thank you again and sorry this bug caused you to have to trash some PCBs. :(

PR #1502, opened 2019-03-26
Added UDFN-8

SchrodingersGat:
e.g. https://www.digikey.com/product-detail/en/on-semiconductor/NLSV2T244MUTAG/NLSV2T244MUTAGOSCT-ND/3487620

Datasheet: http://www.onsemi.com/pub/Collateral/NLSV2T244-D.PDF

image

image

The KLC check fails for the center of the footprint as pin-1 is slightly larger (as per the datasheet!)


evanshultz:
1. Let's make this footprint ON-specific.
1. Please add a description instead of just a link.

Otherwise looks good to me.

SchrodingersGat:
Ok, changes made as requested

PR #1505, opened 2019-03-27
Add footprint for RTC RV-1805-C3

jneiva08:
Created footprint for RTC RV-1805-C3 from MicroCrystal

Symbol: https://github.com/KiCad/kicad-symbols/pull/1687

Datasheet: https://www.microcrystal.com/fileadmin/Media/Products/RTC/Datasheet/RV-1805-C3.pdf

Footprint:
rv_1805-c3_footprint

Didn't know where to put, have created RTC.pretty because the footprint did't fit in the others

---


jneiva08:
Was seeing other model and appear the package is the same.

RV-8523: https://www.microcrystal.com/fileadmin/Media/Products/RTC/App.Manual/RV-8523-C3_App-Manual.pdf

Probably the footprint name should be: RTC_SMD_MicroCrystal_C3_2.5x3.7mm ?

evanshultz:
Since it's specific to this company and it appears they call this package C3, I think what you wrote above is fine and the dimensions are even optional. Up to you. Please put the dimensions in the description and update the keywords too.

1. Courtyard height should be +/-2.1mm.
1. Perhaps push the silk down closer to the fab lines and make it shorter to fit between the pads.

jneiva08:
Updated footprint

rtc_c3

evanshultz:
Thanks! All but the silk looks good. Silk needs to clear the pads by at least the line width, or 0.12mm in this case (note that the line will then lay 0.18mm away to account for half of the line width). It moved too close to the pads now but it's placed nicely just outside the fab lines. Perhaps shorten the bottom silk line between the pads and make the top right do the same. Then on the top left the silk could "dogleg" up and over pin 1 to mark it? I can make an example if words aren't so easy to understand.

jneiva08:
Have changed the F.SilS
In the bottom maintained in the same place but reduced width.
In the top moved a bit (0.06mm) and made a slope to indicate pin 1.

jneiva08:
Moved back the slope, to mask pin 1

how it were was under the component

rtc_c3

evanshultz:
We would still like to get 0.12mm clearance from silk to pads. Here is an idea:
1. Move the X position ends of the top and bottom silk lines to +/-0.56mm.
1. Move the top silk line down to 1.85mm in the Y direction (same as the bottom line).
1. The line above pin 1 should be 2.03 in the Y direction to clear the pad by 0.12mm.
image

Lastly, did you not find a suitable existing library for this footprint to live? I think Package_DFN_QFN would be fine.

jneiva08:
@evanshultz have made the modifications to have the slope 0.12mm of the pad.

C3

Still didn't changed yet the location of footprint.
I returned to see the datasheet and not the manual application, and refer in the point "terminations and processing" that the footprint is SON-10 and processing IPC / JEDEC J-STD-020C.

Probably should change to Package_SON\SON-10_3.7x2.5mm_P0.8 ?

evanshultz:
Silk looks good!

Because this is a specific footprint and probably not generic, let's keep the footprint name. I'm OK to move it to the SON library, though, and please note this package name in the description.

Lastly, there are now quote marks around all strings in the footprint name. I suppose this is because you are using a new nightly? The footprint opens fine in 5.1.0 for me but syntax changes like this are something librarians need to keep in mind. In this case it's seems backwards-compatible.

evanshultz:
Thanks!

jneiva08:
Thanks for your help

PR #1510, opened 2019-03-28
TSM-SH Series

gcormier:
https://www.samtec.com/products/tsm
http://suddendocs.samtec.com/catalog_english/tsm.pdf

Reference:https://github.com/pointhi/kicad-footprint-generator/issues/319

---


gcormier:
Would apprieciate initial feedback on the overall footprint before getting down to the nitty gritty clearances & etc. (Although I think they are quite close in this case)

poeschlr:
We had specialized footprints for some pin headers in the past. We however decided that it might be a bit much as they typically are compatible with each other. How large are the differences in the dimensions of the part itself compared to the one used for generating the current footprints in the lib? (I ask about the part because that is what was used to determine how the footprint looks like. The suggested footprint can vary a bit depending on which standard was used to generate it. And also by the manufacturing tolerances input into the calculation.)

----

See discussion at: https://github.com/KiCad/kicad-footprints/issues/130

gcormier:
Hmm.

I've used the footprint in this PR for a component. The pads are slightly larger then the Harwin, and I don't see any benefit to the smaller pads as you aren't gaining any space between the pins.

They aren't the same size though, so not sure how to decide which sizing "wins" :) Given the nature of the headers, I suspect they are almost identical?

PR #1513, opened 2019-03-30
Fix footprint for SOT-353_SC-70-5_Handsoldering

stm32f1:
*Fix footprint for SOT-353_SC-70-5_Handsoldering*

---

The outline of the footprint for SC70_5 was not correct, so DRC did not work.
Thus I corrected the outline of it.

bofore
after


evanshultz:
@stm32f1 
I see this is your first contribution here. Thanks for taking the time to do this. Hope to see you contribute more in the future.

Good catch and thanks for the fix.

PR #1514, opened 2019-03-31
Package_DFN_QFN: Add Texas_S-PWQFN-N32_EP2.8x2.8mm

zonque:
Used by this part:

  https://www.ti.com/lit/ds/symlink/bq25703a.pdf (page 91)

image


zonque:
Generator PR: https://github.com/pointhi/kicad-footprint-generator/pull/330

zonque:
Oh, the script generates footprints that fail the Travis checks? Hmm.

poeschlr:
The travis check simply does not deal well with thermal vias. It throws false positives for it. (This is because it does not realize that these are vias but assumes them to be normal tht pads which of course have different layer requirements compared to what we use them for when creating thermal vias.)

poeschlr:
Looks good now. Thanks

PR #1521, opened 2019-04-02
Package_DFN_QFN: Add Texas_S-PVQFN-N20_EP3.05x2.05mm

zonque:
This adds a rater exotic footprint used by this part, for instance:

  https://www.ti.com/lit/ds/symlink/tlc59108.pdf (page 42)

It was orginally scripted but then manually modified.

image


zonque:
As discussed here, it's probably not worth adding support for such a thing to the generator, so I followed the manual approach.

evanshultz:
Please supply the device entry to the script that generated the original footprint for verification. Then list the changes you manually made.

Note that silk runs over the left and right pins, is missing between them, and the pin 1 mark isn't in the right place. So some manual silk work remains to be done.

evanshultz:
That all being said, since this is such a specific part it may be best to adhere strictly to the datasheet. Doubly so since you've used a Texas_ prefix but the part doesn't actually match their datasheet. I've not fully made up my mind if this is required, but you won't get any complaints if you make a TI-specific footprint and match their datasheet. Easiest way to go.

zonque:
Oh, yes. Thanks for noting. Fixed the silk and fab layers now.

FWIW, the script entry I used is this:

```
Texas_S-PVQFN-N20_EP3.05x2.05:
  device_type: 'QFN'
  #manufacturer: 'man'
  #part_number: 'mpn'
  size_source: 'https://www.ti.com/lit/ds/symlink/tlc59108.pdf#page=42'
  #ipc_density: 'least' #overwrite global value for this device.
  custom_name_format: 'Texas_S-PVQFN-N{pincount}_EP{ep_size_x:g}x{ep_size_y:g}mm{vias:s}'
  body_size_x:
    minimum: 4.35
    maximum: 4.65
  body_size_y:
    minimum: 3.35
    maximum: 3.65
  overall_height:
    minimum: 0.8
    maximum: 1.0

  lead_width:
    minimum: 0.18
    maximum: 0.3
  lead_len:
    minimum: 0.3
    maximum: 0.5

  EP_size_x:
    nominal: 3.05
    tolerance: 0.1
  EP_size_y:
    nominal: 2.05
    tolerance: 0.1
  # EP_paste_coverage: 0.65
  EP_num_paste_pads: [2, 2]

  thermal_vias:
    count: [3, 2]
    drill: 0.3
    # min_annular_ring: 0.15
    paste_via_clearance: 0.1
    paste_between_vias: 1
    paste_rings_outside: 1
    # EP_num_paste_pads: [2, 2]
    # EP_paste_coverage: 0.7
    grid: [1, 1]
    # bottom_pad_size:
    # paste_avoid_via: False

  pitch: 0.5
  num_pins_x: 8
  num_pins_y: 2
  #chamfer_edge_pins: 0.25
  #pin_count_grid:
  #pad_length_addition: 0.5
  #suffix: '_Pad{pad_x:.2f}x{pad_y:.2f}mm_HandSolder'
```

I then changed the position and size of all outer pins to match the recommended landing pattern in the datasheet. The pins on the left and right side where completely off, the rest I had to amend a tiny bit. Did I miss anything or which detail are you referring to when you say it doesn't match?

evanshultz:
The pad shape, paste pad dimensions, etc. Many things are just a bit off.

@poeschlr 
How do you want to deal with this. Since this has a Texas_ prefix do you want to match the datasheet exactly? Or since the doc says IPC 7351 is an alternate design, and 7351C is saying to use round rect instead of D-shape, do you want to accept this as is? Or accept this as a general (non Texas_) footprint?

zonque:
Hmm, okay, the pads could be rectangular, changed that now. However, I don't see which pads are off, tbh. Could you elaborate? See the measurements below, and compare them with the "example board layout" on page 44 of the datasheet.

image
image
image
image

Am I missing anything?

I've pushed the change to make the pads rectangular.

evanshultz:
There were two things I mentioned above. Let me give more details:
1. Pad shape: The datasheet shows D-shape pads, not rectangular or round rect.
1. Paste pad dimensions: They were close to the datasheet but not quite correct in X and Y size.

Let's wait to see if @poeschlr wants a `Texas_` part to exactly match the datasheet. Even with an allowance here that IPC 7351 is an alternate design as the doc says. To me, matching the datasheet exactly eliminates any questions but I want to elevate this question before merging anything.

poeschlr:
Regarding pad shape: IPC addresses this. They suggest to use rounded rectangle even if the manufacturer suggests something else as they say it has the best soldering result. (oval was mainly used because rounded rectangle was not really supported by tools until very recently. D-Shaped is not even a viable option for us as kicad does not support it other than using complex pads whos downsides might not make it worth it.)

Regarding pad dimensions: It seems this is a modified IPC generated footprint that is not based on the manufacturer suggestion. So it should be ok as long as it is done right. (If pads 1, 20 shown in the screenshot are only moved in y direction not in x direction)

PR #1523, opened 2019-04-03
Murata SFECF

joemadeus:
Adding footprints for Murata SFECF 10.7MHz ceramic filters. Associated symbols are in this PR: https://github.com/KiCad/kicad-symbols/pull/1708

Product page/Datasheet: (apparently there are no PDF datasheets available?) https://www.murata.com/en-us/products/filter/cerafil/sfecf

Murata SFECF
Murata SFECF HandSoldering

---


evanshultz:
@joemadeus 
Thanks for taking the time to contribute to the official KiCad library!

Where did you come up with the pad design? I assume the image at https://www.murata.com/en-us/products/productdetail?cate=luCeramicFilters&partno=SFECF10M7DA0001-R0 (I picked one part from the list) is showing top, side, and then bottom views?
image

Nowhere is the footprint that I can see.

Review notes:
1. Can you please add a description for the footprint in addition to the link?
1. The fab layer should be a rectangle of the nominal part body size and with a beveled corner.
1. The courtyard should be a rectangle around the of the part body and pads, with 0.25mm clearance from the extents of those elements.
1. The silk layer will avoid pads by at least the line width (0.12-0.2mm is typical) and go outside the part body offset by a small amount (0.11mm is common) with a pin 1 mark of some type. This one is a little tough but perhaps between pads 3 and 4 and then a smaller line just by pin 6. This may take some experimentation. Silk may not lay under a SMD part body.
1. There is a small gray rectangle centered between pins 1, 2, 5, and 6 in the image above. Do you know what it is?

Some of the above sounds awkward so please review some existing footprints, read  
http://kicad-pcb.org/libraries/klc/, and ask questions.

joemadeus:
Hi @evanshultz !

Thanks for reviewing this! I interpreted that lower-most drawing as the footprint, not the bottom view -- I can see that wasn't correct. I don't have a footprint drawing for this part (and I don't know what the central rectangle is) but just sent an inquiry to Murata for a diagram. I'll let you know what I get back.

I do have a board with this part's footprint on it, but it isn't populated yet (in fact, I have yet to order the parts.) I'll be happy to get these parts in and see how they fit, for what that's worth.

I'll be happy to make the other changes in the mean time.

joemadeus:
Murata got back to me this morning with this footprint diagram (way to go, Murata! better than 24 hours turnaround.)

SFECF_Footprint

I'll turn it into a footprint over the weekend or early next week.

evanshultz:
The the image above you can take the 6 copper pad sizes (1.05x0.8mm) and placement (1.4x2.45mm pitch). The rest will come from the body drawing I inserted above. The traces will be created by the user.

joemadeus:
@evanshultz I think that's everything -- except for a 3D model, which I'm still figuring out how to create.

joemadeus:
@evanshultz Updated based on your review (I added the description in the same field as the document link -- I wasn't sure where this was supposed to go?)

evanshultz:
Thanks!

Here is what we have now:
image

Just two things left and then I will merge:
1. Please swap the silk ref des and value positions. Also ensure the silk ref des is far enough away from the pads as described above.
1. Add another ref des on the fab layer as described at http://kicad-pcb.org/libraries/klc/F5.2/, clause 4.

SchWeInefleisch:
@evanshultz The examples from KLC 5.2.4 have the ref des below the component in the examples there; it's 0.12mm from the pads, which is what I think you specified above. And the ref des is already on the fab layer as well.

evanshultz:
@SchWeInefleisch 
Ref des on top is most common throughout the library. This isn't in KLC as far as I know, but since the footprint needed other changes I threw in this request for consistency.

On a 0.01mm grid the ref des is just short of 0.12mm from the bottom pads. It might be OK, though I often see the board not quite match the Gerbers with text. So a bit more margin wouldn't hurt.

You are absolutely right about the fab ref des. It is under the silk ref des and I missed it. Thank you.

@joemadeus 
The fab layer ref des should be centered. And please shrink the text, while keeping the same ratio, so that the entire "REF**" string fits within the body outline. You may rotate the text 90 degrees if the part is taller than it is wide.

SchWeInefleisch:
@evanshultz Swapped the ref des on the fab layer and the part number. Rotated the fab layer ref des, put it inside the body outline and scaled appropriately. Changed the pin-1 marker on the silk layer so it doesn't conflict with the ref des and brought it to within 0.12 of the pad (but this now conflicts with KCL F5.1.2, doesn't it?)

joemadeus:
@evanshultz Swapped the ref des on the fab layer and the part number. Rotated the fab layer ref des, put it inside the body outline and scaled appropriately. Changed the pin-1 marker on the silk layer so it doesn't conflict with the ref des and brought it to within 0.12 of the pad (but this now conflicts with KCL F5.1.2, doesn't it?)

evanshultz:
Current FP:
image

Fab ref des looks good and fab value and silk ref des position are good as well. Thank you.

There should be at least the line width in gap between pads and silk lines. Not the line center but the gap, so you need to consider half of the line width when figuring the gap. The pin 1 left pad edge is 1.75mm, so to get 0.12mm gap the vertical silk line should be placed at 1.93mm. It's at 1.87mm now. And 0.2mm gap is better for cheap board vendors so that would scoot the line farther left. Similarly, the silk ref des needs to go up a bit more to give the same 0.12mm min clearance since it's silk too.

If we can just get a bit more clearance between silk and pads then this is ready.

joemadeus:
> If we can just get a bit more clearance between silk and pads then this is ready.
Sounds good. Give me a day or two and I'll get it finished up

joemadeus:
@evanshultz I noticed that other footprints in `Filter` are named `Filter_xxxx` -- should this be as well?

joemadeus:
@evanshultz Updated per your code review. I understood your comment "And 0.2mm gap is better for cheap board vendors ..." as "you should move it even farther than the required 0.12mm", so that's what I did: the vertical line's position is at -2.00 (-1.75 + -0.20mm + -0.06mm, then rounded) and the ref des on the silk layer is at -2.5.

joemadeus:
I just realized a number of footprints that shouldn't have been pushed up, were. I'll remove them and push. Sorry 'bout that. I'm returning to this after quite some time and lost track of what's where.

evanshultz:
@joemadeus 
Thanks!

The SFECF footprint silk is now good. The 3D model name needs adjusted per the CI output.

For the MSA-0505 footprint, it was never discussed above so if you want it reviewed please add a screenshot and datasheet link and fix the CI errors. Alternatively, you can remove it from this PR. Up to you.

evanshultz:
Thank you!

PR #1524, opened 2019-04-03
Added the HVSON8 footprint for the NXP PCF8523

johnfwhitmore:
This footprint was auto generated with the kicad-footprint-generator
https://github.com/pointhi/kicad-footprint-generator/pull/335

Datasheet - https://www.nxp.com/docs/en/data-sheet/PCF8523.pdf

Screenshot from 2019-04-03 20-55-58


evanshultz:
Thank you!

PR #1525, opened 2019-04-04
Add Texas_RLR0018A_B3QFN_7x9x4.1mm_2EP

cp-aquila:
Add Texas_RLR0018A_B3QFN_7x9x4.1mm_2EP

---

---

* Datasheet: http://www.ti.com/lit/ml/mpqf481b/mpqf481b.pdf
* There are quite a few F6.3 violations because pads are stacked to create the required SMD style pads.
* I choose not to add the big copper pours because they would connect pins that have different pin numbers ("6, 7, 8" and  "5, 18"). It is up to the user to connect them in the schematic and add a appropriate copper pour. However, there are some lines on Dwgs.User to indicate where that copper-pour should go.
* Zone-connection is set to solid
* Because the copper-pour is not part of the FP I also did omit the thermal vias.

---

grafik

grafik


evanshultz:
Wow. This is a cool footprint!

I selected http://www.ti.com/lit/ds/symlink/lmzm33602.pdf as a representative datasheet so help me understand the electrical connections. From that datasheet, the pins are connected internally together and so they can have the same pin number. However, perhaps in other packages pins 5, 14, 15, and 18 aren't all PGND (just one example). Do you know if the pinout of the part varies, meaning we need to have all unique pin numbers on the footprint or if they can be combined?

They only shown a soldermask-defined footprint. Usually this is bad practice because registration issues increase the tolerances. Since the outer pad "fingers" are 0.25mm in mask width these small dimensions would likely benefit from a NSMD construction. Oh well. We can still follow the datasheet.

A couple quick initial notes:
- I would think adding copper and vias per the footprint is good. If it's there the user doesn't have to worry about adding it and we can be sure it's as TI says. But this is intertwined with the first paragraph above. If there are variations with other pinouts, perhaps each variation gets a unique footprint?
- Please add a description in addition to the datasheet.
- Mask and paste pads should not have a pad number.
- It would be nice to see custom pads instead of overlapping copper pads. You can see some funky Infineon PQFN footprints at https://github.com/KiCad/kicad-footprints/pull/981 for an example.
- If you can set mask oversize in the footprint, instead of on each pad, it's much more clean.
- The zone connection is set on the whole footprint so it may not need to be set on each pad. (I haven't checked if this works but I would think so.)
- Paste pads don't need their zone connection set.

cp-aquila:
Hi,

The lmzm33602 was the part I had in mind when making this FP.
Here is a list of similar parts.
http://www.ti.com/product/LMZM33603#parametrics
Out of those, only the lmzm33602 and lmzm33603 share the same FP. I do assume that the FP "RLR0018A" is unique, for those two parts.
The other devices have different package sizes with a different pin-out and landing pattern.

Thanks for the review. I will fix those asap:
* Zone connection and Mask-Size only as global setting
* Remove pad-numbers on mask- and paste-only pads

On these points I dont agree 100% and are open for discussion:
* I choose soldermask-defined because of the general rule 'stick to the datasheet'. Personally I prefer NSMD as well. How do you feel about making a minimal foorprint with just the correct copper for the pads (no vias or additional copper) and alls mask/paste set zo zero. 
* Custom-pads don't support arcs, so round-rect-stype pads are not possible. Thats why I did the stacking. Please correct me if I am wrong!
* Adding copper and vias will create some other problems
    * either 'short circuits' or pin-numbers not identical to datasheet since multiple islands need to be connected together
    * the device-datasheet says ```Switch node. Connect these pins to a small copper island under the device for thermal relief``` for pins 9,10 and 11; I don't see that in the recommended footprint, but in the device-datasheet (Page23). So the idea was to have the user figure the it out.

evanshultz:
I hope you don't mind that I added numbers to your post to make it easier to respond.
1. We are in agreement to follow the datasheet.
2. You can also place a polygon on a copper layer and choose the line thickness of the polygon to get rounded corners of a size you choose. Custom pads can be created in FreeCAD with any shape and exported into KiCad. But we can accept overlapping pads as you've done since the result in Gerber should be the same. It truly would have been nice to support custom pads with corner radius in KiCad, but hopefully sometime later...
3. I would then prefer to have footprint variations for each device so that all copper and vias in the datasheet is included. Because this is tricky and specific, taking care of this the footprint will make implementation easier for the user. Another variant of RLR0018A could be created for other devices if they have the same package but a different pinout. I'm not sure exactly what to name them in this case, though. Any ideas?

cp-aquila:
I will go with the custom pads and vias. But it will take some time, I am working on some scripts that take a .svg to create the footprint. Which is usefull here since I can extract the vectors from the datasheet.

cpresser:
New approach:
- using evince to convert the datasheet .pdf to .svg
- use inkscape to clean up that svg and add missing parts
- use https://github.com/mtl/svg2mod to generate a kicad footprint file

3d
inkscape

The .svg contains the original annotations and is easy to review.
https://shells.ccc.ac/~cpresser/lmzm33602.svg

There are a few things left to improve/patch with svg2mod:
- support for vias
- support for arcs as part of a path

evanshultz:
Nice! I'm assuming you're working to finish this off and I should review when you make a new commit. If that's not correct and there's something I can do now, please let me know.

BTW, if you could document the steps you used it might be helpful for other folks too!

cpresser:
Just a quick update: I am still alive and working on this. But its more complicated than expected. SVG-Path-Segments are Bezier-Curves. Converting them to KiCad Arcs is not straight forward.
Once I get this working I will write a blog-post with the complete workflow.

cp-aquila:
I am closing this PR since most likely I will never finish it.
Getting the toolchain to work properly with arcs is a PITA and my motivation to rewrite the scripts is basically zero.

PR #1531, opened 2019-04-06
Added SiliconLabs WES0077 434MHz printed anntenna

lorenz:
Adds the SiliconLabs WES0077 printed antenna, part of the [SiliconLabs 434MHz antenna development kit](https://www.silabs.com/products/development-tools/wireless/proprietary/434-mhz-antenna-matrix-development-kit). Design is based on the Gerbers downloaded from there minus the passives.

SiliconLabs_WES0077-01_434MHz
---


evanshultz:
Welcome and thanks for contributing to the KiCad official library!

First, you will need to sign the CLA before we can review and accept your submission.

lorenz:
That is weird. It said I all committers have signed the CLA last time I checked. I've told it to recheck and  now it's saying everything's ok again.

evanshultz:
1. The ref board has square corners. Perhaps use polygons or a custom shape to match this? Note that this is outside my expertise so maybe it doesn't matter?
1. There is no ref des on the fab layer. The TI antennas don't have them but I believe they should. Would you please add it?
1. There is some amount of misalignment on both ends of the polygon. Changing the entire thing to a polygon may help.
1. I don't understand the courtyard size. It's far away from copper on the left side and touches the pad on the right side. What was your design methodology?
1. Did you take into account the line endcaps when picking trace lengths? For example, in the Gerber the leftmost horizontal line is 5mm long. Yours is also 5mm long but the endcaps on a 1mm thick line end up with a total 6mm length.

lorenz:
Thanks for your feedback!

I've redone the main part as a single polygon which should address issues 1, 3 and 5.

2. I've added a ref on the fab layer.

4. The matching network and the dipole (bottom and top layer RF ground) need to be on the input side, so I can't extend the courtyard on the input side. You still shouldn't put anything other than RF matching components there, but that's hard to enforce.

evanshultz:
Thanks!

I checked the dimensions against the eval board Gerbers and it appears quite close. Perhaps off a little bit in some areas and maybe each "U" section is slightly inconsistent, but it could perhaps be my measuring as well. So I'm OK with that if you are.

There is one place where edges of the poly are not quite straight so would you please fix up that part?
image

I'll look at the other antennas and think about the courtyard tomorrow. Does anyone have some relevant thoughts about the courtyard of PCB antennas? @poeschlr @Misca1234 @diegoherranz @myfreescalewebpage

evanshultz:
Here is what IPC 7351B says about the courtyard:
image

There is no specific courtyard excess for PCB antenna, so let's stick with the 0.25mm used as the default elsewhere in the library.

Also, the pad is rather long. Should it be shorter? If a 0603 component, for example, connects to the antenna horizontally the long pad will run over the 0603 pad.
image

myfreescalewebpage:
> Does anyone have some relevant thoughts about the courtyard of PCB antennas?

I'm not an antenna expert but I think there is some distance to respect depending of each antenna (with no components and no gnd plane on the pcb). If the author is confident with that we should trust him.

evanshultz:
IPC 7351 tells us what a courtyard means, if we choose to go with that. Here and on other antennas in the library already, the User.Dwgs layer is used to show a keepout area.

poeschlr:
Sadly kicad does not support copper keepout in footprints so we need to go with the workaround as described in http://kicad-pcb.org/libraries/klc/F4.5/

The courtyard is not really useful here as that one only checks for other things that have courtyard not for copper. (And as evan mentioned the courtyard is defined to a specific usecase by IPC. We will not do anything that would violate the meaning specified in industry standards.)

evanshultz:
@poeschlr 
Regarding the keepout, since there is no defined area should be the size of the antenna copper with no excess? The antennas in the lib already vary so much there's no discernible guideline.

In the screenshot I posted IPC says the intent of the courtyard is to capture the minimum area consumed by the part and/or land pattern. And it uses "and/or" which would seem to include footprints which are only copper. I don't have a strong personal preference, and maybe I'm reading into IPC 7351 too much, but it appears to me PCB antennas should get a courtyard. But none of the tables cover the courtyard excess for copper-only footprints so IPC doesn't tell us how to construct the courtyard.

poeschlr:
I think the IPC standard does not say copper but pads when it defines the keepout. Pads are typically not only copper but also free of solder mask and have paste applied. The reason for pads to be included is such that rework is possible. (And possibly also to provide some minimum clearance for the paste stencil.)

In this case courtyard does not even provide all we need. We really need a copper keepout here which sadly is only possible using the workaround we define in the KLC.

lorenz:
I removed the courtyard altogether (since it doesn't do what we need and there seems to be some disagreement on its size and necessity). I also perfected the dimensions by hand (everything is now 0.5mm  aligned like in the Gerbers) and shortened the pad to 0.5mm. Is there anything else I should change?

lorenz:
@evanshultz Could you have a look at my changes?

evanshultz:
@lorenz 
Yes, but it appears some extraneous commits were added. Can you remove them from this branch?

lorenz:
@evanshultz Sorry, I merged something else in and accidentially pushed it when pushing for another PR. It's now fixed.

evanshultz:
@lorenz 
I took a look at the latest commit. The copper is perfect now. Thanks!

The pin's Y position is -0.25mm, and I think it would be best if it was 0mm. That will make routing to the pin easier. Can you move everything down by that amount?

The Dwgs.User box represents the keepout so let's be explicit. Is there any guidance about the size of copper keepout on the top layer, copper keepout on lower layers, component keepout, etc.? Let's actually use the word "keepout" as shown at http://kicad-pcb.org/libraries/klc/F4.5/, and remove `except for matching` to be more explicit. Also, please crosshatch this area as shown instead of an "X".

@poeschlr 
I included a snip of the IPC doc above. It uses the term "land pattern" which includes pads and any other elements (mask, other copper, etc.). And the component body. All of that makes up the extents of the elements around which the courtyard is built. I'm sure you know that.

The question in my mind is: within the scope of KiCad's capabilities as an ECAD tool, and considering the target use case of the official library, how should the courtyard be defined for PCB antennas (this probably applies to other footprints where no physical component is placed)?

I vote for adding a courtyard clearance as is done with all other parts, with 0.25mm (the default) being fine. (Keepout clearance is a different issue for which KiCad doesn't have a proper solution right now.)

evanshultz:
Also please add a link in the footprint description. Use the link above if that's the best one.

evanshultz:
@lorenz 
Can you please make the couple changes above? This footprint is looks like it's almost ready to merge.

lorenz:
@evanshultz I'll have a look at it on the weekend

PR #1532, opened 2019-04-07
Add TI TO-PMOD-11 / NDY0011A

ottojo:
Add TI TO-PMOD-11 / NDY0011A

This adds the footprint for the TO-PMOD-11 package used by TI "SIMPLE SWITCHER Power Module" DC/DC buck converters.
PR for symbol at https://github.com/KiCad/kicad-symbols/pull/1716.

I couldn't figure out in which footprint library this belongs, and would greatly appreciate if someone could point me towards a more fitting one.
Same with footprint naming: It seems to be both referred to as "TO-PMOD-11" and "NDY0011A".

The corresponding 3D Model is available for download from the manufacturer [here](https://webench.ti.com/cad/dlbxl.cgi/newstep/NDY0011A.stp).

Datasheet is [here](http://www.ti.com/lit/ml/mmsf025/mmsf025.pdf).

Texas_NDY0011A

---


evanshultz:
Welcome and thanks for contributing!!

Review notes:
1. As shown at http://www.ti.com/lit/ds/symlink/lmz10505.pdf, and inspecting the package, this is a TO-style footprint so let's put it in the `Package_TO_SOT_SMD` library.
1. Please add to description with the size and other info along with the URL.
1. Consider the footprint and pads together, not just the body size to pick the footprint center.
1. Courtyard should be 0.25mm bigger than the extents of pads or board.
1. The fab layers needs a pin 1 mark. A triangle sticking into the fab outline would work nicely here. See many of our connector footprints for an example.
1. The silk also needs a pin 1 mark. Perhaps a vertical line next to pin 1 and meeting the existing silk outline, keeping at least a line width clearance (0.12mm) to the pad.

Also, we can't accept third party models for licensing reasons. The 3D model is OK for your library but we won't be able to accept it into the KiCad library. Just FYI.

ottojo:
Thanks for the feedback, i think i resolved all your comments.
Also thanks for clarifying the 3D model situation, i anticipated it would be like that, but wasn't sure.

evanshultz:
Thanks!

FYI, the three-letter name ("NDY" here) is the most accurate TI package naming system. I'm not sure why they make it so confusing...

Dimensioned drawing:
image

The only problem is excessive courtyard offset where I circled above. Then it's ready to merge!

You could also consider flipping the pin 1 mark on the fab layer so it touches the pin 1 pad. Or adding a bevel to the corner of the fab outline nearest pin 1. But I think any way is OK so you choose whichever version you prefer.

evanshultz:
Thank you!

The corner of the fab layer for the flipped pin 1 mark now sticks out past the part boundary:
image

Perhaps make it 1mm wide instead of 2mm wide?

evanshultz:
Nice. Thank you!

PR #1533, opened 2019-04-07
New LGA-14 Package for Analog ADXL343

nonninz:
According to the reccommendations of the mantainer of the corresponding symbol pull request (@evanshultz),
I created a new footprint that follows the chip datasheet, starting from the existing LGA-14_3x5mm_P0.8mm_LayoutBorder1x6y.kicad_mod

Datasheet: https://www.analog.com/media/en/technical-documentation/data-sheets/ADXL343.pdf
Symbol PR: https://github.com/KiCad/kicad-symbols/pull/1700
Screenshots:
Screen Shot 2019-04-07 at 17 04 46
Screen Shot 2019-04-07 at 17 04 57


nonninz:
It seems like the CLA signing system doesn't work. I tried already 5 times since I opened the PR and it won't recognize it...

evanshultz:
@poeschlr 
Please see the info above and at https://github.com/KiCad/kicad-symbols/pull/1700#issuecomment-479652533. How would you like to deal with this? Merge this PR or another way?

poeschlr:
The datasheets of both components are quite bad. The ST datasheet at least gives tolerances for the body size. I however doubt there really is 0 tolerance for the lead width. 

The differences between these packages in the long direction of the pads vanishes because of  tolerances. (the analog devices pads are inside the st pads with their tolerance)
The lead width is equal for them so i suspect that they might be the same package but dimensioned differently. (We can not say for sure without full tolerance information)

Another observation: ST does not give us a suggested landpattern. Analog would but i am in favor of ignoring it and using IPC (the generators). That would however require us to find a source for the tolerances of that package.

evanshultz:
@poeschlr 
If there is direction in the above I've missed it. Would you like to use a common footprint?

poeschlr:
In general i think they can share a footprint. But the datasheets are so bad that we might never know for sure. If somebody is over motivated then asking the manufacturers for better drawings might give us more info to be truly sure.

evanshultz:
Decided the existing one was close enough and we couldn't find detailed info to verify either package anyway. So ultimately, we don't need this new footprint.

PR #1534, opened 2019-04-07
Add SM5S10AT

pbn4:
Add SM5S10AT

Adding SM5S10AT recommended footprint for DO-218 compatible diode as per [documentation page 4.](https://pl.mouser.com/datasheet/2/427/sm5s-102180.pdf)

2019-04-07-171255_1088x676_scrot


cp-aquila:
Hi,
just a quick look, no real review:
- The Center should be somewhere in the Pin2 Pad
- The Courtyard should be adjusted
- Silkscreen does extend to much to the right

evanshultz:
@cp-aquila 
Thanks!

@pbn4 
Welcome and thanks for contributing! You will find many guidelines for the library at http://kicad-pcb.org/libraries/klc/.

Review notes:
1. Please add a link to the official datasheet at https://www.vishay.com/docs/87609/sm5s10atthrusm5s36at.pdf in the description.
1. Note that the datasheet I linked is newer and has a rather different copper layout. Please update.
1. Update the description and keywords to be relevant for this package.
1. Some of the text, a couple lines, and the pads are at incredibly bizarre locations. Can you move them to a much more coarse millimeter location?
1. Just from the screenshot above I believe the fab outline above doesn't match the part body placement relative to the pads. Also, you can remove the inner horizontal lines and just draw the body outline.
1. Courtyard should be 0.25mm outside the extents of the package body and pads.
1. Correct the 3D model reference as caught by CI.

Once that's wrapped up we'll come back and review all over again.

evanshultz:
Oh, and since this is a generic footprint let's call it `DO-218AC`. The library is fine.

PR #1535, opened 2019-04-07
Footprints for pololu Step-Down Voltage Regulators

jstjst:
D24V5Fx family
https://www.pololu.com/file/0J1498/d24v5fx-step-down-voltage-regulator-dimensions.pdf
grafik
grafik
grafik

D24V10Fx family
https://a.pololu-files.com/picture/0J5237.1200.jpg?871e92934fd5c93cd85d78aee5264736
grafik
grafik
grafik

D24V22Fx family
https://www.pololu.com/file/0J1031/d24v22fx-step-down-voltage-regulator-dimension-diagram.pdf
grafik
grafik
grafik
grafik
grafik

Symbol PR https://github.com/KiCad/kicad-symbols/pull/1721


jstjst:
> Violating F7.3

I did make the GND Pins Rectangular to match the Pololu board
> Violating F6.2

I think it makes more sense to match the Footprint anchor of the SMD parts with the THT parts for easy swapping

evanshultz:
Welcome and thanks for submitting these footprints!

A couple starter questions just so I'm following you before an official review:
1. How are the SMD version to be attached to the PCB? Simply laying the board on the PCB and filling the through holes with solder?
1. Do you know why pin 1 isn't square in these boards? it baffles my mind why pin 1 isn't square, as can clearly be seen at https://www.pololu.com/file/0J792/d24v5x-schematic.pdf, since that's common convention.
1. Can you explain why it makes more sense to keep the same origin? Using pin 1 for THT and in the middle for SMD footprints is our convention and it I think it is fine here.

jstjst:
>     1. How are the SMD version to be attached to the PCB? Simply laying the board on the PCB and filling the through holes with solder?

Yes, I created this footprint if someone needs to build a low profile board.
>     2. Do you know why pin 1 isn't square in these boards? it baffles my mind why pin 1 isn't square, as can clearly be seen at https://www.pololu.com/file/0J792/d24v5x-schematic.pdf, since that's common convention.

It also surprised me and I have no idea.
>     3. Can you explain why it makes more sense to keep the same origin? Using pin 1 for THT and in the middle for SMD footprints is our convention and it I think it is fine here.

If someone start to layout the board with the THT variant and the decides to switch to SMD, the footprint don't jump all over the place. Thats something, tthat annoys me when I switch footprints.

evanshultz:
For the second and third items, we will stick to KiCad guidelines. Pin 1 is square and the SMD origin is in the center. I understand your point but our library has common rules for this exact reason, which we know doesn't match what every manufacturer chooses. Unless another librarian feels differently, of course.

jstjst:
@evanshultz Should I remove the 3d model paths?
Is this ready to be merged?

evanshultz:
Thanks!

Just with a quick look again and not a full review, but let's get this done to move on:
- Please supply a link in the description and make the description consistent; I guess the product page will be OK since there are many relevant docs.
- Only one 3D model should be linked, and it needs to be just for the model of the footprint and must have the offset, rotation, and scale set to the guidelines of this library. See the CI test results.
- All footprints should have `_THT` or `_SMD` in the name.

myfreescalewebpage:
@jstjst any news following the last comment ?

myfreescalewebpage:
@jstjst ping

myfreescalewebpage:
No news of the author, indicate this is Abandoned

PR #1538, opened 2019-04-07
Oscillator: Add Oscillator_SMD_DiodesPericom_FN-4Pin_7.0x5.0mm

gmsotavio:
Screenshot_2019-04-07_19-00-16

Manf. Doc: https://www.diodes.com/assets/Datasheets/FN_3-3V.pdf

---


evanshultz:
1. Get rid of the "Pericom" part of the footprint name.
1. The fab layer needs a pin 1 mark. This would be easy enough to do with straight sides and one bevel instead of rounded corners, but you can keep the rounded corners too. What ideas do you have?
1. Courtyard should be 0.25mm bigger than the part body or pads. It seems you used 0.5mm? http://kicad-pcb.org/libraries/klc/F5.3/ mentioned canned crystals but this can use the normal SMD rule.

gmsotavio:
@evanshultz, thanks for the suggestions!

>Get rid of the "Pericom" part of the footprint name.

Done.

>The fab layer needs a pin 1 mark. This would be easy enough to do with straight sides and one bevel instead of rounded corners, but you can keep the rounded corners too. What ideas do you have?

I have kept the rounded corners for a while, but something says me that you prefer straight sides with no rounded corners. Drop me a line and I will change this. I have no personal preference. You decide!

>Courtyard should be 0.25mm bigger than the part body or pads. It seems you used 0.5mm? http://kicad-pcb.org/libraries/klc/F5.3/ mentioned canned crystals but this can use the normal SMD rule.

Fixed! In my mind I was drawing a crystal but it is and SMD Oscillator.

Screenshot_2019-04-08_23-33-27

evanshultz:
Thanks! The pin 1 fab line mark varies a lot in this library so we'll go with it. Eventually there could be a standard but that time is not now.

PR #1539, opened 2019-04-09
Added Johanson Balun 1.6 x 0.8 mm filter footprint

ppielorz:
Added Johanson Balun 1.6 x 0.8 mm filter footprint.
Datasheet: https://www.johansontechnology.com/datasheets/0850BM14E0016/0850BM14E0016.pdf

image
image
image

Symbol pull-request: https://github.com/KiCad/kicad-symbols/pull/1725

---


ppielorz:
`Footprint center calculated as (-**0.0025**,0.0)mm`

Really... I can't change it.

aewallin:
silk under component is discouraged by KLC - does the real component have a mark according to the red square in the datasheet?
in any case some (more) silk indicating pin1 outside the component f.fab could be useful?

poeschlr:
baluns should go into the RF_Converter lib.

the footprint name should use the prefix Balun

Silk under SMD parts especially ones this small is a no go as it negatively impacts reflow yield. Look at DFN footprints for inspiration on how to make a pin 1 mark for such parts. (IPC compliant) Example:
Screenshot from 2019-06-01 20-13-58


---

Something is strange with your courtyard outline. Marked with red arrow.
The pad positions in x direction are not really specified in the datasheet. How did you come up with your measurement. Also notice that the pads are not on the same symmetry axis as the part body outline.

Screenshot from 2019-06-01 20-23-35

ppielorz:
Here is package with additional information: https://johansondielectrics.sharefile.com/d-s33398a8d9b243669. I received this link directly from Johanson support. 
There are Gerber files from TI reference design, from where I took X measurement.

poeschlr:
The body is not centered in x direction
Screenshot from 2019-06-03 19-29-54

Pin 1 mark on fab layer should be a chamfer of the top left corner with sidelenght 25% of shorter body side.

I measured the outside to outside dimension in the gerbers as 1.275 (But i do not fully trust that measurement as my tools are not that well suited to measure gerbers.) Meaning the dimensions are generally ok.

Screenshot from 2019-06-03 19-34-33

ppielorz:
One question about courtyard clearance which should be 0.25 mm according to KLC. Which of these is correct? Clearance to Fab layer or to pads?
image

poeschlr:
Clearance is both against pads and the nominal body dimensions (the union of the area taken up by these different polygons.)

ppielorz:
Hi,
I'm sorry, I forgot a bit about this PR...
Last remark corrected I think.

poeschlr:
Thanks

PR #1540, opened 2019-04-10
Add Honeywell_CSLW

cp-aquila:
Add Honeywell_CSLW

---

* Datasheet:  https://sensing.honeywell.com/honeywell-sensing-cslw-series-product-sheet-005861-1-en.pdf
* The Datasheet has barely any measurements and states that its for reference only!
* I could not get additional Information from Honeywell support (even tried calling^^)
* The Footprint is proven to be working on one of my designs
* Dimensions for the Fab/Silk layers were taken by counting pixels in the drawing

grafik
grafik
grafik


poeschlr:
Silkscreen should be outside the part body (the outer outline)
As fab represents the nominal outline this means silk (on the outside) must be outside the fab outline (but still close enough to be helpful. I typically draw it such that the fab and silk outlines are just touching.)

---

A lot of dimensions are missing from the datasheet. How did you determine the measurements i marked orange?
Screenshot from 2019-06-02 14-54-40

cp-aquila:
Hi,

since its a THT component I did follow KLC F5.1-4. The component is not 'massive', you can see trought the coils. The Silkscreen reflects that.

The missing measurements are a mixture of measured with caliper and by counting pixels in the PDF.

As for the 0.6mm diameter of pins, I honestly do not fully remember. I think it was following F7.6 so 0.2mm plus the 0.381mm from the datasheet, rounded up the the next nice drillsize. My 'proven' design used 0.5mm and I was able to assemble the part without hassle.

poeschlr:
It does not matter if the body is see through. All silkscreen must be outside (this is not only for visibility. A robot checking correct alignment might not see the same way as a human. Also the picture in the datasheet does not look like anything is see through. (I am not talking about the inner details. They are specificially allowed for tht components. I talk about the outermost outline.)

cp-aquila:
Potato-quality picture. The board is mounted inside an enclosure, so thats the best picture I can get without tons of effort.
CSLW mounted
This picture does not not show exactly the same FP as discussed here.

poeschlr:
This is not see through. You need to look from an angle which a robot will not do (It looks from directly above. Which is the exact reason for this requirement)

Again move the outside silkscreen towards the outside. End of discussion. (You start to get me annoyed for wasting my time explaining a rule that is set in stone. Not just by us but by industry standards.)

poeschlr:
The courtyard line bottom middle has only 0.12mm clearance to the fab outline (which i assume represents the nominal body dimension as close as we can get it from the limited information provided)
Screenshot from 2019-10-21 20-18-48

cp-aquila:
Yep, that was incorrect. fixed in the above commit.
I also made some changes to silkscreen, it was to close to the exposed copper next to the pins 4&5.

There is no Pin1 marker on F.Fab as required by KLC (F5.2.2). But most other THT-Parts also have none. For example `TO-220-2_Vertical` which falls in the same class of symmetry (along y-axis).
Do you want a dedicated Pin1 marker added to this footprint?

poeschlr:
thanks

PR #1546, opened 2019-04-11
Alps EC11B Horizontal

joemadeus:
Adding footprints for Alps EC11B rotary encoder, horizontal version (plus a circular holes version)

I marked the board edge on the schematic with a datum marker, and (more importantly, I supposed) assumed that this edge falls at the edge of the body of the device. There's a 0.8 x 2.0mm tab "in front" of this edge, that I assumed hung over the side of the board. Nothing on the Alps web site says how we should treat that.

Datasheet/Product sheet: https://www.alps.com/prod/info/E/HTML/Encoder/Incremental/EC11/EC11B15242AZ.html

Alps EC11B
Alps EC11B CircularHoles

---


poeschlr:
There are a few problems with the footprints measurements

Screenshot from 2019-06-04 21-45-03

The copper restring for the mounting pads is a bit small to be honest. (0.15mm is the absolute minimum allowed but i doubt this would be easy to solder. I suggest doubling it.)

---

The hole diameter for the circular version is a bit small. Should be 2.8 to have the equivalent size of the rectangle indicated in the datasheet.

Screenshot from 2019-06-04 21-50-56

PR #1550, opened 2019-04-13
Myrra 48022 AC/DC Converter Footprint

johnfwhitmore:
The Footprint for the Myrra 48022 PCB Mount AC/DC converter.

Datasheet - http://www.farnell.com/datasheets/2339616.pdf

Screenshot from 2019-04-14 00-06-14


johnfwhitmore:
Ping! Sorry wondered about this pull request's status?

poeschlr:
The body outline is slightly wrong.

Screenshot from 2019-06-04 21-09-17

johnfwhitmore:
Screenshot from 2019-06-05 18-24-03

poeschlr:
The chamfer was not the only thing wrong with the footprint. The pink measurements are also wrong. Should be 25mm not 24.7

johnfwhitmore:
Sorry I made the silk screen line the correct length and the other two a bit smaller and a bit bigger. Should have been the inner line was the correct length 25mm.

poeschlr:
The "chamfer" is quite wrong after your latest change. And the courtyard clearance is wrong as well.

PR #1552, opened 2019-04-17
Add Texas_QFN-41_10x16mm

tpambor:
This adds a footprint for the QFN-41 package used by power modules made by TI.

A datasheet of a module using this footprint can be found at: http://www.ti.com/lit/ds/symlink/lmzm33606.pdf

The datasheet of the footprint is available at: http://www.ti.com/lit/ml/mpqf506/mpqf506.pdf

As this footprint is quite complex it was created by hand taking into account the manufacturer recommendations and the KLC.

qfn41_kicad

For comparison I made some screenshots with the layers overlayed on the datasheet. The layers as defined in the footprint of this PR are illustrated using a solid black fill.

Copper:
qfn41_copper_overlay

Soldermask:
qfn41_soldermask_overlay

Solderpaste:
qfn41_solderpaste_overlay
---


tpambor:
The footprint is actually centered. The tool fails to calculate the center correctly because on the top side of the symbol are only custom shaped pads.

poeschlr:
Sadly all my review tools break down for this complex part (Stepup can not import custom pads with more than one primitive)

Also for future endeavors in designing something this complex: https://forum.kicad.info/t/kicad-stepup-the-sketcher-for-footprint-generation/9582 (Might just make your live a lot easier)

Does TI provide anything that could help us? (dxf version of the suggested pad layout, a footprint in eagle format, ...)

If not then i can only make a limited approximated review with kicad internal measurement tools.

tpambor:
@poeschlr The footprint is provided by TI in .bxl format at http://www.ti.com/product/LMZM33606/pinout-quality. I think it can be converted to Eagle or even KiCad but I haven't tried that myself so far. Additionally, at the same page they provide a STEP model if that helps you.

For reviewing I used http://www.ti.com/lit/ml/mpqf506/mpqf506.pdf. The drawings should be to scale if I remember correctly. I imported the PDF into Inkscape and exported the footprint layers to DXF with KiCad, then imported them into Inkscape to create the comparison images. You could also import the PDF with Inkscape and convert them to DXF.

easyw:
@tpambor your fp seems very good to me :smile: 

@poeschlr 
> Sadly all my review tools break down for this complex part (Stepup can not import custom pads with more than one primitive)

what you can do is to import the fp in KiCAD, add it to an empty board and then export it as DXF (aligning the aux origin to the center of the fp).
Then you will be able to import the DXF inside FreeCAD and check all measurements...
and as @tpambor already suggested, this can also checked vs. the STEP model
PS I noticed that KiCAD exports DXF always in decimils, so you need to scale by 25.4 importing the dxf to FC
[Texas_QFN-41_10x16mm-F_Cu.dxf.zip](https://github.com/KiCad/kicad-footprints/files/3245210/Texas_QFN-41_10x16mm-F_Cu.dxf.zip)

.
fp-dxf-to-freecad

As a final note to @tpambor I also would suggest you to have a try to StepUp in doing your complex footprints... I saw your works inside the kicad fp and it is an extremely hard work...
In FC you can import your pdf as a reference image and create sketches for the pads.... all your complex pads would be just a closed poligon do be converted by StepUp to a fp.   

> The footprint is provided by TI in .bxl format

Unfortunately not having the Ultra Librarian sw it is not an easy conversion...
I remember that @erichVK5 has done something at his repo, but I never tested it.
https://github.com/erichVK5/BXL2text
@erichVK5 could you join this conversation?

easyw:
@tpambor @poeschlr 
I've done a deeper look at this fp and I think we should full review it ...
I've tried to compare this fp with the one offered by ultralibrarian tool...
here a pad overview comparison
ti-fp-comparison

The UL fp has all Rectangular Pads (they are instead suggested from Ti as RoundRects), plus some Circular TH Pads for thermal vias.

From the DS I think all the pads should be Roundrects and only Mask and Paste should be shaped as the in the actual fp.
Blu colored -> Metal Pads & Thermal Pads
Green colored -> Solder Mask opening
Red colored -> Paste pattern

complex-fp

What do you think?

poeschlr:
I think your observation is correct.

erichVK5:
> Unfortunately not having the Ultra Librarian sw it is not an easy conversion...
> I remember that @erichVK5 has done something at his repo, but I never tested it.
> https://github.com/erichVK5/BXL2text
> @erichVK5 could you join this conversation?

The BXL converted easily enough with 

translate2coralEDA - which exports a pcb-rnd  .lht  format footprint

and also by

translate2geda - which exports a gEDA PCB .fp footprint

both tools use the same BXL parsing code, which is also used by BXL2text.

gEDA PCB automatically generates solder mask and stencil apertures for pads, whereas pcb-rnd supports both automatic generation or custom polygonal apertures on any layers, but I have not written the code to export custom apertures to pcb-rnd format from BXL, as I have not come across the need previously.

The screenshot below shows a .lht exported footprint on the left, alongside a gEDA PCB .fp footprint on the right. The geometries match the screenshot shown from ultralibrarian in this thread.

kicad  should be able to import gEDA PCB .fp footprints; I attached the gEDA PCB .fp output in case the geometry is helpful for comparison in kicad.

[RLX0041A.fp.zip](https://github.com/KiCad/kicad-footprints/files/3265728/RLX0041A.fp.zip)

RLX0041A-pcb0rnd-lht-from-translate2coralEDA-and-geda-pcb-fp-from-translate2geda

If I find the time, I will have a look at the solder mask aperture definitions in the BXL file, and look at implementing support, which could then be folded back into the BXL2text code as well.

Cheers,

erichVK5

poeschlr:
You guys know that converting a file to a diffent file format does not mean it looses its original license right? Meaning we will most likely not be able to include such a converted file in the official library. (Unless it was under an copy left license that allows re-licensing.)

erichVK5:
> You guys know that converting a file to a diffent file format does not mean it looses its original license right? Meaning we will most likely not be able to include such a converted file in the official library. (Unless it was under an copy left license that allows re-licensing.)

I am not a lawyer, but Bruce Perens, who consults in the field of GPL compliance, commented on this in 

https://hackaday.com/2017/12/21/exporting-eagle-libraries-to-foss-tools/

Conversions are rarely lossless. I think if the pad and aperture geometries are as per the spec from the manufacturer, and that is all that is retained from the conversion, then it would be hard to argue infraction. At any rate, I would not blindly add a converted footprint to the library anyway as it would likely need refinement and fine tuning.

Cheers,

erichVK5

easyw:
@poeschlr @erichVK5 
quoted Bruce Perens:
> By the way, the part libraries are probably not copyrightable. It’s a consequence of 17 USC 102(b):
(b) In no case does copyright protection for an original work of authorship extend to any idea, procedure, process, system, method of operation, concept, principle, or discovery, regardless of the form in which it is described, explained, illustrated, or embodied in such work.
If there is artwork, rather than following of dimensions of a physical part, that could be subject to copyright.

Anyway, 'blx' is available on Ti site to get a library/footprint model to be included in a design... then any information included can be used to check/compare our footprint... 

@erichVK5 
> but I have not written the code to export custom apertures to pcb-rnd format from BXL

 having a bxl converter also for mask and paste would be very useful... 
also a direct conversion to kicad would be very appreciated by many users (me included :smile:)





> At any rate, I would not blindly add a converted footprint to the library anyway as it would likely need refinement and fine tuning.

BTW the UltraLibrarian model is not really accurate...
1) Ti is suggesting to have all Pads as roundrect... UL fp has only rect
2) Ti is detailing solder mask and paste apertures, UL doesn't have this info in the converted fp
3) Ti is detailing the thermal pads, UL fp doesn't have those

From Ti **mpqf506.pdf** document it is possible to get 3 DXF files representing the metal Pads, Solder Mask and Paste apertures... from those it should be possible to complete the fp and check the result comparing this to the **TI_BXL/LMZM33606_RLX_41.bxl** fp.
As I said, it would be nice to have the BXL converter available as a open source converter to kicad format, instead to have to install a proprietary tool, available for windows only.

erichVK5:
> **Edit:** kicad cannot import gEda 'sym' library 

https://github.com/erichVK5/BXL2text

can export a kicad symbol lib. There is an outstanding pull request which improves the slotting in exported kicad libs, but I have not had a chance to test it properly.

easyw:
@erichVK5 would it be possible to add the -o folder option as in your other tool?

erichVK5:
> @erichVK5 would it be possible to add the -o folder option as in your other tool?

OK, done. It seems to work on testing.

easyw:
> OK, done. It seems to work on testing.

it works, only  a minor fix... it requires a '/' at the end of the folder path... it could be automatically added if missing...
 `java BXLDecoder ~/Downloads/LMZM33606_RLX_41.bxl -o ~/Documents/`
desired behavior accept both
`java BXLDecoder ~/Downloads/LMZM33606_RLX_41.bxl -o ~/Documents`

tpambor:
@poeschlr What changes are you requesting? Making the pads RoundRects on the copper layer? From my side it does not make much difference because this footprint is used for power modules and as such most of the pads are nevertheless part of solid copper planes. For example see 
qfn41_example_layout
But if you like I can change them to RoundRects.

I intentionally didn't include the dashed (thermal) copper planes because that would result in violating _F4.3 - Connected copper elements have the same pad number_. As a consequence I also didn't include the thermal vias as many of them would not even be connected to one of the pads.

Any further points I missed from your comments?
Thanks for the review!

poeschlr:
I agree with the review made by @easyw meaning the pending changes tag was added to represent that somebody wants changes.

The roundrect copper pads simply make dealing with the footprint easier for users. And it is suggested that way by TI.

tpambor:
@poeschlr I changed the copper pads to roundrects

poeschlr:
Sorry it took so long. Looks good for me. (I generally trust the reviews made above, the requested changes were made.)

PR #1554, opened 2019-04-20
added Littelfuse FLR series 178.6165 fuse holder

jstjst:
added Littelfuse FLR series 178.6165 fuse holder
https://www.littelfuse.com/~/media/commercial-vehicle/datasheets/automotive-fuse-holders/ato/littelfuse-fuse-holder-ato-flr-pcb-datasheet.pdf (p. 2)
grafik


poeschlr:
Footprint name: replace Automotive-ATO with Blade_ATO.

The silkscreen must be outside the body. (but still close enough to be useful)
As the fab layer represents the body silk must therefore be outside of it. (I typically place silk such that the silk and fab lines just touch. This is not critical but looks kind of nice in kicad and also ensures a consistent clearance between silk and part body)

Courtyard clearance should be 0.25mm all around.

Screenshot from 2019-06-02 15-42-49

PR #1558, opened 2019-04-24
add footprint for Lumberg 1613 14/15 SMT Barrel Jack socket

zgyarmati:
add footprint for Lumberg 1613 14/15 SMT Barrel Jack socket
---

- data sheet:
Data sheet: https://downloads.lumberg.com/datenblaetter/en/1613_14.pdf

- screenshot:
dc_barrel_smt

- 3d model is coming soon


zgyarmati:
Added a commit to fix up the messed up module reference text and 3d model name, please feel free to squash it with the prev. commit

zgyarmati:
And the corresponding 3D model PR is:
https://github.com/KiCad/kicad-packages3D/pull/544

gmsotavio:
Hi @zgyarmati,

It looks like that you have violated the KLC F5.1 (http://kicad-pcb.org/libraries/klc/F5.1/) which states: "Silkscreen must not be placed over pads or areas of exposed copper".

56696736-fc235600-66ec-11e9-9f77-3bd3d41d9605

zgyarmati:
Thanks for the review! Apparently one shouldn't blindly believe in the check_kicad_mod.py's results...

Here is the fixed version: 
dc_barrel_smt_fixes

Please let me know if you notice any other places to improve the footprint.

evanshultz:
Welcome and thanks for contributing!

Indeed, we do manual checking because the scripts can only reasonably do so much. They are usually a good rough test, but we know the limitations and either improve the scripts or live with it.

1. What does the `_15` in the footprint name mean?
1. Please include the datasheet link in the description.
1. Courtyard must go around the pads and part body.
1. For connectors, courtyard offset is 0.5mm.
1. You can leave the silk ref des above the part but please move the fab layer value to the bottom and center it horizontally.
1. The fab layer ref des can be centered.
1. The vertical fab layer line inside the part body isn't needed.
1. The fab layer needs a pin 1 marker.
1. The silk marker is not on pin 1. You may remove silk instead of add it to mark pin 1.
1. There are several dimensional errors:
image

PR #1562, opened 2019-04-27
Added TI SWRA416 PCB antenna footprint

ppielorz:
Added Texas Instruments SWRA416 PCB antenna footprint.
Datasheet: http://www.ti.com/lit/an/swra416/swra416.pdf
Screenshot: 
image

---


poeschlr:
I think dimension H is violated (This should specify where there is no copper allowed if i understand it correctly. Right now it is 11.5mm instead of 12mm)

Screenshot from 2019-06-02 15-56-40



I am also unsure if this even needs a fab layer.

ppielorz:
You are perfectly right.
I've corrected dimensions of keep-out zone and Courtyards. Also, I have removed Fab layer as you pointed.

poeschlr:
looks good, thanks

PR #1564, opened 2019-04-28
Add everlight irm h238 t tr2

roger225:
add Everlight_IRM-H238T_TR2 footprint
image

image

image

datasheet: http://www.everlight.com/file/ProductFile/IRM-H238T-TR2.pdf


roger225:
Wait for 3D model PR, https://github.com/KiCad/kicad-packages3D/pull/545

roger225:
Gnd pins are pin number "1"
image

evanshultz:
The 3D model doesn't need to be merged first. They are separate, and if there's no footprint the 3D model won't have any "anchor" so it actually makes the most sense to merge the footprint first.

- Can the footprint name be more general? Do more than just the IRM-H2XXT/TR2 parts use this footprint? For sure, the description shouldn't mention a more generalized range of parts using the footprint than the footprint name indicates.
- Let's change the description to `Everlight IRM-H2XXT/TR2 Infrared Receiver Module, 4.6x5.1x1.45mm, http://www.everlight.com/file/ProductFile/IRM-H238T-TR2.pdf`. This depends on the item above.
- The text positions are odd; could you move them to a more coarse grid? They can all be horizontally centered and the fab layer ref des and be centered in the body outline.
- Keep pin numbers 1-4. This particular part has GND for pins 1 and 4, but Everlight may use this package for other parts where that is not true. This goes along with the first item above.
- You can remove the square on the fab layer inside the body outline. Because the bevel is already there this doesn't help to orient the part during placement.
- Let's get rid of the silk line sticking out from the body. Instead, draw silk lines around the part edge, avoiding pads by at least 0.12mm and offsetting them from the fab lines so silk isn't under the part, and then you can choose to either add silk to mark pin 1 (similar to what you've done) or remove silk around pin 1.
- Round rect pads should be the minimum of 0.25mm radius or 25% of the pad edge; yours are too big.
- Would you please rotate the pads so they're 0 degrees?

roger225:
Hi @evanshultz,

Please help to review it.

Thanks!!
Roger

evanshultz:
Closing and re-opening to refresh the CLA check...

evanshultz:
1. Can you please answer my first bullet above? Is there a more general name you can see?
3. The silk ref des can be horizontally centered. Fab ref des can be bigger. All can be on a more coarse grid (like 0.5mm) but they don't have to be.
6. Please clean up the silk as mentioned above.

roger225:
> 1. Can you please answer my first bullet above? Is there a more general name you can see?
The IRM-H236T_TR2 shares the same footprints, I think IRM-H2xxT_TR2 would be fine.
> 2. The silk ref des can be horizontally centered. Fab ref des can be bigger. All can be on a more coarse grid (like 0.5mm) but they don't have to be.

- Ref des centered.

- Fab ref des change size to 1mm.

- "All can be on a more coarse grid (like 0.5mm) but they don't have to be.", I can not understand this, could you explane more about this? Thanks!!

image

> 3. Please clean up the silk as mentioned above.
My vender think these silk are helpful for engineering, may I keep these silk? 
image

Sorry for the late response.

evanshultz:
@roger225 
If you're still around and interested in getting this done, please let me know.

By the coarse grid, I meant that the value is at [0 4.445] and [0 4.5] would be fine.

And yes, please remove those silk lines.

PR #1565, opened 2019-04-28
Added AA battery holder - Keystone 2460

ppielorz:
Added complete AA holder footprint.

Datasheet:
https://www.keyelco.com/product-pdf.cfm?p=1025

Screenshot:
image

---


poeschlr:
The datasheet requests two additional non plated holes (approximately added in screenshot).
The body is too wide and the courtyard clearance in y direction is not quite right (will be fixed by fixing the body. Unless i am already too tired to do math in my head.)

Screenshot from 2019-06-02 22-23-07

ppielorz:
I have corrected to your remarks. However the courtyard clearance was too small in x (0.15 instead of 0.25). After correcting y the courtyard was ok.

poeschlr:
You changed the previously correct x direction and did not quite hit the y direction. (pink to show wrong right now) The courtyard in y direction will become ok when fixing the body (hence orange) The courtyard in x direction will become wrong once fixing the x body direction (hence the use of dark red)

Screenshot from 2019-06-03 19-05-32

---

The outer silk outline must be fully outside the nominal body dimension. This means outside the fab outline as it represents the nominal body. The offset is not really critical should however be small enough to be still useful. (I typically draw silk such its edge just touches the fab outline as it is easy to draw that way and looks nice in kicad.)

ppielorz:
Footprint ready for review.

poeschlr:
The last commit only changes one end of one courtyard outline.
The errors in the body dimension are still as reported above. The silk screen is still as reported above.

poeschlr:
Why did you move the non plated holes around? They were correct in the past now the hole to hole distance in x direction is wrong.
Size in y direction is still wrong.
Screenshot from 2019-11-03 10-16-35

ppielorz:
Those non plated holes were a not exactly in right place so I moved them, but moved right one too mush to right - I corrected it.
Also now size in y direction is ok - I had to add 0,005 mm. This is why courtyard has 0,26 mm in y direction because it has to be on 0,01 mm grid.

poeschlr:
The courtyard would have 0.255 not 0.26 otherwise it would not be on grid

poeschlr:
looks good now. thanks

PR #1566, opened 2019-04-28
Add XT60PW

ottojo:
This adds the XT60PW (Horizontal PCB-mount variant of the XT60 plug) footprints.

---


evanshultz:
Two questions before I really review:
1. Are there other pin counts in this family?
1. How did you come up with these? It looks like a DXF import or something.

Also, please don't force push commits since that makes reviewing harder. Thank you.

poeschlr:
Rotate the -F version by 180 degree to get pin 1 to the top left corner.

The silk outline must be outside the nominal body outline. As the nominal body is represented by the fab outline this means silk must be outside the fab outline. (It is not too critical but the outline should not be too far away such that it is still useful. I typically draw the silk outline just touching the fab outline. This makes it easy to draw, looks nice and has for sure enough clearance.)

Is there a reason for the silk outline to stop where it stops?

Add pin 1 markers on fab and silk.

A few dimensions are off. Particularly the hole sizes. See: http://kicad-pcb.org/libraries/klc/F7.6/ and http://kicad-pcb.org/libraries/klc/F7.7/ for details

PR #1571, opened 2019-05-03
Added LED footprint

SchrodingersGat:
Footprint for LiteOn LTST-C19HE1WT LED

https://www.digikey.com/product-detail/en/lite-on-inc/LTST-C19HE1WT/160-2162-1-ND/4866310

https://optoelectronics.liteon.com/upload/download/DS22-2008-0044/LTST-C19HE1WT.pdf

image

In action with aligned 3D model 

image


poeschlr:
The chamfer on fab looks a bit large (should be 25% of shortest dimension)
Could you move the silk lines a bit outwards? The clearance to copper should be at least the same as the silk line width.

Courtyard clearance in x direction is too large.
Screenshot from 2019-06-02 21-54-36

SchrodingersGat:
Changes made, thanks for the review Rene

SchrodingersGat:
Bump - this footprint has now been used in production FYI

poeschlr:
I seem to have missed that the footprint name does not include the manufacturer. Could you add it?
Looks good otherwise.

SchrodingersGat:
Good catch - I have fixed this now. Cheers

poeschlr:
Thanks

PR #1572, opened 2019-05-03
Added UDFN-9 package

SchrodingersGat:
See https://media.digikey.com/pdf/Data%20Sheets/Semtech%20PDFs/RCLAMP3328P.pdf

Reference part: https://www.digikey.com/product-detail/en/semtech-corporation/RCLAMP3328P.TZT/RCLAMP3328P.TZTCT-ND/5423770

Symbol PR to come after this is merged.

image


evanshultz:
- A bit unnerving that Semtech doesn't have the datasheet on their website...
- I can't reconcile the pitch on page 1 with the two drawings on page 7. The pitch of each column, and if that pitch is consistent for the entire column, doesn't add up for me. Am I misunderstanding? (And as there may be two different pitches I'm unsure about the `P0.5mm` part.)
- Courtyard should be +/-0.98mm and +/-2.15mm.
- The body is +/-0.5mm wide, so the silk needs to be a bit wider than +/-0.6mm.

evanshultz:
OK. All of the other datasheets in the symbol PR have better dimensioning. Could you please change the link? And perhaps review the recommended footprints while you're at it.

The pin placement isn't right. Pins 1 and 5 should be Y=+/-1.7mm. And pins 6 and 9 should be Y=+/-1.3mm. Adjust the silk to fit the new pin placements, please.

SchrodingersGat:
Using the https://katalog.we-online.de/pbs/datasheet/824014881.pdf datasheet as it specifies a land pattern as well as the package.

evanshultz:
Travis puked. Weird.

Courtyard X should be +/-1.03mm.

Here is the footprint now:
image

The pad clearance is 0.12mm in the pic above and you can see the silk-to-pad distance isn't consistent and sometimes is too tight. Can you please update that too?

SchrodingersGat:
Changed courtyard and silk as requested

evanshultz:
Clearance is ~0.18mm. Could be smaller but this should be OK. Thanks!

PR #1576, opened 2019-05-04
Adding footprint for relay Fujitsu FTR-B3S.

persus:
Adding footprint for relay Fujitsu FTR-B3S.

---


antoniovazquezblanco:
Please, add a screenshot of the footprint and a link to the datasheet.

Thanks!

persus:
image

https://www.fujitsu.com/downloads/MICRO/fcai/relays/ftr-b3.pdf

poeschlr:
The pin 1 mark on silk does not quite fit with IPC suggestions that we try to use now. (For parts with non protruding smd leads pin 1 is marked by a missing line. As an example see the DFN footprint)
Screenshot from 2019-05-31 18-00-41

The courtyard in x direction is defined relative to the pads because they protrude from the body in this direction. Courtyard must be 0.25mm all around the part (rounded to 0.01mm grid)

The fab pin 1 mark should be 0.25% of shortest side or at most 1mm (In your case 1mm)

Screenshot from 2019-05-31 18-03-50

persus:
Hi. Sorry I abandoned this pull request. If there is a way to reactivate this pull request, I would like it. Two of your changes I did correct (I think).
- [X] For parts with non protruding smd leads pin 1 is marked by a missing line.
- [X] The fab pin 1 mark should be 0.25% of shortest side or at most 1mm (In your case 1mm).
- [ ] The courtyard in x direction is defined relative to the pads because they protrude from the body in this direction. Courtyard must be 0.25mm all around the part (rounded to 0.01mm grid)

The last one I do not understand. Maybe it is possible you explain it to me in german?

persus:
Like this?

image

poeschlr:
Courtyard clearance in x direction is now ok. It is however still 0.4mm in y direction instead of the required 0.25.

persus:
Like this?

image

poeschlr:
Yes looks good now, thanks

PR #1581, opened 2019-05-06
Add Texas DRB

evanshultz:
http://www.ti.com/lit/ds/symlink/ucc24610.pdf#page=37

Since it's quite unique I built it just the way Texas says in their doc, by hand.

image

So that the outer pin 9 pads overlap I enlarged and moved them slightly. This means there is no radius to the big pin 9 pad but I think this should be OK.

image

Unfortunately, KSU can't import the paste layer so that's not included above.

@poeschlr 
Did you notice that FreeCAD can't load the DrawingDimensioning workbench (or I'm doing something wrong) so it can't be used to do the dimensioning you figured out?

---


poeschlr:
The drawing dimension workbench is not really maintained. I think the tech draw workbench will be the future. (But i do not really like it in its current implementation as it is a bit strange from the interface side. Also i kind of liked the option to use colored dimensions to highlight problems.)

The footprint looks good, thanks.

PR #1585, opened 2019-05-09
Variant of the HSSOP-16-1EP part

SchrodingersGat:
- Based on the LM43603 datasheet
- http://www.ti.com/lit/ds/symlink/lm43603.pdf
- Mask aperture 2.4x3.3mm
- Stencil aperture is 65% of mask size

image


poeschlr:
I would highly suggest the use of the IPC script for creating footprints for such a standadized package: https://github.com/pointhi/kicad-footprint-generator/tree/master/scripts/Packages/Package_Gullwing__QFP_SOIC_SO

serkanxselcuk:
Hi @SchrodingersGat ,

I think pad positions does not match with datasheet by HSSOP-16 Serie. (Distance between pad 1 and 16 should be 5,8mm, but it is 5,55 on footprint)
image
Can you fix them too, if you are still working on this fp?

poeschlr:
@serkanxselcuk  We will use the IPC generators anyway so the suggested footprint will be ignored.

myfreescalewebpage:
@SchrodingersGat are you still working on this ?

myfreescalewebpage:
Indicate Abandoned because no news of the author (on this PR and others)

PR #1588, opened 2019-05-09
Fix hole recommendation

hildogjr:
Second the manufacture datasheet https://omronfs.omron.com/en_US/ecb/products/pdf/en-g5q.pdf the recommend hole are 1.3mm D and not 1/0.76mm D (with difference between coil and power pins).


evanshultz:
Same deal. Please use 2.3mm diameter pads.

Issue #1560, opened 2019-04-26
[WIP] Xilinix BGA footprints


herostrat:
This is a tracking issue to compile the work of footprints for Xilinix BGAs.

**1. Artix-7**
 - [ ] Script PR: https://github.com/pointhi/kicad-footprint-generator/pull/343
 - [ ] Footprint PR: https://github.com/KiCad/kicad-footprints/pull/1557

**2. Spartan-7**
 - [ ] Script PR: 
 - [ ] Footprint PR:
_Other Initiatives:_
Script PR: https://github.com/pointhi/kicad-footprint-generator/pull/344
Footprint PR: 

**3. Kintex-7**
 - [ ] Script PR: 
 - [ ] Footprint PR:

**4. Virtex-7**
 - [ ] Script PR: 
 - [ ] Footprint PR:

**5. Zync**
 - [ ] Script PR: 
 - [ ] Footprint PR:
_Other Initiatives:_
Script PR: https://github.com/pointhi/kicad-footprint-generator/pull/240
Footprint PR: https://github.com/KiCad/kicad-footprints/pull/1167

Additional Work:
 - [ ] Remove old and/or wrong packages that reference to Xilinix FPGAs

evanshultz:
Thanks so much!

Is there any more renaming that needs to take place to ensure all applicable package names are captured in the footprint names?

herostrat:
Afaik those were all BGA packages Xilinx uses for the 7-series chips, so no, I don't think so.

The remaining work is cleaning up the "old" packages (as you previously stated according to https://github.com/KiCad/kicad-footprints/pull/1167#issuecomment-447585654) and close similar PRs.

evanshultz:
@herostrat 
https://github.com/KiCad/kicad-footprints/pull/1167 and https://github.com/pointhi/kicad-footprint-generator/pull/240 have a section named "Duplication of existing parts". Are you open to checking that out to see if those are footprints that are unneeded? It appears they are not covered in the list above.

herostrat:
Yes, I will, this will be the next workpackages for me.
The ones listed above are just the starting point I already identified.

PR #458, opened 2018-03-27
Bga676

vlunyov:
BGA-676 was added (based on BGA-672).
All links to 3D model, PDF document etc were changed to correct state.
It is usefull for XILINX FPGA Artix 7 and Kintex 7 users.
https://www.xilinx.com/support/documentation/package_specs/fg676.pdf

PR #366, opened 2018-02-11
added 256-ball caBGA footprint for iCE40-HX8K-CT256

cyber-murmel:
according to Lattice PCB Layout Recommendations for BGA Packages (March
2017; Technical Note TN1074) page 4 and Package Diagrams (November 2017)
page 82

Thanks for creating a pull request to contribute to the KiCad libraries! To speed up integration of your PR, please check the following items:

- [x] Provide a URL to a datasheet for the footprint(s) you are contributing
* [Lattice PCB Layout Recommendations for BGA Packages](http://www.latticesemi.com/~/media/LatticeSemi/Documents/ApplicationNotes/PT/PCBLayoutRecommendationsforBGAPackages.pdf?document_id=671)
* [Package Diagrams](http://www.latticesemi.com/-/media/LatticeSemi/Documents/PinPackage/PackageDiagrams.ashx?document_id=213)
- [X] An example screenshot image is very helpful
screenshot of PcbNew
- [x] If there are matching symbol or 3D model PRs, provide link(s) as appropriate
https://github.com/KiCad/kicad-packages3D/pull/262
screenshot of the 3D model
- [x] Check the output of the Travis automated check scripts - fix any errors as required

PR #1597, opened 2019-05-15
move BGA footprints that are a dublicate of Xilinx BGAs to obsolete 

herostrat:
This PR moves multiple footprints that are a dublication of Xilinx BGA footprints to the obsolete folder.

Reasons, why they are not kept but removed:
 - The description mentions Xilinx footprints
 - I could not find a generic device that fits the specification of the footprint
 - There are no symbols using the footprints

The footprints are not generated with the script generator.
I will try to find the original PR for each footprint and verify that they are only meant for Xilinx BGAs and are therefore obsolete.

Tracking issue with additional information:
https://github.com/KiCad/kicad-footprints/issues/1560


evanshultz:
@herostrat 
It seems nobody looked at this. Did you get a chance to find the PR where these footprints were originally merged?

evanshultz:
It was ugly to find out the history of these, with several name changes, so I did a quick look:
- BGA-1156 / BGA-400 / BGA-484: originally from https://github.com/KiCad/SMD_Packages.pretty and present at the first earliest commit visible in the web interface (https://github.com/KiCad/SMD_Packages.pretty/commit/96de441ab61fdcf404354867f4f51c962b2175bc) but they're been renamed many times.
- BGA-256: For Lattice FPGA from https://github.com/KiCad/kicad-footprints/pull/366
- BGA-256*NSMD: Xilinx-specific from https://github.com/KiCad/kicad-footprints/pull/45.
- BGA-676: For Xilinx from https://github.com/KiCad/kicad-footprints/pull/458.
- FB-BGA-484: Xilinx-specific from https://github.com/KiCad/kicad-footprints/pull/219

IMHO if these aren't used by any current symbols they aren't currently needed by the library from a myopic perspective. We aren't using the `NSMD` suffix since that's the default for our library. And any existing designs will have the footprint integrated into the board file so we aren't at risk of breaking any existing designs. This PR makes it clear why this was done and how to re-create any footprint desired (using the generator) so I'm actually in favor of just deleting them to clean things up. The Lattice one might be needed by a pending symbol PR since there have been some symbols submitted but not merged, but I didn't dig into that. And if so, a generated footprint would be better.

PR #1615, opened 2019-05-29
Wuerth horizontal smt mounting hardware

poeschlr:
Made manually (There are only two of them which differ only in the hole size of the 3d model.)

See: https://katalog.we-online.com/en/em/2ND_TAXONOMY%24WP-SMRA_WOPINS
- https://katalog.we-online.com/em/datasheet/7466300.pdf

Screenshot from 2019-05-29 20-47-57

Screenshot from 2019-05-29 20-43-01


chmorgan:
@chschlue this looks like that part that has no obvious library. Maybe 'Mounting' as a library?

chschlue:
Mounting_Wuerth doesn't seem unreasonable to me, they manufacture more than enough of the stuff to justify an extra lib.
That lib has been added in #1612, so why not use it?

chmorgan:
@poeschlr it looks like this is a part not yet in the Mounting_Wurth library. Would you be interested in rebasing against master to confirm the library warning is gone after the creation of the Mounting_Wurth library and that the latest travis checks pass?

poeschlr:
Rebased and here is also a dimensioned drawing of the footprint
Screenshot from 2020-04-01 19-06-15

PR #1616, opened 2019-05-29
added NHD-C0216AZ

rs-development:
https://www.mouser.de/datasheet/2/291/NHD-C0216AZ-FN-GBW-18835.pdf

I will add all other things when done.


poeschlr:
A screenshot would give us at least a first impression.

poeschlr:
The silk outline is a bit far away for my taste.
Screenshot of the footprint for future reference.
Screenshot from 2019-06-04 22-15-35

The arrow as pin 1 marker does not quite fit the style we typically use. (Especially as the fab layer is somewhere in the middle of the part far away from the pin) Here an example (The fab layer shown is IPC compliant with size 1mm)
Screenshot from 2019-06-04 22-20-32


---

On the right side there should be a semicircular protrusion of 0.8mm to the body. At least make the body size larger by that amount or include a rectangular protrusion. (The optimal solution would of course be the use of an arc but it is not required to get it merged.)

---

The courtyard outline has the wrong clearance (should be 0.25 all around)
Screenshot from 2019-06-04 22-27-16

PR #1617, opened 2019-05-30
DFN_QFN: Add DHVQFN-{16,24} packages

smunaut:
https://assets.nexperia.com/documents/outline-drawing/SOT763-1.pdf
https://assets.nexperia.com/documents/outline-drawing/SOT815-1.pdf

Signed-off-by: Sylvain Munaut <tnt@246tNt.com>


antoniovazquezblanco:
In the PR template there were instructions on how to collaborate. It would help us if you could follow them in order to make the process easier.

Please post an image of your footprints, thanks!

smunaut:
Sorry. there your go :

2019-05-30_843x793_scrot
2019-05-30_1019x801_scrot

smunaut:
I had missed the KLC requirement of having pin 1 on the top left. Updated footprints and images.

2019-05-30_591x854_scrot
2019-05-30_601x858_scrot

poeschlr:
IS there a reason to using square pin 1? You might want to know that using rounded ends is not done to make it prettier but to increase reflow solder behavior. (Meaning changing the pad shape to rectangle will negatively impact solderability)

---

Also IPC (industry standard that defines land patterns) suggests the use of rounded rectangle with 25% round radius (maximum radius of 0.25mm which will not apply here) instead of oval. It even acknowledges that many manufacturers will continue to suggest oval. (Which was used in the past as there was more tool support for this shape. rounded rectangle where introduced quite recently to a lot of tools. KiCad included.)

smunaut:
* I made pin 1 square because of some suggestion I saw at kicon that making pin 1 different in copper was the best way to make it super obvious of the orientation. But if that's an issue, I'll change it back to be consistent with the others.
* I used oval mostly because I based it off an existing footprint, but it might have predated rounder rectangles support.

smunaut:
Updated with all pads being roundrect 25%

2019-05-30_583x857_scrot
2019-05-30_728x891_scrot

poeschlr:
I checked the smaller of the two footprints. Sadly there are a lot of differences between the footprint and datasheet. Unless you have good reasons for this please fix it.

Also the courtyard clearance is too close to the pads in a lot of places.

Please double check the large footprint as well as i fear it will be in a similar state.

Screenshot from 2019-06-04 22-06-09

PR #1618, opened 2019-05-31
7 seg footprint

Solartraveler:
Add footprint for SA39-11, SC39-11, SA39-12, SC39-12 10x13mm Kingbright 7 segment LED displays 

Symbol https://github.com/KiCad/kicad-symbols/pull/1896

---

The footprint is based on an other 7segment display from the same lib. 
http://www.kingbrightusa.com/images/catalog/SPEC/sa39-11ewa.pdf

The outline allows placing the segments on a 2.54mm grid without overlap.

7segment-footprint


poeschlr:
Your body size on the fab layer does not agree with the datasheet.
Screenshot from 2019-06-02 11-42-22

Solartraveler:
Okay, I changed the fab layer, by reading the guidelines, it was not clear to me, fab must fit the size exactly. I assumed the fab layer must represent the geometry somehow (rectangle, round...) and it would be good not to be overlap it with fsilk.

poeschlr:
Fab measurements are now ok but the courtyard clearance must be 0.25mm relative to the nominal body size (as fab represents the nominal body size this means you need 0.25mm relative to fab)

The silk outline must be outside the nominal body size. Again as fab is the nominal body this means it must be outside the fab outline (i typically draw it such that silk is just touching fab, i do this because it is easy to draw that way. In reallity silk is completely uncritical it should be outside the nominal body but close enough to be still of use for manual assembly and end of production checks.)

Solartraveler:
Okay, I read through the datasheet and there are indeed tolerances of +-0.25mm. So having a countryard clearance of 0.25 makes sense. I will find time to fix this next week.
However they manufacture with less tolerance (I measured three spare ones with a caliper, and all have a width within 10.00mm and 10.05mm). 
When building my PCB, I placed them on a 2.54mm grid (=10.16mm -> 0.16mm space between two digits) and was happy there is no bad looking large gap between them. I still wonder how with such tolerances you can make a large display out of a 5x7 dot matrix display such as
http://www.kingbrightusa.com/images/catalog/SPEC/TA07-11CGKWA.pdf
because with a countryard of +0.25mm, there would be the choice between a bad looking display or ignoring the complains of the DRC check.

poeschlr:
The courtyard clearance has nothing to do with tolerances of the part. It comes from industry standards that ensure automatic processes can assemble your pcb and that it can be repaired.

And an answer from a similar question over on the forum:
A DRC violation does not necessarily mean something can not be produced. It can also mean: “I think this could be a problem would you please make sure this is ok?” Of course a flag for “Do not tell me this one error again” would help here but that is something for the next iteration.

I do not think that this in general will ever change. (How would i tell DRC that courtyard validations are ok if the neighboring part is one of a subset of allowed footprints.)
So no matter what changes in the future there will be either a false positive or you need some special footprints. For special footprints that avoid the false positives you have the following choices.
Option 1 would be to see a combination of multiple such terminal blocks as a single component (makes sense as it will most likely be assembled before being placed onto the pcb. -> will however limit the information you get via the bom. A special house part number could however encode which exact parts to by to achieve a connector of this specific length.)

option 2 is to have footprints for left end, middle and right end that define the courtyard such that they do not overlap with each other.

option 3 is to only draw the “middle” option from above and accept that the ends are not properly checked with courtyard DRC.

Also i feel this would have been a good question to ask in the open as i think you will not be the only one who has this specific question.

Solartraveler:
Okay, I hope, I finally fulfilled the requirements.
I came across option 2 too, and while it may work with 7segment displays with digits left and right, resulting in 4 combinations, it would get confusing with dot matrix displays having 16 possible footprints to select the right one.
I wonder why there is a countryard with clearance drawn at all. Why not store the clearance as extra value or category 0603, connector, canned capacitor,... and let Pcbnew calculate the adjusted countryard. This should be easy for polygons and the user could overwrite the clearance per part when needed or select an alternate clearance for parts with the same case. It would also make creating parts more easy and less work for the reviewer. But you are right, this might be better done in a general discussion.

poeschlr:
I am not aware of any tool that allows generation of courtyard definitions at design time. Every tool i used so far had it defined as a fixed polygon in the footprint. This is also how IPC describes it making me believe that a vast majority of tools do it that way. You can however always create a wishlist bug report. I think two iterations would be feasible. Iteration 1 is a way for the designer to tell kicad "this violation is ok. do not list it again" and iteration 2 could be the automatic courtyard definition calculated from the fab outline and pad definitions. Iteration 3 could then even allow for different clearance values depending on direction.

poeschlr:
It seems you used a mail address for the commits that is unknown to github. This trips up the licence agreement script we use. The footprint would look good to go now.

Solartraveler:
Okay, is the dummy commit enough, or do I need to repeat all commits to be associated with the useraccount? Also cla-assistant.io seems not to work right now. I just get an endless "login with github" loop.

Solartraveler:
Sorry for hitting one problem after the next one. But I simply don't get it. The webpage says signed, but the check here says no.

cla-footprint-not-signed

chschlue:
Not a cla-assistant expert, but I think all commits need to be linked to the Github account cla-assistant connects to and it looks like your first six commits are not (telling by the Octocat instead of your avatar next to them)

Solartraveler:
Obviously, by using
https://help.github.com/en/articles/changing-author-info
successfully to comply the license check, I messed things up another way...
I will look into another bunch of tutorials the next days in the hope to find a solution.

chschlue:
Can't you just add the email address on your first commits to your Github account?

chschlue:
The easiest way out of this mess might be a new branch (and new PR), you only want to add one file after all.

Closing for now, reopen if you really want to clean this up instead.

Solartraveler:
Okay, I just created a new branch with a new PR: https://github.com/KiCad/kicad-footprints/pull/1866

PR #1619, opened 2019-06-01
Footprint for 3 pin 5.08mm Pitch connector

johnfwhitmore:
This is a simple wire to board connector bit it might be a bit "fat" I've taken it from a part on farnell:

Screenshot from 2019-06-01 13-19-12


poeschlr:
This is a terminal block. The Connector wire lib is for directly soldering wires to the board.

johnfwhitmore:
Oops sorry, I'll close this and check that other library, The necessary component is probably already in there. If it isn't I'll create a Pull Request for that library. Sorry again.

PR #1623, opened 2019-06-03
Con 1x3 p5.08 drill1.4

johnfwhitmore:
Addition of the Multicomp 3 position 5.08mm Pitch terminal block footprint

Screenshot from 2019-06-03 16-19-51


poeschlr:
The KLC has a specific section about terminal block naming. Please take a look at it. (You can also use other terminal blocks as guidance)

johnfwhitmore:
Sorry about the name, I was moving it across from where I had the footprint previously in the Wire library. Think it's good to go now.

poeschlr:
Same as with the 2pol version,

Name should be `TerminalBlock_Multicomp_MC000049_1x03_P5.08_Horizontal`

Either add more detailed silk drawings or indicate on which side the wire will enter in some other way (The part is slightly asymetrical so it kind of matters. And it will definitely matter if ever a 3d model is added.)

Screenshot from 2019-06-04 21-32-30

johnfwhitmore:
Screenshot from 2019-06-05 18-46-37

johnfwhitmore:
Screenshot from 2019-06-06 21-37-20

johnfwhitmore:
Does that kinda suggest where the wire inserts into the terminal block? To be honest I wasn't even sure looking at the data sheet which side the wire accessed the part. Let me know what you think. Some of the other terminal blocks used a lot more lines but not sure they'd add any information.

poeschlr:
Sorry for the long wait. Yes the wire entry markings look good. I am however not convinced of your way or marking pin 1. We do not use writing to do that in the lib. It is also not visible after assembly. So i suggest to remove it and place an L shaped second silk line outside the top left or bottom left corner plus a chamfer of the fab outline on the same corner.
Example:
Screenshot from 2019-09-20 13-37-11


The courtyard clearance should be 0.5mm all around. It is 0.3mm right now.

PR #1624, opened 2019-06-03
Added Multicomp 2 position 5.08mm Pitch terminal block

johnfwhitmore:
Addition of the Multicomp 2 Position 5.08mm Pitch Termainlblock

Screenshot from 2019-06-03 16-47-45


poeschlr:
Name should be `TerminalBlock_Multicomp_MC000048_1x02_P5.08_Horizontal`
Screenshot from 2019-06-04 21-19-13

pin 1 markers on fab and silk are missing.

Either add more detailed silk drawings or indicate on which side the wire will enter in some other way (The part is slightly asymetrical so it kind of matters. And it will definitely matter if ever a 3d model is added.)

johnfwhitmore:
I'm not sure what part of that diagram is asymmetrical. The terminal block can be soldered on so that the wire is coming in from the top or bottom, depending on rotation of the component?
Screenshot from 2019-06-05 18-40-04

poeschlr:
the part has a width of 10.2 mm but the center of the lead is at 5.4mm from the back (slightly off center making it assymetrical. And again the 3d model will definetly be assymetrical as the part can only accept the wire to enter from one side)

johnfwhitmore:
Aragh!!! Sorry, I drew it symmetrical but it ain't :( I'll correct both related symbols.

johnfwhitmore:
I've added a possible change on the 3 position version and I'll copy into this symbol if suitable.

evanshultz:
@poeschlr 
Would this be best scripted?

johnfwhitmore:
I'm sure it probably would be. 

Screenshot from 2019-06-11 21-09-19

poeschlr:
I am sure it can be scripted. But as we do not have a ready made script that only needs dimensions entered i would not go so far as to require it.

poeschlr:
The courtyard clearance should be 0.5mm all around.
Screenshot from 2020-06-11 20-15-37

PR #1627, opened 2019-06-05
added qfn32 with 3.5mm EP

rs-development:
i tried the internal wizard but i seems to be not creating footprints similar to the existing qfn32 footprints, so i derived the footprint from the 3.6mm EP version.

[https://infocenter.nordicsemi.com/pdf/nRF52811_PS_v1.0.pdf](https://infocenter.nordicsemi.com/pdf/nRF52811_PS_v1.0.pdf)

Since no QFN32 with EP size of 3.5mm existed i created one to be able to merge the nrf.

image

image

[https://github.com/KiCad/kicad-symbols/pull/1874](https://github.com/KiCad/kicad-symbols/pull/1874)


poeschlr:
Footprints for packages like these should be generated using the script found here: https://github.com/pointhi/kicad-footprint-generator/tree/master/scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON

rs-development:
So i was creating the footprint and i recongnised in the config file that the EP 3.6x3.6mm Version is actually designed by the reference for the nRF52810 QFN32 package. This is the exact same package as the nRF52811. So i guess we should use the same package for the nRF52811 or we should change both packages, can you please check!

http://infocenter.nordicsemi.com/pdf/nRF52810_PS_v1.1.pdf#page=468

rs-development:
Any feedback for this?

poeschlr:
Yes these are the same packages. My guess is that 3.6mm overwrite is used because there is a application note that suggests it that way or because it simply was that way before it was scripted. (the manufacturer suggestions vary quite a bit for how large the pad should be in relation to the package lug. Everything within minmum and maximum dimensions of it is reasonable.)

PR #1630, opened 2019-06-06
Fix courtyard clearances for scripted no lead footprints with pull back pins

poeschlr:
script fix: https://github.com/pointhi/kicad-footprint-generator/pull/369


evanshultz:
Assuming script is good and merging. Thanks!

PR #1631, opened 2019-06-07
Added HRO TYPE-C-31-M-12

lorenz:
Adds the HRO TYPE-C-31-M-12 receptacle. This is a Type-C receptacle without the high-speed lines designed for USB2.0 and USB-PD. [Datasheet](https://datasheet.lcsc.com/szlcsc/1903211732_Korean-Hroparts-Elec-TYPE-C-31-M-12_C165948.pdf)
USB_C_Receptacle_HRO_TYPE-C-31-M-12

---


lorenz:
A few things that still need to be fixed:
- [ ] Handling joined pads (A1/B12, ...)
- [ ] Fix fab layer
- [ ] Check center (I'm pretty sure it is centered, but on the physical part, not the footprint, unsure what IPC says about connectors)

diegoherranz:
Thanks for your contribution. I hope it's the first of many!

Things I've checked that look OK:

- Overall dimensions on Fab OK
- Pads locations and sizes including slots OK

Things which need some changes:
- [ ] The X location of the 2 NPTHs seems to be wrong.
- [ ] Can you add a link to the datasheet/drawing at the end of the description separated by a comma. Is there a link to the manufacturer rather than a distributor?
- [ ] The reference designator on the Fab layer should have the value `%R`. See http://kicad-pcb.org/libraries/klc/F5.2/ Can you place it at `(0,0)` too?
- [ ] The clearance on the courtyard should be 0.5mm around the combined component body and land pattern boundaries. See http://kicad-pcb.org/libraries/klc/F5.3/
 - [ ] The silkscreen gets too close to the S1 pads. Clearance should be at least 0.2 mm. See [KLC F5.1](http://kicad-pcb.org/libraries/klc/F5.1/). 
- [ ] Regarding the center, on [KLC F6.2](http://kicad-pcb.org/libraries/klc/F6.2/) we say "Footprint anchor should be placed in the middle of the component body". The Travis checks use [padMiddlePosition()](https://github.com/KiCad/kicad-library-utils/blob/master/pcb/kicad_mod.py#L722) which doesn't return the centroid of the connector so we can ignore it this time. However, I think we should use either the centre of the main 7.3 mm x 8.94 mm body or the centre of the body plus the pins, 7.60 mm x 8.94 mm. And it doesn't seem to be either at the moment? Could you use one of them?
- [ ] Regarding the stacked pins A1/B12, etc. I'd like to handle it in a way not to need a new symbol, be able to use the current [USB_C_Receptacle_USB2.0](https://github.com/KiCad/kicad-symbols/blob/489265dc083c33092b6e2c312f1cb92bdeb9ff29/Connector.lib#L15010) and still have the footprtint filter working (same number of pins). Could you place copies of the pads on top of the existing ones with the other names? Does that sound acceptable, @KiCad/librarians?

Thanks again for your contribution!

lorenz:
Thanks  for your review :smile:

I've fixed issues 1, 2, 3, 5 and 6 (by centering on the connector).

Regarding issue 4: I'm not sure if I agree that we need a 0.5mm courtyard on the whole connector. I've put 0.5mm on the mechanical sides, but I've kept 0.25mm on the pad side since 0.5mm would waste space there.

I've thought of putting pins on top of eachother, but it renders very badly. We could have touching separate pins if KiCAD does not complain. They are stacked on the symbol anyways.

poeschlr:
The KLC requires equal courtyard to both the pads and body. For connectors it is set at 0.5mm. There will not be a discussion about it at this point in time.

diegoherranz:
Thanks for the changes @lorenz.
I think it is all OK now except for the courtyard. Could you make it 0.5mm all around to both the pads and body? The current clearance is small at the moment on the north side and a tiny bit too big on the other sides.

If you find 0.5mm to be too much, you can always ignore it and get closer when designing a board, but it really helps with the consistency of the library.

Also, a very minor point (optional): The reference designator on the Fab layer is not exactly at `(0,0)`. This is not a requirement and it is OK as is but if you're making other changes it could be tweaked.

Thanks!

lorenz:
@diegoherranz I've fixed the courtyard, should now be 0.5mm on all sides (+- alignment). I've also centered the designator.

diegoherranz:
Thanks for the changes.

The north side looks good now.
The other 3 sides have a clearance a tiny bit too big (0.53mm left and right and 0.55mm at the bottom side) because of the grid/alignment as mentioned by @lorenz. I'm not sure reading the convention whether it has to be exactly 0.5 mm or at least 0.5 mm. @poeschlr?

Everything else is good to go.

Thanks!

diegoherranz:
This looks basically good so it would be nice to include it in the library.
Would you fix the courtyard comment, @lorenz? Or, @poeschlr, is it OK to accept a slightly bigger clearance?

Thanks!

lorenz:
@diegoherranz Courtyard is now exactly 0.5mm, so should be good now.

diegoherranz:
Thank you for your patience. It all looks good to go now.
I know we can be very picky sometimes, but I believe that in order to maintain a big library, used by many users, consistency is really a plus.

Thanks for your contribution!

evanshultz:
I don't think the "USB Type-C"-ish symbol is needed. We don't do that on other USB footprint and I think it falls into clause F5.2.1.b. of KLC at http://kicad-pcb.org/libraries/klc/F5.2/.

Issue #462, opened 2018-03-31
KLC: Silk under SMT part bodies


evanshultz:
@poeschlr @jkriege2 @Ratfink @Misca1234 @Shackmeister 
From  https://github.com/KiCad/kicad-footprints/pull/442.

I think KLC needs a bit of clarification about silk going under pads. Perhaps it can be allowed if the package body does not touch the PCB. If the purpose is to prevent silk affecting coplanarity, this does not apply if the part body does not touch the PCB surface where the silk is.

If we still don't want silk under the body, even if the body doesn't touch the PCB, that's OK. But would you also look at the silk pin 1 mark for BGAs and see if there's a standard style we want to adopt? Please take a look at the IPC doc I linked in the PR above.

Ratfink:
I agree that it shouldn't be a problem to have silkscreen under a surface mount device where the body doesn't touch the board.  This could also allow silkscreen lines under gull-wing leads in some footprints.  Frankly, I thought this rule was a bit too strict ever since KLC 1.0.

With regards to BGA ball A1 style, I'd prefer the suggestion of removing the silk lines on both sides at the corner, but I wouldn't object to putting a chamfer under the corner if others prefer.

evanshultz:
For ease of reference, here is the IPC doc: http://www.ocipcdc.org/archive/What_is_New_in_IPC-7351C_03_11_2015.pdf#page=17 (pages 17, 19, 22).

I like removing both sides in the pin A1 corner too. If the other librarians agree I'll volunteer to make the changes to those footprints.

And KLC should also be updated for both items: silk can be under a SMT body if the body doesn't touch the PCB surface, and also provide an example of our preferred BGA pin 1 silk mark. @poeschlr how do you want to handle KLC updates? Oliver would usually make the updates himself, sometimes after soliciting feedback for proposed wording. I feel there are quite a number of KLC updates discussed here and there but few updates have been made recently. What method do you prefer to make KLC updates?

poeschlr:
I am not really sure i understand the proposed silk marker for bgas. could one of you post a screenshot that shows the newly proposed look?

Ratfink:
@poeschlr I'm at the wrong computer right now to make an example footprint, but it would be like this image, except both silkscreen lines would be cut back at the top left corner instead of just the one.

![](https://user-images.githubusercontent.com/4781841/38120284-75f8bb9a-3413-11e8-98ce-87a61a77e9fb.png)

[...]

PR #1634, opened 2019-06-08
Added new NTC Varistor footprint

johnfwhitmore:
Not 100% sure about location in library. All existing footprints have pins offset from centre line whereas this footprint has the two pins inline. 

http://www.farnell.com/datasheets/2131309.pdf

Screenshot from 2019-06-08 09-03-15


johnfwhitmore:
Screenshot from 2019-06-08 09-12-14

poeschlr:
Sorry for the late review.
Move silk outside the nominal body dimensions.
The silk reference must be outside the outline (at the place where the fab value field is right now)
fab value field should be on the bottom
Add the datasheet link to the documentation. Is there a datasheet available at the manufacturers website?

evanshultz:
@poeschlr 
On your first point: why? http://kicad-pcb.org/libraries/klc/F5.1/ does not diallow silk under THT parts. Or it's unclear or I misunderstood. I thought I had written an issue about this but I can't find it now. So in case somebody else can find it, I won't create an new issue just yet...

chschlue:
@evanshultz 
https://github.com/KiCad/kicad-footprints/issues/462?
Partly about THT silk.

poeschlr:
Additional silk can be under the body. but the main outline must still be outside.

PR #1636, opened 2019-06-10
added FFC molex series 502231

Shackmeister:
Datasheet: https://www.molex.com/pdm_docs/sd/5022312400_sd.pdf

OBS: The lock is sticking out on one side, since I need the space under I just added it in cmts.user.
If prefered I can add this to the fab layer.

Travis might complain about the center of the footprint, the center is estimated from the datasheet and measured using FC

Skærmbillede fra 2019-06-10 15-40-48

---


evanshultz:
@Shackmeister 
There were handmade, I presume?

Shackmeister:
Yes b)

evanshultz:
The ground pads aren't really a shield, right? More like a mounting point. So 'MP' instead of 'SH'?

All dimensions check out for me:
image

@poeschlr 
Are you OK using Cmts.User for this purpose? Makes sense to me since there's no an immovable element there that must always be avoided, but only something to watch out for at a height above the PCB surface that allows small components to fit. With the current limitations of KiCad, visual 3D checking might help but I don't think it's right to use F.Fab.

Shackmeister:
The pads are connected to shield of the flat Flex cable

evanshultz:
Ahh. OK.

Two more things:
1. The center is shown as, I guess, 2.21mm below the edge of the pin tips in the datasheet. I don't know exactly where the pin tip lays on the pad, but the top edge of the pins in the footprint are 1.73mm (1.045mm pad center + 1.37mm tall pads / 2). So it seems like if you were targeting that center from the datasheet the pin tips will stick off the pads. Have I misunderstood?
1. Pad corner radius should be 25% up to 0.25mm. The SH pads have too much corner radius.

Shackmeister:
1. I used the image in page 4 showing the pick place center, did some measurements in FC on the 3D model, and moved it from there.

2. Fixed the corner radius to 0.25mm

evanshultz:
OK. We'll go with that. Thanks!

PR #1639, opened 2019-06-10
Add switch KMR2 with ground pin

mibe:
CK KMR2 tactile push button with ground pin

There's a version of the (already existing) KMR2 push button in the `Button_Switch_SMD` library with ground pin, which is added here. 
[Datasheet, see page 2](https://www.ckswitches.com/media/1479/kmr2.pdf?page=2)

I wasn't sure which name to use for that pin, so I left it empty, just like in footprint `SW_SPDT_PCM12`.

KMR2xxG

---


antoniovazquezblanco:
I believe that the ground pin is shielding. We should probably use SH for the ground pad name in order to be able to connect that pad to any other signal. Maybe other librarians may prefer MP for mounting pad. When a reviewer is assigned to your PR, the reviewer will help you with the naming.

Thanks!

mibe:
Should the footprint without the ground pin renamed to `SW_Push_1P1T_NO_CK_KMR2xxNG` as per datasheet?

antoniovazquezblanco:
I think so but it would probably be a good idea to do it in a separate PR just in case lead librarians want to merge it in the V6 release cycle.

Thanks!

poeschlr:
The ground pin is indeed closer to a shield pin than a pure mounting pin. So yes SH as the pad number is preferable.

As name i would then suggest `SW_Push_1P1T-SH_NO_CK_KMR2xxG ` (include SH in the pin definition section)

The existing footprint might indeed be best kept with the legacy name till we start with the version 6 reorganization.

mibe:
Thank you for your inputs @poeschlr.

> The ground pin is indeed closer to a shield pin than a pure mounting pin. So yes SH as the pad number is preferable.

I've numbered that pad SH now.

> As name i would then suggest `SW_Push_1P1T-SH_NO_CK_KMR2xxG ` (include SH in the pin definition section)

I renamed the footprint (and also fixed that reference to the 3D model, which was still using the original).

> The existing footprint might indeed be best kept with the legacy name till we start with the version 6 reorganization.

Is there a tracking issue or something like that to add the original footprint to?

BTW, does anybody know where this center offset Travis complains about comes from?

poeschlr:
The switches (and buttons) will receive a major update with v6. (not only naming convention but also pin numbering) Meaning there really is no need to separately document which footprints violate the KLC right now as we will go through this whole lib during the v6 upgrade anyways.

---

The center is calculated using the pads which leads to false positives for non symetric parts. This is a known weakness of our checks.

---

Thanks for your contribution.

PR #1646, opened 2019-06-13
Cmc

evanshultz:
Thai Lin TL22
https://www.thailin.com.hk/wp-content/uploads/2013/02/029-032-Encapsulated-Common-Mode-Filter-Rated-250Vac.pdf
image

On this one the pins used don't start with 1 (the four outer pins are totally missing), so I kept that and skipped pin 1.

Kaschke RDS 1200
https://www.kaschke.de/en/products/inductive-components/common-mode-toroid-chokes/
image

On this one, the drawing is a bit hard to read but the horizontal pin spacing is 15mm.

On both, the outer radius of the plastic body is rather small so I just used a corner.
---


evanshultz:
Dimensioned drawings:
image

image

evanshultz:
image

image

poeschlr:
The drawing you have on fab looks like a fuse. Is this intentional?

evanshultz:
Yes, it does look like a European fuse. This is symbol is shown in the TDK datasheet, for example, and is also common on several CMC footprints in our library.

Is it necessary? It any kind of drawing whatsoever desired? It's really unnecessary IMO so if you'd rather not have them I'll take it away from this PR and happily clear out similar stuff from existing footprints. IMO it falls under the "no excessive elements on the fab layer" clause of KLC.

poeschlr:
Sorry for the delay i kind of missed that you responded.

On the symbol side we use the coil like symbols for inductors so it might be better to use something similar here as well.

evanshultz:
@poeschlr 
Would you object to removing this silk? That would be my preference. If so, I'll create another PR removing similar silk from other footprints.

poeschlr:
What do you mean with removing silk? Do you mean the coil markings because they seem to be on fab not on silk (in your screenshots)

I am not opposed to remove these markings but can see the benefit of including them (might help while laying out the board as well as when debugging.)

evanshultz:
Yes, I mean fab. (At least for this one; some footprints may use silk.) Many of them are made of tons of polylines instead of just a few arcs or have different styles. In a separate PR, I'd put in the work to commonalize them if a style was agreed upon.

That all being said, it shouldn't stop this PR from being merged if it's ready.

PR #1649, opened 2019-06-16
update Hirose DF63 connector series

jhalmen:
This updates the [Hirose DF63 connector series](https://www.hirose.com/product/document?clcode=&productname=&series=DF63&documenttype=Catalog&lang=en&documentid=D31622_en).
Committed changes:
 - correct single pin footprint width
 - add mounting hole to single pin footprint
 - correct footprint orientation in the name (vertical)
 - specify `DF63M` designation because Hirose started introducing this instead of the `DF63`
 - introduce the `DF63R` version, which has the mounting holes on the opposite side

This was done by working on the footprint generator script in https://github.com/pointhi/kicad-footprint-generator/pull/376.


evanshultz:
Thank you!

We want to handle things just a bit differently because keeping the libraries continuous between major releases (5 to 5.1 to 6) is a mandate. So...
1. Please fix the DF63 single-pin footprint, if you will. (This could be done by making a new 'M' footprint and manually editing it.)
2. Add the new R and M footprints using the script you've fixed and updated.

This will allow existing designs using the DF63 to continue without losing any footprints while fixing the truly broken single-pin footprint. Then the new part names are introduced so users looking specifically for them will be covered. Does that sound OK?

jhalmen:
Does this look like what you had in mind?

evanshultz:
Almost. For item 1 above the manually fixing means removing the `M` inside the file so that it matches the file name. See the CI errors.

evanshultz:
Also, the datasheet shows the `R` version only goes up to 5 pins, not 6.

evanshultz:
@jhalmen 
Are you able to wrap up the couple things above?

jhalmen:
should have thought a bit about what i'm doing before i did it.
of course, i'm just a bit swamped currently, i hope i can get to it today

jhalmen:
that took some tries :sweat_smile: 

hope this is good now

evanshultz:
Thanks!

One question left is that the connector is Vertical but the error in the script makes the footprint name say Horizontal. Due to the script's current state, it's only generating non-suffix footprints for pincounts of 5 and 6. DigiKey shows the 5 and 6 pins as Active while smaller pincounts are Last Time Buy, so is this intentional?

jhalmen:
yes it's intentional, all footprints for the vertical active parts are being generated this way.

the footprints currently in kicad that have the erroneous 'horizontal' in their name would have to be deprecated somehow. the script has been updated to for the correct orientation

evanshultz:
Yay! Only `Hirose_DF63-1P-3.96DSA_1x01_P3.96mm_Horizontal` seems to need work as it's fixes have been reverted.

jhalmen:
all fixes? that `top entry` i did by hand, because i didn't want the 1P footprint to now differ from all the currently included footprints in the tags. the rest should still be fixed though

evanshultz:
Oh. Gotcha.

I think this is good then. Thanks for all your time and effort!

PR #1650, opened 2019-06-18
Schurter 0031 8201 fix

evanshultz:
https://us.schurter.com/bundles/snceschurter/epim/_ProdPool_/newDS/en/typ_OGN.pdf

Changes:
- Corrected fab dimensions (were 0.1mm too small on all sides).
- Silk offset to 0.11mm.
- Removed curved courtyard lines.
- Though KLC says periods are allowed in footprint names, I did not correct the `0031-8201` to `0031.8201` (change hyphen to period) because that would break designs using this footprint.

Old:
image

New:
image

image

---


cpresser:
This looks like a easy one, I checked with the footprint editor:
- [x] Pad distance 22.5, Drill 1.3
- [x] Body Dimension 9.6x22.5
- [x] Courtyard 26x10.1 (26 = 22.5 +3 + 0.5; 10.1 = 9.6 + 0.5)
- [x] Center-Hole drill is 2.7mm nominal
- [ ] Silk Clearance to copper 0.5mm, should be 0.2mm
- [ ] No Pin1 marker on F.Silk or F.Fab. But that is rather an issue with KLC since they are not required for non polarized components.
- [x] Datasheet link works
- [ ] Description could be more verbose "fuse block with insulation wall, 500V, 16A"
- [ ] One thing I am unsure about is the 10.8mm Dimension from the Datasheet. What does this refer to?

poeschlr:
Having larger silk to copper clearance is ok (For smaller parts going down till clearance equal line width is also allowed.)

This is a non polarized part so it does not require a pin 1 mark

I checked the 3d model. There is no feature that would fit the 10.8mm.

evanshultz:
@cpresser @poeschlr 
- Moved silk closer to pads (~0.2mm).
- Changed description to `Fuseholder horizontal open, 5x20mm, 500V, 16A, Schurter 0031.8201, https://us.schurter.com/bundles/snceschurter/epim/_ProdPool_/newDS/en/typ_OGN.pdf` (what is above is not the description I see).
- I understood that the body is 9.6mm wide and the cover is 10.8mm, but the drawing is badly done so that's not for sure. (I believe cover is aligned with the body but should be moved to the right.) At any rate, the body is 9.6mm and that appears to be correct.

Thanks for looking! Anything else?

PR #1651, opened 2019-06-19
Add Kycon kdmix-sl1-ns-ws-b15

evanshultz:
http://www.kycon.com/Pub_Eng_Draw/KDMIX-SL1-NS-WS-B15.pdf
image

Dimensioned drawing (the front mounting hole location is wrong, 1.25mm instead of 1mm, but I didn't want to redo the whole thing just for that):
image

And here is the latest dimensioned drawing showing that corrected location:
image

---


evanshultz:
CI errors:
- Travis is checking for 0.2mm silk-to-pad clearance, apparently. I used just >0.12mm here.
- The annular rings are sized as the datasheet suggests. They could be bigger, but I counter that if a user if designing an HDMI board their PCB fab requirements are higher and this isn't a problem.

poeschlr:
I suggest increasing the silk clearance for such a huge component. (It really is not necessary to go this close with it and an increased clearance means more people can use it as is)

Should we introduce a special orientation specifier for connectors oriented like this? (phoenix for example uses right angle for MCO 1,5/ 3-GR-3,81 vs the normal horizontal for the ones we have in the lib like MC 1,5/ 3-GF-3,5)

Dimensions look good.

evanshultz:
I assume you mean silk-to-pad clearance. Begrudgingly backed off to 0.2mm.

A special modifier sounds fine to me. Do you have any ideas? Maybe `_VerticalRightAngle`?

evanshultz:
@poeschlr ?

poeschlr:
Sorry for missing this. Yes the suggested suffix is ok by me

evanshultz:
@poeschlr 
Sounds good. Done.

I think the one remaining Travis error is fine as I described above.

poeschlr:
thanks

PR #1652, opened 2019-06-19
Add TE 826576

evanshultz:
https://www.te.com/commerce/DocumentDelivery/DDEController?Action=srchrtrv&DocNm=826576&DocType=Customer+Drawing&DocLang=English
image

This is a pretty basic connector, so it can be more generic instead of manufacturer-specific if you want. Let me know.

It's also so simple and script-based so I didn't dimension it. See https://github.com/pointhi/kicad-footprint-generator/pull/384.

Edit: https://github.com/KiCad/kicad-footprints/issues/1141 says the series name can be omitted if the MPN starts with the series name. These do if the pin count is less than 10, so for consistency I left it off.

---


poeschlr:
Most of these are considered obsolete. Should we really include them?
(For reference only 2, 3, 5, 6, 7, 8, 9, 20 and 36 pin options are considered active)

There is a contradiction between the dimension L in the table and what the formula would suggest (for 9 pol: table=35.1 while formula would result in 35.65)

evanshultz:
I am using the 3 pin version. There is nearly zero overhead to make the entire series so I just did that. Many other components are similar and could share this footprint (https://www.te.com/commerce/DocumentDelivery/DDEController?Action=srchrtrv&DocNm=640383&DocType=Customer+Drawing&DocLang=English and http://suddendocs.samtec.com/prints/fwj-xx-xx-tx-s-xx-mkt.pdf and https://www.molex.com/pdm_docs/sd/026602091_sd.pdf), but since it's a manufacturer-specific one maybe it's not very discoverable? It's up to you if you want to reject all or some of them.

Good catch on the dimension conflict. Naturally, I used the equation to determine body width. Since the difference is small and the equation is bigger, maybe that's OK?

poeschlr:
As you have the 3 pin version maybe you can measure it and see what is closer.

Also i would really not include the obsolete versions in the official lib. Maybe you could add a flag to the generator that allows also generating the obsolete versions should a user desire them.

evanshultz:
Sure. I can get the measurements and report back here tomorrow.

evanshultz:
So on the 3-pin version, the table shows 11.3mm and the equation gives 11.88mm.

On the two parts I have, including the breakaway burrs remaining on the edges, they measure 11.06mm and 11.08mm, which fits within the +/-0.3mm tolerance of the table. And excluding the breakaway burrs I measure ~10.9mm which would support the table values with that tolerance being realistic and correct values.

There is no tolerance on the equation so maybe it's a max? Since the drawings seems to include some of the breakoff in the pincount * 3.96 equation that perhaps support it being a max?

We get roughly the table values if we use an equation of pincount * 3.96mm - 0.59mm.

I'm not sure where that leaves us. What do you prefer?

poeschlr:
I would use the table values to be honest. I guess the formula is simply wrong 

To me it looked strange that they had pin pitch times pins as there is a small section between the symmetric pin "sections" that is missing at either end. (the reason why i even looked at the table to begin with)

evanshultz:
@poeschlr 
OK. Done. Easiest to check the last commit in the script PR.

poeschlr:
thanks

PR #1657, opened 2019-06-21
Add Adafruit_Feather footprints

matthijskooijman:
One version with mounting holes, one version without.

See https://learn.adafruit.com/adafruit-feather-m0-basic-proto for the basic info on an example Feather board, and https://learn.adafruit.com/adafruit-feather/feather-specification for details on the layout of Feather boards in general. The corresponding symbol is added [in another PR](https://github.com/KiCad/kicad-symbols/pull/1933).

Pin numbering is
fairly arbitrary (counterclockwise with the USB connector down, like the
similarly sized Arduino Nano), since the part does not have its own pin
numbering (in the Adafruit schematic, both pin headers are numbered separately).

Since there is no well-defined pin 1 on the part itself, I have left out
pin 1 markings from Fab and Silk layers. I did put a square pad on the
copper layer, though. The assymmetric nature of the pin headers should
make it clear how to rotate the board when inserting it, even without an
obvious pin marking.

Checking gives one warning: The version with mounting holes has the same 3D model name as the one without, since I'd expect that the mounting holes would not be represented in the 3D model when it is created (just the pin headers).

image

image


matthijskooijman:
I've rotated the footprint 180°, so the USB connector is up, as suggested in https://github.com/KiCad/kicad-symbols/pull/2473#issuecomment-659327413

myfreescalewebpage:
Hello @matthijskooijman starting the review of this PR:

- [X] No comment on the pads, holes and Fab lines
- [ ] The clearance between the F.SilkS and F.Fab layer should be 0.11mm (there is another Adafruit ESP8266 in the lib for example)
- [ ] The clearance between the F.CrtYd and F.Fab layer should be 0.25mm (there is another Adafruit ESP8266 in the lib for example)
- [ ] Even if there is no real pin 1 on the real device, we need to indicate pad 1 in Kicad on F.Fab layer with a triangle and on F.Silk layer with a single small line for example. See "BeagleBoard_PocketBeagle" for a good example.

Joel

myfreescalewebpage:
Personal note: symbol waiting this PR : https://github.com/KiCad/kicad-symbols/pull/2473 and https://github.com/KiCad/kicad-symbols/pull/1933

matthijskooijman:
@myfreescalewebpage, thanks for your review so far.

>  The clearance between the F.SilkS and F.Fab layer should be 0.11mm (there is another Adafruit ESP8266 in the lib for example)

Where is this specified? I couldn't find anything in https://kicad-pcb.org/libraries/klc/F5.1/ or https://kicad-pcb.org/libraries/klc/F5.2/

Also, the Adafruit_HUZZAH_ESP8266_breakout symbol you refer to (I think) seems to use 0.12mm center-to-center between F.SilkS and F.Fab? So haven't fixed this yet. If you can confirm the value to use, I'll fix this as well.

>  Even if there is no real pin 1 on the real device, we need to indicate pad 1 in Kicad on F.Fab layer with a triangle and on F.Silk layer with a single small line for example. See "BeagleBoard_PocketBeagle" for a good example.

Ok, done.

>  The clearance between the F.CrtYd and F.Fab layer should be 0.25mm (there is another Adafruit ESP8266 in the lib for example)

Ok, done.


There are a few Feathers that use a slightly different footprint
 - https://learn.adafruit.com/adafruit-feather-m0-wifi-atwinc1500/pinouts is slightly longer. Pins have the same position, but two mounting holes are moved.
 - https://learn.adafruit.com/adafruit-feather-32u4-fona is even longer. One mounting hole is moved, another is omitted.
 - https://learn.adafruit.com/adafruit-feather-m0-radio-with-lora-radio-module has four extra pins.
 - https://learn.adafruit.com/adafruit-feather-32u4-radio-with-lora-radio-module/ has three extra pins.
 - https://learn.adafruit.com/introducing-the-adafruit-wiced-feather-wifi/board-layout has one extra pin.

These should be added as separate footprints, but I'd rather wait until the primary one is reviewed (and maybe merged).

myfreescalewebpage:
> Where is this specified? I couldn't find anything in https://kicad-pcb.org/libraries/klc/F5.1/ or https://kicad-pcb.org/libraries/klc/F5.2/

You're right, it's not a strong value, 0.11mm is the one used when we are generating the footprints with scripts. I request you to change because you have actually 3 decimal digits (resolution 0.001mm) which is too much). You can ue 0.12mm if you want, I just indicated 0.11 because it's more common generally.

> Also, the Adafruit_HUZZAH_ESP8266_breakout symbol you refer to (I think) seems to use 0.12mm center-to-center between F.SilkS and F.Fab? So haven't fixed this yet. If you can confirm the value to use, I'll fix this as well.

0.12mm is also accepted of course

> Ok, done. There was little space for the triangle, so it overlaps the pad, is that ok?

F.Fab can overlap the pads, so it's ok. Thanks

> There are a few Feathers that use a slightly different footprint

The current one proposed in the PR seems to be the most common one, so the naming is not a trouble for me here, I suggest to have a specific naming for the next Feather board. What do you think ?

matthijskooijman:
> You're right, it's not a strong value, 0.11mm is the one used when we are generating the footprints with scripts. I request you to change because you have actually 3 decimal digits (resolution 0.001mm) which is too much). You can ue 0.12mm if you want, I just indicated 0.11 because it's more common generally.

Ah, I see. Would be good to put this in the KLC maybe. I'll see if I can provide a PR for that.

I updated the PR with 0.11mm clearance now.

> The current one proposed in the PR seems to be the most common one, so the naming is not a trouble for me here, I suggest to have a specific naming for the next Feather board. What do you think ?

Yes, that's what I was thinking as well.

matthijskooijman:
> Ah, I see. Would be good to put this in the KLC maybe. I'll see if I can provide a PR for that.

See https://gitlab.com/kicad/services/kicad-website/-/merge_requests/507 and https://gitlab.com/kicad/services/kicad-website/-/issues/275#note_419749587

myfreescalewebpage:
You have two travis errors to be corrected: the F.SilkS indicating pin 1 should have a width of 0.12mm

myfreescalewebpage:

matthijskooijman:
> You have two travis errors to be corrected: the F.SilkS indicating pin 1 should have a width of 0.12mm

Sneaky that travis shows as succesful even with warnings, I'll remember that for the next time. Also, fixed my default silkscreen width to be correct (I see that the KiCad nightly already has its defaults fixed as well).

Anyway, footprint updated again, should be good now, but the travis build seems to have checked a lot of other footprints as well, maybe because I force pushed? Ah, I see you forced a refresh, thanks :-)

myfreescalewebpage:
Maybe because of force push I'm not sure, but for the futur please do not fore push except if you are locked because of git. Force push prevent to make the  review step by step.

myfreescalewebpage:
No more on the footprint, merging.

matthijskooijman:
> Maybe because of force push I'm not sure, but for the futur please do not fore push except if you are locked because of git. Force push prevent to make the review step by step.

Yeah, I know, but in this case I had "fixed" the 3D model and then just after pushing realized it was intentionally different, so I removed the commit again. Could have reverted it, but did not seem relevant here.

Thanks for your extensive reviews and merging!

PR #1658, opened 2019-06-23
New Footprint Converter_DCDC_Murata_MGJ2DxxxxxxSC_THT for MGJ2DxxxxxxSC

atari91:
New Footprint for the Murata MGJ2 DCDC Converter series

---

[datasheet](https://power.murata.com/pub/data/power/ncl/kdc_mgj2.pdf)
grafik
https://github.com/KiCad/kicad-packages3D/pull/574
Checking footprint 'Converter_DCDC_Murata_MGJ2DxxxxxxSC_THT' - No errors
No rule violation


evanshultz:
@atari91 
Thanks for contributing to the official KiCad libraries!

1. Please add some extra info to the description, along with what you have, like `19.5x9.8x12.5mm, 5.2kVDC Isolated`.
1. Have you used this footprint? Even at the max lead sizes 1mm holes might be OK, but the datasheet gives a range. I'm wondering if the middle or top of the range is better than the bottom.
1. Oddly, the chamfer on the fab layer is 1.01mm. Can you fix it?
1. Can you make the silk offset from the fab layer consistent? Usually we use a 0.11mm offset (so fab and silk lines just touch) even though KLC allows silk under THT footprints.
1. Courtyard should be offset 0.25mm from the fab lines or pins (in this case the fab lines will dominate).

evanshultz:
Thanks for the updates.
1. The left vertex of the top fab line is -1.12mm and it should be -1.13mm in the X direction.
1. Courtyard offset is still too big as noted above.

atari91:
I finished the footprint.
Thanks also for the second review. Next time, I will place some more information into the title of the commit,  because https://github.com/KiCad/kicad-footprints/pull/1658/commits/29e24ab9fbb4b8b600261902f918cadeddc8c4d1 was an intermediate result as mentioned in the description.

About the hole size: I implemented this footprint also in Altium Designer and made some PCBs with these settings. I never have had problems therewith.

grafik

`Checking footprint 'Converter_DCDC_Murata_MGJ2DxxxxxxSC_THT' - No errors`

evanshultz:
Looks good to me. Thanks for this contribution! Hope to see more from you.

evanshultz:
Probably should have requested to name it just `Converter_DCDC_Murata_MGJ2`. I guess this is OK too.

PR #1659, opened 2019-06-24
Added 9V battery holder BA9VPC made by MPD

mibe:
[Datasheet](http://www.memoryprotectiondevices.com/datasheets/BA9VPC-datasheet.pdf)

The datasheet is unfortunately not really complete, e.g. the dimensions on the top and bottom right notches are missing completely (top and bottom left in the screenshot). I used a caliper to measure them.
For maximum compatibility with the PCB manufacturers I used normal THT pads instead of oval / milled slot ones.

BA9VPC

---


poeschlr:
It might be better to use oval holes for the pads. You can additionally add the one with the round ones but give it the suffix CircularHoles.

I would further suggest to slightly increase the clearance between the pads and silk. (no need to be this close for such a large component)

The datasheet has quite a few missing dimensions. Where did you get them from?
The outline and courtyard are not quite right
Screenshot from 2019-09-24 19-58-07

mibe:
First of all, sorry for replying so late! 😬 

> It might be better to use oval holes for the pads. You can additionally add the one with the round ones but give it the suffix CircularHoles.

Done. I'll happily make a footprint with circular holes, as soon as my fix for the layer problem below is approved.

> I would further suggest to slightly increase the clearance between the pads and silk. (no need to be this close for such a large component)

Done that, too.

> The datasheet has quite a few missing dimensions. Where did you get them from?
> The outline and courtyard are not quite right

Indeed. I had the part at hand, so I could measure the missing dimensions with a caliper. To be honest, I'm a bit in a Love-Hate relationship with the Fab / Silk / CrtYd layers. I hope the newest commit fixes that.
In the meantime, I've had a PCB made with this footprint, the fit is looks good to me. I've had no problems placing the part:

poeschlr:
Oval drills are best combined with oval pads (or elongated rectangle pads) instead of circles and squares.

Add a mark for pad 1 on silk such that it is visible after assembly (ease of debugging) and one on the fab layer.

Hole sizes are not quite right.

Screenshot from 2020-02-05 21-29-46

mibe:
> Oval drills are best combined with oval pads (or elongated rectangle pads) instead of circles and squares.

Fixed.
 
> Add a mark for pad 1 on silk such that it is visible after assembly (ease of debugging) and one on the fab layer.

Also fixed. I used an arrow symbol on the fab layer, since it is not a IC, but more a connector (where the battery is connected to).

> Hole sizes are not quite right.

Fixed, too. The motivation behind the rounded up hole sizes was that the PCB manufacturer probably doesn't have a 2.29 mm drill bit, but a 2.3 mm bit no doubt...

poeschlr:
thanks

PR #1662, opened 2019-06-25
Jst b2p3-vh

evanshultz:
B2P3-VH is popular as an AC connector where earth ground is not required so I wanted to generate it
http://www.jst-mfg.com/product/pdf/eng/eVH.pdf
image

See page 4 for post-omitted variants of this header. This version, 3 pin with pin 2 omitted, falls into 3) which is named "when the pitch is set again" so I changed the pitch. I also used pin numbers 1 and 2 so that it will work with the standard 2-pin symbols in KiCad. Let me know if either choice is undesirable.

Since there was a minor correction in an external file which the script uses and affects all footprints, I pushed all that come from this script.

Script PR: https://github.com/pointhi/kicad-footprint-generator/pull/389

---


poeschlr:
pad 1 should have a different shape than circle (roundrect or rect)

evanshultz:
@poeschlr 
You are right. Updated footprint and screenshot above.

poeschlr:
thanks

poeschlr:
thanks

PR #1664, opened 2019-06-26
TE 928492 series

evanshultz:
I can't find the datasheet online but here is the one I got from TE:
[ENG_CD_928492_B3.pdf](https://github.com/KiCad/kicad-footprints/files/3331008/ENG_CD_928492_B3.pdf)

Here is the smallest and largest members of the family:
image

Why? Well, I'm documenting an older product that used this family of connectors.

@poeschlr 
Maybe you, or somebody else, can check my understanding of the datasheet to make sure I didn't misunderstand the German.

---


herostrat:
Hallo evanshultz,

sorry for the late reply...

One question regarding the so called variants. 
I am not able to identify where in the datasheet a assignment of component and "leiterplattenvariante" (PCB-variant) A or B is made.
Did you use your PCB as a reference?

Note:
In triangle 15 and 16, there is a reference to A and B. Those are not references to variants, but to chamber types as defined in "Teilschnitt C-C".
A translation would be:
15: 2-6 poles: all chambers type "A"
16: 7-12 poles: starting from position 1, chamber "A" and "B" alternating.


In addition, the F.Fab outline should imo be 16.4 mm (compared to your 17.4 mm), defined in:
grafik

evanshultz:
Thanks for looking!

Cross-section (Teilschnitt) C-C is rotated 90 degrees from A-A and B-B and thus couldn't show the pin stagger. Only A-A and B-B can capture pin stagger. Note that all three cross-sections are of the drawing at the far right. I'm not sure how else to interpret this but now I'm concerned this is wrong.

I know the 4 pole version comes with inline pins. Or at least an inline pin version was available.

The 16.4mm dimension is height, which isn't captured by fab lines. Look at Schnitt D-D to find the body depth which is 17.4.

Issue #1307, opened 2019-01-17
Templates vs Footprints


herostrat:
I think in issue https://github.com/KiCad/kicad-footprints/issues/864 the problem stems from a misunderstanding of the use cases of templates and footprints.

In my understanding:
 - Footprints are used to add components and modules to a PCB. If the component is a PCB itself, it can be seen as a daughterboard mounted to the mainboard that is the project PCB

 - templates are used if something gets mounted to the module, in other words if the projects PCB acts as a daughterboard and gets mounted to the motherboard that is the templates outline.

To avoid confusion:
A arduino footprint would be mounted on the PCB, e.g. this arduino nano breakout board https://store.qkits.com/arduino-nano-breakout-board.html

A arduino template would be a shield, e.g. the https://store.arduino.cc/arduino-ethernet-shield-2

Is my understanding correct?
Is there a official definition of this?
Should we add this definition to the KLC and the docs to avoid confusion?

----

In order to track progress, here are working points I plan to update:

 - [ ] Verify the differences and usecases of "footprints" vs "templates", define rules with the maintainers
 - [ ] Update the documentation accordingly
 - [ ] Recheck if all templates and footprints are correctly designed in such a manner
 - [ ] Monitor and push the template functionality
   - Folder for templates
     https://bugs.launchpad.net/kicad/+bug/1808011
   -

herostrat:
Ok interesting.
Thats unfortunate, because that means that the confusion remains - at least I think it is not clear what is meant for what.

----

I do not understand what you mean with block?
You mean something like a blueprint?

calebreister:
As a first-time contributor working on [#1304](https://github.com/KiCad/kicad-footprints/pull/1304), I can confirm that the particular use cases can be confusing. Perhaps it would be a good idea to add a section on templates to the KLC.

poeschlr:
There simply is no clear cut rule as to what should be a template and what should be a footprint. (For things that can be a template)


The main problem might simply be that we use templates in a way they where never really intended to be used. This is because there is no way to define a block (or blueprint) in kicad which would in most cases be the better option. So we use templates as such.

herostrat:
Ok that is at least a bit more understandable.

As it stands now, templates are completely useless.

Is there any discussion as to what should templates, blueprints and module footprints do where you can point me at?
I understand that the current solution is a workaround, but imho without rules we get suboptimal results where users use all tools wrong, each with his/her own understanding and therefore wrong usage.

---

[...]

PR #1668, opened 2019-06-28
Add rapsberry pi footprints

fvollmer:
Specification: https://github.com/raspberrypi/hats/raw/master/hat-board-mechanical.pdf

grafik
grafik

---


herostrat:
Hey fvollmer,

fyi, there are templates in KiCad (although often not well known or used).
There are multiple Raspi templates (i.e. https://github.com/KiCad/kicad-templates/tree/master/raspberrypi_hat for the HAT)

As your PR is a HAT and not a footprint, I think it would be better suited in the templates, probably as an additional HAT template without the (required and/or specified).

fvollmer:
I'm aware of the templates and think they are a good idea, but wouldn't it be better to use this footprint with the templates? 

I actually used the raspberry pi hat template and think it need some work:
 * The board outline wasn't closed (warning at 3d viewer)
 * The PoE header isn't mentioned
 * Some raspberry pi hat specific symbols couldn't be found, because the library file was renamed by the wizard (?)
 * No bypass capacitor for the eeprom

If I would use my footprint for the template that would fix the first two entries. Additionally it would allow to change easily between the hat without cutouts to the hat with cutouts by just changing the footprint.

herostrat:
You are right in that the templates rework.
In general they are in a grey-zone adoption wise and that unfortunately results in a lack of maintenance.

They problem - in general - with footprints like this is:

 - they need special symbols (you provided this)
 - the BoM does not mirror the needed materials for placement
   say e.g. you have a Arduino footprints, you need at least two pinheaders to mount it to the daughterboard/motherboard. A footprint only shows as one element in the BoM whereas the templates does include all elements.

The problem was actually discussed quite a lot in the recent past, unfortunately we could not solve the problem (yet).

fvollmer:
I see your point. How should we proceed with this PR? I think this PR would fit to the other footprints in `Modules.pretty`. All of them suffer from the limitations you mentioned.

I guess the best solution would be to have "grouped footprints" or footprints with multiple BoM entries, but this would require additional development and a file format change.

herostrat:
Honestly I am not totally sure.
The discussion template vs footprint came up a lot of times (see e.g. https://github.com/KiCad/kicad-footprints/issues/1307).

I am a new member of the team, so I do not see myself in the position to decide on this more "meta" stuff.
I hope it gets clearer once v6 implements a better template/blueprint functionality.

Just by looking at your footprint and the module.pretty library I can that your usage of the edge.cuts layer differs from the rest of the modules.
In my understanding (and note again that this is no definite statement or official opinion) a module is used primarily similar to a footprint, in that it not (necessarily) defines the outline of the board.

If it is used as a HAT, meaning it has the exact same dimension as in your case the Raspi, one should add the edge layer or change e.g. the F.Fab to Edge.Cuts.

I guess the best would be to wait for a seasoned librarian...

Issue #1663, opened 2019-06-25
Connectors with wrong orientation in footprint name


evanshultz:
There are a number of connector footprints with the wrong orientation because of a typo in the scripts. This issue is to track them so for v6 the old ones can be removed and new, corrected ones can be added.

- [x] Hirose DF63 (remove `_Horizontal`)
- [ ] Samtec HLE, SMD and THT libraries (should be `_Vertical`)

poeschlr:
We are allowed to rename footprints. The reasoning is that they are embedded in designs and do not rely on the temperamental rescue feature. (Which was the reason given by @stambaughw for not changing symbol libs too much during a stable release.)

Also: if a wrong name can mean users get confused about which part it is then it would be classified as a bug and can even be changed for symbols. (A typo in the part number would be a typical example in the symbol side. the wrong cleartext naming of a footprint another. Even if the part number is included as well as users will not really parse that as easily.)

antoniovazquezblanco:
Added some changes to fix this to https://github.com/pointhi/kicad-footprint-generator/pull/392 I will now generate the corresponding footprints.

If this is too much I can revert the changes.

Thanks!

evanshultz:
This should add duplicate footprints (`_Horizontal` suffix for horizontal footprints that were incorrectly suffixed with `_Vertical` before). I understand we are allowed to change/remove footprints at any time since they're sucked into the board file.

I started a list above about which connector series have dual suffixes and which suffix is to be removed. Can you please add to the list?

antoniovazquezblanco:
Ohh! I though that this was to keep track for removal during v6 but was finally moved to v5.3. I've added the removal of the old DF63 footprints to my PR and checked all the Hirose connectors. That PR should tick all the `Connectors_Hirose` footprints when merged. But I do not have time to go all over all the connectors available.

evanshultz:
Yeah... well... I still feel nervous to delete footprints at will but I understand Rene is saying that's OK. I don't think waiting until v6 is a big deal, but if this can be done sooner with Rene and the dev team understanding and approving then so be it.

antoniovazquezblanco:
I am also nervous about it but footprints will not disappear from your board if no longer in the library and poorly named footprints can lead to confusion and errors when designing new boards.

Footprint renaming has been done in previous stable releases (e.g. v5.1.1) without any complaining from the devs.

poeschlr:
[...]

PR #1670, opened 2019-07-03
Connector_Hirose: Add DF11 series

antoniovazquezblanco:
Datasheet: https://www.hirose.com/product/document?clcode=CL0543-0517-5-08&productname=DF11-4DP-2DSA(08)&series=DF11&documenttype=2DDrawing&lang=en&documentid=0000986679

Script PR: https://github.com/pointhi/kicad-footprint-generator/pull/392

imagen

---


herostrat:
Hey antonio,

I checked the dimensions and could not find any error there.

Why did you name the footprints not according to the datasheets table?
There they use the total pins as the number before the DP, you used the number of colons.

antoniovazquezblanco:
Sorry, I missed that one. Now corrected. Thanks!

herostrat:
Awesome, than I do not see any more problems.
Should I merge or do you think we should wait for another feedback?

evanshultz:
Hang on please. I'm making comments on the script.

@herostrat 
How did you check dimensions? And on the footprints or the script?

herostrat:
Understood.
What would be the best order, first script checks, than footprint merge, than script merge?

I checked on the footprint, I measured the dimensions for two in FreeCAD and crosschecked with KiCads internal ruler tool.

evanshultz:
A couple of my script comments will affect the footprints. Since this is often the case, I review the script first. The order you mentioned works for me and has seemed to be successful.

After checking the script, I usually just do FreeCAD unless KSU can't import some elements and I then need to use KiCad. I also post the dimensioned drawing so that any issues, or verification of correctness, are clear. After taking the time to review pasting a screenshot is usually minimal extra effort.

antoniovazquezblanco:
Fixed scripts and updated fps

antoniovazquezblanco:
Fixes https://github.com/KiCad/kicad-footprints/issues/1663

evanshultz:
There is no 2 pin version. The minimum number of pins is 4. Did I misunderstand, or is there a connector that shouldn't belong here (if so please update the script).

antoniovazquezblanco:
> There is no 2 pin version. The minimum number of pins is 4. Did I misunderstand, or is there a connector that shouldn't belong here (if so please update the script).

In DF11 series there is a 4 pin connector (2 pins per row) `DF11-4D...`. Or are what footprint are you talking about? Thanks

evanshultz:
I thought I saw a `DF11-2DP` in one commit. Maybe I lost my mind. Sorry.

Changes that I note, with questions in italic:
- FH12: Corrected vendor name and updated orientation keyword
- DF11: @herostrat checked it out above and the script was reviewed as well
- DF12: Updated orientation keyword
- DF13: Corrected vendor name and updated orientation keyword. _Some had added silk lines, like `Hirose_DF13-03P-1.25DS_1x03_P1.25mm_Horizontal`. What's up with that? Also in the same footprint, should a 0.9x1.4mm pad have a roundrect_ratio of 0.25? The criteria of 25% the side up to 0.25mm max would seem to indicate to me that roundrect_ratio should be only 0.225. What am I missing?_
- DF13C: Corrected vendor name and updated orientation keyword
- DF63: Removed `_Horizontal` footprints that are actually vertical (new footprints are already present).

Can you check out the DF13 questions above?

And because of the last item, I'd like to get a sign-off from Rene when he gets back. Just to be safe.

antoniovazquezblanco:
> Some had added silk lines, like Hirose_DF13-03P-1.25DS_1x03_P1.25mm_Horizontal. What's up with that?

There are no added silk lines. Only removed. There were overlapping silk lines and I removed those. There was a particular case for the 2 pin version that required an "if-else" block added.

> Also in the same footprint, should a 0.9x1.4mm pad have a roundrect_ratio of 0.25? The criteria of 25% the side up to 0.25mm max would seem to indicate to me that roundrect_ratio should be only 0.225. What am I missing?

imagen

imagen

I have not changed anything regarding that in the script. I don't know why in that specific model the diff tool has highlighted line changes.

http://kicad-pcb.org/libraries/klc/F7.3/

Should I change anything?

> And because of the last item, I'd like to get a sign-off from Rene when he gets back. Just to be safe.

https://github.com/KiCad/kicad-footprints/issues/1663#issuecomment-510801145

> I would still require good reason to do so as to keep the library from being too volatile. (Something like this is a clear case where the name change can be easily explained.)


Thanks

evanshultz:
Ah. I meant deleted, not added. Sorry. Everything still looks fine after viewing this branch so it seems OK.

My brain was thinking about the round rect pads backwards. It's good.

Let's do it!

antoniovazquezblanco:
Thank you!

PR #1679, opened 2019-07-10
Added JDY-08 in RF_Module

Trubochkin:
Added JDY-08 in RF_Module
Dataheet for footprint: https://fccid.io/2AM2YJDY-08/User-Manual/User-Manual-3511895.pdf
PR for 3D model: https://github.com/KiCad/kicad-packages3D/pull/594

---


poeschlr:
The linked datasheet does not contain a suggested footprint. How did you determine the pad sizes and positions?

Trubochkin:
> The linked datasheet does not contain a suggested footprint. How did you determine the pad sizes and positions?

Physically from a real board.
If something is wrong, then I will not change anything, because I have no motivation to do new pullrequests, because you still don’t consider them (or consider them for a very long time), especially 3D models 😕

poeschlr:
Well we are only volunteers and do not have the time we would like to have (we are also desperately short on manpower. Remember the review of any manually made asset can take 15 minutes for something easy that is first time right but up to a few hours for something complex especially if there are errors to document. And then multiply that by the amount of contributions we get and you will learn that we simply can not come around to everything within a short time.)

If you want your stuff to be considered faster then give us all the information required (for example how you determined pad sizes. If we need to ask anything then this will introduce delays and in a lot of cases simply push your contribution back in the priority list.)

Also to speed it up even more make a dimensioned drawing of your contribution. Example on how to do it: https://forum.kicad.info/t/how-to-check-footprint-correctness/9279

poeschlr:
Now for the review:

The heel fillet (for castellated towards outside) is a bit small as it is defined as 0.55mm by IPC (yours is at most 0.5mm but i suspect less as the part for sure has some tolerances). 

There should not be silk below the part as it negatively impacts solderability.
Add pin 1 markers on silk and fab.

Do not set the soldermask clearance in the footprint (unless there is a document by the manufacturer requesting it. I kind of doubt it as there is not even a suggested footprint)

add the datasheet link to the footprints description field

I am convinced that there is some area where there should not be copper. My guess is that it is under the antenna. We typically mark this area in the footprint. But we would really need more documentation to determine this.

I need to use IPC as a guide as there is no suggested footprint. Sadly i can not even really use IPC as the datasheet does not even the necessary information for that (There is no information about size tolerances and the castellations are not really dimensioned at all.)

Courtyard clearance must be 0.25mm all around.

The body size in x direction is wrong.

Screenshot from 2019-10-27 21-56-01


---

Note: this review took me 21 minutes

PR #1680, opened 2019-07-10
Infineon GAN footprints

evanshultz:
Adding packages used by parts at https://www.infineon.com/cms/en/product/power/wide-band-gap-semiconductors-sic-gan/gallium-nitride-gan/.

https://www.infineon.com/cms/en/product/packages/PG-LSON/PG-LSON-8-1/
image
image

https://www.infineon.com/cms/en/product/packages/PG-HSOF/PG-HSOF-8-3/
image
image

https://www.infineon.com/cms/en/product/packages/PG-DSO/PG-DSO-20-85/
image
image

https://www.infineon.com/cms/en/product/packages/PG-DSO/PG-DSO-20-87/
image
image

Thermal via info is at https://www.infineon.com/dgdl/Infineon-Board+Assembly+Recommendations-Discrete+Leadless-P-v06_00-EN.pdf?fileId=5546d462580663ef01581a365f1347b7#page=10, showing that a 0.3mm via in a 1x1mm grid is good.

As is allowed now, I have chosen the G, D, S, SS pin names since that makes it easy to copy a symbol and assign it to another footprint without adding/removing/changing any pins.

Symbol PR: https://github.com/KiCad/kicad-symbols/pull/1946

---


evanshultz:
Dimensioned drawings added above. Note that I fixed the HSOF courtyard after making the drawing. Sorry KSU doesn't capture paste pads so those aren't dimensioned above.

I think the CI errors are fine.

evanshultz:
Oh, and the PG-DSO-20-85 drawings on the website are missing two dimensions. Here is the updated drawing from Infineon:
image

poeschlr:
Since when do we use this type of pin numbering? And are there symbols fitting them? (I would prefer these have the normal pin numbers as suggested by infineon and the symbols are created to fit that. This is because symbols with named pin numbers do not make sense for most transistors as they typically use standardized packages that can also house other component types. Which would mean the symbols for these would need to be fully specified anyway so why not use our normal pin numbering scheme.)

evanshultz:
Since you accepted pin names at https://github.com/KiCad/kicad-symbols/pull/592.

There is a PR for symbols linked above.

You would know better than I about Infineon packaging, but it appears that is a real competitive advantage and more and more packages are becoming proprietary. And I also shared above about symbol reuse.

poeschlr:
If we use pad names like that then these packages should not be in the same library as the generic packages. Possibly make a library for transistors (or just for GANs)?

evanshultz:
OK then. I will change to pin numbers.

You can see at https://github.com/KiCad/kicad-symbols/pull/1946/ it becomes obvious what the pins are since they don't have an arbitrary number but a name. And a single symbol pin can connect to any and all pins in the footprint. And a single symbol will work regardless of the pin numbers in any footprint.

poeschlr:
thanks

PR #1682, opened 2019-07-10
Littelfuse 372

evanshultz:
https://www.littelfuse.com/~/media/electronics/datasheets/fuses/littelfuse_fuse_372_datasheet.pdf.pdf
image

Not sure why FreeCAD selects the inner and outer edges of the line instead of the centerline, so I captured both diameters of the body and courtyard.
image

---


poeschlr:
The reason why the drawing dimension workbench select the inner or outer line is because stepup does import footprint features that way (It imports the outline of graphical features not its center.)

Might it be a good idea to at least add the body diameter to the footprint name.

Dimensions are correct.

evanshultz:
@poeschlr 
Sure thing. Done!

poeschlr:
thanks

PR #1683, opened 2019-07-10
Lodestone vtm950 6

evanshultz:
http://www.lodestonepacific.com/CatKpdf/VTM950-6.pdf
image

Why the weird pin numbering? Well, I know of two major uses for this bobbin:
1. And inductor. In this case, for easy insertion the pins across from each other are used so that the part can be inserted into the PCB in either orientation and it will work. This was done with the other Lodestone bobbins I created that have been merged.
1. A common mode choke. In that case, the following symbol will work:
image

In both of those cases, the middle two pins aren't used so I numbered them 5 and 6.

image

---


poeschlr:
My understanding of this thing is that it is meant to hold self made inductors (or at least the inductor is not part of this product)
Meaning the user can choose freely which pins to use for which function. Your pin assignment therefore was selected as the one that makes most sense in your opinion (and users can make their own symbols if needed)

If my assumptions are correct then this is good to go.

evanshultz:
You are absolutely correct, similar to the other Lodestone inductor base footprints in the library already.

poeschlr:
Ok thanks.

PR #1691, opened 2019-07-12
Add NEG-12 package

drid:
Add NEG-12 package  
https://www.idt.com/package/neg12  

image

---


cpresser:
I did not do a full review, so just a few points that popped up while looking at the symbol:
- [ ] The datasheet suggests to put vias into the exposed-pad
- [ ] The Name of the footprint is not according to KLC. Please take a look at KLC 2.x. Something like "IDT_NEG12_1EP_2x2mm_P0.5mm_EP1.1x1.1mm_ThermalVias"
- [ ] Please use this as datasheet link https://www.idt.com/document/psc/neg12-package-outline-20-x-20-mm-body-epad-11-x-11-mm-050-mm-pitch-qfn

cpresser:
- [ ] Courtyard distance is 0.5mm. It should be 0.25mm
- [ ] The Pad-Size seems to be good (0.25*0.475), but none of the positions I checked are okay.
- [ ] The Exposed-Pad should have a 0.2mm chamfer
- [ ] The vias should not have a solder-paste margin set
- [ ] B.Mask is not to be enabled on the B.Cu pad so mask is present

drid:
Where did you measure 0.5 courtyard? It was 0.25 and adjusted when i moved the pads.
image

Pad chamfer was estimated from package chamfer and adapted to pad size (45 deg angle)

cpresser:
Regarding the courtyard - my bad, I measured from the body outline. Which is wrong.

The name originally suggested by me is not good either. Sorry about that. Similar footprints do have the pin-count in separated by dashes.
For example ``MLF-8-1EP_3x3mm_P0.65mm_EP1.55x2.3mm_ThermalVias``.
Following that scheme, this FP should be named ``IDT_NEG-12-1EP_2x2mm_P0.5mm_EP1.1x1.1mm_ThermalVias``


I found two more issues with the pads:
image
- [ ] Pad-Dimension is 0.47mm, should be 0.475mm (this will influence the courtyard position as well)
- [ ] The chamfer is not 45° (45° line on Silk as reference)

Please take a look at other FP with thermal vias. One example is "QFN-12-1EP_3x3mm_P0.5mm_EP1.65x1.65mm_ThermalVias". It has SolderMask on the bottom copper. You can check that with the 3D-Viewer.


I will do a complete check again once

poeschlr:
I would suggest this to be scripted with https://github.com/pointhi/kicad-footprint-generator/tree/master/scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON (Uses IPC standard to generate a footprint from the package dimensions https://www.idt.com/document/psc/neg12-package-outline-20-x-20-mm-body-epad-11-x-11-mm-050-mm-pitch-qfn)

It can chamfer your pads. I am however not yet certain that it is really a requirement for this part.

drid:
> I found more issues with the pads:
> image

Can you please check the pad properties, pads are 0.475 (2 * 0.2375)
image

drid:
> And there is the topic of licensing. The source repo is under a license that does not allow commercial use. And it does not allow relicensing (not even a symbol exception) This would be violated by the license of the kicad library.

We have changed the license to CC BY-SA 4.0

poeschlr:
I can measure the pads in kicad. This is not the question. How did you get from package dimension to pad size? This is why we typically use the IPC generator scripts as that script uses industry standard rules to determine that. And it also makes future maintainance easier to have stuff scripted. (A change to the standard is easily implemented by just changing a central script parameter, regenerating all footprints and updating them here. The same for changes to the library convention.)

drid:
I used the recommended land pattern from page 2 of datasheet.
I've also extracted missing dimensions from PDF using inkscape as suggested.
I've also tried the generator script, it does create the pad chamfers but pad size will differ from datasheet.

poeschlr:
If this is considered a qfn than it must use the generator. If not then it must be placed in a more specialized library. (And i am not yet certain that you will get me to agree that this is indeed such a specialized package as to not be best served with the IPC compliant version of the footprint. Only option would be that this is a very high frequency device or a MEMS that really need exact compliance with manufacturer footprints as otherwise mechanical or HF behavior differ from their assumptions. The former would then go into one of the RF_xxx libraries and the later into one of the Sensor libraries.)

drid:
Ok, so generated footprint looks like this
image
Should I force square pads?
Is there a way to achieve R 0.102 and chamfer for EP or leave it as it is?

poeschlr:
Right now there is no support for chamfering the center pad. I do not believe this to be necessary. (Nearly all qfn packages have a chamfered lead to indicate pin 1, but a vast majority of datasheets suggest a non chamfered pad. This could of course be because of a lack of software support but if there would be a major benefit then they would still suggest it.)

There might however still be something not ideal with exposed pad creation right now. The more i think about it the less rounded corners for it make sense. So i kind of decided to roll this back https://github.com/KiCad/kicad-footprints/issues/1798 (benefit of the script. I just need to change the parameters and regenerate and even such a major decision is easily implemented.)

drid:
Script https://github.com/pointhi/kicad-footprint-generator/pull/418

cpresser:
I checked at the script and the generated FP. 
- The Pads are 0.2mm wide. I would expect 0.2 + 0.05 (tolerance) - which is the same as the value recommended by the datasheet.
- The leads extend slightly longer than 2.4mm (2.7mm). 
- The radius of the exposed-pad seems way bigger than the EP of the part. That might be an issue. But I cant see the radius defined in the datasheet. It it already mentioned as a possible issue above (https://github.com/KiCad/kicad-footprints/issues/1798)

That's most likely intended by the IPC generator. 
Unsure how to handle this. Perhaps @poeschlr has some time to look at this.
image

drid:
Fixed pad width, can't find a way to control how much a pad will extend past the body size, any clue?

evanshultz:
All fillet goals are controlled by the IPC document which governs the generator. IPC 7351B is not freely available but you can search for it online and reading the script contains the math part of it. You don't adjust anything, like the pad size above, manually (OK, maybe in rare cases) because the footprint is generated the way it's supposed to be according to IPC. The footprint in the datasheet is discarded in favor of the scripted one.

The inputs to the script are the package drawing, not the footprint drawing. So use page 1 of https://www.idt.com/document/psc/neg12-package-outline-20-x-20-mm-body-epad-11-x-11-mm-050-mm-pitch-qfn, not page 2 as shown in the screenshot above.

cpresser:
@evanshultz 
The change of the lead-witdh to 0.25mm ist most likely a reaction to my previous post. I am still unsure why a lead-with of 0.20mm and tolerance of 0.05mm just gives a pad of 0.2mm. Is that the correct behavior of the IPC generator?
Would you mind taking a look and handling this PR? I am not experienced enough with the generator scripts (yet).

poeschlr:
Flat no lead has negative fillet sizes on the side (i can double check that but it kind of makes sense. If it is wrong then all our qfns are wrong right now.):
```yaml
# Flat No Lead (DFN/QFN) with toe fillet
#           | Minimum | Median    | Maximum |
#           | (Least) | (Nominal) | (Most)  |
#           | Density | Density   | Density | round
# Lead Part | Level C | Level B   | Level A |  to
# ----------+---------+-----------+---------+-
# Toe       | 0.2     | 0.3       | 0.4     | 0.05
# Heel      | 0.0     | 0.0       | 0.0     | 0.05
# Side      | -0.04   | -0.04     | -0.04   | 0.05
# Courtyard | 0.1     | 0.25      | 0.5     |
```

evanshultz:
Confirmed the table Rene shared above is correct.

PR #1693, opened 2019-07-12
L smd tdk

evanshultz:
https://product.tdk.com/info/en/catalog/datasheets/inductor_commercial_standard_nlv25-ef_en.pdf
https://product.tdk.com/info/en/catalog/datasheets/inductor_commercial_standard_nlv32-ef_en.pdf
image

I had to guess on the body widths a bit because the lead ends are defined and not the body.

---


evanshultz:
Dimensioned drawing:
image

Note that other 2.5x2mm and 3.2x2.5mm SMD inductors that can fit into these footprints, but I know TDK and they make both sizes so I went with them. Let me know if there's another way these should be named.

poeschlr:
NLFV25-EF where did you get the body length of 2.1mm from? (datasheet only dimensions including terminals and that measurement is 2.5mm)

Similarly with NLFV32 and the 2.7mm body length (datasheet gives 3.2 including terminals)

evanshultz:
I noted in the first post that I guessed those dimensions by measuring from the document. I also measured actual parts.

poeschlr:
It might be better to draw the body outline such that it includes the terminals instead. We do it similarly for tantal caps. And this would remove the need to guess a measurement.

evanshultz:
@poeschlr 
OK. Done. I thought this is easy enough to check in the commit that I didn't re-dimension.

Also, I updated the datasheets above to the ones used by the footprints (I used the basic NLV## instead of the commercial or "special" NL*V## versions).

poeschlr:
We decided in similar pull requests that we should include the body size in the footprint name (to be KLC compliant and also to make it easier for users to see what the footprint represents)

Dimensions are now ok.
The change in mpn is ok with me.

evanshultz:
@poeschlr 
You got it. Footprint names updated.

poeschlr:
thanks

PR #1694, opened 2019-07-13
Added Molex 48037-2200 USB plug.

koenvandenelsen:
Added Molex 48037-2200 to Connectors_USB footprints.

Datasheet: https://www.molex.com/webdocs/datasheets/pdf/en-us/0480372200_IO_CONNECTORS.pdf

image

---


antoniovazquezblanco:
@koenvandenelsen Please, sign the CLA so that we can take into account yout submission.

Thanks!

koenvandenelsen:
> @koenvandenelsen Please, sign the CLA so that we can take into account yout submission.
> 
> Thanks!

I've signed the CLA, but it will not update the status. If you need me something more to do, please let me know!

antoniovazquezblanco:
It seems that the email you have configured in git may not be the same as the one you have used for your github account. Can you verify this? Thanks!

poeschlr:
The silk outline must be outside the nominal body (=fab outline). (The two short sections are not.)
I would also suggest that the fab to silk offset is kept constant all around (or at least approximately constant.)

Please link a datasheet that contains a dimensioned drawing of the part and if available the suggested footprint layout.

koenvandenelsen:
> The silk outline must be outside the nominal body (=fab outline). (The two short sections are not.)
> I would also suggest that the fab to silk offset is kept constant all around (or at least approximately constant.)
I've updated the silkscreen location and checked whether it is now at a constant space around the fab outline.

> Please link a datasheet that contains a dimensioned drawing of the part and if available the suggested footprint layout.
I think the datasheet is updated from Molex and therefore the drawing is not available within the datasheet. Just to be clear here is the drawing according to molex:
https://www.molex.com/pdm_docs/sd/480372200_sd.pdf

poeschlr:
Move the edge cut indicator to the edge cuts layer (the line only).

Rotate the footprint such that pad 1 is nearest to the top left corner

A few dimensions are wrong see drawing
Screenshot from 2020-02-05 22-59-54

PR #1697, opened 2019-07-13
Add Hirose CX90M-16P

0xdec:
Add Hirose [CX90M-16P](https://www.hirose.com/product/jp/products/CX/CX90M-16P/)
- [Drawing](https://www.hirose.com/product/download/?distributor=digikey&type=2d&lang=en&num=CX90M-16P)
- Did not include pin one designator because none of the other USB C footprints have them, and pin one is unclear (A1?)
- Did not include any silkscreen: due to the pcb edge cutout, the silkscreen would be very far from the actual outline of the component

USB_C_Receptacle_Hirose_CX90M-16P

---


poeschlr:
Move the pcb edge outline to the edge cuts layer

A few dimensions are not quit right.

Screenshot from 2019-10-13 21-30-09

0xdec:
@poeschlr fixed the dimensions. Couldn't move the edge outline to Edge.Cuts because Edge.Cuts is disabled in footprint editor.

poeschlr:
Moving to edge cuts requires the use of a text editor. See https://bugs.launchpad.net/kicad/+bug/1251393

This should be an easy "search and replace" operation. Replace "Dwgs.User" with "Edge.Cuts" for the lines (do not move the user text to the edge cuts layer)

0xdec:
If the suggested cutout is on edge cuts, wouldn't it be hard/impossible for someone to use a different size cutout? For example, it's fine if the connector overhangs the board edge (bringing the S1 pads closer to the edge), but if the cutout were on edge cuts, this wouldn't be possible?

poeschlr:
Yes having it on the edge cuts layer fixes it for the user. This is in this case necessary as the connector really interferes with the board in this area.

This connector really leaves no option for the user to place the edge cut anywhere else. It can not get closer towards the body as it would then no longer fit and it can not really get closer to the pads as it would then be no longer producible.

0xdec:
@poeschlr I've made a drawing to show what I mean. The green parts are required on edge.cuts, but the purple parts can be extended downward or pulled in closer upward, depending on how much the user wants the connector body to protrude from the PCB edge.

poeschlr:
So how about making what you marked in green on the edge cuts layer and leave the rest to the user?

0xdec:
@poeschlr moved board cutout to Edge.Cuts layer

poeschlr:
I just double-checked our USB-C symbols. None of them have a pin fitting the pad number "MID". Could you look into this? (Do we need some special symbol or do you need to change the pad number?)

---

This is also only including the USB2 pins so we might want to make that clear in the footprint name.

---

The dimensions look good now. Only the body width is not dimensioned in the datasheet so i can not really be sure about it.
Screenshot from 2020-06-11 19-55-02

0xdec:
@poeschlr the MID pins seem to be connected to the stainless steel mid plate that stiffens the diving board that the contacts sit on. I believe it is mainly a mechanical connection, but it's unclear from the drawing.

poeschlr:
16P is definetly not clear enough. It is at least a massive violation of knowledge domains as understanding what the connector represents requires understanding of the USB standard. (The reason we use human readable names for footprints is to introduce an abstraction layer. Otherwise we could just use IPC naming conventions.)

0xdec:
@poeschlr in that case, we should also rename the following footprints as they are also USB 2.0 only:
- USB_C_Receptacle_GCT_USB4085
- USB_C_Receptacle_HRO_TYPE-C-31-M-12
- USB_C_Receptacle_Palconn_UTC16-G
- USB_C_Receptacle_XKB_U262-16XN-4BVC11

PR #1522, opened 2019-04-03
Add SKY65605-21

pbn4:
Add SKY65605-21

---
URL: http://www.skyworksinc.com/uploads/documents/SKY65605_21_203118F.pdf (page 6)

2019-04-03-090919_1507x500_scrot

I have run `check_kicad_mod.py` but I don't know what to do with the output:

```python
Traceback (most recent call last):
  File "check_kicad_mod.py", line 117, in <module>
    rule.check()
  File "/home/michal/kicad-library-utils/pcb/rules/G1_7.py", line 17, in check
    if 'linux' in platform.platform().lower() and not checkLineEndings(self.module.filename):
  File "/home/michal/kicad-library-utils/common/rulebase.py", line 75, in checkLineEndings
    LE1 = ord(chr(filecontentsraw[-2]))
TypeError: an integer is required
```

PR #1698, opened 2019-07-14
Add Lumex SML-LX0404SIUPGUSB

0xdec:
Add Lumex SML-LX0404SIUPGUSB (big sibling to #1990)

[Datasheet](https://www.lumex.com/spec/SML-LX0404SIUPGUSB.pdf)

LED_Lumex_SML-LX0404SIUPGUSB

---


cpresser:
Hi,
it looks like the silkscreen between the pads might violate KLC F5.1 (0.2mm clearance silk to pad)
Did you try to fit the F.Fab refdes into the body?

poeschlr:
We allow smaller silk clearances for small parts like these. (The lowest allowable is clearance equal to line width.)

0xdec:
@cpresser it's a bit smaller bc otherwise basically no silkscreen would fit. And yeah the refdes really has no hope of fitting inside the outline.

evanshultz:
1. Add the official datasheet at https://www.lumex.com/spec/SML-LX0404SIUPGUSB.pdf to the description.
1. The courtyard is too small. It should be a 0.9mm square.
1. No silk will fit between pads. Pads 1 and 2 are 0.4mm apart. 0.12mm clearance requires a line end 0.18mm away to account for the line endcap. That means a line length of 0.04mm. No way.
1. Similarly, the pin 1 silk lines need to be 0.83mm away from the origin and they are only 0.8mm now.

@poeschlr 
Why is text limited to 0.5mm and not smaller? If the assy layer is printed to PDF, and I'm not sure what else it's used for, vector text can be zoomed without a problem. If printed on paper, even 0.5mm could be a problem.

0xdec:
@evanshultz fixed all 4 points

evanshultz:
The silk is at 0.825 instead of 0.83 as it should be.  Practically, that won't make any difference.

And as I found at https://github.com/KiCad/kicad-footprints/pull/1522#issuecomment-701053128, fab ref des are smaller on some parts. But this is OK with me.

PR #1702, opened 2019-07-15
Sot 89 custom pads

evanshultz:

image
image

SOT-89-3 had no datasheet so I picked http://ww1.microchip.com/downloads/en/DeviceDoc/3L_SOT-89_MB_C04-029C.pdf and updated the footprint to match.

image

---


evanshultz:
Regarding the body size above. I edited the comment. At https://github.com/KiCad/kicad-footprints/pull/981#issuecomment-451771137 Rene said that the body size in the datasheet should match our footprint. I noted the governing body has the same size at https://github.com/KiCad/kicad-footprints/pull/981#issuecomment-452130164. The part body was already changed but as I started at the top and made my way down the large number of comments in the old PR I didn't catch that earlier.

Hopefully this isn't a problem if there are wider parts that I didn't find since the existing body size is 5mm.

poeschlr:
Thanks for this great enhancement.

PR #1703, opened 2019-07-15
No number paste pads

evanshultz:

Change paste pads to have no pin number. Also fixed fab line width to 0.1mm and removed mask from THT pads (vias in this case).

---


evanshultz:
image

cpresser:
I took a look at this. To make the changes easier to review I created a small script based on some others in kicad-library-utils
https://gist.github.com/cpresser/f733167b703bb32cae773e68c69862ca

Output looks like this:
(The Pad-numbers get formated as float, so that looks weird)
ATPAK-2 kicad_mod

The Rohm part has more changes, I can't fit it on a single page to screenshot it.
I need to figure out a way how to copy&paste the output somewhere with the colors. Any hints?

Anyway, with this output its easy to see that the ATPAK-2 part has exactly the changes as described in the first post. I also looked at the ROHM part - no weird stuff found.
I am not sure if this counts as a review though :)

poeschlr:
> I need to figure out a way how to copy&paste the output somewhere with the colors. Any hints?

pipe it into a log file. Editors like atom can show terminal colours for these. (program > output.log should do the trick under linux, not sure if windows has something similar)

---

Regarding this change: we typically do not add mask at all for thermal vias. This allows the large pad in front to control where mask is. And the back side should stay covered http://kicad-pcb.org/libraries/klc/F4.4/

evanshultz:
@poeschlr 
Good catch. That is what I intended to do (my note above) but I did it wrong. Fixed now.

I believe the other CI errors are OK.

poeschlr:
thanks

PR #1707, opened 2019-07-15
Correct Tdson-8-1

evanshultz:

https://www.infineon.com/cms/en/product/packages/PG-TDSON/PG-TDSON-8-1/

Nearly all dimensions were wrong so they got fixed.

image

image

Removed hand soldering footprints since the X dimension was +/-3.5mm and the fixed reflow one is +/-3.325mm. No substantial difference.

---


poeschlr:
Infineon suggests a simple rectangle for copper. Only soldermask would get the complex outline you chose.

evanshultz:
Oh yes. Thanks. The original one was like that so the documentation changed or I missed it when I did the original update. Also corrected the description to match just the Infineon metadata.
image

evanshultz:
@poeschlr 
Is this ready? I hope I fixed what you brought up.

poeschlr:
Infineon suggests your old detailed shape for the mask layer, you now have only a rectangle pad. I am convinced it would work this way but would for sure require a bit more paste that way so i would prefer to follow the suggestion more closely.

evanshultz:
That's not what I see. Here is the website's image:
image

The mask is cut out the same as the copper pad. Only the stencil (paste) is broken up.

Have I misunderstood?

chschlue:
Mask looks "pinnacle" shaped to me while copper is rectangular.

evanshultz:
Alright. I finally get it. Mask has 'teeth' at the top. I was afraid I was misunderstanding. Will fix...

evanshultz:
Corrected:
image

poeschlr:
thanks

PR #1708, opened 2019-07-15
UDFN-4_1x1mm_P0.65mm update

evanshultz:

http://ww1.microchip.com/downloads/en/DeviceDoc/MIC550x-300mA-Single-Output-LDO-in-Small-Packages-DS20006006A.pdf

Added datasheet and updated footprint to match, using custom pad shapes.

image

Note that the courtyard was fixed after this was done.
image

---


evanshultz:
Regarding the courtyard above: Rene said at https://github.com/KiCad/kicad-footprints/pull/981#issuecomment-455718226 to use 0.25mm. The footprint was already updated but I didn't see this comment before posting above.

poeschlr:
Should we raname this footprint to make it clear that it includes an exposed pad? So `UDFN-4-1EP_1x1mm_P0.65mm_EP0.48x0.48mm`

evanshultz:
That probably would have been a good idea. But it's an existing footprint, so is that reason enough to rename right now? Let me know if you want it that way.

poeschlr:
It does not look like there is any symbol in the lib that has this footprint assigned as its default footprint. And because footprints are included in the pcb file we have a lot more wiggle room here.

evanshultz:
Okey dokey. Done.

BTW, "Remove" in the commit message above should have been "Rename".

evanshultz:
@poeschlr 
Sorry that I missed pinging you above. Is this ready?

poeschlr:
thnaks

PR #1709, opened 2019-07-15
Merge overlapping pads

evanshultz:

image
image
image

The ST UQFN might be supported by the footprint generator script, but I already did the work here so it seems potentially an easier path to merge into the library.

I didn't touch the other LFPAK56 because they're non-standard and perhaps should be removed?

---


evanshultz:
Shall we remove LFPAK56 footprints with vias since they're not from the datasheet and seem like a specific implementation? All vias are way outside the footprint boundary. It seems to me this should be done on the board or a local library but not in our official library. I propose to remove them.

poeschlr:
The ST-UQFPN would indeed be supported by the script. But i accept it as a manual edit as it existed in the past (new contributions would need to be supplied as scripted.)

I overlaid these with the original footprints and everything seems to match up. So this is at least not a step back. Thanks for your work.

PR #1718, opened 2019-07-19
Add TRACO_TEL3 arv NP variant

Salamandar:
Add TRACO_TEL3 arv NP variant.
Those are DCDC converters, and the footprint is kinda weird.

# TRACO_TEL3-xxxx
- https://www.tracopower.com/products/tel3.pdf
- Capture d’écran du 2019-07-19 11-27-44

- No 3D Cad models have been found :(

# TRACO_TEL3-xxxx-NP
- https://www.tracopower.com/products/tel3.pdf
- Capture d’écran du 2019-07-19 11-27-17

- No 3D Cad models have been found :(

- The F7.2 rule is violated (warning only) as the footprint does not contain Pad 1.


poeschlr:
Which symbols are intended to be used for these footprints?
The drill diameter is a bit strange. I would suggest 0.75mm (0.5+0.05+0.2, see [KLC F7.6](http://kicad-pcb.org/libraries/klc/F7.6/). Pad diameter of 1.5mm might also be a better choice to what you currently have. (Nice metric values.)

Pin1 should be 2mm from the edge in y direction (is 2.03)
Courtyard clearance must be 0.25mm is 0.2mm
Add pin 1 marker on the fab layer (Chamfer nearest corner with chamfer lenght = 1mm acourding to IPC, see KLC )

For the part without pin 1 the same corner is to be marked in the same fashion. Also make pad 2 square in this case.
I would further suggest the use of rounded rectangle instead of square to mark pin 1 (25% round ratio or in this case maximum 0.25mm round radius. So 16.7% for 1.5mm pad size.)

Move the value field and the silk reference to the outside (fab reference should stay where it is)

PR #1720, opened 2019-07-19
Add Bourns inductors

Salamandar:
Add Bourns inductors

# SRR7045
- http://www.bourns.com/docs/Product-Datasheets/SRR7045.pdf
- Capture d’écran du 2019-07-19 11-00-18
- I found the official STEP file here : https://www.bourns.com/engineering/SRR7045/srr7045.stp

# SRP7050TA
- http://www.bourns.com/docs/Product-Datasheets/SRP7050TA.pdf
- Capture d’écran du 2019-07-19 10-58-12
- I found the official STEP file here : https://www.bourns.com/engineering/SRR7045/srr7045.stp


poeschlr:
Replace the dash between part number and manufacturer name with an underline.
Include the body size in the footprint name.


The radius of the SRP part body is not specified in the datasheet, so we can not really trust it for the courtyard outline.

---

A few dimensions are wrong

Screenshot from 2019-10-27 09-10-00

PR #1721, opened 2019-07-19
Inductor_SMD: Add Sumida CDMC6D28 SMD Power Inductor Series

gmsotavio:
Datasheet: http://products.sumida.com/products/pdf/CDMC6D28.pdf
L_Sumida_CDMC6D28_4

L_Sumida_CDMC6D28
L_Sumida_CDMC6D28_2
L_Sumida_CDMC6D28_3

L_Sumida_CDMC6D28_5
---


gmsotavio:
I'm aware that I need to append the body dimensions to the footprint name.

gmsotavio:
I have changed the courtyard style and appended the body dimensions to footprint name.
image

poeschlr:
The round corner is not dimensioned in the datasheet so i would not suggest to include it in the courtyard layer (for the others it is ok to indicate that this corner is rounded.)

gmsotavio:
Sorry, I didn't understand. You are suggesting to remove rounded corners in Fab and CrtYd layers?

gmsotavio:
Ok. I agree with you. Great that you have pointed out this mistake. The same is done in this PR, from another contributor: https://github.com/KiCad/kicad-footprints/pull/1720
image

gmsotavio:
Ok. Good catch, I agree with you. Thank you for pointing out that mistake. 
In another PR, the contributor did the same mistake: https://github.com/KiCad/kicad-footprints/pull/1720. I kept the contoured style, but squared the corners. If you prefer, I can return the style to rectangular. KLC is not clear about the preferred style (both rectangular and contoured is allowed), but I have been moving my footprints to contoured style because it allows more density and is preferred in IPC-7351C.

image

Note: I messed up my branch, then a force push was unavoidable.

poeschlr:
thanks

PR #1722, opened 2019-07-19
Fischer FK224 ... 220 1 heatsink

evanshultz:
https://www.fischerelektronik.de/web_fischer/en_GB/$catalogue/fischerData/PR/FK224_220_1_/datasheet.xhtml?branch=heatsinks

I used silk lines to help indicate where the TO220 device should be placed. Pads are placed at the center of the legs. There is a keepout noted to help catch approximately where the legs touch the PCB surface and could scratch the mask to create a short.
image

image

---


poeschlr:
Might it be a good idea to somehow separate 224 and 220 in the name? for example by a dash or underline character?

The drill size might be a bit small. The diagonal of the nominal "lead" dimension is already 1.77mm. So even a bit of tolerance would mean that the part would no longer fit into a 1.8mm hole. To be honest i would probably go with 2mm.

The silk outline should be moved outside (the outermost outline).

evanshultz:
Thanks for looking.

I think the full PN includes a two-character surface finish so I used `xx` as a wildcard. Would you like this footprint specific to a finish for 3D model uniqueness? There also appears to be a space before the trailing number which I have omitted.

Good catch. Holes are now 2mm.

I have never understood this rule. You did clarify it a bit that only 'extra' silk can be under THT parts but I really don't understand why.

image

poeschlr:
The outermost silk outline should be visible after assembly this is why it must be outside. This is because the outermost lines are in part used for quality control. (Checking if the assembly was successful and everything is aligned. This would be impossible if the silk line is covered by the body.)

Normally no silk is allowed under parts as it negatively impacts solderability. But THT parts are not affected by this and silk under it can be used to help identify the orientation for manual assembly which is why we allow it.

evanshultz:
Is this ready to be merged or is there still work to be done?

poeschlr:
looks good now, thanks

PR #1729, opened 2019-07-22
Fix Wago 236 courtyard

chschlue:
Fix Wago 236 courtyard

Issue https://github.com/KiCad/kicad-footprints/issues/1730
Datasheet: http://www.farnell.com/datasheets/2079105.pdf

Violating F5.4: no change

236foot

---


evanshultz:
Thanks for making your first contribution here, and especially for making a bug fix.

Yeah, the courtyard should be 0.5mm offset from the body for this footprint. Oops.

As can be seen from the footprint description, these footprints are generated by a script. It appears you made the changes by hand. Would you be willing and able to update the script? It is at https://github.com/pointhi/kicad-footprint-generator/blob/master/scripts/TerminalBlock_WAGO/make_TerminalBlock_WAGO.py. That is not required to accept a bug fix, but it should be obvious that keeping the script updated is highly desirable.

chschlue:
Sure, I'll look into it.
Sorry, I'm not all that familiar with footprint editing yet.

Also, AFAIK 236-10? terminals are 1-pin versions which are out of production. I think all these footprints should actually be called -40?, -50? and so on.
https://www.wago.com/gb/search?f=%3ASerie%3A236

How to fix that without breaking anything?

evanshultz:
@chschlue 
Can you clarify the above? Sorry I'm not getting it.

Looking at the link you sent, 236-401 is 1pole but 236-402 is 2pole. But then 236-744 and 236-746 are both 1pole. And then back to 236-101 is a 1pole and 236-102 is a 2pole. I'm not seeing a family datasheet that brings the entire PN system together. Can you help me understand?

chschlue:
Not really.
236-4xx, -5xx, -6xx are gray with 200, 300, 400 mil pitch respectively and xx referring to pole count.

The colored versions have random part numbers as far as I can tell. 🤷

chschlue:
Closing. Superseded by #1752 and #1753.

PR #1733, opened 2019-07-23
Added Bourns SRN8040

gggfred:
[Datasheet](https://www.bourns.com/docs/Product-Datasheets/SRN8040.pdf)
imagen

---


poeschlr:
The fab layer must represent the nominal dimensions of the part (8x8.15mm).

The silk outline must be outside of that and also far enough away from exposed copper (at least same clearance as line width, ideally 0.2mm). Right now the silk as you chose to draw it does not help a lot anyways as the real part has rounded corners. So i suggest to only have two short lines like this:
Screenshot from 2019-09-20 14-34-22

The courtyard is drawn with a clearance of 0.25mm all around (relative to the nominal body or pad whatever results in the larger courtyard)

The silk reference and value should exchange places.

Add the link to the documentation to the footprint description.

gggfred:
@poeschlr thank you for your feedback, I have already applied the changes

poeschlr:
The courtyard clearance must be 0.25mm all around (In this case relative to the pads as they are outside the nominal body dimensions. Rounded to nearest 0.01mm grid but that is not a concern for a footprint with such nice metric dimensions)

Screenshot from 2019-09-21 13-27-27

Also i suggest to move the silk a bit closer in y direction and make it a bit shorter in x direction (such that there is still enough clearance to the pads) Something like this: 
Screenshot from 2019-09-21 13-30-00

poeschlr:
thanks

PR #1735, opened 2019-07-23
Murata LQH55DN inductor footprint

evanshultz:
https://search.murata.co.jp/Ceramy/image/img/P02/JELF243A-0045.pdf

image

I measured the bottom section's height right at 4mm, so I included that into the fab lines. However, I just took the overall 5.7x5mm rectangle for the silk lines.

image

---


poeschlr:
Should we possibly include the body size as a suffix as documented in the KLC? (I know we do not do it for other inductors right now but i fear this is simply an oversight from version 4 to 5 reorg.)

evanshultz:
Sure! Can you point me to where this is mentioned in KLC? I don't see any section specific for inductor footprints and searching the term "inductor" only turns up one hit in the symbol section.

Are you suggesting `L_Murata_LQH55DN_5.7x5.0mm`?

evanshultz:
@poeschlr 
Added body size as I wrote above. Please advise if this is not what you intended.

poeschlr:
looks good, thanks

PR #1736, opened 2019-07-23
Infineon TISON footprints

evanshultz:
This PR includes all four current Infineon PG-TISON packages that have footprints.

The pin numbering is a bit tricky. I chose to start with the designated pin 1 and then go counter-clockwise, as usual, and finish with pads under the part. I'm open to suggestions about a better scheme.

Also, I squared off the paste pads on PG-TISON-8-2 and -3.

PG-TISON-8-2 / PG-TISON-8-3
https://www.infineon.com/cms/en/product/packages/PG-TISON/PG-TISON-8-2/
https://www.infineon.com/cms/en/product/packages/PG-TISON/PG-TISON-8-3/
image
image

PG-TISON-8-4
https://www.infineon.com/cms/en/product/packages/PG-TISON/PG-TISON-8-4/
image
image

On this one I caught an incorrect courtyard during dimensioning and fixed it before committing.

PG-TISON-8-5
https://www.infineon.com/cms/en/product/packages/PG-TISON/PG-TISON-8-5/
image
image

This one doesn't import the custom shape with two primitives. And it's missing the inner radius so I know somebody will say I should have used FreeCAD. And you'd be right.

---


evanshultz:
TISON-8-3 appears to be identical to TISON-8-2. I have asked Infineon if there is a difference.

poeschlr:
Did infineon ever get back to you on that?
Otherwise, i would say this is ready to merge.

evanshultz:
@poeschlr 
Yes, sorry about that. The internal clip is different to hold a bigger die size. But from the outside, where we care about the footprint, they're identical.

poeschlr:
thanks for the update and of course the work you did here

PR #1739, opened 2019-07-24
8devices Carambola2 footprint added

EpeR1:
I have created a Carambola2 footprint.
Documentation is here:
https://www.8devices.com/media/products/carambola2/downloads/carambola2-datasheet.pdf

## Note:
~(Violating F5.2)~
~There is a significant bevel on the board's view, so I drove that on Silk, and not PIN1's bevel.~

The module:
cara1
Carambola_shield-800x600

The footprint, that I have created:
cara2

The footprint in the documentation:
cara3

~Note:~
~(Violating F5.2)~
~There is a significant bevel on the board's view, so I drove that on Silk, and not PIN1's bevel.~

---


poeschlr:
This should be in the RF_WiFi library.
The description should be a bit more detailed.
Is there a symbol for this contribution?

---

Rotate the part such that pin 1 is in the top left corner (IPC zero orientation A)

---

Silk outline is quite far away from the body meaning it is kind of useles for helping to align the part. Move it in closer. (I typically draw it such that the silk outline touches the fab outline. But there can be a bit of space or overlap so you do not need to use an extremely fine grid to draw it.)

We do not allow silk under SMD parts so remove the detailed silk and leave only the fab. (silk under parts negatively impacts solderability)

We do not typically include numbers in the footprint so please get rid of them. (numbers do also not play nice with rotating footprints in pcb_new so there is a technical reason not only preference)

Use IPC style pin 1 marking (look at QFN-12-1EP_3x3mm_P0.5mm_EP1.65x1.65mm as an example)

---

There is a suggested paste layout which i assume is done this way as there is a lot more paste needed to fill up the castellations. Please create paste only pads to fit this requirement. (remove paste from the normal pads)

---
Dimensions of the copper layout look good.

Screenshot from 2019-10-27 08-53-45

myfreescalewebpage:
@poeschlr I do not agree about the library, Module is great for this one because it's not only a WiFi chip. This is a Linux module with multiple interface including WiFi. Joel

EpeR1:
Updated.

Is there a way to create a special paste layers? 
Such as:
sold

EpeR1:
Here is the symbol:
https://github.com/KiCad/kicad-symbols/pull/2015

EpeR1:
(The question about paste layers, is standing still.)

The status now is:
fp

poeschlr:
Yes a special paste layer like that can be made. The first step is to disable the paste layer from the current pads. Then create a complex paste only pads 

Paste only means no copper or any other layer than paste active plus no pad number assigned.

Complex means a pad comprised of an anchor pad plus a number of pad primitives in this case most likely a polygon. To do this draw the shape you require on any layer by use of the polygon tool. To get the radius select a line thickness that fits. Place a circular pad overlapping with the shape (you can even place it such that this pad takes care of the half circle in the drawing) and then select the polygon(s) plus the pad -> right click -> pad from selected shapes.

Or use KiCad stepup to create the pad. https://forum.kicad.info/t/kicad-stepup-the-sketcher-for-footprint-generation/9582

In any case after you have the pad once use either the array tool or just duplication plus the grid to get the pads where you need them.

myfreescalewebpage:
@EpeR1 any news of the pads after the comment of @poeschlr ? Thanks for the update

EpeR1:
Done.

xx

myfreescalewebpage:
@poeschlr :)

myfreescalewebpage:
@EpeR1 I continue the review here:

- [ ]The top F.Fab line is made of two lines, only one expected:

Capture

- [ ] Clearance between F.Fab and F.SilkS is usually 0.11mm
- [ ] Clearance between the pad and the F.CrtYd should be 0.2mm and I think this should take into account the Paste, so the current F.CrtYd should be enlarged

Joel

EpeR1:
Updated

myfreescalewebpage:
- [ ] Current clearance between F.Fab and F.SilkS is still too large (0.225mm instead of default 0.11mm). and silkscreen is too much closed to the pad (0.1mm but required at least 0.2mm, see screenshot below).

Capture

Else ok for the other points.

Joel

EpeR1:
Can you explain the clearance issue between F.Fab and F.Silk?
I measure 0.115mm everywhere.

ddd

The clearance between the F.Paste, and F.Silk is updated.

ccc

myfreescalewebpage:
From F.Fab to F.SilkS the 0.11mm value is common (this is the one used in generator when we create footprint with python) and it is measured from center to center, not from border to border. anyway there is no strong value, the minimum is 0.1mm to have the silkscreen visible when the device is on the PCB. Do you want to keep your clearance or do you want to set to 0.11mm ?

EpeR1:
I am not exactly sure the correct position of F.SilkS edges, because there are technical notches at the corners, so the panels usually, are a little bit bigger than the sizes in the documentation. 
It's because the 8devices uses panelization.
So, the bigger values are more helpful at manual soldering, I think the 0.11mm center-center distance could be unuseable in the most cases.

carambola2_top

( Original picture is from: https://www.8devices.com/products/carambola-2 )

I made a quick print out and checked on paper, so I think the present value (0.23mm center-center) can be appropriate.

cara

Or should I set to 0.11mm?

myfreescalewebpage:
Great detail, in  this case it's not bad to have something a little bit larger :)
This is not an issue, just a recommended value commonly used.

myfreescalewebpage:
No more comment on the fp, merging, thanks for the great job on this footprint :)

PR #1743, opened 2019-07-25
Te ms5525 dso

cp-aquila:
Add TE MS5525DSO Pressure Sensor

Datasheet: https://www.te.com/commerce/DocumentDelivery/DDEController?Action=srchrtrv&DocNm=MS5525DSO&DocType=DS&DocLang=English

grafik
grafik

---


gmsotavio:
@cp-aquila (@cpresser),

First thoughts about your contribution

- [ ] I realized you have used imperial units even with metric available in documentation. Metric is preferred because it avoids round errors and mixed units with those given by KLC, IPC and so on. Taking into account the dimensioned drawing (imperial), all dimension looks fine, except for the courtyard clearance, because KLC give it to us in metric units. So you will need to ensure at least 0.25mm (now it is 0.247).
## Imperial
image
## Metric
image

image

image

---
- [ ] The ends are not touching each other

image
---
- [ ] As per KLC F5.2, bevel should be 1mm or 25% of package size (whichever is smaller). In this case 1mm is smaller.

image
---
I have no problem to put "portX" in the Fab layer, but I would personally put it in Cmts.User. Actually I don't really know what would really be right. I will merge anyway.
image

cp-aquila:
Hi & thanks for the review.
Quite a lot of stuff I missed :/

I used imperial because my understanding is that values in square-brackets are for reference only (https://en.wikipedia.org/wiki/Reference_dimension) and thus should not be used for manufacturing.

Bevel and Silk issue have been addressed.

I moved the Port? designators to Cmts.User. The Circles indicating the port locations are still on F.Fab. They are usefull for the CAM operator to indicate that its not a simple flat part and a proper nozzle has to be selected.

poeschlr:
If the datasheet mentions metric then metric has to be used. (The KiCad library only uses imperial if it is completely unavoidable)

I am however a bit uncertain about what the numbers in brackets mean here. 0.02inch is nowhere near 0.4mm. and 0.1 inch is not 2.4mm. (So either this part is sold in two distinct measurements depending on target region or the datasheet is simply wrong. I guess the later -> sadly means we will need to use imperial here.)

gmsotavio:
> I used imperial because my understanding is that values in square-brackets are for reference only (https://en.wikipedia.org/wiki/Reference_dimension) and thus should not be used for manufacturing.

Sorry for my ignorance on the subject. I think this content was not covered in my technical drawing class. I am learning this in the school of life.

> Bevel and Silk issue have been addressed.

Perfect!
image

> I moved the Port? designators to Cmts.User. The Circles indicating the port locations are still on F.Fab. They are usefull for the CAM operator to indicate that its not a simple flat part and a proper nozzle has to be selected.

I agree with the ports on the F.Fab. I was talking about designators only. You got the point.

There is remaining a minor issue. You have forgotten to provide the description. There is the datasheet link only.
image

cp-aquila:
Thanks for having another look 👍 
I am not sure what to put into the description, its more or less a copy of the keywords. I am open for suggestions if you don't like it.

gmsotavio:
The description is perfect. I going to check the PR of the symbol soon.

myfreescalewebpage:
Hi @cp-aquila , I try to continue/restart the review from this point

- [ ] You can remove "SOIC-16" in the description, first because the footprint has only 14 pads and then because it's not really SOIC :)
- [ ] F.Fab: the dimensions of the footprint should be 7.9x12.5mm, your footprint is not correct at all
- [ ] F.SilkS: 0.11mm offset from F.Fab is usual, to be fixed after the previous comment
- [ ] F.CrtYd: 0.25mm offset from F.Fab is usual, to be fixed after the previous comment
- [ ] Position and size of the pin to be rounded because actual you have 4 decimal digits!!! Same comment for the position of the circles (remember you have to set the position with mm, not inch)

Will check again after those changes

Cheers,
Joel

cp-aquila:
Hi @myfreescalewebpage Thanks for taking over.

I will fix the keywords ASAP.

The coordinates seem wrong since you measured in metric. But the footprint was drawn in imperial.
> I used imperial because my understanding is that values in square-brackets are for reference only (https://en.wikipedia.org/wiki/Reference_dimension) and thus should not be used for manufacturing.

The Silk and Courtyard are again drawn on the metric grid. Thus the courtyard distance is slightly above 0.25mm.

The metric dimensions are most likely not correct, you can check by doing the math yourself. Also see @poeschlr comment above.

I also have another non-public datasheet for that part that I cannot share.
grafik
grafik
The drawing in that non-public datasheet is 1:1 identical to the datasheet linked here. So there is no difference and sadly no improvement.

myfreescalewebpage:
I was pretty sure of that but the problem is that the kicad_mod file is written with metric values (https://github.com/KiCad/kicad-footprints/pull/1743/files) and contains many digits. Metric is preferred in the library (https://kicad-pcb.org/libraries/klc/G1.9/) and I'm pretty sure we have forgotten to use so much digits (you have resolution of 0.0001mm !!!!!)

@poeschlr  @evanshultz @chschlue  maybe your advice on this metric/imperial question in the footprint ?

chschlue:
Note: I haven't checked the datasheet in depth, the following is just a general opinion.

If a manufacturer only gives metric reference dimensions which are either very coarsely rounded or plain wrong, I see no point in using them.

myfreescalewebpage:
He provides metric in parenthesis and imperial values

cp-aquila:
Parenthesis are for reference (from Wikipedia)
> Reference dimensions are not intended to be used directly to define the geometry of an object. Reference dimensions do not normally govern manufacturing operations (such as machining) in any way and, therefore, do not typically include a dimensional tolerance (though a tolerance may be provided if such information is deemed helpful). Consequently, reference dimensions are also not subject to dimensional inspection under normal circumstances

And: metric is not good. At least some values definitely fucked up the rounding. If it were just in the range of 0.01mm, that would be acceptable. But some measurements are off by 0.1mm.

|imperial | metric | imperial*25.4| rounded to two digits|Location|
|--|--|--|--|--|
|.49|12.5|12.446 |12.45|BodySize|
|.37|9.5|9.398 | 9.40|Body2Barb|
|.12|3.0|3.049| 3.05|Body2Barb |
|.21|5.3|5.334|5.33|Body2Barb |
|.31|7.9|7.874|7.87| BodySize|
|.024|0.6|0.6096|0.61|Pad|
|.087|2.2|2.2098|0.21|Pad|
|0.02|0.4|0.508|**0.51**|Pin|
|0.1|2.4|2.54|**2.54**|Pin2Body|
|0.1|2.6|2.54|2.54|Body2Barb|

I can redo this in metric. But then the questing is which values should I take? The datasheet ones? Or manually converted and rounded? Or a mix of both?

chschlue:
Definitely not the datasheet ones if they're obviously wrong.
This looks good to me the way it is now.

myfreescalewebpage:
OK for me in this case, I follow your advice @chschlue 
Just made last checking of the dimensions with imperial values, no more comment, merging.
Joel

PR #1744, opened 2019-07-25
Add Atmel_ATSAMR21G18-MR210UA_NoRFPads

cp-aquila:
Add Atmel_ATSAMR21G18-MR210UA_NoRFPads

Datasheet: http://ww1.microchip.com/downloads/en/devicedoc/atmel-42475-atsamr21g18-mr210ua_datasheet.pdf
grafik
img

This will fail the Travis-check since I choose the center to in the actual components center, not the center of the pads.
This Footprint does not have the additional Pads for routing the antenna on the base-pcb. Instead there is a keep-out zone according to section 2.2.1.

Please note that the datasheet has rather poor quality. For example section 2.2.1 and section 2.2.2 have the same header, but cover different topics. And there are contradictions in the pinout.. this will be fun for the symbol :)

---


cp-aquila:
with this additon https://github.com/KiCad/kicad-library-utils/pull/260  to the travis-checks the F6.2 would not fail for this FP.

poeschlr:
I think this should go into RF_Module.

the courtyard clearance towards pads is too large.
Screenshot from 2019-07-28 15-28-12

cp-aquila:
Thanks for the review.
Moved to RF_Module. This now also matches the Symbol-PR https://github.com/KiCad/kicad-symbols/pull/2021
Couryard is fixed. Turns out I measured it from the mask, not the copper. Stupid me.

cp-aquila:
I made another commit that removes the per-pad setting of paste and cream clearance and sets them for the whole symbol.
According to the datasheet they all have the same values, so setting local values is fine according to F9.2 (but generate another error).

poeschlr:
looks good, thanks

PR #1749, opened 2019-07-26
Inductor_SMD: Add Vishay IFSC-1515AH Low Profile, High Current Inductors

gmsotavio:
Datasheet: https://www.vishay.com/docs/34295/sc15ah01.pdf
IFSC-1515AH_2

Screenshot_2019-07-26_10-26-41

---


gmsotavio:
Enhanced Fab. drawing
IFSC-1515AH_3

cpresser:
Seems good to me.
I checked:
- Courtyard distance = 0.25mm
- Body size = 4mm
- Pad size = 1.15x3.6
- Pad distance = 1.8 and 4.1mm
- Silkscreen2copper clearance > 0.2mm

evanshultz:
Thanks!

One possibly important thing and two minor things:
1. Should PN have `-01` at the end, as in the datasheet? It looks like that's really part of the MPN.
1. The value can be at -3 in the Y direction instead of 3.35.
2. Can the fab refdes be the typical 1x1x0.15mm and still fit in the fab outline?

gmsotavio:
> 1. Should PN have `-01` at the end, as in the datasheet? It looks like that's really part of the MPN.

Hum. I had this thought. But lets analyze the current Vishay SMD inductors in the latest version of KiCad library:
Screenshot_2019-07-26_23-49-02
Note that the others omit the remaining part of the MPN. For example the IHLP-1212 has for the following variants:

IHLP-1212AB-11 - https://www.vishay.com/doc?34299
IHLP-1212BZ-11 - https://www.vishay.com/doc?34289
IHLP-1212AE-11 - https://www.vishay.com/doc?34300 
IHLP-1212AB-5A - https://www.vishay.com/doc?34527
IHLP-1212AB-A1 - https://www.vishay.com/doc?34528
IHLP-1212AZ-A1 - https://www.vishay.com/doc?34532
IHLP-1212BZ-5A - https://www.vishay.com/doc?34535

All this parts share the same land pattern. This open two possibilities. The first is drop the "AH" from "L_Vishay_IFSC-1515AH" to be consistent with other footprints in library. The second is keep the "AH" and append the "-01" as you've suggested and review the other footprints in library. What do you think?

> 2. The value can be at -3 in the Y direction instead of 3.35.
Done!

> 3. Can the fab refdes be the typical 1x1x0.15mm and still fit in the fab outline?
The result would be the following
Screenshot_2019-07-27_00-10-28
Up to four characters, it still remain inside fab outline limits. Should I change this?

poeschlr:
It seems that the two letters indicate the height of the component. So it might be a good idea to include it as it means that they require different 3D models. I am not sure what the two numbers after the two letters represent. (They call it the code for the series.)

The reference looks good inside as is.

---

The courtyard clearance must take the pads into account.
Screenshot from 2019-07-27 08-28-43

gmsotavio:
Courtyard has fixed
Screenshot_2019-07-27_10-45-22

> I am not sure what the two numbers after the two letters represent. (They call it the code for the series.)

For example IHLP-1212AB-5A (https://www.vishay.com/doc?34527) and IHLP-1212AB-A1 (https://www.vishay.com/doc?34528) have the same body dimensions, same land pattern, but different code for the series.

Screenshot_2019-07-27_10-59-40

Appending the code for series implies that we will need to provide a different footprint for each part with different code series. I think which is reasonable ignore this code.

> Edit: we might even consider appending _4x4x1.8mm (To fully fit with the documented naming convention. And yes I am aware that this means that nearly all inductors need renaming. Will be quite a bit of work.)

I agree. Done!

evanshultz:
@poeschlr 
Is this good with you? It appears the pad edge-to-edge isn't dimensioned in the latest drawing above, but I can see it is correct. So I think it just comes down to the naming convention you want, and as I noted at https://github.com/KiCad/kicad-footprints/pull/1735#issuecomment-516054543 I do not see a convention for SMD inductors in KLC.

evanshultz:
@gmsotavio 
Can you look into the CLA? Obviously you agree so I'm not sure what happened here.

evanshultz:
@chschlue 
Are you OK if I merge this? I think we can safely expect @gmsotavio agrees (he's a member).

chschlue:
@evanshultz Go ahead.

PR #1751, opened 2019-07-27
HRO_TYPE-C-31-M-12: remove USB symbol on Fab layer

diegoherranz:
USB type-C symbol removed from Fab layer which would go against
F5.2.1.b of KLC at http://kicad-pcb.org/libraries/klc/F5.2/

Pointed out on https://github.com/KiCad/kicad-footprints/pull/1631#issuecomment-515557184
by @evanshultz

Also, "%R" moved to exactly 0,0 for simplicity. It was close but not quite.

@evanshultz: I've taken the liberty of assigning this to you given it was pointed out by you. Hope that's OK. Let me know otherwise. Thanks!

hro


poeschlr:
thanks

Issue #1730, opened 2019-07-22
Wago 236 terminal block wrong courtyard


chschlue:
Wago 236 single pole courtyards are too wide, as these terminal blocks are supposed to be stacked.
~PR https://github.com/KiCad/kicad-footprints/pull/1729~

PR #1752, opened 2019-07-27
Add Wago 236 series 2-pin terminal blocks

chschlue:
Add Wago 236 series 2-pin terminal blocks

TerminalBlock_WAGO_236-401_1x01_P5
TerminalBlock_WAGO_236-602_1x02_P10

236-[4-6]xx double pin terminal blocks.
~Left and right courtyard of single pole versions = fab for stackability.~

Replaces PR #1729 
Script PR https://github.com/pointhi/kicad-footprint-generator/pull/403
Issue #1730 

Datasheets:
https://www.wago.com/de/products/datasheets/236-401.pdf?product=236-401&lang=en
⋮
https://www.wago.com/de/products/datasheets/236-624.pdf?product=236-624&lang=en

---


poeschlr:
The drill hole diameter is given as the minimum (tolerance is given as -.../+0.1 which i would assume means the same as -0/+0.1, In the catalogue they only give positive tolerance for drill size so this also seems to suggest 0 tolerance towards smaller.) 

I suggest to center the drill size in the tolerance range (1.15mm drill size would then be correct)

poeschlr:
looks good, thanks

PR #1753, opened 2019-07-27
Fix Wago 236 series 1-pin terminal blocks

chschlue:
Fix Wago 236 series 1-pin terminal blocks

TerminalBlock_WAGO_236-101_1x01_P5
TerminalBlock_WAGO_236-302_1x02_P10

236-[1-3]xx single pin terminal blocks.
~Left and right courtyard of single pole versions = fab for stackability.~

Replaces PR #1729 
Complements PR #1752 
Script PR https://github.com/pointhi/kicad-footprint-generator/pull/403
Issue #1730 

Datasheets:
https://www.wago.com/de/products/datasheets/236-101.pdf?product=236-401&lang=en
⋮
https://www.wago.com/de/products/datasheets/236-324.pdf?product=236-624&lang=en

---


poeschlr:
I am not so sure about this. This kind of feels like a dirty workaround for the missing "yes this is on purpose do not show again" feature of DRC. I also feel that with only the reduced courtyard one now has the opposite problem of false positives. 

The "proper" workaround would require 3 footprints. The one you have here for connectors in the middle of the stack and then also one for the left and one for the right ends that have the courtyard normally on the appropriate side.

chschlue:
Extra left and right footprints for the same part aren't exactly what I'd call beautiful either, but I see your point.

poeschlr:
So how about we make only one that sadly generates false positives that the user can easily ignore instead of having one that will not complain if parts are too close to the end?

chschlue:
That's what we've got now, isn't it?
One or two are easy to ignore, the problem I have with the stackable ones is that it's not uncommon to stack 20 or more and 20 or more false positives are pretty ugly.

poeschlr:
In this case we either need to supply 4 footpirnts as follows:
- one with crtyd clearance all around (for using it alone)
- one with crtyd clearance only on top and bottom (middle of a stack)
- one with crtyd clearance all around except on the right side (left end)
- and one for the right end (mirror of left end crtyd)

Or we accept that kicad can not do everything and that some workarounds are left for the user. In that case only the first listed option (the one we already have) is needed.

And as i assume v6 will bring a proper solution i wonder if it is a good idea to introduce a workaround into the library. Maybe an issue where we discuss this might be a good idea including a querry to the devs if there will be a proper solution in v6. (If not then we might as well properly codify the workaround and document it in the KLC.)

chschlue:
Alright, how about fixing the wrong naming (and adding the two pin variants) for 5.1.5 and leaving the courtyard discussion for v6?

poeschlr:
Sounds good to me.

poeschlr:
thanks

PR #1758, opened 2019-07-30
Add Cromatek 0.56in 7-segment display footprint

RockyTV:
Required for https://github.com/KiCad/kicad-symbols/pull/1464

Datasheet: https://archive.org/download/CTKD1x8K/Cromatek%20D168K.pdf


RockyTV:
Hm, I have no idea how to set the footprint origin to match pad 1's location. Any help is appreciated.

RockyTV:
Okay, fixed it. I have only one question remaining:

In the datasheet, pins 3 and 8 are connected and they are the cathode pin. Does the [F4.3 rule](http://kicad-pcb.org/libraries/klc/F4.3/) apply in this case?

myfreescalewebpage:
@RockyTV starting the review here.

> In the datasheet, pins 3 and 8 are connected and they are the cathode pin. Does the F4.3 rule apply in this case?

It's something we do for SMD, but not for THT in my memory so the current proposition is good

- [ ] Datasheet link has changed it's now https://ia800903.us.archive.org/24/items/CTKD1x8K/Cromatek%20D168K.pdf
- [ ] Name should be simply "D1X8K"
- [ ] 3D model name to be modified according to previous comment and to fix travis error
- [ ] Pad 1 should be rectangular (see travis error)
- [ ] Position of the pin is not correct, pin 2 is at 2.54mm of pin 1, pin 10 has no X offset, etc.
- [ ] Size of the pin is too small, from the other footprint I recommend you size=1.6mm and hole=0.8mm
- [ ] You should have a marker for pin 1 on F.Fab and F.SilkS layers
- [ ] F.Fab is not placed horizontally has expected, pin 1 should be at 1.13mm from F.Fab on the left
- [ ] Offset from F.Fab to F.SilkS: 0.11mm
- [ ] Offset from F.Fab to F.CrtYd: 0.25mm

Joel

myfreescalewebpage:
@RockyTV ping

RockyTV:
Sorry for the delay, I'm trying to fix these issues right now. I have some questions:

> -  Position of the pin is not correct, pin 2 is at 2.54mm of pin 1, pin 10 has no X offset, etc.
Well, according to the datasheet, the distance between each pin is 2.54mm. About the pin 10 X offset, I don't understand, it's aligned with pin 1, so what's the problem here?


> Size of the pin is too small, from the other footprint I recommend you size=1.6mm and hole=0.8mm
The datasheet says the radius of the pin is 0.5mm but making the hole 0.8mm shouldn't cause any problems.

> Name should be simply "D1X8K"
I'd suggest CTK-D1X8K, as another manufacturer might have the same model name.

myfreescalewebpage:
> Well, according to the datasheet, the distance between each pin is 2.54mm. About the pin 10 X offset, I don't understand, it's aligned with pin 1, so what's the problem here?

In the design of your footprint, it is not aligned with pin 1

> I'd suggest CTK-D1X8K, as another manufacturer might have the same model name.

"Cromatek_D1X8K" in this case, we do not use short name for manufacturers

RockyTV:
> In the design of your footprint, it is not aligned with pin 1
Yeah I noticed it after I was fixing the offset between the outlines.

How should I add the pin 1 marking in F.Fab and F.SilkS?

myfreescalewebpage:
On F.Fab we usually have a corner at 45° or a triangle
On F.FilkS various possibilities exist, you can check "CA56-12CGKWA" in the lib, which is with a small line in front of pin 1

RockyTV:
I don't know why but it's complaining about the origin not being set to pad 1, but I've already set it to the same position as pad 1.

myfreescalewebpage:

RockyTV:
I think it's because I renamed the footprint file.

myfreescalewebpage:
You still have this issue:
- [ ] F.Fab is not placed horizontally has expected, pin 1 should be at **1.17mm** from F.Fab on the left, currently it's at 1.15mm. Same comment on the right, should be at 11.33mm, currently it's 11.43mm. Same comment at the bottom, same comment at the top. Please do it carefully.
- [ ] F.SilkS pin 1 marker is very small and will be covered when the part is on the PCB, which is not permitted. I suggest:
Capture

Joel

RockyTV:
>  F.Fab is not placed horizontally has expected, pin 1 should be at 1.17mm from F.Fab on the left, currently it's at 1.15mm. Same comment on the right, should be at 11.33mm, currently it's 11.43mm. Same comment at the bottom, same comment at the top. 

How do you measure it? The previous listing said it should be at 1.13mm, which one should I use?

myfreescalewebpage:
1.13 was an error sorry.

(12.5mm - (4*2.54) ) / 2 = 1.17mm

RockyTV:
I think it's solved but now there's a problem about the courtyard lines not being on a 0.01mm grid.

myfreescalewebpage:
Checking...

myfreescalewebpage:
Top part of F.Fab has also a problem. Total height of your device should be 19mm.

Capture

When F.Fab is fixed, then F.SilkS distance from F.Fab should be 0.11mm (center to center) so it's easy to fix Silkscreen. And then F.CrtYd distance from F.Fab should be 0.25mm (center to center).

myfreescalewebpage:
Just type the right values using a calculator...

RockyTV:
I just pushed a new commit and double-checked F.Fab and the width is 12.5 and height is 19.0 as it should be. Sorry for the confusion!

myfreescalewebpage:
Thanks, clearly better! Pads OK, F.Fab OK, External F.SilkS OK and F.CrtYd OK.

Now about the internal drawing: The horizontal segments of the digit should be horizontal according to the datasheet. No need to respect the exact size of the digit itself but just the general design is great to have.

Joel

RockyTV:
Is it a problem if I make the vertical segments actually vertical, like this?

footprint

myfreescalewebpage:
I think this is acceptable. Is it your final version ?

RockyTV:
Yes it is, unless there's something I missed. Commit a14970a has the changes to the internal drawing.

myfreescalewebpage:
Checking

myfreescalewebpage:
Looks great, thanks. I have a last query, sorry, I have learned today pin 1 should be at the top left position of the footprint. Can you rotate everything by 90° around pin 1 please ? Like that:

Capture

RockyTV:
It is done, everything rotated.

myfreescalewebpage:
Thanks, great job you have done on this PR @RockyTV ! Again sorry for this last change. Merging your contribution now :)

PR #1602, opened 2019-05-16
Add a few custom THT transformers

evanshultz:
A few discussions have brought up how to name custom transformers (those that are custom-designed and not an off-the-shelf part) and so I want to kick things off. Right now I've just added one so that we can iron out the details, but there are more I will contribute.

Here are some discussions I could find:
- https://github.com/KiCad/kicad-library/issues/1447#issuecomment-325499139
- https://github.com/KiCad/kicad-symbols/issues/146 (for symbols, but related)

For these parts the winding construction is captured elsewhere so the core and bobbin info are the only documents needed to verify footprint correctness. Notes:
1. KLC doesn't have a suitable section for footprint naming. Thoughts?
1. While the core and bobbin are selected independently (they do rely on each other but various vendors make compatible parts) a 3D model may be specific to each bobbin vendor. Should the bobbin vendor and MPN then be called out as part of the footprint name?
1. Pin numbering is taken from the bobbin to make life simple and avoid specific implementation things that could happen when starting with a symbol that result in the pin numbers not matching the bobbin. The actual transformer design can vary so this footprint can be used as a basis if the user desires.
1. Footprint naming with missing pin(s) is/are an unresolved issue.
1. I kept the courtyard offset at 0.25mm since KLC doesn't mention this type of footprints but perhaps that's too small?
1. It is common on these parts to use oval pads because you need considerable heat for rework on the bottom but big pads aren't needed on other layers (during wave solder they're actually harmful). This is my attempt to implement this pad style.

Core: http://www.e-coretech.co.kr/pdf/cores/ferrite/EE.pdf (page 2)
Bobbin: http://www.pinshine.com/upload/product/20090824092203_link.pdf
image

---

All contributions to the kicad library must follow the [KiCad library convention](http://kicad-pcb.org/libraries/klc/)

Thanks for creating a pull request to contribute to the KiCad libraries! To speed up integration of your PR, please check the following items:

- [ ] Provide a URL to a datasheet for the footprint(s) you are contributing
- [ ] An example screenshot image is very helpful 
- [ ] If there are matching symbol or 3D model pull requests, provide link(s) as appropriate
- [ ] Check the output of the Travis automated check scripts - fix any errors as required
- [ ] Give a reason behind any intentional library convention rule violation.

PR #1760, opened 2019-08-01
[WIP] add Block THT transformers

lorem-ipsum:
This is the first revision of my attempt to script the collection of THT transformers from Block. Currently it contains the packages used in the FL series. The website is not really good, the best overview for the dimensions is the [complete catalogue](https://www.block.eu/fileadmin/Blaetterkatalog_und_ebook/eBook.pdf) (~60MB), the corresponding dimensions are on page 281. My corresponding python mess can be found at https://github.com/lorem-ipsum/kicad-footprint-generator/tree/block (no PR yet, will post it once the path is clear).

At the current state I have two questions:
1. Any comments about the footprint in general? Currently it is very simple, containing only the required markers, but no extra drawings. Some of the currently existing footprints have a lot of extra bells and whistles (e.g. marking the coils on fab layer, writing Pri and Sec on the Silk, etc.). Should I add something like this?
2. Probably more important: How to name them correctly? The current naming uses the core type as they appeared to have a unique housing each. However I already found the obviously existing exception from the rule as the FL 52 and the FLD 48 (on page 285) share the same core, but have slightly different dimensions. The obvious alternative is to use the Type name (e.g. FL 52). This is definitely unique, but will create many duplicate footprints when adding all transformers (which I intend to do).

Example screenshot of one of the footprints:
Screenshot_20190801_225902

---


evanshultz:
1. I prefer simplicity. No need to add this extra stuff. As you can see some interpretations vary, but KLC does not require this (in fact it goes against extraneous markings). IMO I would personally remove this stuff from all existing footprints unless somebody objects.
1. Since the 3D model is unique for each, I don't see any duplicated dimensions (LxWxH). Since footprints are relatively small files, if having all of them in the library is useful to KiCad users then I would include each one. I would also think basic dimensions would be nice since a generic symbol could fit many footprints and so the user will be able to get a sense of the size (and VA) if that's in the footprint name. Something like `Transformer_BLOCK_FL2_53x44x17.6mm`. https://github.com/KiCad/kicad-footprints/pull/1602, and the issues I linked, contain some discussion about transformer footprint naming. As these are off-the-shelf and not custom transformers, though, not all the rules apply.

When you add the URL, please include the page number of the PDF so that a navigating a big file like the one you linked isn't tough. Append `#page=281` to the URL, for the FL series.

In addition, two other comemnts:
1. I don't see fab or silk pin 1 marks.
1. Typically the value is below the part and not inside.

lorem-ipsum:
I now changed the naming scheme to `Transformer_Block_{series}-{power}[-outputs]_{L}x{W}x{H}mm` and created all series which share the dimension labeling with the FL and the VB series. I also added pin 1 makers on fab and silk and moved the value below the package. I had it inside as the components are rather large, so there is enough of space.

I also got the unique datasheet links for each package using a small script. I always used the 12V variant as this exists for all transformers (output voltage has no effect on the package).

Below a reference screenshot of one of the smallest transformers:
Screenshot_20190808_132825

lorem-ipsum:
another question which I stumbled upon when reading this issue: https://github.com/KiCad/kicad-symbols/issues/146. How should the pins be numbered? The datasheet gives no numbering scheme. Currently they are just numbered counter clockwise. The suggestion in this issue is to separate the numbers between primary and secondary side. Should I do this?

PR #1762, opened 2019-08-02
add Mini-Circuits DB1627

lorem-ipsum:
[Datasheet](https://ww2.minicircuits.com/case_style/DB1627.pdf).

Screenshot_20190802_155924

---


evanshultz:
The generator you used is for no no-lead packages. Your package has leads. I think you will need to make this footprint manually as I don't believe we have a generator for this package type.

lorem-ipsum:
meh, you're right. I viewed the bottom as a continuous plane, but then looking closely one can see the leads. Will build it manually off the Mini-Circuits land pattern.

evanshultz:
I removed the script PR from the first post since it's not valid any longer.

1. Clean up the description and keywords so they're not the now-irrelevant script-generated ones.
1. You rounded the courtyard width up to get it on grid, so the height should change to +/-2.67mm. Or math has failed me.

Can you also post a new screenshot?

lorem-ipsum:
Updated the description and fixed the courtyard. Checked the numbers again, you're right.
Screenshot_20190808_101558

evanshultz:
Thank you once again!

PR #1766, opened 2019-08-04
 Added Samtec_FTSH-105-xx-x-DV-007

pegasus17:
Added Samtec FTSH .050" Micro Terminal Strip 5 pins pre row with polarized position at pin 7.
Datasheet: http://suddendocs.samtec.com/catalog_english/ftsh_mt.pdf, http://suddendocs.samtec.com/prints/ftsh-1xx-xx-xxx-dv-xxx-footprint.pdf

This footprint is used for the ARM Cortex Debug Connector (http://infocenter.arm.com/help/topic/com.arm.doc.faqs/attached/13634/cortex_debug_connectors.pdf)

3D model PR: [602](https://github.com/KiCad/kicad-packages3D/pull/602)

Samtec_FTSH-105-007-DV
Samtec_FTSH-105-DV


diegoherranz:
Without jumping in (too much), 
may I suggest adding "PinHeader_2x05_P1.27mm_Vertical" or similar to the footprint name so that it shows up for footprint filters like that used for "Conn_ARM_JTAG_SWD_10"?

Also, I wonder whether the 7th pin should be present still even if not present in the connector. That gives you better flexibility populating alternatives.

Thanks!

evanshultz:
1. This can be in the generic Samtec library. So undo the lib table change and move the footprint.
1. Include both units of pitch in the description, or ditch the inches in favor of millimeters.
1. I think changing `5 Pins per row with polarized position at pin 7` to `2x5 with pin 7 removed` is shorted but even more informative.
1. I would remove the ARM and debug stuff from the keywords. This is a generic header that could be used for many applications.
1. Generally we will put refdes on top and the value on the bottom.
1. The silk is quite far from the fab lines. Offset them by 0.12-0.2mm.
1. Courtyard should be 0.5mm from the pads or body on a 0.01mm grid.
1. You need to add a bevel to the fab rectangle corner by pin 1.
1. I'm not totally sure if the highest pin number should be 9 or 10. Thoughts?

evanshultz:
@diegoherranz 
Good points. Thanks for chiming in.

Unfortunately [that symbol](https://github.com/KiCad/kicad-symbols/blob/master/Connector.lib) uses question marks and not asterisks in the footprint filter. It is `PinHeader?2x05?P1.27mm*`. So maybe something like `PinHeader_2x05_P1.27mm_Samtec_FTSH-105-xx-x-DV-007`? Feels a little awkward. That symbol also includes pin 7 and I'm not sure if that's the best approach (I'm not too familiar with this topic). Yeah, probably some change here can help and there's lots to think about.

Not if the footprint is specifically for a connector without a 7th pin, I don't think. That's what the trailing `-007` of the MPN means. That being said, perhaps it wouldn't stick out if all FTSH footprints were present. Right now it's a little odd to have this one specific version but not footprints to fit all the common versions of this header family.

diegoherranz:
The main reason (the only?) why this connector without pin 7 exists is the ARM Cortex debug connector. We've got to make it work with the symbol Conn_ARM_JTAG_SWD_10: that is both pin numbering and footprint filters.

With that in mind, replying to a few of the comments/questions:



> Unfortunately that symbol uses question marks and not asterisks in the footprint filter. It is PinHeader?2x05?P1.27mm*. So maybe something like PinHeader_2x05_P1.27mm_Samtec_FTSH-105-xx-x-DV-007? Feels a little awkward.

It's probably the best option given the above.
On suppliers like Mouser and Digikey, these connectors without pin 7, still show up as 2x05 so I think it is OK.

Would it make sense to replace the question marks by asterisks on the symbol? We could do that on a separate PR.

> I'm not totally sure if the highest pin number should be 9 or 10. Thoughts?

We need to go up to pin 10, skipping pin 7, to match the ARM standard definition of the connector and the symbol in our library.

> Not if the footprint is specifically for a connector without a 7th pin, I don't think.

I get your point. Being specific for this connector, it may make sense to have no pad on 7 position.

However,  even if ARM defines the connector for this purpose without pin 7 serving as a key feature, from my experience in the real world, it is more common to have all the 10 pins (I know it's not the right thing to do, but I guess it makes it much easier supply chain wise).

Maybe we could create another footprint for the identical connector with the 10 pins (FTSH-105-xx-x-DV), which should be trivial once this one is complete/approved. Personally, and for flexibility reasons, I would use that footprint even when populating the correct connector without the pin 7, but having both footprints in the library, the user can decide.

Thanks!

evanshultz:
I'm happy to merge this as-is, but then adding the full set of FTSH footprints gives this application and many others another solid footprint choice.

I would feel best to update the symbol with specific footprint filters for the headers recommended. And keep generic footprint filters, of course. So this 9-pin version, the full 10-pin version, and any other headers ARM suggests. That allows the current flexibility but specifically guides users to suggested headers. What do you all think?

pegasus17:
I have reworked the footprint (description, fab, silk, courtyard).
When this footprint is ok, I will add the Samtec_FTSH-105-xx-x-DV as well to give the user the choice of the 9-pin and the 10-pin version.

diegoherranz:
> I'm happy to merge this as-is, but then adding the full set of FTSH footprints gives this application and many others another solid footprint choice.

Agree. That sounds like a scripting job if anyone wants to do it :)

> I would feel best to update the symbol with specific footprint filters for the headers recommended. And keep generic footprint filters, of course. So this 9-pin version, the full 10-pin version, and any other headers ARM suggests. That allows the current flexibility but specifically guides users to suggested headers. What do you all think?

That would work but requires updating the symbol every time a new compatible connector/family is added.

I find it easier to keep the footprint filter as "header, 2x05, 1.27mm", and include that information even on specific brand footprints like this Samtec. I think that's more similar to what we do with audio jacks, for instance. Adding new jack footprints doesn't require updating the jack symbols. What do you think?

Anyway, having said that, I would be OK with that approach too if considered the best one.

Thanks!

pegasus17:
Should the footprint name be changed to following (regarding the footprint filter)?

Samtec_FTSH_105_DV_Header_2x05_P1.27mm
Samtec_FTSH_105-007_DV_Header_2x05_P1.27mm

diegoherranz:
> Should the footprint name be changed to following (regarding the footprint filter)?
> 
> Samtec_FTSH_105_DV_Header_2x05_P1.27mm
> Samtec_FTSH_105-007_DV_Header_2x05_P1.27mm

I think we normally go the other way around. Generic description first and then the specific brand and MPN. For example: `USB_C_Receptacle_HRO_TYPE-C-31-M-12.kicad_mod` or `SW_Push_1P1T_NO_CK_PTS125Sx43PSMTR.kicad_mod`.
 
Also, given the filter is `PinHeader?2x05?P1.27mm*`, "PinHeader" has to be at the beginning.

poeschlr:
@diegoherranz the KLC disagrees with you. See http://kicad-pcb.org/libraries/klc/F3.6/

for manufacturer specific connectors the the series and mpn come first (as they are the main identifyng thing. If i search for an FTSH series connector i want all of them listed together. A usb connector is different because i am likely searching for a mini usb connector not for a molex series x connector that just happens to be a mini usb)

diegoherranz:
Thanks, @poeschlr, for stepping in.

This connector, despite having a custom-ish footprint, it's a very standard connector. It's basically a pin header and there are tons of manufacturers making esentially the same thing.
It's not like other very specialised Samtec connectors, e.g. https://www.samtec.com/connectors/high-speed-board-to-board in which case you would search based on the series name.

I think I still find the value of this starting with `PinHeader_2x05 since that is what I would be looking for when trying to find a footprint for an ARM debug connector, for instance.

Regarding the footprint filters: if this doesn't start with `PinHeader_2x05`, the footprint filter of Conn_ARM_JTAG_SWD_10 would need to be updated.

Thanks!

diegoherranz:
I've just seen your "edit 2" comment.
Maybe this belongs better in the `Connector_PinHeader_1.27mm` library. 
We've got different styles within that one (SMT, TH, vertical, horizontal, etc) just by using suffixes, so I wouldn't think we need a separate `pin_header_p[pitch]mm_shrouded`.

Maybe just call it `PinHeader_2x05_P1.27mm_Samtec_FTSH-105-007-DV_Shrouded` inside the `Connector_PinHeader_1.27mm` library?

I would be happy with that. Would that work?

Thanks!

pegasus17:
Putting it into _Connector_PinHeader_1.27mm_ library will work form me.
In my opinion, it should be just called _PinHeader_2x05_P1.27mm_Samtec_FTSH-105-007-DV_ because of the Samtec -DV is normal pin header with no shroud.

poeschlr:
why include the samtec order number if you claim that this is a generic connector? If it is not generic then it has no place in this lib (would then go into the manufacturer specific lib), if it is generic then the manufacturer name/order number have no place (at most as an example order number in the description).

pegasus17:
It is a generic connector for 2x05 without pad 7. Would the name _PinHeader_2x05_P1.27mm_Vertical_SMD_Without_Pad7_ be ok for it?
Or wouldn't it be a generic connector at all due to the fact one pad is not available and an own library is needed?

PR #1767, opened 2019-08-05
LED_Osram_LxP47F reverse mount LED

openelectron:
Its been a while since I did a footprint but I have checked the KLC and I think this is ok.
Not sure how to do the thermal copper pads, so i used two overlapping pads, one with no resist layer.

datasheet:
https://dammedia.osram.info/media/resource/hires/osram-dam-5910133/LR%20P47F_EN.pdf

image

---


openelectron:
The rules check fail because the silkscreen crosses the thermal pad, which is covered in resist and therefore not exposed.  KLC says a silk width of 0.1 is allowed on compact designs. 

I will await comments before I change anything.
Thanks.

cp-aquila:
Hi openelectron,

Thanks for your contribution.
It seems like the linked datasheet is redacted or missing parts. At least for me some Items are just grey boxes. I can't find the actual landing-pattern in the osram pdf.
I did find a pdf which does render correct for me: https://www.mouser.com/datasheet/2/311/LR%20P47F%20-%20PointLED-318874.pdf
I checked your FP against that .pdf, page12, top drawing.

Here are my thoughts
1. Naming: Why "LxP47F" instead of "LR_P47F"?
2. Pad1 and Pad2 y-locations are 0.35mm, but should be 0.7/2 + 0.025
3. The big part of the body is 1.2 but should be 1.25
4. The two arcs of the body in F.Fab radius should be 1.9mm/2, but their radius is 1mm. I personally think that is okay, its pretty hard to get the arcs correct for a fractional radius. You can do it with a little bit of trigonometry if you like :)
5. Thickness of the REF** could be 0.1mm 
6. A small part of F.Silk is below the part. But all silkscreen should be visible after assembly. How about this: 
grafik
7. Naming again: KLC F3.1 likes the orientation after the dimension. "LED_Osram_LxP47F_2.1mm_ReverseMount"
8. You can set the copper-layers of the 'small' pads to none, then there will be no overlapping copper.


Silkscreen-clearance to copper is always bigger than 0.1mm, so thats fine.
The Pin1 marker in F-Fab looks good.

openelectron:
@cp-aquila 

- I named it like that as the R is the colour of the led.

- Changed datasheet, The one I had from Farnell is a bit clearer, I tried to link to the MFR site instead of a vendor but, as you found that datasheet is broken.

- If the radius on F.Fab is acceptable, I'll leave it.!

- All other suggestions are incorporated.

CI still fails as the silkscreen lint for pin 1, crosses the pad (although covered in solder resist, so I don't think this is a problem)

LED_SMD.pretty/LED_Osram_LxP47F_2.1mm_ReverseMount.kicad_mod
```diff
@@ -1,41 +1,38 @@
...
   (fp_line (start -1.7 0.3) (end -1.5 0.1) (layer F.Fab) (width 0.1))
   (fp_line (start -1.5 0.1) (end -1.3 0.1) (layer F.Fab) (width 0.1))
   (fp_line (start -1.7 0.3) (end -1.7 0.6) (layer F.Fab) (width 0.1))
   (fp_line (start -3.55 -1.5) (end 3.55 -1.5) (layer F.CrtYd) (width 0.05))
   (fp_line (start 3.55 -1.5) (end 3.55 1.5) (layer F.CrtYd) (width 0.05))
   (fp_line (start 3.55 1.5) (end -3.55 1.5) (layer F.CrtYd) (width 0.05))
   (fp_line (start -3.55 1.5) (end -3.55 -1.5) (layer F.CrtYd) (width 0.05))
-  (pad 2 smd rect (at 1.55 -0.35) (size 0.5 0.7) (layers F.Cu F.Paste F.Mask))
-  (pad 1 smd rect (at -1.55 0.35) (size 0.5 0.7) (layers F.Cu F.Paste F.Mask))
+  (pad 2 smd rect (at 1.55 -0.373) (size 0.5 0.7) (layers F.Paste F.Mask))
```
cp-aquila: should  be at "-0.375"

LED_SMD.pretty/LED_Osram_LxP47F_2.1mm_ReverseMount.kicad_mod
```diff
@@ -1,41 +1,38 @@
...
+    (effects (font (size 0.6 0.6) (thickness 0.1)))
   )
-  (fp_line (start -1.35 -0.2) (end -2 -0.2) (layer F.SilkS) (width 0.1))
-  (fp_line (start -2 -0.2) (end -2 0.8) (layer F.SilkS) (width 0.1))
-  (fp_line (start -2 0.8) (end -2 0.9) (layer F.SilkS) (width 0.1))
   (fp_line (start -2 0.9) (end -1.05 0.9) (layer F.SilkS) (width 0.1))
-  (fp_arc (start 0 0) (end 1.099999 -0.799999) (angle -135.5239235) (layer F.SilkS) (width 0.1))
-  (fp_arc (start 0 0) (end -1.05 0.9) (angle -130.9717287) (layer F.SilkS) (width 0.1))
+  (fp_arc (start 0 0) (end 1.099999 -0.799999) (angle -107.9452625) (layer F.SilkS) (width 0.1))
+  (fp_arc (start 0 0) (end -1.05 0.9) (angle -100.1092985) (layer F.SilkS) (width 0.1))
```
cp-aquila: Those two arcs don't look alike.

poeschlr:
pads without copper are not allowed to have a pin number.

openelectron:
@poeschlr 
Thanks, I have removed the pad numbers.

openelectron:
@cp-aquila 
The arcs should now match and are as you calculated.

cp-aquila:
- [ ] Some lines do in F.Fab do not properly intersect:
grafik
Those should all end at 0.715 to match the arcs.
grafik
(please also check the other lines on F.Fab not shown in this screenshot)

- [ ] The arcs on F.Silk are still not the same.

LED_SMD.pretty/LED_Osram_LxP47F_2.1mm_ReverseMount.kicad_mod
```diff
@@ -31,6 +26,11 @@
   (fp_line (start 3.55 1.5) (end -3.55 1.5) (layer F.CrtYd) (width 0.05))
   (fp_line (start -3.55 1.5) (end -3.55 -1.5) (layer F.CrtYd) (width 0.05))
   (fp_arc (start 0 0) (end -0.715 0.625) (angle -97.6) (layer F.Fab) (width 0.1))
+  (fp_line (start 0.714073 0.626059) (end 1.315 0.626059) (layer F.Fab) (width 0.1))
```
cp-aquila: Those are weird numbers.

cp-aquila:
- [ ] The two arcs on F.Silk are different.

LED_SMD.pretty/LED_Osram_LxP47F_2.1mm_ReverseMount.kicad_mod
openelectron: I agree. tried again.. I wish I had never changes those arcs!

cp-aquila:
Looks good. No further comments. I think this is ready to be merged.

evanshultz:
I could more things with just a quick look:
1. My math says the paste/mask pads should be +/-1.8mm in the X direction. 2.6mm / 2 + 0.5mm = 1.8mm.
1. One of those pads has a Y location of 0.373mm and it should be 0.375mm.
1. The description should have more info than just a link. Perhaps `OSRAM, reverse-mount LED, SMD, 2mm diameter, <link>`. Usually the keywords are more basic and often info is duplicated (probably not good practice but KLC doesn't constrain this.)
1. I also wonder why it's not `Lx_P47F` with an underscore. There is a space in the MPN.
1. I could 13 elements on the fab layer but I think only 12 should be required.

openelectron:
@evanshultz 
1. I think it is correct (2.6/2) + (0.5/2) = 1.55
2. corrected.
3. Is this the keywords field? I'm happy to change just need to clarify to where you are referring.
4. renamed
5. I also count 13, I'm not sure how to do it in 12?

evanshultz:
1 You are right.
3 No, description. Sometimes the keywords duplicate something in the description, but let's add some nice info to the description that describes the part as I showed above.
5 I missed the pin 1 mark, which makes it 13.

Good to me!

cp-aquila:
> 
> 
> I could more things with just a quick look:
> 
>     1. My math says the paste/mask pads should be +/-1.8mm in the X direction. 2.6mm / 2 + 0.5mm = 1.8mm.
> 
>     2. One of those pads has a Y location of 0.373mm and it should be 0.375mm.
> 
>     3. The description should have more info than just a link. Perhaps `OSRAM, reverse-mount LED, SMD, 2mm diameter, <link>`. Usually the keywords are more basic and often info is duplicated (probably not good practice but KLC doesn't constrain this.)
> 
>     4. I also wonder why it's not `Lx_P47F` with an underscore. There is a space in the MPN.
> 
>     5. I could 13 elements on the fab layer but I think only 12 should be required.

Dammit. I also mentioned 2 and 4 earlier, but did not check if they are fixed. My bad!
I need to get better organized.
@evanshultz is there a checklist maintainers use to review footprints? Or do you make that one up on the fly?

openelectron:
@cp-aquila You did pickup on points 2 and 4, they are both already fixed.

@evanshultz I'm still unclear which field is the description field...
image

cp-aquila:
That would be http://kicad-pcb.org/libraries/klc/F9.1/
It has a good example picture.

That UI and the field-names have changed with Kicad 5.1.x. Your screenshot shows the old UI, the KLC already has the new UI. The datasheet field is now replaced with Description.

Please note that one has space-seperated fields, the other is comma seperated.

openelectron:
@cp-aquila  That explains why I couldn't find the description!

Should be done now.  Thanks for your help.

evanshultz:
@cp-aquila 
Not that I know of. I usually start by looking at the diff in general, then comparing with the datasheet (pin numbers, types, etc.) and then finish by looking at the symbol and pin placement. But others might do it totally differently. In general the check scripts help to catch many things that would be easy for humans to miss, letting us focus on the more subjective elements or things tricky to rigidly check.

Is this good to merge for you? I'm ready but don't want to merge without your approval.

Thanks for another great review! I believe we would all be happy for you to join to team, if you like.

cp-aquila:
> 
> 
> @cp-aquila
> Not that I know of. I usually start by looking at the diff in general, then comparing with the datasheet (pin numbers, types, etc.) and then finish by looking at the symbol and pin placement. But others might do it totally differently. In general the check scripts help to catch many things that would be easy for humans to miss, letting us focus on the more subjective elements or things tricky to rigidly check.
> 
Since I am easily distracted and miss things (see above), a checklist will be of great use for me. I will start to draft one.

> Is this good to merge for you? I'm ready but don't want to merge without your approval.
> 

I checked the recent diffs, seems fine to me.
There is one thing though where I am not sure how to handle it. That is the "_2.1mm_" part of the name.
There is no leading letter or indicator telling what is 2.1mm. It could be body diameter, height, drill diameter. So I would rather specifiy what it is.
" LED_Osram_Lx_P47F_Drill2.1mm_ReverseMount" (according to KLC F2.2)

> Thanks for another great review! I believe we would all be happy for you to join to team, if you like.

Yep, I think I will give it a try. But I would rather join with my private account @cpresser, the current one ist my company account. Right now work is slow and I can spend some time on KiCad, but once vacation time is over that won't be the case anymore.

openelectron:
@cp-aquila  what about B2mm, as I suspect the led size is more important than the drill size?

cp-aquila:
> 
> 
> @cp-aquila what about B2mm, as I suspect the led size is more important than the drill size?

That sounds reasonable. Not sure if B2 or D2 is better, "D2mm" sounds best to me.

poeschlr:
The B prefix would require giving more than one dimension (Syntax: B[size x]x[size y]{x[height]}mm; height is optional.). Use D for the body diameter.

cp-aquila:
- [ ] 3D-Model Filename is not consistent
- [ ] Diameter should be 1.9mm (or am I wrong here?)

openelectron:
The datasheet describes it as a 2mm led, however the dimension is actually 1.9mm with a tolerance +/- 0.1mm. 
My vote is D2mm but I'm happy to change it to whatever is most appropriate 

I just spotted the missing _ in the 3d model name..

cp-aquila:
Page 11:
Correct, the drawing shows 1.9mm, the second drawing on the left has 1.8..2.0mm
Thats already bad style, I would have failed my university drawing class for giving two contradicting values for the same dimension.

So I think that 2.0mm is fine. Then it also fits the Description/Keywords and the title-page of the Datasheet.

@evanshultz I am okay with this being merged.

@openelectron Thanks for you patience. This was my first 'real' review and had some hickups.

evanshultz:
KLC isn't completely clear if dimensional info is required in the footprint name since this is a manufacturer-specific footprint. But it doesn't hurt either. The existing manufacturer-specific SMD LED footprints are mixed in this regard. So let's do it.

Thanks all!

evanshultz:
The F5.1 violations caught by CI should be addressed but that's all that's left from me.

openelectron:
@evanshultz  The error in question is referring to the silkscreen in crossing a pad. The pad in question is covered in solder, resist and therefore is this not valid?

evanshultz:
Yes, but silk line widths are 0.1mm and should be 0.12mm.

evanshultz:
Thank you!

PR #1770, opened 2019-08-08
add SMA connector Molex 73251-2120

lorem-ipsum:
This adds an edge mount SMA connector from Molex [73521-2120](https://www.molex.com/pdm_docs/sd/732512120_sd.pdf). Used the existing 73251-1153 as a reference. Chose the board egde as reference for the origin. This makes it easier to align the component with the board edge, and there is no clear center of the component. Violations due to missing layers on the back pins which are on purpose.

Screenshot_20190808_091513

---


cpresser:
This is pretty close to "SMA_Molex_73251-1153_EdgeMount_Horizontal" which is good.
But there are some things I would like to discuss
- [ ] Add Pin1-Marker on F.Fab (like SMA_Amphenol_132289_EdgeMount)
- [ ] Having copper directly on the board-edge might violate a lot of DRCs. Afaik there should be ~0.25mm between board-edge and copper. Depending on your Fab that produces the board. tbh. I am unsure how to handle this.
- [x] According to KLC (F4.2) Pin1 should be on the top. But this rule does not seem to apply to connectors.
- [ ] Move REF on F.Fab to the center of the connector so it does not overlap other Fab lines
image

cpresser:
@lorem-ipsum I am 'cleaning up' old PRs that I am involved. Should we try to get this one done?

evanshultz:
I'm confused why the mask layer is turned off on any pad.

The rotation angle isn't too clear here, and I'm not surprised this is inconsistent in the library, but Tom (on the IPC 7351 board) has a post at https://blogs.mentor.com/tom-hausherr/blog/tag/ipc-standards/ and it only uses Left. So may flip this thing around?

chschlue:
Each SMD pad 2 is defined twice per side, copper only and then again copper, paste and mask.
The upper and lower copper-only ones on top and bottom should be removed, I guess (is that sentence remotely intelligible?).

To offer a third opinion on orientation: The way I read F4.2, it leaves no room for interpretation but top (LCC-style), even if coaxial connectors weren't on anybody's mind when that rule was written.

lorem-ipsum:
I flipped it now with pin 1 to the left, but I can also rotate it upwards if desired.

I also moved the REF marker on the fab layer, added a PCB edge label and added a pin 1 designator on the fab layer (even though I find it a bit pointless on a symmetric coax connector).

The copper to the board edge is most likely an error with almost all PCB manufacturers, but AFAIK there is no clear standard on how close you can get to the PCB edge. I can retract an arbitrary amount like 0.25mm, but not sure if thats a good solution.

I am not sure about the comments regarding the layers: Only or the extension with the through hole the silk layer is turned off. This is how it is done in the existing similar footprints where it is based upon. Other than that I see every pad only defined once (for front and back).

Screenshot_20200514_113633

cpresser:
@chschlue TBH I don't understand your comment. Can you elaborate?

Here is my observation:
* Pad1 has copper, mask and paste
* Pad2 (each copy) consists of
   * The 'actual' Pad with with copper, mask and paste. This exists twice, one for top and bottom. 
   * The via
   * A small strip with connects the pad to the via. Just copper. Also on top and bottom.

Currently it is not possible to select 'only outer copper layers' for a pad. Thus, having a pad for each side of the board seems like a good choice.

Regarding the orientation, the library is already really inconsistent. We should do an overhaul anyway. In this case I personally do not have a strong opinion. IPC conformity seems like a good idea. That would be left.

**Off-Topic:** Does someone have real-world-experience how the 0-orientation affects the assembly process? My current assumption is that the operator loading the reels/tapes/tubes has to manually check the orientation anyway.
As a 'consumer' that orders assembled boards from a fab I never had any trouble with the pick'n'place position file.

chschlue:
> @chschlue TBH I don't understand your comment. Can you elaborate?

TBH I don't understand my comment either. Don't know what I was thinking. Looks fine to me.

> **Off-Topic:** Does someone have real-world-experience how the 0-orientation affects the assembly process? My current assumption is that the operator loading the reels/tapes/tubes has to manually check the orientation anyway.
> As a 'consumer' that orders assembled boards from a fab I never had any trouble with the pick'n'place position file.

I wondered about this, too.
In case of "normal" packages, the manufacturer specifies the orientation of pin 1 in the tape, so a specific zero orientation combined with a relative rotation in the pos file totally makes sense.
With connectors and such, manufacturers usually just graphically specifiy how they're lying in the tape and there often is no notion of "pin 1" at all. So, where that pin 1 we just made up is seems pretty much meaningless to me.
Note that I'm also just just a customer of assembly houses and don't run one either.

PR #1771, opened 2019-08-09
Add footprint Texas_VQFN-15_2.5x3_Pitch0.5mm

drid:
Added Texas_VQFN-15_2.5x3_Pitch0.5mm
This is not a standard footprint. As you can see Mask and Paste is smaller on some pins.
This is the reason for the warnings.

Land pattern and solder mask
image

Solder paste
image

Datasheet: http://www.ti.com/lit/ds/symlink/tps63070.pdf

---


poeschlr:
Could you add a bit more detail to the description field? The datasheet alone is not enough.
You seem to have set the pad clearance for some pads. I do not think this is necessary. (example pad 9)
Screenshot from 2019-09-05 21-29-32

I did not yet find time to double check the measurements of the part.

PR #1773, opened 2019-08-09
Add RF Testpoint MS-156C3

drid:
Add RF Testpoint MS-156C3

Datasheet https://www.hirose.com/product/document?clcode=CL0358-0340-0-00&productname=MS-156C3&series=MS-156C&documenttype=Catalog&lang=en&documentid=D49421_en

image

---


poeschlr:
This should go into `Connector_Coaxial`
Document where no copper is to be placed. See http://kicad-pcb.org/libraries/klc/F4.5/
Use (paste only) pads to define the paste layout.
Which symbol is intended for this part?
The outline on fab must agree with the nominal size of the component body.
Courtyard clearance must be 0.5mm for connectors.

The footprint name should follow other coaxial connectors.
The anchor pad of the complex pads could be increased in size (would make the net name and pad number easier to read)

Some pad dimensions are not quite right
Screenshot from 2019-10-06 16-24-15

drid:
I will add a switched tespoint symbol for this footprint

drid:
Symbol is https://github.com/KiCad/kicad-symbols/pull/2068

poeschlr:
Please do not force push after a review! it makes our process much harder than necessary. Also use one commit to move a file and a separate commit to make changes to it as it allows us to use the github difftool to check what you did. Updating the screenshot also helps a lot.

---

The clearance between silk and the pads must be at least the same as the silk line width.

I suggest marking keepout with simply having a text KEEPOUT inside the area (size 0.4 would fit) i also added a bit of additional text with size 0.2 mm ("No Traces" and "No Resist") We can go this small as this text is meant to be used during design time where there really is no minimum size (must only be readable at reasonable zoom levels)
Screenshot from 2019-12-27 10-37-23

Exchange the position of the silk reference and the value field (reference must be above outline while value must be below it.)

We might want to add a prefix to the footprint name to catch similar additions later on. Possibly TestSwitch, CoaxialSwitch or something along these lines.

Copper pad dimensions look good.
Screenshot from 2019-12-27 10-56-57

Please use normal rectangle pads for the paste only pads also these pads should have no other layer active. The 4 corner pads are at the wrong position in x direction (possibly also wrong size)
Screenshot from 2019-12-27 11-09-31

poeschlr:
The paste pads are still custom pads and they still include copper.

poeschlr:
The land pattern now looks ok but you deleted the keepout area marking on the drawings layer.

poeschlr:
Thanks

PR #1774, opened 2019-08-10
Inductor_SMD: Add Murata LQH2MCN_02 Chip coil (Chip Inductors) Series

gmsotavio:
## Murata LQH2MCN_02 Chip coil (Chip Inductors) Series

| | |
| :--- | :---: |
|Datasheet | https://search.murata.co.jp/Ceramy/image/img/P02/JELF243A-0053.pdf |
| Proposed footprint | L_Murata_LQH2MCN_1 |
| General settings | L_Murata_LQH2MCN_2 |
| 3D settings | L_Murata_LQH2MCN_3 |
| 3D view | L_Murata_LQH2MCN_7 |
Mechanical drawing | L_Murata_LQH2MCN_6 |
| Recommended land-pattern | L_Murata_LQH2MCN_4 |
| Dimensioned drawing | L_Murata_LQH2MCN_5 |

---


cpresser:
Thanks for this great PR. Nice screenshots and annotations.
So far I only found one issue:
- [ ] Silkscreen clearance to copper is ~0.15mm. It is okay according to KLC (its > silksceen width). But 0.2mm would be a nicer value here. It still looks fine after this change.
image

gmsotavio:
Hello cpresser. Thank you for the time spent to review my contribution. My considerations about your review.

> Silkscreen clearance to copper is ~0.15mm. It is okay according to KLC (its > silksceen width). But 0.2mm would be a nicer value here. It still looks fine after this change.

I have adopted the convention 0.20mm starting from Silk **node** to pad/Fab outline. I think all footprints have been following this convention. Confirm that you want to change this.

> The name should not include the "_02", the footprint will also match other inductance values.

The naming convention adopted by Murata for this inductor series is odd. Actually, the "_02" is not related to inductance value, but it is part of the series name. Take a look in the picture bellow.
image
I will check the Murata catalog in order to investigate if the "_02" can be removed from the footprint name.

cpresser:
I think the line-width has to be considered when talking about clearance.
However, changing the silk-screen clearance is not mandatory. I am fine merging it like it is since it conforms to KLC. This was just a suggestion.

Regarding the name, I suggest using xxxx for the vaule if you want to keep the "02" suffix.
``LQH2MCNxxxx02``

I checked a few part numbers they all have 4 characters for the value, they seem to conform to the datasheet:
* LQH2MCN 4R7M 02L
* LQH2MCN 100K 02L
* LQH2MCN 1R0M 02L

poeschlr:
please also include the body size in the name (similar to other inductors we merged recently)

gmsotavio:
@cpresser @poeschlr sorry by the long delay. I have accepted and provided the changes as suggested by you.

I ended up with the following name: **L_Murata_LQH2MCNxxxx02_2.0x1.6mm**

PR #1775, opened 2019-08-10
Button_Switch_SMD: Add 1188E SMD Tact

gmsotavio:
Datasheet: http://www.switronic.com/uploads/images/A01.gif
SW_SPST_1188E_1
SW_SPST_1188E_2
SW_SPST_1188E_3
SW_SPST_1188E_4
SW_SPST_1188E_5
SW_SPST_1188E_6
SW_SPST_1188E_0
SW_SPST_1188E_7

Notes:
- I used the center of the datasheet rather than suggested in KLC.
- There are several manufacturers that produces this Tact Switch: Diptronics (P/N: 1188E-1K2-V-TR), Switronic (P/N: IT-1188E-DD),  ChipDip (P/N: SWT-1188E-1K2), Homyet (P/N: HY-1188E-H38), then I have neglected the manufacturer in footprint name.
- The switch pads are the number 3 and number 4, but the current symbol in KiCad library uses the pins 1 and 2. I need some guidance to make this footprint useful. For my personal use, I created a new symbol
SW_SPST_1188E_9

SW_SPST_1188E_8
- The dimension 7 mm is missing in the aforementioned datasheet. This measure was taken from the following link: http://www.yizhijia.com/Upload/30/IT-1188E-DD.jpg
IT-1188E-DD

---


poeschlr:
To me it seems the pins 1 and 2 are really just mounting pins so how about a symbol similar to how we do it with the mounting pin connector symbol? (pin/pad number MP)

According to https://github.com/KiCad/kicad-symbols/issues/580 the names would be:
`SW_1P1T-MP_NO_Horizontal_Switchtronic_IT-1188(E)`

gmsotavio:
> To me it seems the pins 1 and 2 are really just mounting pins so how about a symbol similar to how we do it with the mounting pin connector symbol? (pin/pad number MP)
> 
> According to [KiCad/kicad-symbols#580](https://github.com/KiCad/kicad-symbols/issues/580) the names would be:
> `SW_1P1T-MP_NO_Horizontal_Switchtronic_IT-1188(E)`

You are right. Pins 1 and 2 are just mounting pins. What are you suggesting to me? Change pins 1 and 2 to MP and renumber pins 3 and 4 to 1 and 2 to fit with current generic switch symbol?

Regarding the footprint name. As I have mentioned in PR, there are dozens of manufacturers that produce this switch. The "IT" from  " SW_1P1T-MP_NO_Horizontal_Switchtronic_IT-1188(E)" is manufacturer specific. Can I drop the manufacturer and "IT" from the name?

poeschlr:
I would keep the full manufacturer and part number in there. My reasoning is that we otherwise need to encode a lot more generic information in the footprint name and everytime we try with such things we discover that it is basically impossible. A user can then still use the footprint for anything that is compatible but we make it much clearer that it is really only checked against this single component.

Also please include the actuator in the courtyard (is done that way with other switches as well).

---

And it seems at least one of your commits does not use the mail address that is known to github for the author indication. This means our licence agreement handling fails and needs to be fixed.

PR #1784, opened 2019-08-18
Buzzer AT-0927-TT-6-R added

Resaj:
Buzzer AT-0927-TT-6-R added.
Datasheet: http://www.puiaudio.com/pdf/AT-0927-TT-6-R.pdf

imagen


cpresser:
Thanks for your contribution.
I could only find one thing that is not according to KLC
- [ ] Description and Keywords need more details. For example "Buzzer Magnetic 9mm"

cpresser:
You only updated the Keywords, but not the description. In this case this is pretty redundant since they will contain basically the same content. Please update the Description as well.

Resaj:
Hi, I didn't find the description field, so reading the KLC rules (header F9.1) I suppose you refer to the "Document link" field in Kicad 5.

imagen

I'm going to change it.

poeschlr:
The graphical interface for the documentation field has changed with 5.1. In the past (including 5.0) it was simply known as documentation and the KLC is based on that. (It might even be considered a bug of 5.1 as the file format did not change but the text in the GUI did. Maybe for v6 we could get a separate field for documentation and datasheet link. Or even better multiple possible links as a footprint can be specified not just by a datasheet but also application notes.)

PR #1785, opened 2019-08-18
Add WQFN-24-1EP_4x4mm_P0.5mm_EP2.45x2.45mm

jonas-schievink:
This has a slightly smaller EP compared to the other footprint. It is used by TS3A27518E.

Generator PR: https://github.com/pointhi/kicad-footprint-generator/pull/412

Symbol PR: https://github.com/KiCad/kicad-symbols/pull/2092

screenshot-2019-08-18-21:01:26

---


jonas-schievink:
I don't know why that failure is happening, this is not a THT footprint. A bug in the generator?

chschlue:
CI script complains about thermal vias, which are of thru_hole type obviously.
FP should be fine.

poeschlr:
Yes sadly our scripts do not handle thermal vias correctly. The script would require quite a bit of additional features to be able to do that.

jonas-schievink:
It looks like `WQFN-24-1EP_4x4mm_P0.5mm_EP2.6x2.6mm_ThermalVias` has the same problem, so I couldn't use it as a reference. I pushed a commit that hopefully fixes the layers for the vias.

chschlue:
I think the problem is with the CI scripts, not with the footprints.

Thermal vias are in the middle of a patch of copper with a fitting soldermask. I don't see why the vias would need their own mask definitions.

jonas-schievink:
Yeah, I don't really understand this either, but it seems to match [KLC F7.4](http://kicad-pcb.org/libraries/klc/F7.4/).

Anyways, I can delete the commit anytime.

chschlue:
Yes, it fixes the Travis error but not the problem.
F7.4 is about actual THT pads, not thermal vias.

[F4.4](http://kicad-pcb.org/libraries/klc/F4.4/) says
> B.Mask is not to be enabled on the B.Cu pad so mask is present.

F.Mask isn't mentioned, but what good would that do?

jonas-schievink:
>  F7.4 is about actual THT pads, not thermal vias.

Hence my original confusion with the failure.

Deleted the bad commit.

poeschlr:
See script comment

poeschlr:
thanks

PR #1787, opened 2019-08-19
added Amphenol ICC SFW26S-2STE1LF

g-gg:
1.00mm Flex Connector, SFW Straight series, 26 Position, Top Entry Surface Mount, ZIF

Datasheet: https://cdn.amphenol-icc.com/media/wysiwyg/files/drawing/sfw_s-2st_e1lf.pdf

The connector is drawn with pin 1 on the bottom, which deviates from KLC F4.2. This was done intentionally to be compliant to the datasheet.

Screen Shot 2019-08-20 at 00 09 42

---


cpresser:
Thanks for you contribution.
This looks like a really good series for scripting (https://github.com/pointhi/kicad-footprint-generator) Do you think you can do it? 

I had a quick lock (no full review) and found a few things we can discuss
- [ ] What is the hatching for?
- [ ] I don't agree with the KLC violation. There is no reason not to conform to KLC.
- [ ] The marker on F.SilkS does not show Pin1 but the FPC orientation. Pin1 must be marked, FPC orientation is optional. I suggest do include both.

g-gg:
I guess this is the opportunity I've been waiting for to have a look into the footprint generator.
 
The hatching should indicate the very body of the connector, I thought KLC F4.5 is asking for this. Am I mistaken?

Ok with KLC compliance, I had a doubt on it, but instinctively went for the same orientation as in the datasheet and felt confirmed by the introduction in the KLC:
> Where the KLC deviates from a particular datasheet or manufacturer recommendation, the datasheet should take preference unless there is a good reason not to do so (which should be clarified by a librarian).
According your interpretation the orientation of the connector is not sufficiently strong argument for deviation?

Pin1 was marked on the F.Fab layer. Should I duplicate the mark to F.SilkS or move it?

poeschlr:
F4.5 only applies if the datasheet requests an area without copper. (this is a workaround for footprints missing support for defining copper keepouts)

The body outline should be on the fab layer. (no hatching, but with polarity mark = pin 1 indicator)

g-gg:
Thanks for the comments and the explanations.

I've scripted now the whole series, it was not too difficult. Before pushing all the footprints, I'd like to have your feedback on the version created.
Screen Shot 2019-08-26 at 22 12 42

The script should go with pointhi's repository I guess?

cpresser:
> I've scripted now the whole series, it was not too difficult. Before pushing all the footprints, I'd like to have your feedback on the version created.

A few issues I can see in the screenshot:
- [ ] There is silkscreen above the pads. Please remove the silkscreen between pads 1-25 and 2-26. Aim for the 0.2mm clearance of silksceen->copper.
- [ ] There is no Pin1-Mark on F.Fab layer
- [ ] If you want to indicate the insertion-direction of the FPC you can also put that as text on Cmts.User

A good check you can do prior to pushing the FPs is to run the check-scripts (the same as travis uses) locally. See https://github.com/KiCad/kicad-library-utils/blob/master/pcb/check_kicad_mod.py

If the check-scripts don't produce any errors (excluding false positives), push one or all the FPs so we can review them properly.

> The script should go with pointhi's repository I guess?

Yes. Please create a PR there and link it to this issue.

poeschlr:
You say it is scripted. Is the script somewhere to be found? If it is using the script repo https://github.com/pointhi/kicad-footprint-generator then please make a pull request there and link to it.

g-gg:
Just created the push request
https://github.com/pointhi/kicad-footprint-generator/pull/432

PR #1789, opened 2019-08-25
Added footprint for OLED 128x64

kajusK:
https://www.vishay.com/docs/37902/oled128o064dbpp3n00000.pdf

2019-08-25-175921_1920x1080_scrot

Added footprint for dirt cheap 128x64 oled displays. Only display "connector" (flex cable soldered directly to the pcb) and short part of the flex cable are drawn. The display itself can be positioned on other side of pcb, mounted on device chassis,...


kajusK:
Failed checks explanation:
- F5.1 F.SilkS have incorrect width: there are two holes in the display flex cable that are intended for the connector alignment before soldering. Therefore I placed two marks on the silkscreen just below these holes. To make these circles filled, I've just set the circle diameter to 0 and set the trace width to required value. Is there a better solution to this?

poeschlr:
Why fill the circle? Might it be better to have a normal circle with normal line width set such that its outer diameter has the required diameter? (Silk printing can behave strangely with large filled areas depending on manufacturing process)

--

Also 0 diameter circles can behave differently in different software products. Might be better to use diameter equal to half the final diameter and width equal to final diameter.

kajusK:
Good point, I just thought filled circle looks better... Anyway, updated the circle to use normal line width.

kajusK:
@poeschlr Any update on this?

poeschlr:
I just answered a question you had. You might notice that i did not assign myself to your contribution.

Generally: Any contribution or review is ready when it is ready. Everyone here is a volunteer and none of us have as much time for kicad as we would like. Want to speed up the process? Create a dimensioned drawing of your contribution and add a screenshot of it to the discussion. Tutorial https://forum.kicad.info/t/how-to-check-footprint-correctness/9279

chschlue:
@kajusK shouldn't the FP pads be slightly longer than the actual FPC pads for better solderability?

kajusK:
@chschlue I'm not sure. The flex cable of the display has small holes in each pin through which the solder can flow and when soldering by hand, I just apply a bit on solder on the pcb pads, place the display cable over these pads and reheat the solder through the display cable pads. This method works really well.

kajusK:
Any updates on this PR? Should I change anything?

chschlue:
I haven't gotten back to this yet, I was pretty busy during the last months.

PR #1792, opened 2019-08-27
Fix pad sizes of Taiyo Yuden NR series SMD inductors

### LINKED ISSUE #1756: Taiyo Yuden SMD inductor solder pads bigger than MF recommended
The L_Taiyo-Yuden_NR-xxxx footprints do not seem to follow the manufacturer recommended landing patterns. All pads seem to be 0.2mm longer (Y axis) than recommended.

E.g, the length of the pads of the L_Taiyo-Yuden_NR-40xx footprint is 3.9mm in [this libraries KiCad module](https://github.com/KiCad/kicad-footprints/blob/cf71819ab37168bfad298de859b1bbd00f5a93f6/Inductor_SMD.pretty/L_Taiyo-Yuden_NR-40xx.kicad_mod), but it is only [supposed to be 3.7m according to the datasheet](https://www.yuden.co.jp/productdata/catalog/wound04_e.pdf).

Is this a mistake, or was this done to improve solderability?

seppestas:
Fixes #1756 for Taiyo Yuden NR series SMD inductors by correcting the pad width so it matches the manufacturer recommended pad width.

Also simplifies the outline on the F.Fab layer. For some reason the "sides" where split in 2:
Screenshot 2019-08-27 at 17 10 46

Note that this does not change the "HandSoldering" footprints. I assume the pads for handsoldering are intentionally bigger and I suspect the wider pad size was accidentally copied to the non-handsoldering components.


seppestas:
@evanshultz 

The datasheet is linked in #1756. It applies to all versions of the Taiyo Yuden NR series, and thus to all footprints touched by this PR.

[KLC rule F9.1](http://kicad-pcb.org/libraries/klc/F9.1/) mentions "Where appropriate, a URL to the footprint datasheet should be included".

Are there any guidelines for when it is "appropriate" to add a link to the datasheet in the description? I'm afraid that by adding a link to the description I will create some inconsistency.

evanshultz:
It's still a fix even if there are no datasheets.

PR #1462, opened 2019-03-04
added xt60 connector

rs-development:
I could not find the so famous XT60 connector, so here is one:

Measurements taken from a real piece since i could not find a datasheet online. I used the footprint before and it fits fine, but maybe someone can validate.

PR #1794, opened 2019-08-31
Add footprints for XT60 male and female connectors.

reportingsjr:
Add footprints for male and female XT60 connectors. Yes, these are technically wire to wire connectors, but lots of people have been using them as PCB mount connectors.

I created the footprints by measuring a handful of physical connectors (purchased through hobbyking so they are legitimate parts) and finding the average pin diameter (~4.3mm) and adding the IPC standard 0.2mm clearance.

Here is the only datasheet really available for it: https://www.tme.eu/en/Document/2d152ced3b7a446066e6c419d84bb460/XT60%20SPEC.pdf

This PR should also close #1462 if it is accepted.

3D model PR: https://github.com/KiCad/kicad-packages3D/pull/603

Male connector footprint:
XT60 male

Female connector footprint:
XT60 female

---


reportingsjr:
Hmmm, it is failing and saying some silkscreen in intersecting with pad 1. By chance is the script checking the pad as if it is a square pad rather than a rounded rectangle pad?

poeschlr:
The fab layer is correctly representing the nominal body size (measured using the kicad internal measure tool -> not persistent so no nice screenshot.) But the silk screen should be fully outside the nominal body (as it will otherwise be covered by the body)

Move the fab reference inside the body outline (center).

reportingsjr:
Ok, I went ahead and bumped the silkscreen out a bit so it won't be under the body of the connector. I didn't go too far as I'm not a huge fan of silkscreen that is way bigger than connectors, but I can go further if that is your preference.

I shrunk down the fab layer reference to the minimum mentioned in the KLC (0.5mm in KLC 5.2) and centered it on the connector body.

poeschlr:
There is no need to shrink the reference for this huge component. Please keep it at 1mm size.

reportingsjr:
Ohhh, I misread KLC F5.2 4C. I didn't see the part where it says to scale the reference only if it is overlapping parts of the same layer. Fixed!

poeschlr:
Thanks

reportingsjr:
@poeschlr would you mind looking at  KiCad/kicad-packages3D#603 so this can have the 3d model as well?

PR #1797, opened 2019-09-01
Added AST1109MLTRQ buzzer

Duckle29:
I made a footprint for the AST1109MLTRQ piezo buzzer

[Datasheet](https://www.mspindy.com/specifications/AST1109MLTRQ.pdf)
3D PR: https://github.com/KiCad/kicad-packages3D/pull/605

Screenshots:


gmsotavio:
Hi @Duckle29. Thanks for your submission.

Some dimensions don't match those shown in datasheets.
image

image

gmsotavio:
Hi @Duckle29. Thanks for your submission

- [ ] The dimensions looks good for me.
image
image

- [ ] Please, provide a better description. I think only the datasheet link isn't enough.

- [ ] The pad rotation must be 0 degrees. The same thought is valid for RefDes and footprint name in Silk and Fab layers.

Duckle29:
I agree that just the datasheet isn't enough. I used the Buzzer_CUI_CPT-9019S-SMT for reference. I'll add a bit more and fix those rotations :)

gmsotavio:
Thanks for the changes!

> I agree that just the datasheet isn't enough. I used the Buzzer_CUI_CPT-9019S-SMT for reference

Unfortunately some old footprints have no description.

Sorry. I forgot two things. The Silk mark and the Fab bevel on pin-1. Remember that bevel should be 1mm or 25% of package size (whichever is smaller) as described at http://kicad-pcb.org/libraries/klc/F5.2/. See the example shown bellow.
image

After you provide these changes, I believe that your footprint will be OK to be merged.

Duckle29:
I was a bit unsure about that, as it's not a polarized design. I guess it does allow one to align the laser etched text on the part. Everything else is rotationally symmetric

gmsotavio:
> I was a bit unsure about that, as it's not a polarized part. I guess it does allow one to align the laser etched text on the part. Everything else is rotationally symmetric. Would you want a pin 1 indicator for the lasered text?

I made a mistake. You are right! Forget all.

- A last minor thing. Can you replace the colon by a comma in the description?

- Just for curiosity. How do you came with this footprint whether datasheet don't provide one? I'm assuming you have this component and have used it in some project.

Duckle29:
I'll get that done. I don't have the component yet but was asked to add an unpopulated footprint for it on some boards so that users could use it if they wanted. 

I just took the pad dimensions and added 0.1mm all around it to allow a bit of a solder fillet around the edge. I didn't add more as the pads have a hole punched in them, so I rely on the majority of the solder fillet to form in that hole. Regardless it's not a heavy component, and it has relatively large pads for its size :)

Duckle29:
I also now realize why I was confused. KiCAD 5.x.x versions has that field called "Document link" where it's now called "Description"

gmsotavio:
> I just took the pad dimensions and added 0.1mm all around it to allow a bit of a solder fillet around the edge. I didn't add more as the pads have a hole punched in them, so I rely on the majority of the solder fillet to form in that hole. Regardless it's not a heavy component, and it has relatively large pads for its size :)

It's hard to review when manufacturer do not provide a footprint recommendation. But I used a component with a similar pad shape, and usually 0.1mm is added.

> I also now realize why I was confused. KiCAD 5.x.x versions has that field called "Document link" where it's now called "Description"

I got confused when this was changed.

PR #1801, opened 2019-09-04
Added battery holder

kajusK:
Created footprint for keystone battery holder

Datasheet:https://ken.keyeuro.eu/cat/590.pdf

2019-09-04-194721_3840x1080_scrot

I'm not really sure about correct way to draw silkscreen and courtyard in this case...


kajusK:
Any updates, notes,...?

cpresser:
Hi and thanks for the contribution.
Since there are >200 open PRs right now it takes some time for each one to get reviewed. There is a limited number of reviewer with limited amounts of free time they wish to spend here :)

I took a look and have a few suggestions/questions:
1. The spring-steel is 0.3mm thick, so I suggest to actually 'model' that with lines 0.3mm apart. 
2. Then have the silkscreen outside of that F.Fab polygon. We want silkscreen visible after board assembly. If its hidden below the part (even if its thin and somewhat visible) is bad for AOI systems.
3. The courtyard is not really consistent and should be smaller. According to KLC F5.3 0.5mm seems a good choice (this is basically a connector).
4. Some kind of indicator/text on where the battery goes would be nice. Just by looking at the screenshot my first impression was that the battery goes into the top pocket. I was not aware that the bottom part is a spring contact. Perhaps you have a good idea for that. It is not mandatory, but would be nice.
5. It looks like you used the metric dimensions. That is good.
6. According to KLC F7.7 the oval-pad-inner-size should be `1.3x0.5mm`

cpresser:
@kajusK Do you want to come back to this PR?

kajusK:
@cpresser 

Ah, sorry for a delay, I completely forgot this PR, will take a look into it next week.

PR #1802, opened 2019-09-05
Add Lutronic 0821 .. T10CP

waschhauser:
Datasheet:
http://downloads.lutronic.biz/datenblaetter/en/0821_t10cp.pdf

FPG PR:
https://github.com/pointhi/kicad-footprint-generator/pull/419

SYM PR:
https://github.com/KiCad/kicad-symbols/pull/2130

image

image

Errors of Travis check
    Courtyard lines are not on 0.01mm grid
     - Arc (1.75,0) -> (9.11,2.76) on layer 'F.CrtYd'
     - Arc (1.75,0) -> (3.04,7.75) on layer 'F.CrtYd'
     - Arc (1.75,0) -> (-4.32,4.99) on layer 'F.CrtYd'
     - Arc (1.75,0) -> (-5.61,-2.76) on layer 'F.CrtYd'
     - Arc (1.75,0) -> (0.46,-7.75) on layer 'F.CrtYd'
     - Arc (1.75,0) -> (7.82,-4.99) on layer 'F.CrtYd'

But they are on 0.01mm grid?!

---


chschlue:
I don't have in-depth knowledge of the travis scripts, but the way it looks to me the violations have to do with rounding errors introduced by https://github.com/KiCad/kicad-library-utils/blob/0c0b0351ac7e137a1437781f1530a563d222d5f1/pcb/rules/rule.py#L30.

poeschlr:
Not a script restriction but a kicad file format restriction. Kicads arcs limit which angles you can use. Because of this there is next to no chance to have both the start and endpoint of an arc on the same grid if the arc does not describe a multiple of 90 degree.

waschhauser:
Do I have to fix the Travis Check Errors?
And if yes, how can I solve the erros?

chschlue:
@poeschlr I get that it's actually a mathematical restriction.
Still, as KLC doesn't seem to forbid courtyard polygons with arbitrary angles (don't know what IPC-7351 says) it doesn't seem useful that getStartPoint() just returns unrounded double precision points.

waschhauser:
Now with circular courtyard

image

image

poeschlr:
The function you quoted is only used for arcs. For any other element there is no need to calculate the endpoints as they are contained in the file format. In other words: There is no problems with polygons having outlines of any angle.

waschhauser:
@penoud in this case which which kind of courtyard do you prefer? The first or the second version?

poeschlr:
I assume you wanted to tag me ;) My previous comment was not really meant as an answer to you but to @chschlue.

I think the circle is a good option.

PR #1805, opened 2019-09-08
Add DirectFET L6

chschlue:
Add DirectFET L6

https://www.infineon.com/dgdl/Infineon-AN-1035-ApplicationNotes-v29_01-EN.pdf?fileId=5546d462533600a40153559159020f76#page=42

DirectFET_L6

---


evanshultz:
1. Courtyard width is too big.

image

evanshultz:
2. Didn't notice the excessive courtyard precision before.

Note that this type of thing is why scripting is nice and was my intent for the DirectFET footprints. One simple tweak to a script and a whole slew of new footprints can be made that don't need individual verification.

chschlue:
It's not that I hate scripting, but:
a) There are 29 DirectFETs now and maybe 35 or so in the not-so-foreseeable future
b) You said yourself that full dimensions are next to impossible to come by (Infineon can't even seem to agree on dimensions for each series of package sizes)
c) Source and gate pads are irregular and would need to be hardcoded
d) I need some practice drawing footprints anyway (although these _do_ become repetitive and boring quickly)

evanshultz:
Just one comment on your first item, and it's a tangent: Where are the 35 new DirectFET footprints coming from? I understood the DirectFET/CanPak was not going to be used in future products after the Infineon acquisition. AN-1035 only defines 26 so I thought that was all there would ever be. Are there ~35 more coming from another vendor? Or something else?

chschlue:
I mostly made that one up. I didn't comb through every press release since the acquisition or anything. My point was just that the number of footprints probably won't be growing exponentially.

AN-1035 defines 29.
I had trouble counting, too. The TOC lists appendices 26, 27, 28 and then 26 again...

chschlue:
@evanshultz ping

PR #1808, opened 2019-09-08
Add DirectFET SB

chschlue:
Add DirectFET SB

https://www.infineon.com/dgdl/Infineon-AN-1035-ApplicationNotes-v29_01-EN.pdf?fileId=5546d462533600a40153559159020f76#page=23

DirectFET_SB

---


evanshultz:
Just the issue already mentioned about body and courtyard height.
image

chschlue:
@evanshultz ping

BTW, I just incremented the timestamps during these fixes, assuming they're really just fancy monotonic revision counters. Anything wrong with that?

evanshultz:
No problem whatsoever with me. I'm not aware that we care when the latest change was done, or if the change was done in a text editor and the timestamp doesn't move.

Thanks again!

poeschlr:
If the timestamp is not updated then the update footprint feature within pcb_new might not work

chschlue:
I thought it would be something like that. Still, incrementing by one should be good enough then?

poeschlr:
Didn't you write you made them by hand with kicad? Because kicad already takes care of handling the timestamp

chschlue:
True, but vi doesn't take care of handling the timestamp while fixing the outline dimensions :-)

Issue #1798, opened 2019-09-02
Rounded rectangles for exposed pads might have been a bad idea


poeschlr:
The more i think about it the more i feel that the soldermask and copper layer for non soldermask defined exposed pad pads should both be normal rectangles. (Similarly the soldermask layer for soldermask defined pads)

What is the opinion of the rest of you? If i do not get any objections then i will update the generator script within a week or so and push the updated footprints to this repo.

poeschlr:
I could also add a parameter to allow rounding for particular parts. (At least one manufacturer suggests it. See https://github.com/KiCad/kicad-footprints/pull/1691)

evanshultz:
I am sure this should be an option, as you mentioned above. All TI datasheets with thermal pads I can recall show rounded corners. As just a couple examples I could find quickly, see http://www.ti.com/lit/ds/symlink/drv8353.pdf and http://www.ti.com.cn/cn/lit/ds/symlink/drv8801.pdf.

poeschlr:
They show rounded corners but do not seem to dimension them. It kind of looks like they reuse the radius of the normal pads. So one possibility would be to offer this as a feature.

Issue #1655, opened 2019-06-20
LFCSP scripted footprints need to be re-generated


evanshultz:
https://github.com/pointhi/kicad-footprint-generator/pull/385 fixed an accidental "double-minimum with no maximum" dimensions of the script. The affected footprints needs to be re-generated since I could see the pad size and/or position changed a bit for the one footprint in the PR.

PR #1821, opened 2019-09-08
Fix EP round-radius for scripted footprints (#1798)

poeschlr:
script changes: https://github.com/pointhi/kicad-footprint-generator/pull/422
Discussion: #1798

This changes all exposed pads from 25% (maximum 0.25mm) round radius to be the same as for the normal pads of the same footprint. (Most datasheets show rounded corners without dimensioning them, they look similar in radius to the normal pads of the same footprint.) 

Tip for review: look at the script ;) Checking the footprints would be infeasible.


poeschlr:
Guys this should have quite a high priority as it has the potential of generating massive merge issues if left for too long. Also every generator footprint that is created between this change and the time this (and the script side) is merged will need to be recreated to fit this change (again something that will get worse with every day)

evanshultz:
It appears this will address https://github.com/KiCad/kicad-footprints/issues/1655, right?

poeschlr:
Yes

poeschlr:
@evanshultz did you forget about this or just no time? (Your last comment indicated that you already looked at it and as there is not "this needs to be fixed" part i also kind of assumed it should be ready after answering your question.)

evanshultz:
In the script PR I have three outstanding comments which were minor and don't affect the footprints, but would you mind addressing them first? I wanted to wait to get more input on the script but I think enough time has passed.

evanshultz:
I spot-checked a half-dozen or so of the footprints and the changes to the EP seem correct (the radius matches the outer pads).

Just by randomly clicking, though, I happened to pick `SOIC-4_4.55x3.7mm_P2.54mm` and noticed a pad number change. Is that intentional? Did you check for noteworthy/breaking changes outside of the EP radius or other benign changes that are incremental script improvements?

poeschlr:
I think the SOIC-4 was one that got modified in the footprint editor when it was contributed originally. We should have never allowed it to be placed in the same size definition documents to be honest.

evanshultz:
So does this pin number change need to be reverted? Do these footprints, to your knowledge, then need to be reviewed for other unintended issues?

poeschlr:
The SOIC-4_4.55x3.7mm_P2.54mm has the exclude pin list set wrong (it excludes 2, 4 but should exclude 2, 5) The exclude pin stuff has been added to the script since it was originally generated (but then again it should not have been placed into the same size definition file as the non hand manipulated definitions)
I will fix that as soon as i find time.

poeschlr:
Fixed the soic-4.

I think this one is named wrong to be honest. it is a soic-6 with missing pins not a pure soic-4. but that would be something for a later date. Issue created https://github.com/KiCad/kicad-footprints/issues/1904

PR #1835, opened 2019-09-09
Fix MicroMELF footprints

evanshultz:
http://www.vishay.com/docs/85597/bzm55.pdf (same as all datasheets at https://github.com/KiCad/kicad-symbols/pull/2137).

I don't know what happened exactly, but I suspect the datasheet changed. This footprint was submitted at https://github.com/KiCad/Diodes_SMD.pretty/pull/39 and the datasheet linked there doesn't exist at that URL. Above is the current datasheet for, ostensibly, the same part, and it has smaller pads than our footprint. So I updated the footprint to match.

image
image

image

Note that there is a typo of `Handsoldering` (missing the `n`) but I haven't changed that here. @poeschlr Would this fall into the realm of a fix that can be done now? I believe so since footprints are merged into the board file, but I still would like to be cautious about this.

---


cpresser:
Is there any other device in that package?
I would like to verify this against another manufacturer since its a 'generic' fp.
 I just found "BAS385-TR" so far which is also from vishay and has the same specs.

chschlue:
I think this is Vishay-specific.

evanshultz:
Good find! I looked at component distributors and Googled but I didn't find any other vendors for MicroMELF except for Vishay.

In this case, Littelfuse doesn't have a recommended footprint. And we need some documentation in order to review a footprint. So if we can find some other footprint we could consider it. On the other hand, each footprint represents a single drawing and this footprint went from an inaccessible drawing to the current drawing by the same vendor for the same part (Vishay BZM55). It is certainly not the only unexplainable footprint that needs correction in the library.

chschlue:
Farnell sells them.
I couldn't find a footprint recommendation either, I presume Littelfuse simply borrows from Vishay there.

cpresser:
I did some more digging and found some more MicroMELF products:
https://firstohm.com.tw/phocadownloadpap/01_RESISTOR_2018/MM102.pdf
image

http://www.koaspeer.com/pdfs/SS-197R6_RN41RM41.pdf
image

The KOA part does not have a recommended fp. 
Firstohm has one which is slightly different. But the part dimension is also slightly different.

Since this FP says "D_MicroMelf" I am fine with going with the vishay specs. None of the other MicroMELF parts are diodes.

evanshultz:
@cpresser @chschlue 
Thanks for all the research! Good info.

Please let me know if there's anything you're looking for to get this merged. I read it to say that there are no outstanding issues, but I could be wrong.

chschlue:
As I understand it, fixing the filename right now is allowed, right?

Other than that, no objections

evanshultz:
There's your `n`.

PR #1804, opened 2019-09-08
Fix DirectFET ME

chschlue:
Fix DirectFET ME

* Wrong gate pad
* Typo
* DS link

https://www.infineon.com/dgdl/Infineon-AN-1035-ApplicationNotes-v29_01-EN.pdf?fileId=5546d462533600a40153559159020f76#page=39

DirectFET_ME_old
DirectFET_ME_new

---

All contributions to the kicad library must follow the [KiCad library convention](http://kicad-pcb.org/libraries/klc/)

Thanks for creating a pull request to contribute to the KiCad libraries! To speed up integration of your PR, please check the following items:

- [x] Provide a URL to a datasheet for the footprint(s) you are contributing
- [x] An example screenshot image is very helpful 
- [x] If there are matching symbol or 3D model pull requests, provide link(s) as appropriate
- [x] Check the output of the Travis automated check scripts - fix any errors as required
- [x] Give a reason behind any intentional library convention rule violation.

PR #1846, opened 2019-09-11
Fix DirectFET ME fab, silk, pad height

chschlue:
https://github.com/KiCad/kicad-footprints/pull/1804

---


evanshultz:
Pad size fix looks good.

chschlue:
@poeschlr 
This is the way @evanschultz and me have been doing the 40 or so DirectFET PRs all week long.
What exactly is suddenly wrong with it?

poeschlr:
The same as at the beginning of the week. I can not really say if evan is as effected by something like this as i have been when somebody did this to me. But i can for sure tell you that it is annoying that the full dashboard is full of repeated "ping" comments (which means it is much more likely that i miss stuff that would be important to see)

evanshultz:
@chschlue 
I wanted to get these footprints done and I had them all opened in tabs so I kept charging ahead. But I agree with Rene, that no ping is necessary because I will see all PRs. I just wanted to get these done first.

Thanks for updating all of them!

chschlue:
@evanshultz 
Ok, so the pings didn't hurt much either? If they did, I apologize.

evanshultz:
No worries. They're just unnecessary as Rene said above.

PR #1853, opened 2019-09-13
[Package_DFN_QFN] Add Microchip_DFN-4-1EP_1x1mm_P0.65mm_EP0.48x0.48mm

maximeborges:
Recommanded footprint layout from Microchip from http://ww1.microchip.com/downloads/en/DeviceDoc/MIC550x-300mA-Single-Output-LDO-in-Small-Packages-DS20006006A.pdf

`Microchip_DFN-4-1EP_1x1mm_P0.65mm_EP0.5x0.5mm`:
image
image
image

vs the existing `DFN-4-1EP_1x1mm_P0.65mm_EP0.5x0.5mm`/`UDFN-4_1x1mm_P0.65mm_EP0.5x0.5mm`:
image

- [x] Datasheet: http://ww1.microchip.com/downloads/en/DeviceDoc/MIC550x-300mA-Single-Output-LDO-in-Small-Packages-DS20006006A.pdf
- [X] Matching 3D package: https://github.com/KiCad/kicad-packages3D/pull/606


poeschlr:
Out of curiosity: what workflow do you use to make such nice drawings of the footprint? (I am currently searching for a replacement for our current workflow that no longer seems to work that nicely)

maximeborges:
> Out of curiosity: what workflow do you use to make such nice drawings of the footprint? (I am currently searching for a replacement for our current workflow that no longer seems to work that nicely)

At first I used SolveSpace to redraw the footprint and then wrote the coordinates of every points directly in the .kicad_mod file (Solvespace give you the coordinates when selecting a point).
After that I thought that it would be cleaner to move the pads at their own centers and just edit the pad's coordinates around their center, so I did that on this footprint, but I think the first way is way faster and simpler.
It's far for being the most effective way to do it (I hope), but it's actually not that bad.

Also I just thought about drawing a graphic polygon in the KiCAD footprint editors and then edit the .kicad_mod file to get the points of the polygon to convert it to a pad. This might be way simpler that what I did since you can easily use the differential coordinate system of the editor to put your points without having to calculate everything from the datasheet.

poeschlr:
Oh so it is not a dimensioned drawing of the final footprint. Good to know.

By the way you can use the freecad extension stepup to get directly from a parametric drawing to a footprint: https://forum.kicad.info/t/kicad-stepup-the-sketcher-for-footprint-generation/9582

maximeborges:
I have had really bad experiences with FreeCAD? but I'm willing to try it again.
I've used it to export some 3D models with StepUP and the https://github.com/maximeborges/kicad-3d-models-in-freecad project already, but I'm still not really satisfied by the workflow.

poeschlr:
Now for the review:
The center pad is missing paste
The name should use the correct center pad size (EP0.48x0.48mm instead of EP0.5x0.5mm)

Clearance between exposed copper and silk must be at least the same as the silk line width.
My suggestion for silk would be something like this:
Screenshot from 2019-09-20 14-48-07

The courtyard clearance must be 0.25mm all around.

Screenshot from 2019-09-20 14-52-36

evanshultz:
Note that the existing footprint mentioned above has an outstanding PR at https://github.com/KiCad/kicad-footprints/pull/1708.

poeschlr:
And that footprint is equal to the one we see here. Meaning is this contribution really needed?

evanshultz:
I was leaning towards no, and that's why I pointed it out.

maximeborges:
I didn't saw this PR. As mentioned, it's missing the EP spec in the name and for compatibility reasons I thought making another one with the drawings from microchip was appropriate.

maximeborges:
Since the 3D package actually as an EP of 0.5*0.5mm, I'm now getting the F9.3 warning ("3D model name is different from footprint name (found 'DFN-4-1EP_1x1mm_P0.65mm_EP0.5x0.5mm', expected 'Microchip_DFN-4-1EP_1x1mm_P0.65mm_EP0.5x0.5mm'), but this might be intentional!")
I guess it's OK since it's the recommanded layout for this package ?

diegoherranz:
Given that https://github.com/KiCad/kicad-footprints/pull/1708 was merged, it is for exactly the same footprint and the exposed pad dimensions were added to the name there, I don't think this is needed anymore so I'll close it. Please re-open or comment if you thought otherwise.

Thanks!

PR #1854, opened 2019-09-13
Fix 7SegmentLED_LTS6760_LTS6780

cpresser:
Body size was 12.5, but it is 12.6mm
Also fixed silk screen width and position to be completely outside if F.Fab

Related: https://github.com/KiCad/kicad-symbols/pull/2051

---


poeschlr:
The courtyard outline would also need fixing.

cpresser:
Thanks for the check. Turns out poking around with the values and moving single lines is not the best approach.
I now redrew the body from scratch.

poeschlr:
courtyard clearance in x direction is 0.26 on both sides

poeschlr:
Ah damn forgot that the kicad measure tool rounds to two digits. https://bugs.launchpad.net/kicad/+bug/1829317
The clearance is 0.255 meaning it is ok.

PR #1856, opened 2019-09-15
Close courtyard of Omron B3F-10xx tactile switch.

jmalloc:
This PR properly closes the courtyard of `SW_TH_Tactile_Omron_B3F-10xx` ([datasheet](https://omronfs.omron.com/en_US/ecb/products/pdf/en-b3f.pdf)), which prior to this fix results in the following DRC error:

drc

**This is the gap before:**

before

**And the completed shape as of this PR:**

after

---

- [ ] ~If there are matching symbol or 3D model pull requests, provide link(s) as appropriate~
- [x] Give a reason behind any intentional library convention rule violation. **Edit: as per the Travis build, square pad for pin 1 is not required for passive/bipolar devices. Either way, I have not modified this aspect of the footprint.**


chschlue:
Thanks for the fix.
Can you also change top and bottom courtyard to ±1.1mm to comply with [library requirements](http://kicad-pcb.org/libraries/klc/F5.3/) while you're at it?

chschlue:
Oh, and perhaps change top row pin numbers to 1, bottom to 2?
1-4 is not very useful for an SPST switch.

jmalloc:
> Can you also change top and bottom courtyard to ±1.1mm to comply with [library requirements](http://kicad-pcb.org/libraries/klc/F5.3/) while you're at it?

I'm happy to do this, but I can't see anywhere in those requirements that demands 1.1mm. I think the "default" clearance of 0.25mm applies here? (which is what the footprint currently has, as best I can tell).

> Oh, and perhaps change top row pin numbers to 1, bottom to 2?
> 1-4 is not very useful for an SPST switch.

Will do.

chschlue:
I also edited your comment instead of writing a new one... sorry.

But I hope the request is clear now.

jmalloc:
Oh, I see - 1.1mm from the grid origin... Got it in this last commit, I think :)

chschlue:
That's it.

Thanks for your contribution and sorry for my idiocy.

evanshultz:
@chschlue 
The pin number change was not correct and should be reverted. Datasheet clip:
image

Because either pin 1 and/or 2 and either 3 and/or 4 can be used, having discrete pin numbers gives the user the flexibility to route the board as they desire. Or both internally-connected pins can be connected together externally to give redundancy.

Note that I don't see a schematic symbol that suits this footprint if pins 3 and/or 4 should be connected, but that doesn't mean the footprint has an issue. It means we are missing a valuable symbol.

chschlue:
@evanshultz 
Fair point. Guess I should have had coffee before touching anything yesterday.

Note that this calls for at least two valuable symbols (e.g. [Panasonic](http://industrial.panasonic.com/cdbs/www-data/pdf/ATV0000/ATV0000CE5.pdf) has odd/odd and even/even connected, others use A/A', B/B'...) and many footprint fixes in Button_Switch_SMD and Button_Switch_THT.

poeschlr:
@evanshultz nearly all buttons in our lib have the same layout and yet still use the pin numbering as the change introduced. We also do not have a symbol that would represent the original numbering.

And yes i am aware that i just merged the revert commit.

PR #1863, opened 2019-09-18
Added Bourns 3362P potentiometer

gggfred:
Added Bourns 3362P vertical potentiometer
[Datasheet](https://www.bourns.com/docs/Product-Datasheets/3362.pdf)
bourns3362p
---


gmsotavio:
It's my first thoughts about your contribution.

Unless it's a bug in my FreeCAD, most of dimensions don't conform with KLC or event match those found in datasheet.

- [ ] Courtyard clearance must be 0.25mm, as per KLC F5.3 (http://kicad-pcb.org/libraries/klc/F5.3/);
image
image


- [ ] For rounded rectangle pads the corner size should be 25% with a maximum radius of 0.25mm, as per KLC F7.3 (http://kicad-pcb.org/libraries/klc/F7.3/);
image

- [ ] It would be better if the Fab elements were closed.
image

- [ ] Just my personal preference, but not required. I prefer the following orientation.
image

- [ ] I'm not sure, but I think that Fab elements shouldn't overlapping other features of the same layer. @poeschlr, can you give me a hint about this?
image

gggfred:
I have already changed the footprint by taking your suggestions, thanks for your comments, please check again and say if there are problems

myfreescalewebpage:
@gggfred I try to continue the review from this point:

- [ ] Silkscreen distance with pad should be at least 0.2mm. You should interrupt the lines when the silkscreen is too much closer to the pad. See for example "Potentiometer_Bourns_3266Z_Horizontal" in the lib
- [ ] Distance from F.Fab to F.CrtYd to be set to 0.25mm

Joel

PR #1879, opened 2019-09-27
Add Molex 47590-0001

jneiva08:
Datasheet: https://www.molex.com/pdm_docs/sd/475900001_sd.pdf

Screenshot:

micro-usb-ab

3d_1

3d_2

---


jneiva08:
F6.2: The footprint is a little forward because of the edge.
F7.5: Need advise to fix this, the pads is composed various shapes, I don't know if it was the best way to do.

poeschlr:
The side pads are not quite right. Might is suggested using a normal oblong (oval) pad plus a rectangle smd pad? (easier to get right and easier to check as stepup can import such a padshape) Also i think the square extension is not necessary on any other layer than the top side.

Screenshot from 2019-10-13 18-58-47

Move the pcb edge stuff to the edge cuts layer. (And close the fab outline properly to represent the part body.)
Screenshot from 2019-10-13 19-17-14

jneiva08:
@poeschlr , did you mean to change like this:
molex usb

poeschlr:
Yes

jneiva08:
@poeschlr , hope didn't wrongly interpreted the change have requestes.

micro_ab

poeschlr:
The two small smd pads with number 6 should not have paste as that would result in paste covering the large slot (could lead to production problems) At most a even smaller paste pad could be added that avoids the slot. (with a bit of clearance)

---

Footprint dimensions look good

Screenshot from 2019-10-21 19-58-56

jneiva08:
Have removed F.Paste from side pad's, have added small on with only F.Paste

Screenshot:
pad

poeschlr:
The paste pads should not have copper.

poeschlr:
thanks

jneiva08:
thanks

PR #1880, opened 2019-09-28
Add D_SOD-123W

chschlue:
https://assets.nexperia.com/documents/outline-drawing/SOD123W.pdf
https://assets.nexperia.com/documents/data-sheet/PMEG40T10ER.pdf

D_SOD-123W
sod-123w

https://github.com/KiCad/kicad-symbols/pull/2213

---


evanshultz:
1. See my comments about fab lines for the SOT1289 footprint.
1. Is this manufacturer-specific even if there are similar packages elsewhere? http://rohmfs.rohm.com/en/products/databook/datasheet-nrnd/discrete/diode/schottky_barrier/rb160m-40.pdf and https://www.taiwansemi.com/products/datasheet/SS14LW%20SERIES_B1906.pdf are both 'SOD123W' at DigiKey, and Taiwan Semi even uses the same package name, but they're subtly different. And the recommended footprints don't match.

Dimensions match the datasheet.

chschlue:
The Rohm SOD-123 appears to be a SOD-123F (thin and flat leads), SOD-123 is thick with gullwings, this SOD-123W is thick with flat leads and TaiwanSemi's SOD-123W is larger than this one.

Not sure what to make of that.

chschlue:
The Taiwan Semi package seems to be the freak among the SOD-123s, maybe that one should be manufacturer specific?

poeschlr:
The rohm one claims to follow JEDEC. maybe a good idea would be to check the standard document, it could be that it gives guidance for how to classify these.

chschlue:
I think JEDEC only specifies gullwing SOD (DO-215). Not sure what Rohm was up to.

The flat lead variants seem to be either JEITA or industry standards.

chschlue:
Renamed to `Nexperia_CFP3_SOD-123W`. Also fixed silk clearance.

myfreescalewebpage:
@chschlue I had a look to this footprint:
- F.Fab ok, pads ok, F.CrtYd ok
- F.SilkS: on the left the silkscreen is a little too much on the left, particularly when I compare to the nomal D_SOD 123 footprint
Joel

chschlue:
Not sure what you mean.
This one has 0.2mm clearance, as recommended (F5.1)
D_SOD-123 only has 0.09mm, which is less than required.

myfreescalewebpage:
See above screenshot, I just measure 0.25mm from the left pad, instead of 0.2mm:

Capture

chschlue:
Now I get it. It's fine nonetheless.

Silk clearance is to be measured from the edge of the line, not from the center (because silk is physical paint on the board, as opposed to fab and crtyd lines, which are virtual and have zero physical width.)

myfreescalewebpage:
ooooh this trouble me and I was wrong since a long time in this case ?
When using generators I don't think this is from the border, it's to the center of the silkscreen line in my memory.
I have just check with some phoenix connectors.

myfreescalewebpage:
New look at the footprint, ok to merge

PR #1881, opened 2019-09-29
Add D_SOT-1289

chschlue:
https://assets.nexperia.com/documents/outline-drawing/SOT1289.pdf
https://assets.nexperia.com/documents/data-sheet/PMEG045T030EPD.pdf

Nexperia_CFP15_SOT-1289
sot-1289

https://github.com/KiCad/kicad-symbols/pull/2215

---


evanshultz:
1. Can this package ever be used for something that isn't a diode with two pins tied together? Page 2 of the package drawing shows that it's not a standardize package. If so, all pins need a unique pin number.
1. Silk should wrap only around the body, and not around the pads, so push it right and clip around the big pad.
1. It appears this is a manufacturer-specific footprint so it can be named as such.
1. Only 10 fab lines are needed, not 11. You can join a pair of them forming the cathode mark.

Verified that dimensions are good.

chschlue:
1. It's called D_SOT-1289 at the moment (like D_SOT-23 which also has "broken" pins). Both share the advantage of being usable with generic diode symbols. It's about **di**odes, after all.
2. Will fix.
3. Nexperia_SOT-1289 or Nexperia_CFP15?
4. Will fix. I think this started out as copy of your SOD-128, so you may want to check that, too :)

evanshultz:
1. We can have diodes in common cathode, common anode, series, etc. configurations. SOT-23 packages include diodes in all configurations. Or some diodes may only connect to one of the pin '2' rightmost pins above. If other configurations or pinouts of diodes are available in this package then all 3 pins need unique numbers. Does that make sense?
3. I like the first one. It's more prominent in the doc and also the name of the doc.
4. D'oh! We need more people to catch all my mistakes!

chschlue:
1. Of course that makes sense. I just thought the single diode specific Diode_SMD:D_SOT-23_ANK and _NKA were there for a reason, so I kind of followed their lead.
2. Does this change point 1? Nexperia seems to consider this AAK only.

chschlue:
2. (original 2) All current SOD footprints have silk around the cathode pad. Definitely helps spotting the mark.

evanshultz:
2. I'm a little confused on this one. For a two-terminal part/footprint, yes, that's shown the 7351C documents floating online. But here we have a three-terminal part so the orientation is encoded in the part body and I don't see it as helpful for assembly or troubleshooting. This needs clarification and I think I've brought this up before but it never moved forward. There are many parts type that, to me, don't need further assembly orientation help because the part itself can only reasonably be installed in one orientation. Having the silk farther away from the pads, unlike the two-terminal footprints, means more chance for this silk to get eaten away if there is exposed copper (like pads from other footprints) near this pad. With two-terminal footprints the silk wraps tightly around the pad without poking away so the chance of this is much less. But OK. Leave it if you like it.

chschlue:
If this becomes a "real" 3-pin symbol, I'm happy to remove it.
If it stays a D_, I'd rather keep it even if it's not actually necessary.

poeschlr:
KiCads strength is that it can reuse the same footprint for different parts. Please do not throw that away without good reason.

poeschlr:
Regarding silk (and fab) pin 1 markers: They are also there for quality control. Just imagine someone does indeed place the part the wrong way round. With a silk screen marking you have one standardized thing to check. If you say the mostly covered pads are enough  for this one then you need additional information that in this case no pin marker means "check against pads" instead of "does not matter".

chschlue:
I can push a regular SOT-1289 or Nexperia_CFP in addition.

Commit https://github.com/KiCad/Diodes_SMD.pretty/commit/4d8b429167a31c0b409da64745dac16256a27622#diff-404a567aabaac071757e7c15176bc076 suggests D_SOT-23 based on regular SOT-23 is in the lib on purpose (otherwise there would have been no point in fixing it), so I don't see why a special D_SOT-1289 can't have its place there, too.

chschlue:
> Regarding silk (and fab) pin 1 markers: They are also there for quality control. Just imagine someone does indeed place the part the wrong way round. With a silk screen marking you have one standardized thing to check. If you say the mostly covered pads are enough for this one then you need additional information that in this case no pin marker means "check against pads" instead of "does not matter".

Not sure what you're trying to say.
Are you asking to leave the marker the way it is or to change it to regular multipin style?

poeschlr:
I don't particularly care about the style of the marker. My comment was meant as a counter to the suggestion not to include markers in asymmetric packages like these. (brought up in the comment by @evanshultz )

cp-aquila:
I also like to see pin1 markers in silk for easy debugging with multimeters/probes/whatever without having to look at the schematic. Its a pretty weak argument, because Anode/Kathode could be anywhere on a 3-Pin package. 
But nevertheless, I think its usefull to have it.

chschlue:
Changed to not-diode-but-vendor-specific SOT package.
Has an anode mark now, but I guess one can't have everything.

myfreescalewebpage:
@chschlue I have reviewed all the dimensions and for me it's ok. Merging

PR #1882, opened 2019-09-30
Add Amphenol-FCI SFCxx connectors

cp-aquila:
Add Amphenol-FCI SFCxx connectors

See: https://github.com/pointhi/kicad-footprint-generator/pull/402
Datasheet: https://cdn.amphenol-icc.com/media/wysiwyg/files/drawing/sfv12b_hlf.pdf

grafik

---


poeschlr:
I checked the 9 pin version against the datasheet
Screenshot from 2020-01-07 11-22-41

cp-aquila:
Thanks for the detailed check.
I changed the body- and actuator size.
grafik

You marked the 1.5mm actuator-move red. I took that value from this part of the datasheet: 
grafik Do I miss something?

poeschlr:
The drawings for R1-ST and R2-ST are a bit different, so we will need to have them handled separately in the script (different insert depth and different shoulder width). So now also the footprint differs between bottom and top contact versions

cp-aquila:
I will redo the generator. Perhaps make it load a `.yaml` file, so one script can generate different series.

poeschlr:
That would be overkill. We have generators that do something similar to what you need. One example is  [conn_te_mate-n-lock_tht_side](https://github.com/pointhi/kicad-footprint-generator/blob/master/scripts/Connector/Connector_TE-Connectivity/conn_te_mate-n-lock_tht_side.py) (there we have a dict that holds the parameters that differ including the parts that differ in the name and description)

PR #1883, opened 2019-09-30
Add MQ-6 Gas Sensor

gggfred:
[Datasheet](https://www.winsen-sensor.com/d/files/semiconductor/mq-6.pdf)

imagen

---


cpresser:
- [x] Pin positions are perfect.
- [x] New library for just one part seems odd; but I think its okay since it matches the symbol library.
- [ ] Courtyard clearance should be 0.25mm

poeschlr:
please put this into the generic sensor lib. i do not expect us to ever have enough gas sensors to require a separate lib

gggfred:
> 
> 
>     * [x]  Pin positions are perfect.
> 
>     * [x]  New library for just one part seems odd; but I think its okay since it matches the symbol library.
> 
>     * [ ]  Courtyard clearance should be 0.25mm

please, check again, from the outside fab line, to the courtyard center, actually is 0.25 mm

cpresser:
> 
> please, check again, from the outside fab line, to the courtyard center, actually is 0.25 mm

I did, and measured the `F.Fab` circle to have a radius of 9.5mm and `F.SilkS` circle 9.8mm. Thats 0.3mm difference to me.

gggfred:
Mmmm, I thought than the clearance should be taken from the external side fab line (I mean, 9.5 + 0.05 half fab line width). I fix it now

cpresser:
The F.CrtYd lines can be considered as 'zero' width. Their purpose is to check if parts overlap. Two courtyard-lines can be exactly on top of each other without generating a DRC error.
It is similar for F.Fab lines. They are the theoretical exact boundary of the part. The width itself has little meaning. I am not sure why 0.1mm was chosen in KLC, perhaps its based on IPC, or just because it looks nice.

gggfred:
Thanks!

PR #1884, opened 2019-09-30
Add RB1-125B8G1A RJ45 jack

joshtyler:
Add footprint for RB1-125B8G1A RJ45 jack with integrated 1000BASE-T magnetics.

[Datasheet](https://datasheet.lcsc.com/szlcsc/1901091107_UDE-Corp-RB1-125B8G1A_C363353.pdf#page=3)

image

---


joshtyler:
[Symbol PR](https://github.com/KiCad/kicad-symbols/pull/2219)

joshtyler:
KLC warnings are due to pin 1 being called R1, not 1

aris-kimi:
Hi @joshtyler , thank you for your contribution. Its time came..

I think before a review starts that you need to make one major change,
εικόνα
It mentions board view, so i think you should change your design and update first image.
Thank you for your time.

joshtyler:
Hi @aris-kimi, thanks for looking at this footprint PR!
I'm afraid I don't quite understand what you want changed. Please can you give more information?
Your screenshot shows the view onto the component side of the board, which matches the image of the footprint that I have made in my original post.

aris-kimi:
Oh, i am sorry. I might totally misunderstood something here.
I'll review this today. Thanks

aris-kimi:
Some comments i have:

- [ ] Silk clearance from fab layer and pads must be 0.2mm
εικόνα

- [ ] Courtyard layer it is better to be like this with 0.25mm clearance from pads and body:
εικόνα

joshtyler:
Thanks for the review!

joshtyler:
Thanks for the review!

1. I've changed the fab->silk clearance to 0.2mm from the fab layer as requested. Is this defined as the standard anywhere? KLC F5.1 just suggests 0.2mm from exposed copper.

2. I haven't changed the courtyard. 0.5mm is the correct clearance for a connector, not 0.25mm (KLC 5.3.7). The courtyard doesn't clear the SH holes because the connector uses a flat metal pin that passes through the centre of this hole (see photo). Therefore the current courtyard is 0.5mm away from every part of the connector body.
image

3. I've added an arrow for the pin 1 indication, rather than a notch (as per KLC F5.2.2c). I tried to follow the style of other arrows in the Connector_RJ library.

New footprint screenshot:
image

joshtyler:
It looks like the Travis check failed to run?
`fatal: Invalid symmetric difference expression`
I'm not sure why...

aris-kimi:
Thanks, 
1) You are correct, 0.2mm clearance only from pads is mandatory.

2) You are correct about clearance. Holes on the other hand are part of the footprint, so courtyard needs to readjust around them.

3) Other symbols need to adjust their triangle shape to that 1mm bevel too. Since this is a new addition it should follow KLC as strict as possible.

joshtyler:
2. I've re-read KLC F5.3 - you are correct about holes being included in the courtyard. I've adjusted the courtyard as requested.

3. I think you're mistaken. [KLC F5.2](https://kicad-pcb.org/libraries/klc/F5.2/) reads:

> 2. Footprint polarisation / location of pin-1 is drawn
> a. For IC packages, bevel is drawn at corner next to pin-1
> b. Bevel should be 1mm or 25% of package size (whichever is smaller)
> c. For connectors, a small arrow indicator drawn next to pin-1 should be used

This is a connector, so the correct polarisation marking is an arrow, not a 1mm bevel.

image

joshtyler:
@aris-kimi can this be merged? Unless you have any other comments?

aris-kimi:
I am sorry for the delay, i do not have time at present but i will check it i hope within the next 2 days. It looks good to me from the image.
Thank you @joshtyler .

myfreescalewebpage:
@aris-kimi ping

aris-kimi:
I am not ready yet. I need to settle after moving out. A little off schedule. Will get back to everything i left. Feel free to move further with everything that holds something back.

I hope tomorrow will be the day.

aris-kimi:
Hi @joshtyler , sorry for the delay.

- [ ] One catch only is that forgoten courtyard line over fab layer line at the left side of the body

joshtyler:
Well spotted! I've moved that line from courtyard to fab :)

aris-kimi:
Thanks, great job here.
Merging.

PR #1885, opened 2019-10-01
Molex 43045-xx2x

evanshultz:
Added a missing part of the family that uses metal clips to secure the header to the PCB.
https://www.molex.com/pdm_docs/sd/430450221_sd.pdf
image

This used the same script as an existing header so it got an updated and corrected description.
https://www.molex.com/pdm_docs/sd/430450201_sd.pdf

Script PR: https://github.com/pointhi/kicad-footprint-generator/pull/437

---


poeschlr:
It might be best to give the mounting pads the pin number "MP" and add the -1MP suffix to the pin layout descriptor.

It might be good to include the keepout area from note 8.

The  dimensions seem to be ok.
Screenshot from 2019-10-26 14-03-31

evanshultz:
Good points. Thank you. Updated script and footprints.
image
image

poeschlr:
We have a KLC rule detailing how a keepout area is to be drawn. http://www.kicad-pcb.org/libraries/klc/F4.5/ As this is so small i would accept it without hatching. (so only a rectangle) A filled polygon would violate this rule.

evanshultz:
Done.
image

poeschlr:
thanks

evanshultz:
@poeschlr 
Thank you.

I believe the KiCad dev team is smart enough to implement keepouts with polygons once they are supported by PCBnew. As we're left with visual checking for keepouts at this time, a polygon (to me, at least) is easier to notice an overlap than an unfilled area. I'm also not sure if PCBnew supported polygons when the KLC clause was added. So using polygons now means that only the polygon layer has to change once keepouts are supported. Would you consider allowing polygons for keepouts? I can make a PR on the website repo with my first draft update for discussion if you would support this.

PR #1889, opened 2019-10-04
add Alps SKRTLAE010 horizontal push button

lorem-ipsum:
This adds a horizontal SMT push button from Alps [SKRTAE010](https://www.alps.com/prod/info/E/PDF/Tact/SurfaceMount/SKRT/SKRT.pdf). If this footprint is fine I would create the SKRTB variant as well which lacks the guidance holes. Designed to be used with the standard SW_Push symbol (however a SW_Push_MP might be nice to have).

I placed the center of the symbol in between the guidance holes, thats how I read the guidance lines in the drawing. I also used roundrects instead of square rectangles, I hope thats ok.

Screenshot_20191004_101655

---


poeschlr:
please inlcude `-MP` in the name (after `1P1T`) I did not review this any further

cpresser:
- [x] The center location is okay, its in the middle of the guide bosses.
- [ ] please add the keepout area as shown here:
image
- [ ] Body size is off by 0.01mm
- [ ] Then Pin1 marker seems wrong. There are two Pin 1, why did you choose the left one? This also raises the question if we need a Pin1 marker at all. If yes, it should also be on F.Fab. @poeschlr Any thoughts on this?

lorem-ipsum:
Fixed the body size and added the keepout as shown below. Also dropped the pin 1 indicator. I added it in the beginning due to the labeling of the original datasheet. Then I did not delete it at the point where I realised that the pads on the side are connected to each other.

Screenshot_20191225_105858

cpresser:
Thanks!

PR #1891, opened 2019-10-05
MDBT42Q (Bluetooth Low energy model)

prajwaltelkar:
Raytac’s MDBT42Q  (Bluetooth low energy or BLE) module designed based on Nordic nRF52832 SoC solution, which incorporates: GPIO, SPI, UART, I2C, I2S, PWM and ADC interfaces for connecting peripherals and sensors.

---


poeschlr:
Add a screenshot of your contribution
Add a link to the datasheet that you used to create the footprint

prajwaltelkar:
1. Screenshot of the footprint.
mdbt42q(screenshot)

2.Url of the datasheet used to create footprint.
https://statics3.seeedstudio.com/assets/file/bazaar/product/MDBT42Q-Version_B.pdf

poeschlr:
From the screenshot:

Add an outline on silk
the chamfer marking pin 1 should be 45 degree

prajwaltelkar:
- I have added an outline silk layer 

- I have also chamfered marking pin1 to 45 degrees

udated figure mdbt42q

poeschlr:
The fab outline must represent the nominal size of the component body.

Move your silk outline closer to the component body (=fab outline) such that it can be used as a help for aligning the part.
(It will not go all around as the pads are in the way.)

Add the ground keepout zone as required by the datasheet. Do it as described in http://www.kicad-pcb.org/libraries/klc/F4.5/.

I can only review the dimensions of the part if the fab outline is drawn correctly as it is used as the reference in the datasheet.

prajwaltelkar:
I have updated the footprint as per the requirements mentioned above

- 
mdbt42q(3)

prajwaltelkar:
silk outline is closer to the component body (=fab outline).

poeschlr:
The fab outline is still not close to the body size. (should be 10x16mm is 7.54x13.36mm) The outer silk outline must be outside the nominal body (but not too far away such that it is still useful with aligning the part.)
Silk must also have a clearance to the pads (At least the same as silk line width. Preferably 0.2mm to allow cheaper fabs to be used) This will mean that the silk outline will not go all around.

Also, please clean up the outline. You have many short lines where a single line would do. (use outline mode to see this)

poeschlr:
Also please look at the rule regarding keepout a bit closer. You are missing the hatched lines.
And well lets say your description for what should be excluded does not really make much sense.

I would suggest to have the word keepout in every one of these and then specify what exactly should be excluded in clear english.

prajwaltelkar:
I have added the hatched lines for the keepout region and also I have made the changes for fab layer and silk layer.
mdbt42q_new(dec)

poeschlr:
I specifically asked for the word keepout to be included. Please do so to fit with the rest of the library. You can place it above the descriptive text.

prajwaltelkar:
final

I have written the Keep-out zone ,within the keepout area.

poeschlr:
please prefix the footprint name with the manufacturer name

The fab outline is not correct (again it MUST represent the nominal size of the part body which is given as 10x16mm in the datasheet you linked.)

clearance between silk and pads must be at least the same as the silk line width. But for a part this large i suggest to go with 0.2mm clearance to allow cheaper fabs to be able to produce the pcb.

Could you be a bit more careful with the hatched lines of the dwgs layer? They should be close to parallel and of near equal spacing such that the footprint looks professional as users will not trust it otherwise. (set the grid origin to one of the corners of the area in question and then select a reasonable large grid to draw them.)

The large keepout areas "south" edge should be exactly at the corner of the upper most pads.
The smaller keepout areas "north" edge should be exactly on the same y coordinate as the "south" one from the large area.

The pads on the left side are misaligned with the rest of the pads (The cyan help lines represent the board under the assumption that the left side pads are on the correct position)

Screenshot from 2020-02-05 22-22-30

prajwaltelkar:
RAYTAC_MDBT42Q
1.I have prefixed the footprint name with manufacturer name.
2.I have rectified the fab outline to nominal size.
3.I have took care to see that pads and silk layer are not too close.
4.I have tried to make them look parallel.
I think so other needful things are also rectified.

poeschlr:
It seems you did not push your changes as the footprint still has its old name and does not look like your latest screenshot.

prajwaltelkar:
I have pushed the updated footprint .
Sorry for the inconvinence.

poeschlr:
From our test script:
Your file must use linux stile line endings ("\n" = line feed (LF)) not windows style ("\r\n" = carriage return plus line feed (CRLF)). This is normally ensured by your git client but i fear not by github web upload (so if you use github web upload you need to fix the line endings with a text editor like notepad++, atom, vscode, ...)

The footprint must be setup to accept a 3d model. So the 3d model path must be setup even if no model is supplied.

You can ignore the warning regarding the centering as this is an asymmetric footprint where our script kind of fails.

---

I made another dimensioned drawing. There are still many mistakes in your footprint.
The pads on the left and right side are too far in and too short. They also are different in size (left is shorter than right) but should be equal. (The current pad arangement would make it hard to solder)

I also suspect the bottom pads are too near to the other pads in y direction but can not really determine that because the outline is not correctly aligned to the pads (see measurement from top rightmost pad to outline top) Which is why i stopped dimensioning after that (the body outline is used in the datasheet as the main reference so it must be correctly aligned for me to be able to check your footprint)

Also notice that the left side pads are not aligned with the right side pads in y direction. (compare dimension for top left pad with top right pad both against top body outline)

Screenshot from 2020-06-13 09-24-10

prajwaltelkar:
Sure I'll do this by the end of the week.

PR #1893, opened 2019-10-05
add SMD Fuse Schurter UMZ250

lorem-ipsum:
This adds a footprint for the [Schurter UMZ250](https://us.schurter.com/bundles/snceschurter/epim/_ProdPool_/newDS/en/typ_UMZ_250.pdf) series of SMT fuses.

Screenshot_20191005_170001

---


cpresser:
- [ ] Body Size is 11.1, should be 11.3
- [ ] Pad-Distance is not correct, should be 5.5mm
- [ ] I suggest to add the 'ears' created by the holder to the body-outline in F-Fab

lorem-ipsum:
I've fixed the pads and reworked the fab layer. I copied the fab from the UMT250 (which is clipped into the "ears") and added the ears as extra rectangles.
Screenshot_20191225_155937

cpresser:
image
- [x] Pad size and distance
- [x] courtyard
- [ ] Description has `3.8 x 11.1 mm`. Where did you get these values? I see `4.2x11.3mm`.
- [ ] The F.Fab lines look to crowded in my opinion. This is highly subjective. I tried several different styles and don't have a clear favorite. This one is the least shitty one I could come up with: 
image

lorem-ipsum:
Not sure where I got the size from the datasheet also says 4.2mm. Fixed it. Also used your suggestion on the Fab lines, I agree this looks cleaner.

PR #1907, opened 2019-10-10
Package_SO fix EP round radius

poeschlr:
Splitting #1821 into smaller bits. This one handles all Package_SO footprints with the new EP radius.
Issue: #1798


evanshultz:
@poeschlr 
Both footprints mentioned above are used by existing symbols, either with or without thermal vias. Since users may desire to choose a via or no-via variant regardless of the default footprint, I'd like to see both included. Can you add them here or in a new PR?

poeschlr:
Added in https://github.com/KiCad/kicad-footprints/pull/2037

evanshultz:
I only notice that MSOP-8-1EP_3x3mm_P0.65mm_EP1.68x1.88mm has pad size and location changed. But that's fine. Thanks!

PR #1908, opened 2019-10-10
Reexport SO footpritns to get script and FPs in sync

poeschlr:
Script state from pointhi/kicad-footprint-generator#422
Another spiting of #1821

This is mainly here to ensure that the script and footprint repo are in sync again. (There should not be major changes)


evanshultz:
Only `SSOP-20_5.3x7.2mm_P0.65mm` had any copper changes and it's not major.

PR #1910, opened 2019-10-10
Fix EP radius and resync of LGA footprints

poeschlr:
Splitting #1821 into smaller bits. This one handles LGA footprints with the new EP radius.
I also included non EP footprint resyncronization in this pull request as this lib has only a few footprints generated with the script.
Issue: #1798

For some reason one footprint is generated but does not exist in the lib:
- Texas_SIL0010A_MicroSiP-10-1EP_3.8x3mm_P0.6mm_EP0.7x2.9mm.kicad_mod


evanshultz:
@poeschlr 
The footprint not in the lib you mention above is not included here. I believe it is used by the `Regulator_Switching:LMZM23601` symbol. Can you include it, either here or in another PR so we don't have a symbol referencing a missing footprint?

poeschlr:
Adding that footprint will require the yaml file to be double checked as we kind of have to assume that it was never checked (otherwise the footprint would be here.) Not sure how it happened that it is in the script but not the footprint repo.

evanshultz:
The thermal vias version of this footprint is present in the lib now, but not the non-thermal vias version. And there is a symbol using the thermal vias footprint. I don't see this footprint included at https://github.com/pointhi/kicad-footprint-generator/blob/master/scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON/size_definitions/lga.yaml, so maybe I just don't understand your point above. Why did you mention this footprint?

FYI, all footprints in the PR look fine to me so it's just the above stopping me from merging.

poeschlr:
The script always creates a non thermal via version, the thermal via version is the one that is optional. There must have been a reason why the non thermal via one was excluded when this was originally merged.

evanshultz:
@poeschlr 
So would you add the missing footprint, in case someone wants to use the non-via version and as there's no reason to exclude it, and then I will merge? If you're accepting the thermal via version as-is I don't get why you require the YAML file to be checked to merge the non-via version.

poeschlr:
See https://github.com/KiCad/kicad-footprints/pull/2036

PR #1912, opened 2019-10-10
Fix EP round radius for TI QFN footpirnts

poeschlr:
Splitting #1821 into smaller bits. This one handles TI specific QFN footprints with the new EP radius.
Issue: #1798


evanshultz:
Branch conflict.

And my question elsewhere is if TI parts should use TI footprints which have rounded corners and all the other details (mask oversize) to exactly match the TI documentation.

evanshultz:
@poeschlr 
Since you want to go generic, I tried to find if any of these packages are exactly an industry-standard spec. Unfortunately, not too much good news.
- Texas_R-PWQFN-N28_EP2.1x3.1mm: References JEDEC MO-220; I don't see it there but variation VFSD is close but 0.5mm pitch instead of 0.4mm.
- Texas_RGE0024C_EP2.1x2.1mm: Should point to page 37, not 34
- Texas_RGE0024H_EP2.7x2.7mm: Page 39, not 43
- Texas_RGV_S-PVQFN-N16_EP2.1x2.1mm: Says MO-220 but I don't see a 2.1mm TP in Table 7A or 7B.
- Texas_S-PDSO-N10_EP1.2x2mm: Maybe the documentation changed, but I don't see `S-PDSO-N10` anywhere in the datasheet.
- Texas_S-PVQFN-N16_EP2.7x2.7mm: I don't see it in MO-220, but it's close to variation VGGC-2.
- Texas_S-PVQFN-N20_EP3.15x3.15mm: Between variations VHHC and VHHC-2.
- Texas_S-PVQFN-N24_EP2.1x2.1mm: Page 108 or 109.
- Texas_S-PVQFN-N32_EP3.45x3.45mm: Page 54 or 55.
- Texas_S-PVQFN-N36_EP4.4x4.4mm: Page 35 or 36.
- Texas_S-PVQFN-N40_EP2.9x2.9mm: Page 111.
- Texas_S-PVQFN-N40_EP2.9x2.9mm: Matches MO-220 variation VJJD-2 and that is stated in the doc.
- Texas_S-PVQFN-N40_EP3.52x2.62mm: Same as above.
- Texas_S-PVQFN-N40_EP4.15x4.15mm: Also same as above.
- Texas_S-PVQFN-N40_EP4.6x4.6mm: Mentions VJJD-2 as above, but in fact that variation allows only up to a 4.25mm thermal pad max. Variation VJJD-5 supports 4.8mm thermal pads with a 4.65mm nominal which is close, but the lead length also varies slightly (0.35..0.4..0.45mm). In addition, should link to page 54.
- Texas_S-PVQFN-N48_EP5.15x5.15mm: Page 111.
- Texas_S-PVQFN-N64_EP4.25x4.25mm: Page 120 or 121. Name `RGC 64`.
- Texas_S-PWQFN-N16_EP2.1x2.1mm: Matches MO-220 variation VGGC. Nmaed `RTY`.
- Texas_S-PWQFN-N24_EP2.7x2.7mm: Very close to variation VGGD-6. Named `RTW`.
- Texas_S-PWQFN-N32_EP2.8x2.8mm: Page 90. Named `RSN`.

Updating the page numbers may not be worthwhile since docs can change over time while the link remains the same, which is probably what happened. But while I was looking, I figured I'd note it. I also could have also gone back and grabbed all the the TI three-letter codes but I didn't think about that at first.

poeschlr:
As noted in my other pull request, going generic will still mean that we use the TI datasheets for the package size source but long term remove the TI specific naming.

Especially as we ignore the suggested footprints with our generators.

poeschlr:
Page numbers fixed, jedec references added where you found a fitting one.

We know since quite some time that ti is in the middle of changing its naming convention for footprints. This is yet another reason to long term get rid of the TI naming in the lib.

evanshultz:
Thanks!

PR #1913, opened 2019-10-10
Fix EP round radius for QFN footprints

poeschlr:
Splitting #1821 into smaller bits. This one handles "normal" QFN footprints with the new EP radius. (I think one or two non EP footprints are also included)
Issue: #1798

For some strange reason the following footprint is generated but does not exist in the lib:
- QFN-14-1EP_1.6x1.6mm_P0.4mm_EP0.74x0.74mm_ThermalVias.kicad_mod


evanshultz:
@poeschlr 
I don't find QFN-14-1EP_1.6x1.6mm_P0.4mm_EP0.74x0.74mm_ThermalVias.kicad_mod used by any symbol now. But I think I figured it out:
- `qfn-1x.yaml` contains this definition on line 62 but it's named wrong as `QFN-14-1EP_0.64x0.64mm_P0.4mm_EP1.6x1.6mm`.
- https://github.com/KiCad/kicad-footprints/pull/1443 added this footprint.
- https://github.com/KiCad/kicad-symbols/pull/1585 is the open symbol PR.

So I think we need this footprint. However, some work may need to be done:
- The symbol PR may need to be updated to match the footprint name.
- The size definition in the YAML file could be corrected.
- You may want to remove or rename the Skyworks footprint from the PR above since it doesn't conform to the typical manufacturer-specific footprint name convention.

evanshultz:
The following footprints have minor copper changes:
- `QFN-12-1EP_3x3mm_P0.51mm_EP1.45x1.45mm`
- `QFN-24-1EP_4x4mm_P0.5mm_EP2.65x2.65mm`
- `QFN-38-1EP_4x6mm_P0.4mm_EP2.65x4.65mm`

`QFN-32-1EP_7x7mm_P0.65mm_EP5.4x5.4mm_ThermalVias` oddly has the location of the thermal vias changed but I think that's OK too.

Should the script be updated to add a suffix if the corner pads are beveled?

Here are other footprints for TI parts which don't have rounded corners even though the footprint drawings shows it:
- `QFN-20-1EP_3.5x3.5mm_P0.5mm_EP2x2mm` (no footprint given in the doc provided, but it's used by bq25570 and one is given at http://www.ti.com/lit/ds/symlink/bq25570.pdf)
- `QFN-56-1EP_8x8mm_P0.5mm_EP4.5x5.2mm` (no footprint given so it's actually OK)
- `QFN-56-1EP_8x8mm_P0.5mm_EP5.6x5.6mm`
- `QFN-64-1EP_9x9mm_P0.5mm_EP6x6mm` (also missing a link to page 42 of the datasheet)

poeschlr:
The skyworks footprint stuff is fixed in https://github.com/KiCad/kicad-footprints/pull/2035 and https://github.com/KiCad/kicad-symbols/pull/2423

evanshultz:
Thanks!

That leaves two things from https://github.com/KiCad/kicad-footprints/pull/1913#issuecomment-554082066:
1. Do you want to add a suffix if corner pads are beveled? I guess no, just asking anyway.
1. Can you add the direct page link to the URL of the footprint missing one?

poeschlr:
Doclink updated

> Do you want to add a suffix if corner pads are beveled? I guess no, just asking anyway.

Not for now. We might want to look into this at a later point. (The script luckily knows which footprints have corner pads so it is easy to add such a suffix later on.)

evanshultz:
Thanks!

PR #1914, opened 2019-10-10
Fix EP round radius of DFN footprints

poeschlr:
Splitting #1821 into smaller bits. This one handles "normal" DFN footprints with the new EP radius.
Issue: #1798


evanshultz:
`DFN-6-1EP_1.2x1.2mm_P0.4mm_EP0.3x0.94mm_PullBack` has some minor pad size/location changes.

PR #1916, opened 2019-10-11
add L_Bourns-SRU1028

nerdyscout:
equivalent to the already existing SRU8043 this does add the footprint for SRU1028

grafik

[datasheet](https://www.bourns.com/docs/Product-Datasheets/SRU1028.pdf)

---


gmsotavio:
Hi. Thanks for your contribution.
- Why have you removed the silk near pad 1? This footprint is not polarized.
- I would suggest you appending the body dimensions to the footprint name. Appending body dimensions is a recent decision, and we are doing this for all inductors, but this is not well documented yet. The name would be L_Bourns-SRU1028_10.0x10.0mm.
- Some dimensions are wrong. Take a look in the dimensioned drawing bellow.
image

nerdyscout:
[SRU1028](https://www.bourns.com/docs/Product-Datasheets/SRU1028.pdf)
image

[SRU8043](https://www.bourns.com/docs/Product-Datasheets/SRU8043.pdf)
image

I made no several changes, as well to the existing SRU8043 to keep them compliant.

gmsotavio:
- [ ] I haven't checked the SRU8043 yet.  As you have renamed it, current designs that use this footprint will break. So, I believe that is a better option to open a new PR to SR8043.
- [ ] There are two incorrect dimensions remaining. Take a look at the dimensioned drawing. The proper value is 0.25mm.
image
-[ ] I realized that you had used the Shape offset X (in red color). We usually do not use this feature. I think it is a better option to calculate the right value for Position X (in blue color), taking into account the pad offset.

nerdyscout:
ok, reverted changes of CR8043 and changed CrtYd as well as pad offset.

gmsotavio:
The dimensions look good for me now.
image

image

- [ ] Accidentally, you have removed the RefDes on the Fab. Layer. Add the missing RefDes and this footprint will be ready to be merged.

gmsotavio:
Great! Thanks.
image

PR #1742, opened 2019-07-25
LED-APA102-2020: Move F.Fab marker

cp-aquila:
LED-APA102-2020: Move F.Fab marker

This is the fix mentioned here:
https://github.com/KiCad/kicad-footprints/pull/1389#issuecomment-481635169

grafik

---

All contributions to the kicad library must follow the [KiCad library convention](http://kicad-pcb.org/libraries/klc/)

Thanks for creating a pull request to contribute to the KiCad libraries! To speed up integration of your PR, please check the following items:

- [x] Provide a URL to a datasheet for the footprint(s) you are contributing
- [x] An example screenshot image is very helpful 
- [ ] If there are matching symbol or 3D model pull requests, provide link(s) as appropriate
- [ ] Check the output of the Travis automated check scripts - fix any errors as required
- [ ] Give a reason behind any intentional library convention rule violation.

PR #1921, opened 2019-10-13
Adds EC15 (1.5x1.5mm) Footprint for SK6812 LED

ottojo:
This adds the footprint for the 1.5x1.5mm SK6812 individually adressable LED:

Datasheet: http://www.newstar-ledstrip.com/product/20181119172602110.pdf

LED_SK6812_EC15_1 5x1

---


cpresser:
Hi and thanks for your contribution.
- [ ] Can you please try to get a real data-sheet.
- [ ] Pin1 marker and the maker on the LED itself  (its at Pin4) do not align. you somehow need to indicate that. See here https://github.com/KiCad/kicad-footprints/pull/1742 Unfortunately we don't have a really good solution for that.
- [ ] There is also a arrow pointing to Pin2, I would include that in F.Fab
- [ ] You did not add silkscreen at at. The minimum requirement per KLC is a Pin1 marker. You can go down to 0.1mm width and 0.1mm clearance for this part since it is so small.
- [ ] The `_P0.9mm` is not required, you can remove it. No other LED in the library has it. And I don't see the value added to the designer.

ottojo:
I think i have addressed all your comments, for the markers on the LED i added a small square on the F.Fab layer in the corresponding corner, which deviates from the solution in #1742 but is a more obvious solution in this case IMO.

<rant> Confusing enough already with 2 markers, none of which mark Pin 1, and a conflict in the datasheet regarding the location of the arrow... It is indeed between pads 3 and 2, like in image 4, and not between pins 1 and 2 like in image 6 </rant>

cpresser:
> I think i have addressed all your comments, for the markers on the LED i added a small square on the F.Fab layer in the corresponding corner, which deviates from the solution in #1742 but is a more obvious solution in this case IMO.

Thanks for the changes. We are getting there :)
You missed the size of the refdes in F.Fab.

I don't really like the square. The library follows IPC, so the marker should be the cutoff corner.
The silkscreen marking should be on the same corner as the marking on the part. Anything else will confuse the hell out of the operator setting up the pick'n'place machine.
The hack from the other PR is to add the `1` on the silkscreen next to pin1.

All of that should give enough information to an operator so they start thinking and hopefully put the LED into the correct orientation.
See https://github.com/KiCad/kicad-footprints/pull/418#issuecomment-372084703

There is no known 'good' solution for this problem. So the least we can do is keep the library consistent and don't introduce different styles.

image

> 
> Confusing enough already with 2 markers, none of which mark Pin 1, and a conflict in the datasheet regarding the location of the arrow... It is indeed between pads 3 and 2, like in image 4, and not between pins 1 and 2 like in image 6

Did you check that physically with one of the LEDs? If yes, can you please put a picture here, for documentation purposes?

Do you plan to have this part on board that gets a production run in a professional factory? If yes, please add feedback to this issue once you receive the first batch of boards.

ottojo:
Sounds good, i made the changes you suggested.

I did produce some boards with this part, but i did assemble them myself.

Since i do not own a camera with a macro lens, this blurry smartphone photo is the best i can currently do:

IMG_20191015_221248

Issue #1095, opened 2018-11-18
KLC vague on dimensions to use w.r.t tolerances


johnbeard:
The KLC documentations is vague on which dimensions to use. As I understand it, the implicit rules are:

* F.Fab: the shape is the *nominal* body outline of the part from the datasheet
* F.SilkS: The silk is just outside the *nominal* shape (so outset from the F.Fab by half the line width plus, presumably, "a bit"). The "bit" is also not well-defined, if the pads are not the outermost elements.
* F.CrtYard: 0.25mm (or more for certain types) from the outermost of the pads or body.

This could cause issues when a part has wide tolerances of the body. For example, imagine a (THT, so no pads outside the body) part that is 10x10mm +/- 1mm. If you draw the silk just outside the nominal, and a courtyard 10.5mm wide, if you get a part that is in-spec at 10.9x10.9mm, you will find it not only covers the silk but also spills out of the courtyard and might not fit when assembled.

I think the KLC should spell out in detail which dimensions one should use for each of Fab, Silk and CrtYd layers, and whether or not to account for tolerances specified in the datasheet.

evanshultz:
You are correct that KLC could use some improvement here. Some of these have been discussed already but I'm too lazy to dig up the relevant conversations.

The only comment to make is that the standard offset of silk lines from fab lines used in most of the librarian-developed (or -inspired) scripts use 0.11mm. This means the edges of the fab and silk lines touch. KLC specifies that silk should not be under SMT parts (and your comments about excessive tolerance may also have an issue with that) and 0.11mm offset is the minimum value that meets this rule.

Do you have a part in mind with dimensions and tolerance as described above, or are you only hypothetically postulating issues and pointing out good areas where KLC can be improved?

johnbeard:
A little bit of both.

I was looking at several parts like this:

![2018-11-20_124107_496x547_screenshot](https://user-images.githubusercontent.com/5386578/48774214-7de80100-ecc1-11e8-9a15-fee348529ac3.png)

The case only as a maximum size defined, and the position of that case WRT pin 1 is defined only in terms of maximum and minimum offsets from pin 1. Thus, in the x-axis only:

* The case edge on the left:
** Leftmost is position is as far left as possible: -2.62mm
** Rightmost is position is as far left as possible: -1.92mm
* On the right:
** Rightmost: when the case is at maximum and as far right as possible: -1.92 + 21.87 = 19.95mm
[...]

PR #1927, opened 2019-10-18
Add Taiyo-Yuden NR 10050 inductor footprint with handsoldering variant.

reportingsjr:
This adds a standard footprint and a handsoldering variant for Taiyo-Yuden NR 10050 SMD inductors.

For the handsoldering variant I extended the pads a similar amount as the Taiyo-Yuden_NR-80xx_HandSoldering footprint vs the normal (1.4mm extra for each pad).

---

- [x] Provide a URL to a datasheet for the footprint(s) you are contributing: https://ds.yuden.co.jp/TYCOMPAS/or/specSheet?pn=NR10050T1R3N
taiyo yuden nr 10050 handsoldering footprint
taiyo yuden nr 10050 footprint
No 3d model created for this


gmsotavio:
Hi @reportingsjr,

I'm not reviewing this right now, but consider to include the body size in the name.

reportingsjr:
@gmsotavio Good catch on the courtyard! I pushed it out to be 0.25mm beyond the pads as it should have been.

I also added the inductor body size in the name of the footprint.

reportingsjr:
New screenshots:
NR10050
NR10050 handsoldering

gmsotavio:
Hi @reportingsjr ,

Some thoughts about your contribution:

- [ ] There are some bad dimensions on the Fab. layer. Check the dimensioned drawing bellow.

image

image

- [ ] Something is inverted here
image
- [ ] Change the X coordinate of the RefDes (Silk) to 0.
- [ ] Datasheet link is missing in the description
- [ ] Usually, the dimensions in the name follow this rule: _WxLmm. Please add `_9.8x10.0mm` to the footprint name, rather than `_10.0x9.8mm`.
- [ ] Remove footprints you do not want from the PR. You are submitting 2 footprints, but there are 4 files in the PR.

reportingsjr:
> Hi @reportingsjr ,
> 
> Some notes about your contribution:
> 
> * [ ]  There are some bad dimensions on the Fab. layer. Check the dimensioned drawing bellow.
I looked at your dimensioned drawing and I'm not sure what is wrong? I put a very rough outline in, with the max dimensions at 10.3x10.3mm. What do you think needs to change?

> * [ ]  Something is inverted here
Yep, I thought the drawing on the right was a bottom view. It looks like it isn't.
> * [ ]  Change the X coordinate of the RefDes (Silk) to 0.
Fixed
> * [ ]  Datasheet link is missing in the description
Fixed
> * [ ]  Usually, the dimensions in the name follow this rule: _WxLmm. Please add `_9.8x10.0mm` to the footprint name, rather than `_10.0x9.8mm`.
Fixed
> * [ ]  Remove footprints you do not want from the PR. You are submitting 2 footprints, but there are 4 files in the PR.
Looks like I forgot to git rm them. Fixed!
> **Edit**
> 
> * [ ]  Change 3D shape names to take into account the body dimensions.
Fixed.

Thanks for being so thorough! I wish I did this more frequently so I could become more familiar with all of the rules.

gmsotavio:
Hello @reportingsjr ,

Thanks for addressing the recommended changes! Asking your question:

> I looked at your dimensioned drawing and I'm not sure what is wrong? I put a very rough outline in, with the max dimensions at 10.3x10.3mm. What do you think needs to change?

Usually, the fab outline should represent the nominal dimensions of the components. Some manufacturers don't provide the nominal values, only the maximum. Less often, they provide a mix of maximum and nominal values, such as seen in this PR https://github.com/KiCad/kicad-footprints/pull/1721. In your case, the nominal values are provided in the datasheet, so i think that nominal values are the right way to go. I have checked the KLC and I found no information about Fab outline, but your question has already occurred before (https://github.com/KiCad/kicad-footprints/issues/1095). The best answer I've found from that thread:

"_The requirement for nominal dimensions comes from industry standards. The reason for this is that maximum tolerances are normally never really reached (very low probability)
It is also hard to determine what maximum means. Especially for long chains of dependent dimensions._"

_Originally posted by @poeschlr in https://github.com/KiCad/kicad-footprints/issues/1095#issuecomment-444194815_

If you do not agree, drop me a line.

reportingsjr:
Ok, that makes sense. I've pushed the fix for that.

gmsotavio:
- [ ] Courtyard isn't quite right
image

- [ ] Please, change the coordinates of RefDes (Fab) to (0,0).

gmsotavio:
- [ ] Courtyard isn't quite right
image

- [ ] Please, change the coordinates of RefDes (Fab) to (0,0).

reportingsjr:
Ok, I changed all of the text to be centered on the X axis and put the Fab ref des at (0,0). I also corrected the courtyard size.

For what it is worth I didn't move the silk which is now just outside of the courtyard, as <0.25mm from the component seems too close. I didn't see anything in the KLC about this and I see other components with silk outside of the courtyard, so hopefully that is ok!

gmsotavio:
> For what it is worth I didn't move the silk which is now just outside of the courtyard, as <0.25mm from the component seems too close. I didn't see anything in the KLC about this and I see other components with silk outside of the courtyard, so hopefully that is ok!

You are right. KLC tells us nothing about silk offset from fab outline or anything about silk lines being inside or outside the courtyard layer. It is allowed to do what you did. Don't worry. I had that same doubt some time ago. I found a discussion (https://github.com/KiCad/kicad-footprints/issues/1095) where @evanshultz says "_...at the standard offset of silk lines from fab lines used in most of the librarian-developed (or -inspired) scripts use 0.11mm. This means the edges of the fab and silk lines touch._" I would like to hear @poeschlr and @evanshultz, what they think about add that rule as a silk requirement to a future revision of KLC. If we apply this rule into your footprint it becomes:

image

But as I said, at this time, this is not mandatory. I checked your footprint, and everything seems good for me now.

PR #1930, opened 2019-10-20
Nrf24l01 module smd footprint

critBit95:
The footprint:
image

I could not find the datasheet for this specific module - The footprint is based on:
* the image all Aliexpress auctions  - ( like [this one](https://www.aliexpress.com/item/32949511170.html))  seem to reference 
* The module I have on-site and a calliper

image

PR for the symbol: https://github.com/KiCad/kicad-symbols/pull/2254

The symbol anchor is a little bit of the center calculated by your scripts. This is due to pads being slightly larger than the module itself. However the anchor sits ideally in the center of the module itself, so in my opinion this is the valid option.


poeschlr:
Are you ok with your contribution being published under the KiCad library licence? If you are please sign the licence agreement.

critBit95:
Oh, thought signing on symbol repo is enough. I'm ok with the license thing.

critBit95:
See https://github.com/KiCad/kicad-symbols/pull/2254

PR #1933, opened 2019-10-21
SD kyocera 145638109* fixed

serkanxselcuk:
Bug:
- SD Card SH pins didn't match with symbol pins.
- One of pin name field was empty.

Fix:
- Shield pins are renamed to 12 and 13, as like in symbol.
- Dimensions are not changed.

image
---


gmsotavio:
@poeschlr Can we accept these changes now? My understanding is that if someone is using this footprint, so the design is broken from the beginning. Case affirmative, I will request to the author of the PR to fix more one thing, such as the improper rotation.

poeschlr:
If the footprint now fits the current symbol then yes this should be accepted. (This is not a review just an answer to the question)

evanshultz:
I'm not on-board with this change. The footprints were added at https://github.com/KiCad/kicad-footprints/pull/862 and it was clear they were intended to be used with named pins as discussed in excruciating detail and finally accepted at https://github.com/KiCad/kicad-symbols/pull/592.

So in my opinion, the footprints shouldn't have been merged without matching symbols that extract away pin numbers from these standardized parts (SD card variants). This PR could be merged to make things work now, but that's not taking into account the decision made which should make these card symbols easier to use in the future. I'm willing to submit the missing symbols.

I would agree to rotate the footprints in this PR or make other corrections.

serkanxselcuk:
Hi @evanshultz ,
thank you for focus on #862 and decision on KiCad/kicad-symbols#592. I didn't know about them. I just tried to fix error on my project as fast as possible. But it is true, common usage is CD and WP. It makes sense to change my process, so that I check merged PR comments before change any implemented library parts.

I will
- [ ] change pin names to CD, WP, SH, MP for this PR
- [ ] rotate footprint 180°
- [ ] center mid point of footprint
- [ ] move RefDes so that it`s visible even if SD card is inserted (below footprint dimensions)
- [ ] make a new PR to match symbol.

Any other change requests?

poeschlr:
My decision (I hate to do it but i need to ensure the best result for users even if it means more work for maintainers, please do not take this personally you all know that keeping the lib stable has a very high priority.): 
Right now all symbols have numbered pins (tag 5.1.4). We can not have footprints in the lib that disagree with the symbols that are meant for them. (all footprints except the 4 SD_Kyocera have only numbered pads)

We can not change symbols in this manner during a minor release. Meaning the footprints should follow the current symbols and also have numbered pins. The switch to named pins can happen after we start development for v6 (after the new file format is somewhat stable. assuming the file format does not offer something better.)

I am very sorry for everyone effected by this bug. And also by everyone invested in getting the lib towards allowing generic symbols for sd cards.

---

Also note that the 4 SD_Kyocera footprints would need fixing even if we would decide to go with named pins as only one shield pin has a pin number right now.

---

TlDr: the correct fix at this point in time is to fix the 4 footprints.

evanshultz:
@poeschlr 
I thought it was clear at the second link I sent that new symbols would be required with pin names. Those new symbols would be used for new footprint contributions (or this one which is broken now} with the old symbols staying as they are for existing footprints. I realize it has been a while since discussion took place but your post above seems to discount the result that I shared.

All that being said, sticking with numbers on these footprints works too. I only want to clarify because I understand your comment doesn't take into account the previous discussion I linked about this topic.

As you can see, I intended this to use a symbol with named pins. There were other card footprints wrongly merged at this same time which created a previous headache because of pin names. As I said above, I'll happily submit the missing symbols with pin names and then they will be available for future use; the pin number symbols can be removed later if the time is right.

serkanxselcuk:
image
Changes for all four:
- rotated 180°
- centered to mid point
- RefDes moved to visible area.

I add @evanshultz , I would like to change both footprint and symbol to named pins as soon as possible.

cp-aquila:
> 
>     * centered to mid point
> 
It seems like the previous center-point was the pick-n-place location shown in the datasheet.
There is an exception for this in the KLC: http://kicad-pcb.org/libraries/klc/F6.2/
The travis-ci error is a false positive in this case, so I would vote to keep the previous location.

gmsotavio:
I agree with @cp-aquila , the anchor should be centered to the pick-n-place point which is not always is in the geometric center of the component.

image
image

serkanxselcuk:
@gmsotavio @cp-aquila ,
thank you for your attention. Embarrasingly p&p slipped my mind.
Pad orientation is also set to zero now.

chschlue:
Thanks.

PR #1935, opened 2019-10-25
Add Maxim S16+5, 16 pin SOIC

manio:
This is a footprint for a Maxim/Dallas Small-Outline IC, Narrow (0.15in), 16 pins.
Listed here: https://www.maximintegrated.com/en/design/packaging/package-information.html

Maxim Package code:  S16+5
Outline Drawing No.: 21-0041 (https://pdfserv.maximintegrated.com/package_dwgs/21-0041.PDF)
Land Pattern No.:    90-0097 (https://pdfserv.maximintegrated.com/land_patterns/90-0097.PDF)

This package is used in e.g.
DS2482-800 (https://datasheets.maximintegrated.com/en/ds/DS2482-800.pdf)
DS2408     (https://datasheets.maximintegrated.com/en/ds/DS2408.pdf)


poeschlr:
You can not simply move the pads outwards. This would change the heel fillet. What you aim to do is increase the toe fillet alone or the toe and side fillet.

manio:
@poeschlr 
Thank you for your first review. It seems I was doing it wrong by modifying existing SOIC-16W_5.3x10.2mm_P1.27mm. I drew it according to the manufacturer datasheets. Fillet is now properly aligned. Please re-review.

manio:
@chschlue, @poeschlr 
Hmmm - indeed it looks very similar, but in my footprint the pads are exactly the same size (wider) and coordinates are as in the original Maxim datasheet, in JEDEC pads are also a little thicker.

I have no idea what are the rules about include similar footprints, but keeping in mind @poeschlr notes about toe/heel fillets, then maybe it would be relevant to have an FP with exact Maxim dimentions...
I just have done my best to make it exactly the same as in the specs and want to share this with community. It is up to you guys what would you do with this PR...

chschlue:
Maxim even says "MEETS JEDEC MS012" in the outline drawing (missed that before) and dimensions match MS-012G exactly.
I would strongly prefer adding this under a manufacturer-specific name and sticking with the script generated FP as default for DS2482 and DS2408 (assuming you are going to add these symbols).

manio:
@chschlue 
It's ok for me but first I need to read a documentation how to contribute this way as I didn't use a kicad-footprint-generator yet... So you mean adding the DS2482 and DS2408 definitions to https://github.com/pointhi/kicad-footprint-generator and using already existing MS012G as a base package?

chschlue:
Not quite.
If you feed Maxim's package dimensions to the generator, you will end up with the same SOIC-16 that's already in the lib.
If you want Maxim's suggested footprint instead of the generated, IPC 7351B compliant one, doing it by hand is OK, I'm just suggesting you give it an MF-specific name (like Maxim_S16_5-16) instead of a generic SOIC-16.

manio:
I like this idea. PR fixed according to @chschlue comments: added under manufacturer-specific name and code.

poeschlr:
I am not really in favour of including such footprints if there is not really a good reason for it. Does maxim specify why their suggested footprint would deviate from the industry standard? If they do not give a very good reason then i will veto the addition of this footprint. (Yes we where lenient in the past about this but we now lay much more trust into ipc footprints)

manio:
I see
No, maxim did not provide such info (at least I cannot find it in the schematics).

I was also planning to add similar Maxim_D6_1_6.kicad_mod after merging this one. Do we have something similar already?
https://pdfserv.maximintegrated.com/land_patterns/90-0321.PDF

poeschlr:
You linked the suggested footprint. This is of no use for finding footprints that are generated using package data. Link the document that has dimensions for the package itself and check if a footprint is already generated for it. If not use the IPC generator to make one.

manio:
Ok - I see your point.

@chschlue 
You are telling it is exact size as MS-012G. I can see that the specs for JEDEC MS-012AB are in
https://www.analog.com/media/en/package-pcb-resources/package/pkg_pdf/soic_narrow-r/r_14.pdf
and for MS-012AC are in 
https://www.analog.com/media/en/package-pcb-resources/package/pkg_pdf/soic_narrow-r/r_16.pdf

But do you know what is the link for MS-012G?

chschlue:
AC is the variant (meaning 16 pin SOIC), G is the current revision of [MS-012](https://www.jedec.org/system/files/docs/MS-012G-01.pdf).
As far as I can tell, the only reason the Maxim footprint deviates from 7351**B** is that Maxim's suggestion is based on 7351**A**.
Like I already said at the start of this discussion, I see no real point in adding this, all I'm sure about is that it shouldn't be called SOIC-16.

The case is different for Maxim's D6 package you mentioned as that one actually seems to be a manufacturer specific J-lead package.

manio:
Thank you.

PR #1936, opened 2019-10-25
Create RJ25_Wayconn_MJEA-660X1_Horizontal.kicad_mod

Pwuts:
Create RJ25_Wayconn_MJEA-660X1_Horizontal.kicad_mod

Added the footprint of a common RJ 6P6C (RJ12/RJ18/RJ25) connector

image
(grid size is 1 mm)

Datasheet: [Wayconn MJEA-660X1XXX](https://www.wayconn.com/rj25-6p6c-pcb-mount-ra/)

I mentioned this earlier [in a comment](https://github.com/KiCad/kicad-footprints/issues/301#issuecomment-545521010), this connector is [widely](https://nl.aliexpress.com/item/32767485682.html) [available](https://nl.aliexpress.com/item/32948988244.html) [online](https://nl.aliexpress.com/item/32826457723.html).
In most places it's just called "90 degrees RJ12 6P6C connector", some sellers call it YT2038B (Black) or YT2038Y (Yellow) but Wayconn has a product with matching specifications called MJEA-660X1XXX, see link above.

Example of seller-provided drawings ([of this product](https://nl.aliexpress.com/item/32767485682.html)):
image

I also do not have a 3D model of this connector.

~Does anyone have suggestions what to name it or is it good enough this way?~

---


gmsotavio:
Hi @Pwuts This is your first PR. Thank you for contributing to KiCad!

Usually we do not encourage the contributors to submit a PR containing AliExpress or eBay stuff, because the lack of reliable documentation, or the worst, the lack of a manufacturer part number (How will other users find this footprint without a solid reference (manufacturer part number)? Fortunately, there is a common reference number for this component (**YT2038Y**), so if no other librarian opposes, I will accept your submission.

Some questions and general remarks:

- What the letter **A** from RJ12_A, mean? I think the correct name for this footprint is `RJ12_YT2038Y_Horizontal`. The **Horizontal** word is part of Connectors convention (See F3.6 http://kicad-pcb.org/libraries/klc/F3.6/)
- Which symbol do you intend to use with this footprint?
- For connectors, the courtyard clearance must be 0.5mm (KLC F5.3 - http://kicad-pcb.org/libraries/klc/F5.3/)
- You must supply the 3D path/filename, even it doesn't exists. If someone adds the 3D model later, we don't need to change the footprint again. The 3D shape path/name should be `${KISYS3DMOD}/Connector_RJ.3dshapes/RJ12_YT2038Y_Horizontal.wrl` (See KLC F9.3 - http://kicad-pcb.org/libraries/klc/F9.3/)
- Provide a more technical and accurate description of your footprint. "_RJ12 connector commonly available on the interwebs_" is too generic.
- Please, check the dimensioned drawing bellow, and fix the wrong dimensions

image

poeschlr:
I am not a fan of including "generic" connectors for these unless it is shown that there are compatible parts available from multiple reliable sources (A reliable source can only be an entity that links an exact part number to reliable documentation that is available directly via the manufacturers website. At least one such exact part number must then be listed in the description as an example such that users can then find one guaranteed compatible part.)

gmsotavio:
@poeschlr Should we close this PR? Finding a reliable source for AliExpress and eBay stuff is almost impossible.

Pwuts:
@gmsotavio Thanks for the feedback! I am a bit confused now though. As you mentioned, the common reference for this part number is YT2038Y, but now you say
> Finding a reliable source or compatible part for AliExpress and eBay stuff is almost impossible

This is the most common RJ12 6P6C socket found on those websites, it only takes one search to find that out. I understand not all sellers name it the same, but the part is still available and not very hard to recognize imo.
Indeed it is hard to find the original manufacturer specification, I haven't been able to either, but the specifications provided by some stores do match.

poeschlr:
> As @gmsotavio mentioned, the common reference for this part number is YT2038B, so there actually is a specific part with specs to which this connector is made.

Then first step: rename your footprint to include this part number. Second step find a document that links this part number to a drawing.

gmsotavio:
Hi @Pwuts,
> Wayconn mentions MJEA-660X1XXX instead of YT2038XX but it has the same specs. Would this qualify as a reliable source?

I've checked the documentation and I agree that the specs are the same. Visually YT2038XX looks like an MJEA. Please, change the footprint name according to and replace the old documentation link to get this merged. Thanks to provide a more realiable documentation.

Pwuts:
Ok. :)

One last question: this is a 6P6C jack, so it can be used for RJ12 [but also](https://en.wikipedia.org/wiki/Registered_jack#Registered_jack_types) RJ18 and RJ25. What should I do with this information? Technically, 6P6C would be a better name than RJ12 but RJ12 is probably more common.

gmsotavio:
Hi @Pwuts,

To be honest, I'm not used to using this kind of connector, but I believe you are right. They are 6P6C Registered Jacks. I vote to follow the datasheet definition (in this case RJ25). Add "RJ25" as a tag.

image

Pwuts:
I think I'm done :)

gmsotavio:
Hi @Pwuts :)

I think your footprint is good. I did a final check and realized that three minor cosmetic changes are needed.
- [ ] Center the RefDes (Silk and Fab). By my calculations, the correct X coordinate is (-2.55).
- [ ] It is preferable the link of the datasheet in pdf format rather than the product page.
- [ ] Change the height of the chevron (Silk and Fab) to 1 mm.
image

These are my considerations.

gmsotavio:
The issues found in the footprint are acceptable for a first-time contributor. Thanks for making the changes pointed out and for your contribution to the KiCad libraries.

No more comments. Merge.

PR #1937, opened 2019-10-26
Add Diodes Inc PowerDI5 package

MatthewStephenson:
Add the PowerDI5 package from Diodes Inc, used for a range of power diodes.

Datasheet: https://www.diodes.com/assets/Package-Files/PowerDI5.pdf

image

I don't have a 3D model, sorry.

Rule 6.2 will fail because the footprint origin is 6 micrometres off - however I was following datasheet dimensions for the pads and outline:

image

---


poeschlr:
Please increase the clearance between the pad and the silk outline (right now silk is touching the pad. No manufacturer will print the silk line if it is this close resulting in the polarity mark being removed.)

Add more details to the description

The dimensions look good except for the courtyard clearance on the right side.
Screenshot from 2019-10-26 20-34-13

poeschlr:
The courtyard clearance on the left side is wrong. On the right side i would say it is wrongly rounded (0.248mm clearance would be the nearest when rounded to 0.01mm grid)

I should have been more precise regarding silk. The clearance must be at least the same as the line width but i would suggest to use 0.2mm as it allows cheaper processes to be used for producing a board with this footprint.

Screenshot from 2019-10-27 20-28-32

MatthewStephenson:
Thanks for the guidance. I havent't done one of these for a while so I'm a bit rusty, and appreciate the help.

I think I am clear on the silkscreen now - I had overlooked the recommended 0.2mm which is in KLC.

I'm still struggling with the courtyard clearances though. KLC says "clearance around the combined component body and land pattern boundaries ... 0.25mm". As it doesn't mention silkscreen, I'm assuming that silkscreen can be outside the courtyard - is that right? If the 0.25mm courtyard clearance needs to include the silkscreen, I'll move it again.

Here are my calculations for the courtyard - maybe you can tell me where I'm going wrong:

Left side:
Pad 1 centred at X=-1.12mm, width 4.86mm gives the left side at X=-3.55mm.
Add 0.25mm clearance gives courtyard line at -3.80mm. This overlaps the silkscreen as per above.

Right side:
Pads 2 centred at X=2.862mm, width 1.4mm gives the right side at X=3.562mm.
Add 0.25mm clearance gives courtyard line at 3.812mm
Needs to be on a 0.01mm grid, so I rounded up to 3.82mm to avoid being under the required 0.25mm clearance. Should I have rounded to the nearest (3.81mm) instead?

poeschlr:
Yes silk can be outside of the courtyard area (but is typically at most partly outside because 0.2mm + 1.2/2 is 0.26 which means that it is only outside by 0.01mm)

regarding rounding: we typically round to the nearest grid point. So the resulting clearance can then be between 0.245mm and 0.255mm
As a tip set the pad clearance to 0.25mm temporarily and the grid to 0.01mm. Zooming in then helps to see which grid line is closest to the indicated clearance line.

MatthewStephenson:
Is there anything more I need to do with this footprint?

poeschlr:
looks good now, thanks

Issue #1009, opened 2018-10-10
PCB edge connector footprint datasheets


evanshultz:
As noted at https://github.com/KiCad/kicad-footprints/pull/996#issuecomment-428649751, none of our footprints have a datasheet attributing their design to a standard. That needs to be added and any dimensional changes to adhere to that standard made.

herostrat:
In addition to this I think it would be worthwhile to define how edge connector should be defined or if there are rules regarding them.

If you look into the PCIexpress one:
![grafik](https://user-images.githubusercontent.com/32066572/49819997-0ca6e580-fd77-11e8-937c-10810b36f51c.png)

Imho the silkscreen is wrong on the button side and should be the same as in the others because it should not produce a silkscreen on the cuts on the final pcb
![grafik](https://user-images.githubusercontent.com/32066572/49820054-3233ef00-fd77-11e8-812f-e0c5fb97dc15.png)

Also, is there a reason the pads are not rounded in that case?

I plan on doing the m.2 board connectors. 
Before I start I wanted to define how they should be done.

herostrat:
I agree, the edge cuts should be defined properly in the footprint.
But isn't this not possible until probably v6? I thought 5.1 would not include this option.

Otherwise I also think making templates for standard boardsizes is also very useful and sufficient for most cases (https://bugs.launchpad.net/kicad/+bug/1808011)

dlharmon:
Here's what I've been using since the 4.x days, seems to work fine with 4.05:

![screenshot from 2018-12-13 14-14-52](https://user-images.githubusercontent.com/6054322/49967939-835ff200-fee1-11e8-8501-40cca6c21dc2.png)

These are generated with this:
https://github.com/dlharmon/footgen/blob/master/examples/pcie.py

[pciex1fingers.zip](https://github.com/KiCad/kicad-footprints/files/2677846/pciex1fingers.zip)

poeschlr:
It is possible to define edge cuts. It however requires a bit of blasphemy (in other words a text editor)

In v4 it was even worse as opening a footprint with edge cuts lines meant they where automatically moved to the silk layer. Now they should at least stay on the edge cuts layer. (Maybe we can get the devs to allow a better workaround for 5.1 than using a text editor. Just help me out on this bug report: https://bugs.launchpad.net/kicad/+bug/1251393)

PR #1938, opened 2019-10-26
Adding PCB Edge PCI Express connectors

ki5libs:
Adding PCB Edge PCI Express connectors:
x1
PCIe_x1
x4
PCIe_x4
x8
PCIe_x8
x16
PCIe_x16

[Datasheet - PCIe Standard](http://www.ritrontek.com/uploadfile/2016/1026/20161026105231124.pdf#page=70)

---

---


gmsotavio:
Hi @ki5libs. I'm not reviewing this. I believe would be acceptable to move the edge outline from Dwgs.User to Edge.Cuts layer. Currently, KiCad does not provide this feature in the user interface, so we need a workaround here. Use an text editor and replace Dwgs.User by Edge.Cuts to do the trick.

image

For reference, there is a previous discussion about that.

- https://github.com/KiCad/kicad-footprints/issues/1009
- https://forum.kicad.info/t/mid-mount-connector-footprint-using-edge-cut-solved/5791/15
- https://bugs.launchpad.net/kicad/+bug/1251393

ki5libs:
Hi @gmsotavio. Got it. In this case sign "PCB Edge" on the `Dwgs.User` makes no sense and should be removed, shouldn't it? But what do you think about another one with PCB Thickness? Is it OK?

poeschlr:
Currently, there is a PCIexpress footprint in the lib. My assumption is that it is the same as the x1 of your contribution. If that is the case then remove the old footprint.

---

The current footprints dimensions look good.
Drawing of x1, all others checked directly in kicad (no nice drawing created)

Screenshot from 2019-10-28 20-40-09

gmsotavio:
> Hi @gmsotavio. Got it. In this case sign "PCB Edge" on the `Dwgs.User` makes no sense and should be removed, shouldn't it?

Yes

> But what do you think about another one with PCB Thickness? Is it OK?

I would keep the PCB Thickness comment. But my suggestion is to move this comment from Dwgs.User to Cmts.User.

ki5libs:
@gmsotavio, @poeschlr, all changes asked by you are made.

poeschlr:
Thanks

PR #1946, opened 2019-10-29
Add DFN-4

DanSGiesbrecht:
Used for Sensirion SHTC3

~~Made using the footprint generator: https://github.com/pointhi/kicad-footprint-generator/pull/450~~

---

image

image

---

From [datasheet](https://www.sensirion.com/fileadmin/user_upload/customers/sensirion/Dokumente/0_Datasheets/Humidity/Sensirion_Humidity_Sensors_SHTC3_Datasheet.pdf):

image

image

---


DanSGiesbrecht:
Updates:
- Implemented manufacturer-recommended footprint (rather than scripted IPC footprint)
- Renamed footprint to: `Sensirion_DFN-4-1EP_2x2mm_P1mm_EP0.7x1.6mm`
- Moved footprint to: `Sensor_Humidity.pretty`
- Renamed 3D model path: `${KISYS3DMOD}/Sensor_Humidity.3dshapes/Sensirion_DFN-4-1EP_2x2mm_P1mm_EP0.7x1.6mm.wrl`
  - **The 3D model has not been submitted**

DanSGiesbrecht:
I've changed the pads to rectangular, to align with https://github.com/KiCad/kicad-footprints/issues/1798

@poeschlr , rather than all footprints square, do you think it makes sense to have the thermal pad be square (as it is 1:1 with the component slug) and the leads be rounded? If so, I'll do that.

poeschlr:
In the fixes for #1798 we decided to keep even the center pad with rounded edges but chose to use the same radius as for the normal pads. (As the drawings in a lot of datasheets look like this is the case)

I leave it up to you what direction you go.

poeschlr:
I would say the courtyard should be relative to the copper pad not the paste pad.

Screenshot from 2019-11-09 16-15-36

DanSGiesbrecht:
Thanks - made the suggested changes:
- Pads have been reverted to round-rect (I prefer this)
- X-axis courtyard is now relative to the copper, rather than solder paste

As an aside: I know I've asked this before, but how do you generate the dimensioned drawings? They're very handy!

chschlue:
https://forum.kicad.info/t/how-to-check-footprint-correctness/9279

poeschlr:
The drawing shown here is made slightly differently to how i wrote in the tutorial as the drawing dimensions workbench no longer works with my freecad version.
I used techdraw to export a dxf and added the dimensions with librecad.

DanSGiesbrecht:
@poeschlr thanks for the tip!

This should be ready to merge.

chschlue:
@poeschlr Out of curiosity, what's wrong with TechDraw dimensions?

poeschlr:
@chshlue Techdraw in freecad 0.18 has no way to reference a center point of a line. It also has a very bad user interface for adding dimensions (i need to select two points, click the add dimension tool, place the dimension, repeat.) 

And it is generally quite unstable. For example changing scaling, line widths, ... all only works sometimes. Luckily with dxf export i do not care about these as librecad does everything for me.
Maybe the 0.19 techdraw workbench will be better especially once 0.19 is a true release. (right now it is a dev snapshot)

chschlue:
I see.
Much of that has gotten a lot better with 0.19. First and foremost the center point issue which was a complete show-stopper.
Have you tried the 0.19pre AppImage? It at least brings its dependencies with it.

DanSGiesbrecht:
@poeschlr can this be merged?

poeschlr:
Thanks

PR #1945, opened 2019-10-28
Add Yuji 5730 LED footprint

MatthewStephenson:
This LED package is used for several of Yuji's high-CRI LEDs.

Datasheet (of a representative LED): https://cdn.shopify.com/s/files/1/0344/6401/files/YJWJ014-1.1_YJ-BC-5730L-G02.pdf

image

---

- [x] Provide a URL to a datasheet for the footprint(s) you are contributing
- [x] An example screenshot image is very helpful 
- [x] If there are matching symbol or 3D model pull requests, provide link(s) as appropriate
- [x] Check the output of the Travis automated check scripts - fix any errors as required
- [x] Give a reason behind any intentional library convention rule violation.

PR #1948, opened 2019-10-30
Correct polarity of Yuji 5730 led

MatthewStephenson:
Realised I'd got the polarity wrong with respect to the LED symbol, so swapped the pad numbers.

Is there a reason the LED symbol has the anode as pad 2, and the cathode as pad 1? It seems counterintuitive - I'd expect the positive connection to come first.

Datasheet: https://cdn.shopify.com/s/files/1/0344/6401/files/YJWJ014-1.1_YJ-BC-5730L-G02.pdf

image

---

### DIFF
--- a/LED_SMD.pretty/LED_Yuji_5730.kicad_mod
+++ b/LED_SMD.pretty/LED_Yuji_5730.kicad_mod
@@ -1,4 +1,4 @@
-(module LED_Yuji_5730 (layer F.Cu) (tedit 5DB9E537)
+(module LED_Yuji_5730 (layer F.Cu) (tedit 5DC5BBBC)
   (descr LED,Yuji,5730,https://cdn.shopify.com/s/files/1/0344/6401/files/YJWJ014-1.1_YJ-BC-5730L-G02.pdf)
   (tags "LED Yuji 5730")
   (attr smd)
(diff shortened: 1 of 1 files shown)


gmsotavio:
A remark: Two terminal footprints should be aligned with pin 1 on the left side (http://kicad-pcb.org/libraries/klc/F4.2/). I'm not reviewing this right now.

cp-aquila:
> 
> 
> Realised I'd got the polarity wrong with respect to the LED symbol, so swapped the pad numbers.

Just to be sure, which LED-Symbol are you refering to? Just `LED`?
> 
> Is there a reason the LED symbol has the anode as pad 2, and the cathode as pad 1? It seems counterintuitive - I'd expect the positive connection to come first.
> 
Because it is the same as a diode. Check the `D` symbol.

One more question: Is this actually a correction? It seems like an addition to the current library.

MatthewStephenson:
Yes, I was referring to the `LED` symbol. If this is the case becuse of the `D` symbol, then why does the `D` symbol have the anode as pad 2? I'm sure there's a good/convention reason - I'm just curious as it tripped me up.

I realise I need to rotate the symbol so that pin 1 is on the left - I will try to do that in the next few days.

Regarding the correction question: yes, it is a correction. The original commit is [here](https://github.com/KiCad/kicad-footprints/commit/30efc63cb535cf466d398f1feb0ef39bbafe9e0d) and the original PR is [here](https://github.com/KiCad/kicad-footprints/pull/1945). I made the correction on the same branch as the original addition, but I now see github is telling me there are conflicts. I guess the best solution is to close this PR and do the correction on a new branch branched off master?

cp-aquila:
It seems like the library is following IPC: 
**Cathode is Pin1** (http://pcbget.ru/Files/Standarts/IPC_7351.pdf#page=76)
grafik
grafik

Then there is also the issue that not all manufacturers seem to mark the cathode:
https://blog.screamingcircuits.com/2013/01/how-not-to-mark-a-diode.html

So fix the roation of the footprint (remember to reset te rotation value for pads) and you should be good for this part.

Regarding the merge-conflict: one option is to rebase this branch on current master, then force-push. That should give a clean history.

MatthewStephenson:
Thank you for educating me, I suspected there would be a standards document behind this somewhere. I guess it does make a certain amount of sense: 

- All ICs have Pin 1 marked
- Diodes have the cathode marked
- Therefore diodes' cathode is pin 1?

I will rotate the symbol as you suggest, and check the pad rotation.

cp-aquila:
> 
> 
> Thank you for educating me, 

I actually had to educate myself first 😄 
Now we are both a little bit wiser^^
 
>
> I will rotate the symbol as you suggest, and check the pad rotation.

Another way to get this done is to just invert the x-coordinates (within the footprint editor, or using a text editor). Thats basically a rotation of 180deg.

MatthewStephenson:
Swapped the x-coordinates as suggested. Realised in the process that my last half-arsed attempt was hopelessly rubbish, as I'd also forgotten to move the polarity markers. I think it's right now.

image

poeschlr:
thanks

PR #1954, opened 2019-11-05
Added footprint for BelFuse SI-60062-F.

cnieves1:
Datasheet: https://belfuse.com/resources/drawings/magneticsolutions/dr-mag-si-60062-f.pdf

imagen

imagen


cp-aquila:
Hi,

just a few quick thoughts, no real review:
- [ ] Which symbol is this for?
- [ ] Why did you name the pins R1.. L1..? Other symbols have just numbers: `Pulse_JXD6-0001NL`
- [ ] Pin13 should be `SH` instead.
- [ ] Then Pin1 markers on Silk differs from similar footprints (`RJ45_Amphenol_RJHSE538X`)

cnieves1:
This is a RJ45 connector with integrated magnetics and LEDs (L1-L4 are LEDs pins). It is similar to Amphenol part number referenced below, but with different dimensions.

Pin numbering and pin 1 marker use the same criteria as:
  https://github.com/KiCad/kicad-footprints/blob/master/Connector_RJ.pretty/RJ45_Amphenol_RJMG1BD3B8K1ANR.kicad_mod

cnieves1:
imagen

myfreescalewebpage:

myfreescalewebpage:
@cnieves1 would like to start the review here and thinking we may have to review the symbol after.

- [ ] Pin name is not so clear. I propose to change names T1-T8 simply "1" to "8". Then pin 13 should be called "SH" (shielding). OK to keep L1 to L4 like that.
- [ ] The size of the mounting holes should be 3.25mm (the max value)
- [ ] Pad for L1-L4 are too small. Hole=1.1mm and diameter=2.1mm is good.
- [ ] F.Fab: The design is not horizontally centered. I have measured 2.345mm from left mounting hole to left side and 2.355mm from right mounting hole to right side. To be moved by 0.005mm (even if it's very small !)
- [ ] F.SilkS: We usually use 0.11mm clearance with F.Fab but 0.1mm is allowed.
- [ ] F.CrtYd: should have 0.5mm clearance from pad/F.Fab lines. To be checked because I have at least measured 0.6mm at the bottom

Joel

myfreescalewebpage:
Thanks, great job !!! No more comment, merging.

PR #1956, opened 2019-11-06
LIS2DH footprint with nominal density

evanshultz:
The existing footprint wasn't generated with nominal density. This one is, as our normal library parts should be.

As discussed at https://github.com/pointhi/kicad-footprint-generator/pull/315. This came from removing the IPC density parameter in the YAML file but not changing the package parameters at all.

Since that hopefully captures everything I haven't included a screenshot.

---

---


poeschlr:
I think the reason i used the high density option is because there is nearly no clearance between the pads with the nominal setting. (~0.05mm clearance to be precise)

evanshultz:
That would be reasonable, and I have seen you give 0.2mm as the target clearance for footprints elsewhere. So I guess you would like to target that clearance? However, the part itself has 0.15mm clearance between adjacent pins on the same side so that wouldn't be possible.

Moreso, I think the script has an issue. https://github.com/pointhi/kicad-footprint-generator/blob/master/scripts/Packages/ipc_definitions.yaml doesn't include a LGA entry. That IPC info is used for all footprints created by https://github.com/pointhi/kicad-footprint-generator/blob/master/scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON/ipc_noLead_generator.py. However, IPC 7351B page 21 only has a single density (B) for LGA packages so setting the density shouldn't be possible for an LGA footprint. And it should be OK with this LGA density because all pad size adjustment from the pin (just called 'periphery' instead of 'fillet') have a 0.00 adjustment. So the pads should be the same size as the pins on all LGA footprints. Please confirm if this is an oversight in the script. And if I'm right, do you want to fix it?

So if the script is corrected, a proper IPC footprint would only have 0.15mm clearance between pads. I would think a user should be aware if they're using dense/complicated packages, like LGAs, they need to know what they're doing and not assume it will work with a sloppy board vendor.

poeschlr:
The 0.2mm clearance is between the exposed pad and the outer pads and comes from IPC. I kind of doubt that 0.2mm will be reachable with this part.

Hm i will need to double check IPC then as i did not notice an LGA definition in the past. Will take a look at that in the next few days and come back to you. (If not send a reminder as i am currently quite busy finishing my theisis. Or rather I should be, somehow nearly everything other than writing it is more interesting.)

poeschlr:
The thing is that what most manufacturers call a land grid array does not fit what IPC calls an LGA. See page 76 point 14.1.4 I think the most fitting for such packages is dfn/qfn pullback as shown on page 85 15.4. And the pullback variant fits the lga suggestion in nominal (but also defines least and most) so even if we rebrand it as lga there will not be an increase in clearance.

evanshultz:
OK. That's fair, but I don't see what the difference should be (ignoring the courtyard). Pullback leads under a QFN or LGA, both from the IPC doc images, have pads totally under the part body. And if our generator is using the IPC docs, is it confusing to have such a mismatch between the package name in the datasheets and the IPC doc package type? Perhaps a comment in the IPC YAML file or in the footprint descriptions would help? Or maybe you prefer to leave it as is.

If the fillets are zero, which is QFN pullback level B or LGA, won't the pads be the same size as the pins? And therefore also have 0.15mm clearance?

I found this footprint used by two symbols. Then I looked at https://www.st.com/resource/en/datasheet/lis2hh12.pdf and http://kionixfs.kionix.com/en/datasheet/KX112-1042-Specifications-Rev-6.0.pdf and it appears they packages match. Just a confirmation here since these packages can have minute differences.

I see, with nominal density, 0.125mm clearance at corner pins (along a side it is 0.15mm). The existing footprint has 0.15mm clearance in the corner, so essentially the same. And both well below 0.2mm. So are you OK to merge this? I will update the symbols affected to use the new footprint if so.

PR #1958, opened 2019-11-08
Create LEM_CKSR.kicad_mod

almghandi:
Push of
LEM_CKSR
https://www.lem.com/sites/default/files/products_datasheets/cksr_series.pdf

Footprint with 3D model linked for the current transducer of the LEM CKSR series.

This is my first time trying to contribute to KiCAD and using git. I tried to follow all the Conventions, please let me know if I did something wrong. The corresponding pull request of the **symbol** is here: [https://github.com/KiCad/kicad-symbols/pull/2291](https://github.com/KiCad/kicad-symbols/pull/2291)

Edit:
~~And the footprint pull request for the **3D model** here: [https://github.com/KiCad/kicad-packages3D/pull/615](https://github.com/KiCad/kicad-packages3D/pull/615)~~

Edit2:
Outdated screenshot removed


almghandi:
The footprint should now fullfill the KLC, with the exeption of the Pin 1 rules from F7.2 and F7.3 as there is no pin 1:
image

So I applied the rules to the next lowest pin, which is pin 2. Any better suggestions?

Here is an updated screenshot:
image

cpresser:
This took more than one year. Thanks for being patient. We are kinda understaffed here.

One big issue for compliance is the missing Pin1.
But there are other `CKSR` devices with Pin1. I suggest we follow this datasheet.
https://www.lem.com/sites/default/files/products_datasheets/cksr_75-np.pdf
It is from the same series and fits all the variants. One could even make a new symbol for the 75Amp variant that uses the same footprint.

- [x] Body outline F.Fab is good
- [x] Courtyard is good. 
- [x] Silkscreen position and clearances are good.
- [x] Please add Pin1 and Pin 10
- [ ] Pin 14 y-position is off (and I am being pedantic here). My calculation gives `(12.7 - 3*1.905)/ 2 = 3.4925`. Please check all pins in that row (14..18). Interestingly enough the datasheet for the series contradicts itself. 3x1.905 is not 5.72. The _75 variant datasheet has that fixed.
- [ ] There is no Pin1-Marking on F.Silk. How about something like this?
image
- [ ] Keywords should include the 75A version and also that datasheet.

One question on the side, I saw you used this for a solar-inverter. Is that a open design? I was thinking of building one on my own and could use inspiration.

almghandi:
@cpresser Thank you for the detailed review. I implemented all your points. Here is the updated screenshot:
Screenshot from 2020-08-26 01-21-54

Yes I used it for a solar-inverter prototype, which was part of my Master thesis. It is not published as an open design and uses a daughter board for the gate drivers, which was not designed by me . But I could share my part of the project with you, if interested

cpresser:
Looks good.
I don't have the time to re-review the next few days. This will most likely have to wait until Sunday.

And yes, I would be interested in looking at any DIY design. I am just starting with that topic and have much to learn.

cpresser:
It looks like something went wrong when moving the part:
image

That x-distance should be 9.17mm.
Also the silkscreen and F.Fab seem misaligned.

My workflow to prevent such move-errors in this case would be
* Add pads 1 and 10
* Change silkscreen by editing the two lines in their properties dialog. You can enter simple math in those boxes. So to move one pint just add `-2.54` to the respective x-coordinate.
* Switch grid to 100mil, set new anchor on pin 1.
* Profit

almghandi:
@cpresser I corrected the misaligned silkscreen. Next time I will use your trick with a new anchor, thanks!
There is another contradiction in the two datasheets: 
Screenshot from CKSR_75, here we get 7.62+2.54=**10.16**, while in the other sheet a value of **10.17** is shown. I decided to follow the CKSR_75 datasheet, as it also has the correct pin14 values.
Screenshot from 2020-08-30 15-24-10

For the x-distance I get the value of 9.25mm with the following calculation:
3.63+15.78-4*2.54 = 9.25
How did you get to 9.17mm?

cpresser:
I also see the discrepancy between 10.16 and 10.17.
For the distance of Pin1 to the edge of the part I just added `7.62 + 1.55 = 9.1`. But your calculation also seems correct.

The data-sheet is just not good. I also found this:
image
Which results in `15.78 + 3.63 - 14.26 = 5.18` which should be 5.08

Since the overall linear tolerances are 0.25mm I think we should not invest to much time into getting the best values. On top of that the datasheet contradicts itself.
The pins are 1mm and the holes are 1.3mm. That is plenty of play to make it fit even if we are off by 0.1mm. I will merge it as it is.

Sorry about all the fuzz. I try to be as exact as possible. But in this case this all is in vain given the datasheet.

PR #1959, opened 2019-11-09
Add SOT-363

chschlue:
WIP

SOT-363_SC-70-6

https://github.com/pointhi/kicad-footprint-generator/pull/460

---

---


plaes:
It seems that you removed the `SC-70-6` from tags and description.
And while you're at it, can you also add `SC-88` (this terminology seems to be used by ON Semiconductor) into tags and description as well.

evanshultz:
@chschlue 
Are the other packages exactly the same? Seems a bit silly to have multiple footprints for compatible packages, but if the dimensions used by JEDEC for SOT-363 don't match other compatible packages then we'll have to decide what to do.

chschlue:
@evanshultz Oops. I kinda forgot about this PR and can't seem to find the discussion that led to it right now.
Call it a request for comments.

Problems I see:
1. You don't have to dig deep to find... variations... among SOT-xxx, compare https://www.nxp.com/docs/en/package-information/SOT523-1.pdf to https://www.diodes.com/assets/Package-Files/SOT523.pdf, for example. We probably have lots of SOT FPs that should actually have MF prefixes, OTOH if there's a JEDEC spec availabe, the lib should also contain a vendor-neutral FP.
2. I don't have access to IPC-7351 but AFAIK it contains explicit exemptions for SOT-23 and the like (with the almost square pads we all know and love). I'd like to know what it actually says there; this one is basically a gullwing package but as the screenshot above shows, heels and toes resulting from the usual calculations seem to be excessive for packages that size.

evanshultz:
1. I was thinking differences between JEDEC, JEITA, etc. for packages which are considered compatible by the industry. Vendor datasheets definitely vary in dimensions even if they state adherence to some governing body's package. And note that NXP uses their own nomenclature which you can't compare with any industry-standard package name. By 'MF prefix' did you mean 'MO'? I'm in favor of a single footprint, or a combo footprint if it comes down to it, but we'd have to see how compatible the governing body packages drawings are.
1. I'm looking at IPC 7351B and I don't see anything of the sort. Do you have more information to what you're referring? The SOT23 package is in section 8.6 on page 58. Table 3-2 on page 16 discusses gullwing packages, and it lists applicability for SOT packages described in section 8.6 (among others). But I see there's a note under the table for packages where Smin is <= Amax to use different heel fillet goals. Perhaps that would affect these footprints? In addition, it may be that Level B (the default) is just too much for these smaller packages even though it works fine for bigger SOIC packages. It may be more inline with what users expect to default to Level C density for physically smaller packages. I supported varying densities for chip parts at https://github.com/pointhi/kicad-footprint-generator/pull/439 and so something similar could be done for the gullwing generator script as well.

You requested comments, so there are mine. :)

chschlue:
1. I know my example was extreme, but still I'm pretty sure it's not to hard to find clashing SOT-xxx packages even if you disregard NXP's nomenclature.
I have yet to read up on the v6 file format drafts but FP aliases or hierachies might help clean up the current mess (getting rid of FP names like SOT-363_SC-70-6_SC-88, especially if nobody has ever actually verified JEITA specs)
2. I thought I had seen this
SOT-23
actually codified somewhere. Density variations don't explain it.
No, I don't have more information right now. Guess I'll buy a copy when I'm back from vacation.

cpresser:
> It seems that you removed the `SC-70-6` from tags and description.
> And while you're at it, can you also add `SC-88` (this terminology seems to be used by ON Semiconductor) into tags and description as well.

There are even more names (Center of the picture is the current FP)
image

PR #1960, opened 2019-11-11
Add USB Micro-B Amphenol 10103592-0001LF Connector

dokthar:
Add footprint for the USB Micro-B connector Amphenol 10103592-0001LF

[datasheet](https://cdn.amphenol-icc.com/media/wysiwyg/files/drawing/10103592.pdf)
usb_Amph
No matching 3D model exists.
Breaks rule KLC F4.2 "Pin 1 should be located at the top left"
This connector is a reverse mount, all the pins are ordered in the opposite direction from "normal" USB Micro-B connector. I think it make sense to keep pin 1 in the top right corner (As done in footprint USB_Micro-AB_Molex_47590-0001).


jneiva08:
Hi,

You forgot to  add the default 3D (KLC F9.3)

dokthar:
latest changes:
 - Removed F.Paste Layer in through hole pad
 - Added default 3D model (F9.3)

poeschlr:
Please rotate the footprint such that pad 1 is closest to the top left corner.
Is there a reason that part of the PCB edge is on the drawings layer instead of edge cuts?

Two dimensions are off
Screenshot from 2020-03-06 22-31-55

The body dimensions and courtyard clearances are also wrong
Screenshot from 2020-03-06 22-39-20

dokthar:
I've fixed the two dimensions that were off, and rotated the footprint such that pad 1 is in top left corner.
I have no idea of what the rules are for body dimension and courtyard, here is what I've done:
For the body dimension I have tried  to follow the maximum body dimension shown in the datasheet,
for the courtyard I've put 0.5mm from the closest pad or body boundary.

poeschlr:
The KLC specifies the use of nominal so please use it not maximum.

PR #1970, opened 2019-11-20
Add generated Molex MicroLock Plus (side entry, pitch 1.25 mm)

kacper-ka:
This merge adds *Molex Micro-Lock Plus* connectors (**505567** series), 2-16 pins, side entry, pitch 1.25 mm. The footprints have been generated and they include keepout area (as recommended by the manufacturer) depicted as a hatched polygon on *Dwgs.User* layer. A center offset has been applied to match the vacuum area as designated by the manufacturer.

The 3D model is available at the manufacturer site, therefore there is no associated pull request.

### KLC violation
The *Travis* automated check scripts report violation of the **F6.2** rule:
> Footprint anchor should be placed in the middle of the component body

However the part datasheet indicates that the vacuum area is not located at the center of the component. Appropriate offset in the Y direction has been applied, resulting in the rule violation.

### Datasheet
https://www.molex.com/pdm_docs/sd/5055670271_sd.pdf

### Screenshot
**Molex Micro-Lock Plus, side entry, 5 pins, pitch 1.25 mm**
Screenshot from 2019-11-20 13-45-02

---

- ~~If there are matching symbol or 3D model pull requests, provide link(s) as appropriate~~


kowlown:
@kacper-ka Cool! I would love to have it already by default

PR #1971, opened 2019-11-21
Add Sumida CDRH103R Power Inductor footprint

timtashpulatov:
Add Sumida CDRH103R Power Inductor footprint

Datasheet: https://products.sumida.com/products/pdf/CDRH103R.pdf

image

image

3d model pull request: https://github.com/KiCad/kicad-packages3D/pull/618


timtashpulatov:
I must confess I am slightly at a loss here, with "Library files must use Unix-style line endings (LF)" warning. Footprint in question is created with KiCad under Windows, and file uploaded to github. I would expect .gitattributes to take care of CRLF to LF conversion on github's side but this apparently is not the case.

chschlue:
Github doesn't mess with what is pushed.
It's your responsibility to commit LF endings.
(By setting core.autocrlf for example.)

timtashpulatov:
> Github doesn't mess with what is pushed.
> It's your responsibility to commit LF endings.
> (By setting core.autocrlf for example.)

Thanks **chschlue**, I have set core.autocrlf to true, that does not seem to help.
I've also tried putting .gitattributes into the folder I'm commiting to, also to no avail.
Neither uploading via web nor commiting via git work; there must be something very obvious I am missing.

(Local files produced with Windows version of KiCad are OK to contain CRLF as I get it. Most official footprint files have CRLF endings in them)

timtashpulatov:
Cough

poeschlr:
We will review your stuff when we have time for it. You making "noise" for sure does not increase the priority to review your stuff for me as i personally filter by "longest without activity first"

PR #1972, opened 2019-11-21
Fixed courtyard

cnieves1:
Fixed courtyard. It was not a closed polygon.


chschlue:
Courtyard was obviously wrong but what's up with the increased mask cutout?

cnieves1:
Sorry, it was not intentional.
Just loaded it into Kicad, modified the courtyard and save it.
I will revert the mask cutout tomorrow.
Any idea why Kicad did this?

poeschlr:
The circles in question should not be there to begin with. My guess is that kicad now does not allow graphical elements with 0 width. My suggestion is to remove all these circles on the mask layers.

PR #1974, opened 2019-11-21
Adding SOT-457T / SC-95 footprint

persus:
Adding SOT-457T / SC-95 footprint

image

My source of specs for this footprint:
https://www.mouser.de/datasheet/2/348/qs6k21-e-1139119.pdf
---


herostrat:
Hey persus,

this footprint would benefit from being generated via the footprint generator

This reduces maintenance in the future for the library.

persus:
Hey there.

OK. Right now I am scripting this.

First I thought if you are really serious about this. If there only was already a script so that I just had to add my package specifics into a configuration file. But no, there is no script in the first place.
Then I saw there is no script for all packages like this at all. Since I am not that familiar with Python I were not sure, if I should script this. But hey, this is a good task to learn Python. So I did some programming last weekend.

Right now I am writing a generic solution for all in principle rectangular and symmetric smd-packages with pins on both sides. Since I am not that familiar with Python I don't really know how polymorphism, injection and such techniques work. But my thought was to make a solution, that one should write specifics in a configuration file for the simple rectangular packages. And that everyone can extend or modify the solution to create other rectangular packages with some specialties like heat-vias or such. Let's see if this will work out. :)

The only thing I hope is, that there is no other guy working on this at the same time. :)

Since we get into the pre-xmas-time I can't say how long it would take, but two weeks I think would be a good guess.

chschlue:
https://github.com/pointhi/kicad-footprint-generator/tree/master/scripts/Packages/Package_Gullwing__QFP_SOIC_SO ?

herostrat:
Hey persus,
I can't quite follow, did you look at the repo I linked in my first post?

The footprint generator is exactly what you discribed, so there is no need to write a own scripting-generator.

For the typical footprint styles, e.g. SOT, it is usually only necessary to add the dimension data in a Yaml file.

Take a look there at scripts/Packages/TO_SOT_Packages_SMD and you will see what I mean.

In fact, many (if not most) of the "standard" footprint are already generated via the kicad-footprint-generator.

poeschlr:
Please do not link to the datasheet hosted at a distributor but to the datasheet on a manufacturers website

persus:
Yes, I looked here: kicad-footprint-generator/scripts/Packages/TO_SOT_Packages_SMD

And there is only a generator for DPAK packages. Did I miss something?

persus:
I found it. it‘s located here:
kicad-footprint-generator/scripts/Packages/Package_Gullwing__QFP_SOIC_SO

PR #1976, opened 2019-11-23
Connector_AMASS: add XT90

Maarrk:
Added XT90 connectors similarly to XT60
Datasheet: https://www.tme.eu/Document/d66b125cc4d03b6a25077f0febb346da/XT90H%20SPEC.pdf

XT90 male screenshot

_Note: The silkscreen has bigger corners in order to pass overlap test with pad 1, but still indicate the plug orientation_

---

---


poeschlr:
Why is the silk this far away at the chamfers? Might i suggest moving it closer and just break it up where it would be too close to the pad?

The type names are given as XT90H-F and XT90H-M so i would use that as the footprint names.
How did you determine the required drill size? The diameter is not given in the linked document.

The courtyard clearance should be 0.5mm relative to the nominal body size (=fab outline), it is 0.6mm

PR #1145, opened 2018-12-05
Add 6.35mm and speakON connectors

evanshultz:
I started developing these footprints some time ago based on https://github.com/KiCad/kicad-footprints/issues/191. While I hope everything is ready, my eyes are blearly and https://github.com/KiCad/kicad-footprints/issues/1012 makes me want to get this pushed now, gathering feedback from others, instead of further delay as they look good to me. There could be some personal preference in that how I designed them so let's start the discussion and progress toward merging these. Certainly there can be some changes or improvements to be made. This is massive number of footprints and I don't want to dimension each one, so perhaps a visual review will suffice for now? If there are "nice to have" things, like contoured courtyard, I'd rather have an issue to capture that and get these merged rather than delay.

Checklist of reviewed and fixed, if needed, series:
- [x] A
- [x] AA
- [x] B
- [x] Combo A
- [x] Combo I
- [x] Vertical Jacks
- [x] SpeakON Combo
- [x] SpeakON Chassis
- [x] STX
- [x] Slim Jacks
- [x] M
- [x] Stacking Jacks

I selected pin numbers to match the corresponding audio jack symbols (see https://github.com/KiCad/kicad-symbols/pull/1076):
- Numbers for XLR pins
- "T", "R", and "S" for barrel 1/4" / 6.35mm pins
- "G" for a dedicated ground pin
- "NO", "N", and "NC" for an unconnected switch

If a pin 1 does not exist, the origin is placed at pin "T" or "T1" or "1+". Pin 1 is in the top left or top middle for all but vertical XLRs where it looks super weird to have pin 1 at the top (it's often in the middle of a side) so I left them with the XLR right side up.

Jan made a good comment about being able to use the footprint filters at https://github.com/KiCad/kicad-footprints/issues/191#issuecomment-359935901, and I think this will work. By allowing filtering of the type of footprint to go with the symbol, with the symbols I made at https://github.com/KiCad/kicad-symbols/pull/1076, filtering by pin names and numbers does the rest.

There are too many footprints to show here all individually, but here are images of several of them. For the vertical XLRs I used Neutrik's DXF drawings and then copied fab elements to the silk layer (avoiding pads where necessary). This places silk on top of the fab lines and under the component, but that isn't disallowed according to KLC (and saved a ton of time due to the complexity of the shapes).
image
image
image
image

CI warnings are acceptable:
1. Rectangular pin 1 is not required because all are keyed (sometimes obviously so but at least by the mounting holes)
[...]

PR #1977, opened 2019-11-23
Adding Cherry MX switches with LED/Diode

ki5libs:
Existing Cherry MX footprints supplemented with versions with LED and diode
Dimensions in [documentation](https://www.cherrymx.de/en/dev.html) shown in inches, but I've used mm as in existing in the library Cherry MX footprints.
Example of 1u switch with diode:
CherryMX_Diode
Example of 1u switch with LED:
CherryMX_LED

---

---


chmorgan:
@chschlue there are a ton of new footprints in this PR. Verified that the originals match the existing parts with a few spot checks. Verified overall dimensions of a couple of parts against the data sheet. Checked that the newly added pins for a couple of the variants are in the correct place according to the data sheet. Nothing looks out of place here but thought I'd like to see what you thought.

chschlue:
Not sure what I'm supposed to think.
You can either check them one by one as time permits or put up a checklist so others can chime in.

Large additions take time. See https://github.com/KiCad/kicad-footprints/pull/1145 for example.

chmorgan:
@chschlue I'm pretty confident this one is correct, especially as its based on existing parts that don't have any dimensional issues when compared against the data sheet. Let me know if that's good enough to merge given the reviews we've done in the past few days. I'd like to avoid this one gathering too much dust and being discarded.

chschlue:
I only glanced over the DS but it looks like there's only six really distinct FPs.
Shouldn't a simple diff be enough to catch most potential problems without causing too much hassle?

PR #1984, opened 2019-11-27
Add Hirose FH48 series

evanshultz:
https://www.hirose.com/product/document?clcode=&productname=&series=FH48&documenttype=Catalog&lang=en&documentid=D49363_en

image
image

Questions:
1. Should the body outline be more complex? One can see that the body is inset between the shield pins but that dimension isn't given. I measure ~0.5mm.
1. Similar to the above, should I show the notches on the corners of the body? Those also aren't dimensioned.
1. Shall I add an area, perhaps on Cmts.User, to show the location of the flip lock in the open position?

Script PR: 

---

---


evanshultz:
Decided to update the body outline on both points mentioned above.
image
image

PR #1989, opened 2019-12-05
Add Package_DFN_QFN:QFN-20_1EP_3x3mm_P0.5mm_EP1.8x1.8mm

0xdec:
[Datasheet](https://www.silabs.com/documents/public/data-sheets/cp2102n-datasheet.pdf) (page 41)
Symbol: kicad/kicad-symbols#2364

QFN-20_1EP_3x3mm_P0 5mm_EP1 8x1

---

---


poeschlr:
Footprints for standardized packages (including QFN) must be generated with our IPC scripts found here: https://github.com/pointhi/kicad-footprint-generator/tree/master/scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON (readme for similar generator will be of help: https://github.com/pointhi/kicad-footprint-generator/blob/master/scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md)

0xdec:
@poeschlr I don't think this is a real IPC-compliant QFN footprint, as you can see from the four small corner pads, so the QFN generator would not work. Should I rename it to be something like `Silicon-Labs_QFN-20_1EP_3x3mm_P0.5mm_EP1.8x1.8mm`?

poeschlr:
I have a dislike of simply adding a manufacturer prefix and calling it quits. It does not really tell a user what to expect. Especially as i imagine that silicon labs will use a normal QFN 20 package for other parts.

A better solution would be to add a suffix to the name indicating that there are corner pads. Something like `_CornerPads` would work. In the long run adapting the script to create such a footprint might be an option but for now we really are stuck with making these by hand.

---

Additionally: the counter of normal pads vs center pad(s) is separated by a dash (or minus) not an underline.

Meaning the name would be: `QFN-20-1EP_3x3mm_P0.5mm_EP1.8x1.8mm_CornerPads`

To be honest one could even argue that this is not a QFN20 but a QFN16 with four additional corner pads so the name could also be `QFN-16-1EP-4CP_3x3mm_P0.5mm_EP1.8x1.8mm` (-nCP would be a new pad count modifier indicating additional corner pads). I however fear this would be confusing to users who read the datasheet and get informed that this is a qfn20.

chschlue:
> To be honest one could even argue that this is not a QFN20 but a QFN16 with four additional corner pads so the name could also be `QFN-16-4CP-1EP_3x3mm_P0.5mm_EP1.8x1.8mm` (-nCP would be a new pad count modifier indicating additional corner pads). I however fear this would be confusing to users who read the datasheet and get informed that this is a qfn20.

I actually think this is the most reasonable solution, followed by a manufacturer prefix.

This is far from a "real" (i.e. industry standard) QFN-20, so IMO either `QFN-16-4CP...` (standard QFN-16 plus additional features) or `Silicon-Labs_QFN-20...` (something a single manufacturer happens to call QFN-20) fits best.
The DS claims [on page 42](https://www.silabs.com/documents/public/data-sheets/cp2102n-datasheet.pdf#page=42) this is MO-220 compiant by the way (which it obviously isn't).

poeschlr:
Ok we can go with the manufacture prefix for now. If we discover more such packages then we will have a larger sample size to use for determining if the corner pads should be counted as part of the normal pads or if we should add them separately.

0xdec:
@poeschlr I just realized that this footprint already exists in the library as `SiliconLabs_QFN-20-1EP_3x3mm_P0.5mm`, so it seems I've reinvented the wheel. Closing this PR, thanks!

PR #1995, opened 2019-12-09
Added footprint for ST VL53L1x distance sensor

cnieves1:
Datasheet: https://www.st.com/resource/en/datasheet/vl53l1x.pdf
imagen
imagen
imagen


cnieves1:
Regarding F6.2: I think footprint center is correct. Travis calculation is wrong in this case due to a bigger #1 pad.

cnieves1:
3D depends on: https://github.com/KiCad/kicad-packages3D/pull/624

Misca1234:
(This is not a review, just a comment)
If you create a new directory, in this case "Sensor_Distance.pretty) you should update the file ....../kicad-footprints/ftp-lib-table with the new "entry" and include it in this push

cnieves1:
Isn't the entry already modified in fp-lib-table? I can see the added entry
in the second file at:
  https://github.com/KiCad/kicad-footprints/pull/1995/files

Misca1234:
Yes, I noticed it and removed my comment immidetly but you saw the comment anyway ^^

poeschlr:
Thanks, for reference dimensioned drawing of the footprint

Screenshot from 2020-02-18 23-46-00

PR #1997, opened 2019-12-10
Molex 44661xxx1 RJ45 footprint

evanshultz:
https://www.molex.com/pdm_docs/sd/446610011_sd.pdf

The jack on the first page has an MPN that ends in '2' and all others end in '1'. I decided to wildcard this one because the footprints will be the same even though the 3D model is different. Let me know if you'd like me to make more unique footprints for the 3D model or make this just for a single MPN.

image

I noticed the left courtyard was at the wrong place when I dimensioned this and so I fixed that before committing.
image

---

---


jstjst:
- I took a look at the dimensions and the description, everything looks expect for the dimension `10.095`, I think it should be `10.09`.
- I suggest to add a `PCB Edge` comment to Dwgs.User to make the positioning more clear (similar to this: USB_C_Receptacle_GCT_USB4085).
- The Pin-1 markings on the other RJ45 footprint is different (at silkscreen and Fab), but I don't know the preferred style. I think the silkscreen arrow should have 90° and than it looks good.
- I think an separate footprint for these on the first page would be good, because they body is smaller and they have no shielding or shielding pins.

evanshultz:
1. Yeah, that was a little weird. I assume the 11.3mm spacing between the big NPTHs and 8.89mm total spacing across the LAN pins are correct. Then I centered the 11.3mm around the LAN pins and that's what you get. That seems best to me but let me know if you have a better idea.
1. Done.
1. I'm not sure what you mean for the silk style. I detest the "L" silk line and especially because this pin isn't in the corner. Check the screenshot below.
1. Already done. The first page PNs end in `2` and this footprint in `1`, which is only the 2nd and 3rd pages. Was there something else or is that OK?

image

PR #2002, opened 2019-12-13
Added footprint for ProAnt PRO-OB-440

cnieves1:
Datasheet: http://www.proant.se/files/user/Documents/Datasheets%20and%20appnotes/OnBoard%20SMD/Application%20Notes/2019-09-25/OnBoard%20SMD%202400%20rev4.3.pdf
imagen

imagen

imagen


cnieves1:
Regarding F6.2, I think the footprint is centered. Maybe it is not calculated correctly due to pins are not symetrical...

fake-name:
Right now, the two mechanical support pads are both pad number 3. 

This means they're connected on the same net, which you don't want.

The mechanical support pads should be pads 3 and 4, so they both get single pin nets.

PR #2003, opened 2019-12-13
Added Abracon APAES868R8060C16-T

cnieves1:
Datasheet: https://abracon.com/patchantenna/APAES868R8060C16-T.pdf
3D depends on: https://github.com/KiCad/kicad-packages3D/pull/623

imagen

imagen

---

---


poeschlr:
Is there no restriction on what can be placed under the antenna? Are there no pads other than the centre one? How is this mechanically connected? (It is quite large i doubt the one pad has the mechanical stability to manage this.)

cnieves1:
This antenna is ceramic. Of course no component can be placed under the antenna, but there is no copper restriction. Ceramic component is not conductive... Usually a patch antenna works with a ground plane underneath, so this is expected, but no other restriction.

For mechanical fixing, it has double side adhesive tape under the antenna. 
Therefore, in order to place it in the PCB, you have to remove its protective tape, pass the pin through the hole, align it properly and push the antenna to the PCB so the adhesive tape starts working.

The F.Fab drawing was already made with maximum dimensions, and courtyard is relative to them.

poeschlr:
Sorry for the delay, your explanation is satisfactory so i am ok with this as is. Thanks

PR #2007, opened 2019-12-17
Added footprint for Bourns SRP6060FA series.

cnieves1:
Generated by script, using PR: https://github.com/pointhi/kicad-footprint-generator/pull/456
Datasheet: https://www.bourns.com/docs/Product-Datasheets/SRP6060FA.pdf
imagen
imagen

---

---


chmorgan:
@chschlue can we drop the pin 1 indicator from parts like this where it can be installed either way? Also the part doesn't appear to have an indicator on it that would let the assembly person configure the pick and place without looking at the datasheet.

PR #470, opened 2018-04-05
Add TO-220F footprint & fix TO-220

jkriege2:
1. fixed TO-220 footprints: the y-offset of pins to metal-back was wrong, it has been modified to match https://www.vishay.com/docs/66542/to-220-1.pdf
2. added footpritns for TO-220F (isolated plastic package), see https://www.fairchildsemi.com/package-drawings/TO/TO220M03.pdf
3. 3D models (TO-220-2/3: full set and TO-220F-2/3: vertical): https://github.com/KiCad/kicad-packages3D/pull/288
4. 3D model sources: https://github.com/KiCad/kicad-packages3D-source/pull/115
5. generator script: https://github.com/pointhi/kicad-footprint-generator/pull/120
6. footprints: https://github.com/KiCad/kicad-footprints/pull/470
7. remaining KLC-violations are due to very small pin-pitch in 5-pin packages and have been present in the former version too. I don't see any way to satisfy the 0.15mm with these footprint, the distance between the hole outer diamaters is only 0.6mm. All other bugs are also not new, as I mainly renamed the footprints ==> Please ignore

2018-04-05 11_23_18-pcbnew d__kicad_devproject_devproject kicad_pcb
2018-04-05 11_40_24-3d viewer

------------

Thanks for creating a pull request to contribute to the KiCad libraries! To speed up integration of your PR, please check the following items:

- [x] Provide a URL to a datasheet for the footprint(s) you are contributing
- [x] An example screenshot image is very helpful 
- [x] If there are matching symbol or 3D model pull requests, provide link(s) as appropriate
- [x] Check the output of the Travis automated check scripts - fix any errors as required

PR #2008, opened 2019-12-17
 Add more Maxim 5x5mm TQFN footprints #470 

matthijskooijman:
This adds all Txx55 footprints from the Maxim [21-0140](https://pdfserv.maximintegrated.com/package_dwgs/21-0140.PDF) datasheet that were not already added. In addition, this addds thermal vias to two existing TQFN footprints that did not have them yet.

These footprints were generated using https://github.com/pointhi/kicad-footprint-generator/pull/470

One of these new footprints is to be used with https://github.com/KiCad/kicad-symbols/pull/2104

For the thermal vias, these are not specified by the datasheet, so I based the number of thermals on similar QFN footprints. For the smallest one (EP1.23x1.23mm), 2x2 thermal vias seems a bit dense, but this is about the same as other footprints, such as WSON-8-1EP_3x2.5mm_P0.5mm_EP1.2x1.5mm_PullBack_ThermalVias

For the new footprints, these match a number of different packaging codes in the datasheet, but all these packaging codes seem to have the same footprint (all dimensions are the same, and the landing pattern is also the same). E.g. T2855-8 and T2855-8C have the same dimension and the same [90-0028 landing pattern](https://pdfserv.maximintegrated.com/land_patterns/90-0028.PDF):

image

Also, the common lead length of the 20-pin and 28-pin versions is 0.55mm. However, some versions have a different lead length of 0.40mm, while all other dimensions are the same. Regardless of the lead length, both versions use the same landing pattern (e.g. T2855N-1 and T2855-8 have different pad length but both use the same [90-0028 landing pattern](https://pdfserv.maximintegrated.com/land_patterns/90-0028.PDF)), which is (nearly) the one generated using a pad length of 0.55mm. For this reason, all the 20-pin and 28-pin footprints have the standard 0.55mm pad length. I was originally planning to add a `lead_length_overwrite` parameter to the generator script (so you could specify a lead length of 0.40mm but generate a footprint with 0.55mm), but then I noticed that for the 28-lead version, both 0.40mm and 0.55mm are used for footprints that are otherwise identical, so there we could not 100% correctly represent all packages, unless we introduced multiple variations of the same footprint for different lead length, which is not currently reflected in the footprint name).

Since the script automatically calculates the pad sizes from the package measurements, the footprint does not exactly match the landing pattern specified by Maxim. However, for the couple ones I checked, they are all within 0.1mm, so that should be close enough.

These footprints have some `check_kicad_mod.py` violations regarding missing layers for the EP, but these seem to occur for the existing TQFN/QFN footprints, so this is a problem in the script (if it is a problem at all)?

Here are the footprints with added thermals from the first commit:

image

Here are the footprints added in the second commit: 

image


matthijskooijman:
I've regenerated these footprints based on some minor changes in the script PR.

chschlue:
Thanks.

PR #2025, opened 2019-12-30
Connector_PinHeader_2.54mm: Add horizontal high-profile pin headers

piotr-wiszowaty:
Connector_PinHeader_2.54mm: Add horizontal high-profile pin headers

Datasheet: http://www.adam-tech.com/upload/PH1RA-XX-UA.pdf

PinHeader_1x01
PinHeader_1x02
PinHeader_1x03

---

---


aewallin:
if these are script-generated, is a link to the generating script stored somewhere?

if the only difference to existing pin-headers is that the high-profile pin is located at a height of 5.8mm from the PCB top-surface, should this be included in the description? or on the footprint cmts.user-layer?

piotr-wiszowaty:
These footprints are indeed generated. Stupid question: where one does 
normally put such scripts?

These pin-headers have different footprints from the corresponding (i.e. 
*_Horizontal) right-angle ones.

poeschlr:
Well all our other footprints including the current pin headers are generated with the script found here: https://github.com/pointhi/kicad-footprint-generator

So i really suggest instead of making a new script to adapt the existing one.

poeschlr:
And we might also need to ask the question if we want to include footprints for different length and if so which lengths should we include. The reason being that one can buy nearly every reasonable length imaginable. (Samtec for example has products where one can order connectors with user specifiable length in quite fine steps. I seem to remember the step size was 1mil)

---

My long term plan would be to include most of our footprint generators for generic parts with kicad. This includes the one for pin headers. (I am however unsure when i will find time to do it as i am also desperately needed for reviews here on the library repos)

piotr-wiszowaty:
I'd say that generating footprints with every possible custom-specified pin length is a low priority task.

piotr-wiszowaty:
I assume that the pin-headers are generated with scripts/pin-headers_socket-strips/make_pin_headers.py. In that case the files produced differ from the ones present in the repository. There are differences in file names (e.g. Pin_Header_Straight_1x02_Pitch2.54mm vs. PinHeader_1x02_P2.54mm_Vertical) and their contents (e.g. the courtyards are of different size).

poeschlr:
The naming difference is simply here because the script might not have been updated with the new naming convention.

However, it should generate the same footprints as i doubt anyone was insane enough to change them by hand. I will take a look at this later.

poeschlr:
@evanshultz you recently looked at the pin header scripts, could you chime in here?

chmorgan:
@poeschlr, @evanshultz how should we proceed with this PR?

Issue #2016, opened 2019-12-22
Raspberry Pi footprints


jstjst:
I started to create the footprints for the RPi symbols. Symbol Issue: https://github.com/KiCad/kicad-symbols/issues/1772
There was a previous discussion here: #1064 

I first looked over the existing RPi Zero footprint (PR #1066 ).
I added the micro SD card outline to the footprint, the rest of it is fine.

I added the Run header, but the outline graphics overlap the GPIO header.
Should I extend the outline graphics from the GPIO header and "merge" them together?

![grafik](https://user-images.githubusercontent.com/12716613/71325430-e456d880-24ec-11ea-8339-0767edeb07c8.png)

jstjst:
List of all symbol PR:
- [ ] https://github.com/KiCad/kicad-symbols/pull/1820
- [ ] https://github.com/KiCad/kicad-symbols/pull/1850
- [ ] https://github.com/KiCad/kicad-symbols/pull/1851
- [ ] https://github.com/KiCad/kicad-symbols/pull/1852
- [ ] https://github.com/KiCad/kicad-symbols/pull/1853
- [ ] https://github.com/KiCad/kicad-symbols/pull/2397

jstjst:
@myfreescalewebpage Can you take a look at the problem shown in the picture above?

myfreescalewebpage:
@jstjst the silkscreen lines should be visible so I suggest you should remove the part of the lines under the connector, just between pins 37/42/39/41 and merge the lines.

PR #2032, opened 2020-01-01
Raspberry Pi 3A+ footprints

jstjst:
Added Raspberry Pi 3A+ footprints
Footprint Issue: #2016
Symbol Issue: KiCad/kicad-symbols#1772
Drawing: https://www.raspberrypi.org/documentation/hardware/raspberrypi/mechanical/rpi_MECH_3aplus.pdf

Raspberry_Pi_3A+_Socketed_THT_FaceDown_MountingHoles
grafik

Raspberry_Pi_3A+_Run_PEN_Socketed_THT_FaceDown_MountingHoles
grafik


aewallin:
are there standardized connectors/sockets that go on the host-board? If the socket height is always the same, then this footprint could have a comment stating the suggested stand-off height? (i.e. what height standoffs are fitted to the mounting-holes in the corners).
some footprints have this kind of helpful info on the Cmts.User layer.

For an experienced user I'm sure it's obvious where the different connectors of the Pi are - but for a casual user it could be useful to also indicate where SD-card, HDMI, USB, etc. are? Maybe not on silkscreen but again on e.g. Cmts.User?

jstjst:
The standoff height varies depending on the Raspberry Pi model and also with the position on the PCB, because it is possible to let the RPi overhang so the USB and LAN connectors are not in the way.
And it also depends on the height of the components on the PCB underneath the RPi.
I think there is no standard height, in the past I used 3D models to determine the height.

I would not add labes for the connectors, because they clutter up the layers and are not really part of the PCB design.

poeschlr:
Disclaimer: not a review just a comment to the discussion.

I agree with @jstjst in all points. Using a footprint simply requires at least looking at the datasheet of the part that it represents. If there are however things that protrude out of the currently indicated outline or if there is some area where there is less height available than in the rest then i would say a marking on the dwgs layer might be a good choice.

jstjst:
@poeschlr @aewallin I generally agree but don't think it is useful and worth the effort in this case, because there are a lot of connectors with different heights on the RPi that someone need to verify it in 3D anyway if he wants to put (bigger) components underneath.

PR #1528, opened 2019-04-05
DFN_QFN: add DFN-6-2EP_2x2mm_P0.65mm_EP1x0.7mm

pboe:
#1455

---

All contributions to the kicad library must follow the [KiCad library convention](http://kicad-pcb.org/libraries/klc/)

Thanks for creating a pull request to contribute to the KiCad libraries! To speed up integration of your PR, please check the following items:

- [ ] Provide a URL to a datasheet for the footprint(s) you are contributing
- [ ] An example screenshot image is very helpful 
- [ ] If there are matching symbol or 3D model pull requests, provide link(s) as appropriate
- [ ] Check the output of the Travis automated check scripts - fix any errors as required
- [ ] Give a reason behind any intentional library convention rule violation.

PR #2034, opened 2020-01-03
Create UDFN2020-6 Type J for DIODES.com Voltage regulators

fhw72:
Create UDFN2020-6 Type J for DIODES.com Voltage regulators
by modifying already existing DFN-6-1EP_2x2mm_P0.65mm_EP1x1.6mm footprint
which is very similar.

Footprint is violating rule F9.3 because these are VERY similar.

https://www.diodes.com/assets/Package-Files/U-DFN2020-6-Type-J.pdf

Part to be added soon using this footprint:
https://www.diodes.com/assets/Datasheets/AP3401.pdf


evanshultz:
Did you verify this footprint? I did quite a bit of research on this (see https://github.com/KiCad/kicad-footprints/pull/1528) and don't have confidence to add anything new unless it was verified. And, sadly, I don't have confidence the existing footprints are correct either.

fhw72:
I did some measurements to ensure the distances etc. are right. The only thing I just copied from the
DFN-6-1EP_2x2mm_P0.65mm_EP1x1.6mm footprint are two pads (Paste layer) below the big pad
in the middle. I thought they are there by convention - at least I don't have an explanation why they are there.

fhw72:
I forgot to add a screenshot:

Screenshot 2020-01-03 19 56 37

PR #2053, opened 2020-01-13
Added Kingbright KPBD-3224, dual LED.

cnieves1:
Datasheet: https://www.kingbright.com/attachments/file/psearch/000/00/00/KPBD-3224SURKCGKC(Ver.20A).pdf

imagen

imagen

---

---


cnieves1:
Pin 1 polarity is different than cathode polarity. Added cathode polarity.

imagen

cnieves1:
3D: https://github.com/KiCad/kicad-packages3D/pull/646

aris-kimi:
Hi @cnieves1 , i am not sure if polarity marking should be placed at fab layer, but i think a `T` pointing at the right direction simplifies the drawing.
Fab layer should be simple and point pin1 location so i think diode symbol and - should be removed, silk shows the polarity.

cnieves1:
@aris-kimi Sorry, but I don't think removing polarization from F.Fab is good. 
F.Fab should contain polarization, as per KLC F5.2.
Wether polarization is diode symbol, or anything else is another question.
Diode symbol is shown in other footprints already in the library, like LED_Cree-XP or LED_Cree-XQ or LED_miniPLCC_2315.
I think it makes no harm if the package is big enough as this case.
In this case, polarity is not pointing to pin 1, but is the same as the marking in the LED package to avoid confusion. 
It could be a line (like shown in the above screenshot), a dot, or whatever... Please advice.

aris-kimi:
Hi @cnieves1 , i will allow this since there are some similar symbols, and probably i am going to propose a complete lib rework after migrating.
I will full review this one later today.
Thanks.

aris-kimi:
I am not sure about name, everything else is great.

@chschlue , does that have to be full part number `KPBD-3224SURKCGKC` or we are ok with current naming `KPBD-3224`?
If it is ok now, i am merging.

εικόνα

aris-kimi:
Hi @cnieves1 , after @chschlue 's input on the linked PR i think you should clear fab layer from internal graphics, and also aligne silk line for pin1 (and shorten it a little).
εικόνα

cnieves1:
Mmm... Assembly houses use to align package marking with fab marking. I am afraid they will get confused with pin 1 marking and mount it reversed... Can we mark the package polarity somehow in fab layer?

chschlue:
That's what https://github.com/KiCad/kicad-footprints/pull/2488 is about. Mark what is marked on the package and change zero orientation such that the mark ends up top left.

PR #2057, opened 2020-01-14
Keystone 2479 Fix: Vertical distance from pin 1 center to top F.Fab line and component length are not correct.

cnieves1:
Datasheet: https://www.keyelco.com/product-pdf.cfm?p=1041

imagen

From datasheet data, distance from pin 1 to component edge (F.Fab) is (37.49-23.5)/2=6,995nm.
Same distance in footprint is 5.95mm

imagen

Moved all drawings, keeping the pins and holes positions, to fix this.

Another problem is that component length is 52.63mm in datasheet, while length in footprint (see screenchot above) is 50.4+2.6mm=53mm.
Fixed this too.

Result with these changes:
imagen

---

---


cnieves1:
Regarding F5.4, it is the original design already in the library.

cnieves1:
3D for fixed version at: https://github.com/KiCad/kicad-packages3D/pull/648

poeschlr:
For reference: the footprint was originally added here https://github.com/KiCad/Battery_Holders.pretty/pull/4

Misca1234:
There is a scrips in 
https://github.com/pointhi/kicad-footprint-generator.git
that create this foot print
kicad-footprint-generator/scripts/Battery/BatteryHolder.yml

Instead of modifying the existing fp, it could be re generated instead

cnieves1:
This file was generated using https://github.com/pointhi/kicad-footprint-generator/pull/490
It seems previous file in master was not generated using this generator...

evanshultz:
@cnieves1 
Do you find the scripted footprint to match the datasheet?

cnieves1:
No, it does not match the datasheet, at least in height. Generated model is 11mm height, while datasheet specifies it as 13mm.

In cq_parameters.py, there is specified H=13.00. I tried changing it to H=15.00, but this has nothing to do with case height.

I don't know where I can change this...

Misca1234:
Let me have a check, it was I who created the scripts originally

cnieves1:
Sorry, I meant the 3D has incorrecta height, not the footprint...

Misca1234:
Ok, the scripts have a number of basic shapes to model with, perhaps they are not flexible enough, let me check

Misca1234:
I have fixed the script, but not pushed it, the problem was related to models that have a "thick bottom"
I hope I can push it this week

chschlue:
Thanks.

PR #2058, opened 2020-01-14
Wuerth MH: Anchor pad margin to avoid DRC errors

poeschlr:
Fixes problem with DRC found by @Shackmeister 

script https://github.com/pointhi/kicad-footprint-generator/pull/488


poeschlr:
problem reported in discussion of https://github.com/KiCad/kicad-footprints/pull/1612 (after merge)

Shackmeister:
I still get a DRC error between the custom pad and the NTPH

poeschlr:
Ok strange my first tests showed no error but now i get one too. Well at least it is now only one error per footprint.

Shackmeister:
merging as it's an improvement, hopefully we can get it fully resolved during V6

PR #2059, opened 2020-01-15
Added NHD-C0220BiZ footprint

aracosta1:
Reopen of PR #1406 
The components doc are available at : [newhavendisplay](http://www.newhavendisplay.com/specs/NHD-C0220BiZ-FSW-FBW-3V3M.pdf)

immagine


poeschlr:
Is there a reason the silk outline is this far out? I suggest moving it much closer to the body outline as it would then be useful as an alignment help.

For non plated pads the pad size and hole size must be equal (the two right pads violate this).

Courtyard clearance is ok all around (checked in kicad so no screenshot available)

The y direction of the display area is wrong every other dimension is ok.
Screenshot from 2020-02-06 20-42-06

poeschlr:
I kind of forgot to ask last time round: Could you elaborate how you determined the drill sizes (both for the oval and circular pads excluding the mounting pads)? 

The circular pads seem small to me as the nominal diagonal is already 0.583mm. So 0.78 is already right at the limit even without position and size tolerance of the pins taken into account. I don't quite believe that these would have the globally given 0.3mm tolerance (would mean they could be nearly twice as large as nominal.) 
If we would take this tolerance then it would result in a required drill size of 1.2mm. If we use a more reasonable tolerance of 1.5mm, then the required size comes down to 1mm. (I used statistical tolerance addition here as I highly doubt there will ever be a pin with both directions at the maximum)

---

Similarly, with the oval hole. I would suggest 1x2.6mm (again tolerance of 0.15mm and statistical addition)

aracosta1:
I don't think the stated tolerance of 0.3mm can be applicable to pins that are 0.5x0.3 mm. Even with a tolerance of 0.2mm there could be pins so narrow they will bend under their own weight. My assumption was the stated tolerance apply to the general dimensions of the display and not to the pins. 
I used a tolerance of 10% for the pins,so about 0.05mm, but 0.15mm allow for more room during the component insertion. 
I will change the holes accordingly.

poeschlr:
thanks

PR #2062, opened 2020-01-17
Replace tht solder wire with scripted and add relieve version

poeschlr:
# general info

script https://github.com/pointhi/kicad-footprint-generator/pull/489

This replaces the original THT solder wire footprints with a scripted options build for real wires found in https://ec.staubli.com/AcroFiles/Catalogues/TM_Cab-Main-11014119_(en)_hi.pdf (I also checked with other manufacturers. It seems the wire sizes are quite common)

I also included a version with one NPTH drill for strain release as well as one for two such holes (the later to ensure the wire ends up at the same PCB side as without relieve) The distance to the strain relieve holes is 6 times the outer diameter of the wire (bend radius is 3xOD as suggested by Stäubli/multi-contact but also lapp)

I got rid of the dual row version as it would be quite hard to script it. (Can possibly be added at a later stage)
The courtyard areas are split up as even for the smallest wires there is enough space for a 0402 footprint in between. -> this means the travis courtyard check fails

Right now all wires are metric. We can add footprints for AWG by adding them to the wire definition file.

---

# open questions

I am not sure about the naming convention. For now i used the prefixes D for conductor diameter and OD for outer diameter. Another option would be to use Conductor and Outer (or Isolation) or even add the word Diameter to them.

Right now i do not include the bend radius in the footprint name. Maybe this would be an option (i however doubt we will add other bend radius options)

I am also not really sure how we should write mm² in the footprint name and documentation fields. In the footprint name and tag field i for now used sqmm while i used square mm in the description.

---

Screenshot of a few options:

Screenshot from 2020-01-17 23-36-55

inspiration
example


chschlue:
> I am not sure about the naming convention. For now i used the prefixes D for conductor diameter and OD for outer diameter. Another option would be to use Conductor and Outer (or Isolation) or even add the word Diameter to them.

`D` and `OD` seems clear to me. I wouldn't make the names unnecessarily long by adding `Diameter`.

> Right now i do not include the bend radius in the footprint name. Maybe this would be an option (i however doubt we will add other bend radius options)

6xOD should be a one-size-fits-all-solution.

> I am also not really sure how we should write mm² in the footprint name and documentation fields. In the footprint name and tag field i for now used sqmm while i used square mm in the description.

`sqmm` is fine in the names, but why not `mm²` in the description?
Semi-related: I think we leave out trailing zeroes pretty much everywhere else, so people are probably used to searching `D1.5mm` instead of `D1.50mm`, for example.
How about prefixing wire cross section with A, BTW? This doesn't really fit future AWG sizes, though.

poeschlr:
I am not sure if mm² would work for everything involved (python script, kicad, travis, ...) And to be honest i not really sure i want to invest the time to find that out.

I am also not sure that adding a prefix to the main measurement is fitting with our general naming convention (after all we do not prefix the body size of components)

chschlue:
> I am not sure if mm² would work for everything involved (python script, kicad, travis, ...) And to be honest i not really sure i want to invest the time to find that out.

`²` (U+00B2) is Latin-1 Supplement, not a combining character or anything fancy.
It's even "easier" than Unicode Greek (U+0370 through U+03FF), which we already empirically found to be working.

> I am also not sure that adding a prefix to the main measurement is fitting with our general naming convention (after all we do not prefix the body size of components)

True. My mistake.

poeschlr:
Ok i will use mm² in the description.

chschlue:
One more thing:
Shouldn't it be `Relief` instead of `Relieve`?

poeschlr:
Yes you are right

poeschlr:
@chschlue i forgot to ping you after i fixed your remarks

chschlue:
Actually, I noticed you pushed changes but then forgot to take a look. Will do later.

chschlue:
Have you actually used these?
The reason I'm asking is that D+0.2mm holes might be a little tight if you're going to tin the wire beforehand (or want to insert it without splaying if you don't).

poeschlr:
I generally do not tin wires before inserting them and i think it would also not really work for the versions with relief as the wire is then no longer flexible enough (the whole reason for using the relief version is to have the inflexible part fixed in place).

I however also did not test them myself so am also not certain that they work as created. The thing is also that 0.2mm oversize is surely enough for most of the wire sizes. I can however easily see how it might get too little for 1.5mm² and up. 
Which is why i now chose to increase by 20% with minimum of 0.2mm.

chschlue:
Looks good to me

poeschlr:
Fuck you were too fast

chschlue:
What did I miss?

poeschlr:
I just realized that i have the pitch specifier in the footprint name for single wire versions
And i also had strange drill diameters now so i introduced rounding. Well will make a pull request for these fixes

chschlue:
Oops.

poeschlr:
Fix in https://github.com/KiCad/kicad-footprints/pull/2274

PR #2063, opened 2020-01-18
Added some Bourns inductors (SDR0302, SRP1770C and SRR1280)

blackvladimir:
- SDR0302
https://www.bourns.com/pdfs/SDR0302.pdf
SDR0302

SDR0302dim

There are pads with two rounded corners in datasheet. I created them by placing two overlapping pads with the same number but I am not sure if it is optimal solution.

- SRP1770C
https://www.bourns.com/docs/product-datasheets/srp1770c.pdf
SRP1770
SRP1770dim svg

The same rounded corners as in previous part. The rounding is small (radius 0.2) to prevent overhanging of part on bigger side of tolerances.

- SRR1280
https://bourns.com/docs/product-datasheets/srr1280.pdf
srr1280scr
srr1280 svg

---

---


aewallin:
did you draw F.CrtYd by hand, or is there a tool/script that makes a F.CrtYd according to some given offset from pads/F.Fab ?

blackvladimir:
I drew it all in freecad in sketcher. I drew pads and F.Fab first. Then I drew around it and constrained it to 0.25mm. I drew only one quarter and then mirrored it twice.

screenshot

poeschlr:
The thing is the radius of the part body is not specified in the data sheets. Which means you have estimated it (at best by measuring a real part). With other similar parts we deliberately made the corners rectangular on the courtyard layer to ensure that there will not be a placement conflict.

---

Also we would have a script for components like these. I am currently in the process of reordering the script directory so if you are interested in converting your footprints to the script then base your pull request on my branch. The script is now found here: https://github.com/poeschlr/kicad-footprint-generator/tree/cleanup_mounting_hole/scripts/Inductors/Inductor_generic (link is to my branch. The repo you would fork is https://github.com/pointhi/kicad-footprint-generator, The pull request for the reorg is https://github.com/pointhi/kicad-footprint-generator/pull/487)

blackvladimir:
I get the radius from manufacturer step file. It makes sense to made the corners rectangular the step file does not need to be correct (and it does not include tolerances).

I will try to use the script to fix it.

What about the rounded corner of pads. Do you think it is ok to do it like this? The radius is also not specified in datasheet. I just made it so even with tolerances the part wont overhang.

blackvladimir:
I fixed the corners manualy. 

I looked into scripts but the Inductor_generic is using StandardBox which always creates marks for pin 1 which I think is misleading for non-polarized components. 

Later I maybe get to creating new script and try to unify some of the inductors.

L_Bourns_SRR1280_12 5x12
L_Bourns_SRP1770C_16 9x16

PR #2069, opened 2020-01-24
Fix TC2050-IDC-NL courtyard and added front silkscreen.

cnieves1:
Datasheet:  https://www.tag-connect.com/wp-content/uploads/bsk-pdf-manager/TC2050-IDC-NL_Datasheet_8.pdf

Top dimensions of this connector are 10.15 x 4.06mm, while courtyard in the footprint was 9.5x4mm.

imagen

This PR adds a front silkscreen with 10.15 x 4.06mm, and changes the courtyard to 10.66 x 4.6mm, complying with 0.25mm courtyard clearance specified in KLC F5.3.

imagen

---

---


evanshultz:
Does the mating connector consume a 6.731x12.446mm area on the board?

cnieves1:
10.15 x 4.06mm up to a 6.35mm height
12.446 x 6.731mm above 6.35mm height.

imagen

cpresser:
For the lulz I checked the tag-connect on my desk:
image
image

It is 13x6mm which does not really match the datasheet (12.446 x 6.731mm).

I think that the courtyard should not be as big as 12.5x6.7mm. **A big courtyard defeats the advantage of the tag-connect-system (a really small footprint).**

Since this is a virtual part without any assembly the courtyard could be as small as the inner part of the plug (10.1x4.06mm).
To indicate the size of the upper part of the plug another polygon could be added on Dwgs.User, with a descriptive text ("Z-Keepout") or similar.
Perhaps like so:
image

What do you think?

cnieves1:
It is ok for me. However, I would keep the dimensions given in the datasheet, since they may consider some tolerances in real life...
It seems you have already modified the footprint. Are you able to share/commit it?

cpresser:
> It is ok for me. However, I would keep the dimensions given in the datasheet, since they may consider some tolerances in real life...

Yes. And 'measuring with caliper' is not an allowed method in this repository.

> It seems you have already modified the footprint. Are you able to share/commit it?

No. It was just a draft to show the idea. All the dimensions are wrong.

cpresser:
Perhaps the "Z-Keepout" text should be on all 4 sides.
And we can think about using a different hatch-style for the z-keepout-area.

chschlue:
> Yes. And 'measuring with caliper' is not an allowed method in this repository.

For even more lulz: Don't tell anyone about https://github.com/KiCad/kicad-footprints/pull/1145 ;)

cnieves1:
Mmmm... one question:
the left side of the datasheet drawing shows an arc, and angled lines, without any dimension other than overall dimensions as a rectangle.
If there are no dimensions in the datasheet to make that, what should we do?  keep the footprint as a rectangle (as I did)? or is there any other way which would be accepted?

cpresser:
Yolo-Workflow (works if the pdf has vectors):
* print the relevant part of the .pdf to .svg in evince
* then load it in inkscape (or open the .pdf directy in inkscape - which might take ages)
* remove all useless stuff
* export as .dxf
* import in kicad
* profit

This is virtually impossible to review, but you can add some in-between screenshots to document the process.

cnieves1:
Instead of using inkscape, I used pcb-rnd, which allows to place a background image and loads kicad footprints.
I loaded the original footprint there, and matched the background image looking at pads and holes:
Captura de pantalla de 2020-03-21 05-52-08
and placed some lines for guidance:
Captura de pantalla de 2020-03-21 06-16-15
Then, adjusted the footprint to match the lines, but:
Captura de pantalla de 2020-03-21 07-14-52
Then, I measured on screen, and I realized that outer outline is not scaled with regarding to inner outline...
So I decided to keep outer outline general dimensions as per datasheet, and make diagonal lines parallel to innner outline. This is the result (Z keepout not done yet):
Captura de pantalla de 2020-03-21 07-37-33
Would this be acceptable?

cpresser:
Yes, that is good enough for me.
I assume the screenshot is still a draft? For example the outer outline should be on Dwgs.User.

cnieves1:
Yes, it was a draft.
I think it is done. Z-keepout hatching is reversed with regarding inner keeput. See below:

Captura de pantalla de 2020-03-22 00-37-45

cpresser:
Thanks, that looks pretty good, but also a little crowded:
I have two ideas how to improve it - unsure how good they will look:
* Less text. Perhaps have the blue text only on top and bottom, not left and right.
* The outer-keepout could end on the courtyard.

While you are at it, could you please also do `Tag-Connect_TC2030-IDC-NL_2x03_P1.27mm_Vertical` to keep them consistent?

cnieves1:
What about this?
Captura de pantalla de 2020-03-23 23-51-58

I'll do the other footprint when this is ok, if you don't mind.

cpresser:
I like it! :heart:

cnieves1:
Updated https://github.com/KiCad/kicad-footprints/pull/2070

PR #2071, opened 2020-01-26
Add Crystal Murata GRCGB 2.0x1.6mm

rleh:
Add Crystal Murata GRCGB 2.0x1.6mm

grafik

Datasheet Link: https://www.murata.com/~/media/webrenewal/campaign/ads/america/timing/p79e.ashx?la=en-us

3D model coming soon. (I will put a link here.)


aewallin:
I don't see how the silkscreen indicates where pin1 is?

rleh:
The crystal is symmetrical (like a resistor or ceramic capacitor), so I don't think a pin1 marker makes sense.

evanshultz:
@rleh 
There is no clause in https://kicad-pcb.org/libraries/klc/F5.1/ that allows pin 1 silk mark to be skipped, so please add it. Thank you.

rleh:
Ok, I added the pin 1 indicator.

Thank you for reviewing!

grafik


---
> There is no clause in https://kicad-pcb.org/libraries/klc/F5.1/ that allows pin 1 silk mark to be skipped, so please add it.

This is funny because some of the components in the example picture don't even have a pin 1 marker. So the footprints of the classic SMD resistors and capacitors are all wrong, aren't they?

rleh:
Foorprint renamed: ~muRata~ -> _Murata_ (According to #545)

---

I fixed the courtyard clearance (0.5mm for crystals).

Silkscreen clearance is 0.19mm (KLC suggestion: 0.2mm) from copper pad to edge of the silkscreen line to fit into 0.1mm grid. Ok?

### Correctness check:
Drawing from FreeCAD

### Drawing from datasheet:
[grafik](https://www.murata.com/~/media/webrenewal/campaign/ads/america/timing/p79e.ashx?la=en-us)

cp-aquila:
> 
> 
> @rleh
> There is no clause in https://kicad-pcb.org/libraries/klc/F5.1/ that allows pin 1 silk mark to be skipped, so please add it. Thank you.

Related: https://github.com/KiCad/kicad-website/issues/422

For this part I personally think the Pin1 indicator is usefull. The Layout of the pads is almost 90° roation-symmetrical. After soldering it is really hard to see if the part is mounted correctly or perhaps rotated by 90°.

rleh:
I'm fine with both choices, it should only be clear in KLC when a Pin1-Parker (or "polarity" marker) is intended.

According to the comments I added the marker some time ago.

PR #2072, opened 2020-01-28
Regenerate missed FPs / #422

chschlue:
Regenerate missed FPs in the wake of https://github.com/pointhi/kicad-footprint-generator/pull/422

Plus OnSemi_VCT, current generator generates differently-chamfered pads.

---

---


evanshultz:
@chschlue 
Thanks for taking care of this!

I notice `SO-20-1EP_7.52x12.825mm_P1.27mm_EP6.045x12.09mm_Mask3.56x4.47mm_ThermalVias` has too-small via annular rings, according to our library. Probably the datasheet used shows smaller ones or it just selected arbitrarily. Could you please make any script changes to avoid errors here?

Other than than, everything looks fine and I'll merge.

chschlue:
Good catch. Missed that among the false positives.

chschlue:
@evanshultz Should be good to go now.

chschlue:
Done.

chschlue:
https://github.com/pointhi/kicad-footprint-generator/pull/493

evanshultz:
Why exclude `SO-20-1EP_7.52x12.825mm_P1.27mm_EP6.045x12.09mm_Mask3.56x4.47mm_ThermalVias`? I can only assume it is in the script if it was updated before.

chschlue:
Of course.
It was introduced in https://github.com/pointhi/kicad-footprint-generator/pull/474/commits/7eeecbc606f976edeb2f7630261a1448fabf2bc2 and is not needed until https://github.com/KiCad/kicad-symbols/pull/1733 is merged.
I'm wondering whether this should actually get its MF-specific name (Texas_PowerPad_SO-20...) and if following TI's AN that pedantically is necessary (we don't usually do that and I'm afraid it's not possible without violating KLC).

TL;DR: I'm not touching that abomination today.

evanshultz:
@chschlue 
It's been a while. It looks like this should be OK to merge, right?

chschlue:
I think so.

PR #2073, opened 2020-01-29
USB A Connector: CNCTech 1001-010-01001

rleh:
USB A Connector: CNCTech 1001-010-01001

Datasheet: http://cnctech.us/pdfs/1001-010-01001.pdf

Preview

The footprint was derived from the existing footprint *USB_A_CNCTech_1001-011-01101_Horizontal*.

No 3D model available yet.


poeschlr:
Silk must have at least the line with clearance from pads. Move the edge indicator to the edge cuts layer (requires a text editor) Add a pin 1 marker on silk and fab layers.

The courtyard clearance looks strange in the screenshot. It looks like the clearance between the body and courtyard is much larger compared to the clearance between pads and courtyard. (It must be 0.5mm for both these, rounded to nearest 0.01mm grid)

rleh:
I updated the footprint:

grafik

PR #2075, opened 2020-01-30
added USB_C_6pins

aster94:
Hello,

I recently bought some usb-c connector on aliexpress. I didn't found any footprint on kicad do I made one based on an existing one, the only difference is that they expose only 6 pins

I didn't found any datasheet but this is the link from where i took them: https://it.aliexpress.com/item/32966491026.html

HTB1oGRBaELrK1Rjy1zbq6AenFXa1


aster94:
I think that is the best i can do `Footprint center calculated as (0.0,-0.0025)mm` without breaking everythink else

poeschlr:
Could you add a screenshot of the footprint from within the footprint editor?

aster94:
image

cp-aquila:
A few quick thoughs (not a review):
* Have you tried asking the seller for a better datasheet?
* Why did you choose the center of the connector like this? Using the pick-n-place center would be the way to go, but since there is no documentation :/
* The silkscreen could be closer to the F.Fab lines
* The courtyard does not seem to have the same distance on all sides. It should be 0.5mm from the F.Fab or Copper
* Some other USB-Connectors do have a line indicating the PCB-Edge in the auxilliary layers. That would also be usefull for this connector.

cp-aquila:
This seems to be the same part. At least we now have a datasheet link and a name.
https://lcsc.com/product-detail/Others_Korean-Hroparts-Elec-TYPE-C-31-M-17_C283540.html

aster94:
> * Why did you choose the center of the connector like this? Using the pick-n-place center would be the way to go, but since there is no documentation :/
> * The silkscreen could be closer to the F.Fab lines
> * The courtyard does not seem to have the same distance on all sides. It should be 0.5mm from the F.Fab or Copper
> * Some other USB-Connectors do have a line indicating the PCB-Edge in the auxilliary layers. That would also be usefull for this connector.

I didn't draw it from zero, I took an existing one from kicad (actually I can't remember which one I took but if needed i can check) and as far as i can remember i just removed some pins and made the reamaining pins ~fatter~ wider

cp-aquila:
Existing footprints can have errors. Just look at the amount of open issues of this repository.

All new submissions must follow KLC as close as possible. There are always corner-cases, but for the points above I don't see that. They can be addressed.
We aim for a really high quality. Thus, it would be advisable to also look at the footprint you used as a reference and see if it needs some fixes.

aster94:
Ok @cp-aquila I can fix the issues you pointed out in the previous post, I really like kicad and I would be happy to help, but please keep in mind that I may don't do it right in the first attempt 😁

cp-aquila:
Don't worry. Just go ahead, I will try to help you through.

If you look at my first contributions you will see that I made a lot of mistakes my self. Even now I make mistakes, but that is what reviews are for.

myfreescalewebpage:
Author indicated it is Abandoned

PR #2083, opened 2020-02-09
Addition of Texas Instruments DC-DC Converter Modules PTN78000 (1.5A)…

jamesrlucas:
Addition of a few DC-DC modules from Texas Instruments

I've only done the Through-hole packages for now.

Datasheets [PTN78000](https://www.ti.com/lit/ds/symlink/ptn78000w.pdf) [PTN78020](https://www.ti.com/lit/ds/symlink/ptn78020w.pdf) and [PTN78060](https://www.ti.com/lit/ds/symlink/ptn78060w.pdf)

I have not yet created the Symbols, I'll link the PR once complete.

image

---


jamesrlucas:
I've done drawings for the footprints as well. Some of the dimensions are out by 0.01mm - due to rounding of some of the dimensions to establish a centre line (otherwise I'd be working on a 0.005mm grid
[PTN78000_Drawing.pdf](https://github.com/KiCad/kicad-footprints/files/4177985/PTN78000_Drawing.pdf)
[PTN78020_Drawing.pdf](https://github.com/KiCad/kicad-footprints/files/4177986/PTN78020_Drawing.pdf)
[PTN78060_Drawing.pdf](https://github.com/KiCad/kicad-footprints/files/4177987/PTN78060_Drawing.pdf)

jamesrlucas:
Done! :)

chschlue:
Thanks.

evanshultz:
I'm not quite sure I agree with these names. They are named by the product, which isn't the package name. The package name could be used for many products.

Take https://www.ti.com/lit/ds/symlink/ptn78000w.pdf for example; you can see in the footprint drawing the package name is `EUT` and `R-PDSS-T5` (for THT version).

I would then name them `/Converter_DCDC_Texas_EUT_R-PDSS-T5.kicad_mod`. Adding a `_THT` suffix wouldn't hurt for clarity. And just using `EUT` should suffice, but it probably doesn't hurt to be a bit more verbose so the footprint is easier to find.

@chschlue 
What do you think?

chschlue:
You certainly have a point there, but I don't think it's quite so easy.

I went with this because it's the usual way of naming footprints in Converter_DCDC (granted, that's mainly because most manufacturers don't have special names for their DC-DC converter packages.)

There are a couple other converters using the same module (according to [TI's package search](http://www.ti.com/packaging/docs/searchproductbypackage.tsp?orderablePartNumber=&packageDesignator=EUT&pinCount=&results=results#resultspage)) but that it's currently only used for DC-DC converters is more or less coincidence.

Long story short: I'm not sure artificially restricting this FP to Converter_DCDC is the silver bullet here either as it is basically "vendor-generic".
How about `Module/Texas_EUT_R-PDSS-T5.kicad_mod` and so on?
(EUT would not suffice BTW as there also is a 6 pin EUT)

evanshultz:
Yeah, a lot of the DC-DC stuff we have are modules inside a plastic box whereas these are bare PCBAs.

Your proposal sounds great to me.

chschlue:
@jamesrlucas Do you have anything to add to this discussion?

jamesrlucas:
I didn't even think about this - I'm more than happy to change these to a more generic footprint name to cover other parts using the same footprints. And I'd agree with keeping it as generic as possible - while all products using that footprint currently are DC-DC converters, that's not to say that will always be the case.

I'll make the changes and submit a new PR over the next few days - and I'll update the symbols I've submitted to refer to the new footprints as well.

PR #2093, opened 2020-02-15
removed Paste layer from molex pcb-edge connectors

xgvargas:
removed Paste layer from molex pcb-edge connectors


chschlue:
Welcome and thanks for your first contribution.

Bug is confirmed. However, these connectors are scripted.
Please fix layers in the script at
https://github.com/pointhi/kicad-footprint-generator/blob/master/scripts/Connector/Connector_PCBEdge/molex_EDGELOCK.py

Perhaps a new definition can be added here
https://github.com/pointhi/kicad-footprint-generator/blob/5243ea508c27b7b0ceff3aebb39a5bc1635eb8b6/KicadModTree/nodes/base/Pad.py#L210

poeschlr:
Could you change the pad type from SMD to connector?

---

I would also argue that these will need renaming to fit the standard naming convention.
The name should be of the form: `Molex_EdgeLock_[part number]_01x[pin count]_P2.00mm_Edge`

Part number syntax 2008900t0n
 - t = board thickness: 1 -> 1.6mm, 2 -> 1.2mm
 - n = pin count (2,4,6,8)

We might want to have one footprint per board thickness which would mean doubling up the footprints and adding a footprint name parameter: `Molex_EdgeLock_[part number]_01x[pin count]_P2.00mm_BoardThickness[1.20 | 1.60]mm_Edge`

Which means an example footprint would be named: `Molex_EdgeLock_2008900102_01x02_P2.00mm_BoardThickness1.60mm_Edge`

---

The documentation could also benefit from a bit more information than just the datasheet link

PR #2098, opened 2020-02-19
new USB Type-C Connector (JAE DX07S016JA1R1500)

adamws:
Datasheet: [https://www.jae.com/en/connectors/series/detail/product/id=91780&type_code=T1010](url)
Screenshot:
image

Comment: very similar to USB_C_Receptacle_HRO_TYPE-C-31-M-12 [http://www.krhro.com/uploads/soft/180320/1-1P320120243.pdf](url) but with different mounting holes and different order of pins


adamws:
@chmorgan, thanks for review! Indeed I might screw up spacing for A1/A12. I will apply review fixes in a spare time. 
Fortunately I was not off by too much and managed to successfully solder it: 
image

adamws:
after fixes:
image

PR #2100, opened 2020-02-20
Add footprints for Cyntec DC/DC modules MUN12AD0x-SH

rleh:
* QFN-8-1EP footprint for Cyntec power module *MUN12AD01-SH*
  * Datasheet: http://www.cyntec.com/upfile/products/download/Cyntec%20MUN12AD01-SH_Datasheet.pdf
  * QFN-8-1EP
* QFN-8-3EP footprint for Cyntec power module *MUN12AD03-SH*
  * Datasheet: http://www.cyntec.com/upfile/products/download/Cyntec%20MUN12AD03-SH_Datasheet.pdf
  * QFN-8-3EP


poeschlr:
The generic library dfn/qfn library is only allowed to hold generic footprints created with the ipc generators found here https://github.com/pointhi/kicad-footprint-generator/tree/master/scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON (suggested footprint is to be ignored)

Specialized footprints should go into more specialized libraries.

And as this really does not look like a DFN package but more like a pcb that happens to emulate a similar land pattern to a no lead package. Which means i would place it into the DC/DC converter lib.

poeschlr:
The fab outline must represent the nominal body size, the courtyard clearance must be 0.25mm all around.
Screenshot from 2020-02-22 18-01-14

Screenshot from 2020-02-22 18-11-33

rleh:
Thanks for reviewing!

I would move the footprints to `Converter_DCDC.pretty/Converter_DCDC_Cyntec_MUN12AD0{1,3}-SH.kicad_mod`, ok?

rleh:
Here are drawings of the fixed Fab-layer outline and courtyard clearances:

Converter_DCDC_Cyntec_MUN12AD01-SH.kicad_mod

Converter_DCDC_Cyntec_MUN12AD03-SH.kicad_mod

rleh:
The 3D model reference for both footprints is `${KISYS3DMOD}/Converter_DCDC.3dshapes/Converter_DCDC_Cyntec_MUN12AD0x-SH.wrl`, because the 3D models are identical.

poeschlr:
I would suggest moving silk a bit closer to the part outline such that it is more useful for aligning the part.

rleh:
Ok.
grafik
grafik

poeschlr:
thanks

PR #2104, opened 2020-02-23
Restore PLCC-32 THT socket footprint

AlexeyBond:
This footprint seems to be correct but it was removed in https://github.com/KiCad/kicad-footprints/pull/675 along with some incorrect footprints.

Datasheet: http://www.assmann-wsw.com/fileadmin/datasheets/ASS_0981_CO.pdf

image


cpresser:
Can you please add a link to a datasheet of a socket that fits this footprint?

AlexeyBond:
Datasheet link added to pr description

cpresser:
I found a better datasheet
http://www.assmann-wsw.com/fileadmin/datasheets/ASS_0981_CO.pdf

image

image
image

- [x] Pin numbering
- [x] Body size and position
- [x] Pin1 designator 
- [ ] URL to the datasheet should be in the description
- [ ] Courtyard distance is 0.47mm, it should be 0.5mm in x direction.
image

chmorgan:
@AlexeyBond can you add the url to the part or footprint into the description?

PR #2105, opened 2020-02-24
Lightpipe_LPF-C012303S is not virtual

cpresser:
The device will be placed onto the board. Most likely manually, but it will be
assembled and thus should not be virtual (KLC F8.1)

No screenshot or other info, this is pretty straight-forward. 
It is not a breaking change.

edit: I did a rebase onto current master and force-push since travis did check some weird ranged.


evanshultz:
Yep. The footprint is for a tangible part.

PR #2106, opened 2020-02-26
Replacing LFPAK56 tab with two rectangles, instead of the custom shap…

charithjperera:
…e as thermal reliefs do not work with custom shapes.

Small change to the LFPAK56 footprint, it was previously made up of a custom shape for the tab. I changed it to be two rectangles as thermal reliefs don't work with custom shapes.

Footprint datasheet: https://assets.nexperia.com/documents/outline-drawing/SOT669.pdf
---

---

### DIFF
--- a/Package_TO_SOT_SMD.pretty/LFPAK56.kicad_mod
+++ b/Package_TO_SOT_SMD.pretty/LFPAK56.kicad_mod
@@ -1,4 +1,4 @@
-(module LFPAK56 (layer F.Cu) (tedit 5BA2E0B1)
+(module LFPAK56 (layer F.Cu) (tedit 5E55BAF5)
   (descr "LFPAK56 https://assets.nexperia.com/documents/outline-drawing/SOT669.pdf")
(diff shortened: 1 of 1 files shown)


evanshultz:
@charithjperera 
Thanks for using KiCad and contributing for the first time.

No thermal ties on custom pads is a known limitation of KiCad. And it's been discussed on the dev mailing list. I can't recall if this was already merged, but I believe Jeff was working on it and we can expect it in v6.

Can you give an example of the situation where this is a problem for you?

The current footprint was merged by the head librarian at https://github.com/KiCad/kicad-footprints/pull/1709. So I'm not inclined to merge this PR. We're in a valley with nice custom pads in KiCad but not all the quite the infrastructure to fully use those pads.

charithjperera:
@evanshultz - the problematic situation was when I tried to connect the tab to a zone, the zone went right over the tab of this footprint and would have made it difficult to solder.
I suppose the tradeoff is whether other people may experience this issue prior to v6 and if there are any potential disadvantages of using two rectangles like this going forward.

PR #2116, opened 2020-03-02
Amphenol m8 s 03 pmmr sf8001

cronJ:
Add Amphenol LTW M8 shielded 3-pin connector

https://www.amphenolltw.com/2012download/2D%20PDF/03_M%20Series%20Sensor%20Connectors/M8S-XXPMMR-SF8001.pdf#page=2

image

---

---


chmorgan:
@chschlue how do we approach adding new libraries like this? Amphenol does make a lot of connectors and we don't have a specific library for those yet.

cronJ:
> * How did you get overall part dimensions? The drawing doesn't seem to have a way to show the face to the pins anywhere...

I measured the dimensions on the real part. This footprint is also on an already manufactured board and fits.

chmorgan:
Hi @cronj. Looks like the left and right courtyard are offset from center of silk not fab

Silk lines look ok, a little inconsistent in the notches vs. the straight parts. Orientation looks good now.

chmorgan:
@chschlue, @cpresser thoughts on adding this part to the this proposed Amphenol library?

chschlue:
Looks good to me.

cpresser:
> @chschlue, @cpresser thoughts on adding this part to the this proposed Amphenol library?

Good idea. Amphenol has lots of connectors.

cpresser:
@chmorgan 
I think you need to check submissions more thoroughly. I checked this one since the screenshot looked wrong. Thus, wanted to see if it got merged like shown in the screenshot.

Then I checked some dimensions, and its not looking good :crying_cat_face: 
image
* this should be 7mm according to the datasheet.
* Another body-dimension is 13.2mm where the datasheet has 12.8
* The courtyard could be smaller (tight around the part), but that is not an real issue here: the biggest part of the device is not going to be on the PCB.
* If dimensions are missing in the datasheet another method of verification is needed. Either ask the manufacturer for more details, search for better docs, or measure in the 3D-Model. There is a step-file available here: https://www.amphenolltw.com/de/p2-products-detail.php?product_series_sn=15&product_classify_sn=33&sn=5151
* My estimation of the length is more like 18.5mm 
image With proper use of FreeCAD (or any other CAD-Tool) there will be a more accurate measurement. But I am quite confident that it won't be 19.8mm

chmorgan:
@cpresser I can appreciate that. I'm trying to be as careful as possible but I'm also not a cad person.

In terms of adjusting I can adjust the width to be 7mm and shorten the body up slightly to match what you've measured if that makes sense. I can bring the courtyard in if you think it won't cause an issue with interference.

cpresser:
**We usually ask the original contributor to do the fixes. Its their contribution after all.**
@cronJ Do you want to make another PR to fix the issues I found with this submission?

Regarding CAD, that is something one can learn :)
The above screenshot is actually garbage. I took the wrong part (Female vs. Male Plug).

I redid the whole thing in Fusion360 (because I know how to use that - contrary to FreeCad).
image
Workflow is:
* Import .step file
* Add offset-plane that does not cut the part (just eyeball the distance)
* Project part onto that new plane
* Use dimension-tool to add dimension (ignore the over constrained warning).

I am pretty sure a similar workflow also exists in FreeCAD, I just didn't bother to learn it yet.

cronJ:
Yes, I'm already working on the fixes. I will create an PR when finished.

chschlue:
I _think_ one of the easier ways to do this in FreeCAD is
* import STEP
* create 2D views in the TechDraw WB as needed (0.19pre; 0.18 has several ugly shortcomings in that regard)
* add dimensions

but I'm not an MCAD native either.
Quite possible this is actually a stupid workflow.

PR #2117, opened 2020-03-02
Create Hirose_FH33-28S-0.5SH_1x28-1MP_P0.50mm_Horizontal.kicad_mod

cronJ:
Add 28 pin Hirose FH33

https://www.hirose.com/product/document?clcode=CL0580-1304-0-10&productname=FH33-10S-0.5SH(10)&series=FH33&documenttype=2DDrawing&lang=de&documentid=0000990713

image

---

---


chmorgan:
Hi @cronJ I think the same courtyard to fab/feature distance applies to this PR as well.

chmorgan:
@cronJ looks like the courtyard needs some slight rounding to align to the 0.01mm grid via the travis output.

chmorgan:
@cronJ courtyard is way out on the top. Per llc F5.3 it should be 0.5mm from the nearest pad or fab line that represents the part itself. Maybe you are pushing it out because you pushed the fab pin 1 indicator out? See Hirose_DF52-15S-0.8H_1x15-1MP-P0.80mm_Horizontal as an example of what yours should look like.

evanshultz:
I would prefer that easily-scripted footprints be scripted, and this is certainly one of them. It might even be possible to use the generic single-row script. Building off of an existing script, or using the generic one, means a lot of the work is done for you and it simplifies the decisions about how to build the footprint while at the same time making review easier. That being said, wWhile we encourage contributors to script we won't reject a valid footprint that isn't scripted.

PR #2119, opened 2020-03-02
Create TE_5767171-1

cronJ:
Add TE_5767171-1 (also known as Mictor38)

https://www.te.com/commerce/DocumentDelivery/DDEController?Action=showdoc&DocId=Customer+Drawing%7F5767171%7FB2%7Fpdf%7FEnglish%7FENG_CD_5767171_B2.pdf%7F5767171-1#page=2

image

---

---


cronJ:
Numbering is defined here:

image

chmorgan:
Hi @cronJ. On the left hand side courtyard did you push that out slightly to keep clearance with the fab line that is sticking out to the left? Is that fab line sticking out to indicate where pin 1 is? I'm thinking that it would make sense to make the fab lines match the part outline, then you can make the courtyard consistent and based on the fab, eg. we don't want to push the courtyard out due to a fab feature for pin 1 indication.

In terms of indicating pin 1 you can use a triangle or some other feature on both fab and silk. If its silk you would put it outside of the fab outline so it can be seen after the part is placed.

I don't have a clear idea of what this connector looks like from real world usage so apologizes for asking dumb questions.

cronJ:
Here is the [part page](https://www.te.com/usa-en/product-5767056-1.html?te_campaign=oct_glo_manufacturer&elqCampaignId=26136&te_bu=Cor&te_type=other) so you can see how the real part looks like.

chmorgan:
@cronJ can you check the changes I made to the fab, centering, courtyard etc? Once you get that stuff down it should take a lot less effort to get your footprints through the process.

chmorgan:
@cronJ note that the travis check on footprint center is incorrect as this part has a plastic tab only on one side. If you check the footprint and packaging guidelines you can see that the pick and place clip is centered on the part itself.

PR #2120, opened 2020-03-02
Create Infineon_PG-DSO-8-24.kicad_mod

cronJ:
Add Infineon PG-DSO-8-24 package

https://www.infineon.com/cms/en/product/packages/PG-DSO/PG-DSO-8-24/

---

---


cp-aquila:
Hi,

unless you changed your color-scheme the screenshot looks wrong:
* Silkscreen is on the same position as F.Fab (Silk should be visible after assembly)
* Courtyard is smaller than F.Fab
* Pin1-Maker on silkscreen does not conform to library or IPC rules

I suggest you use the footprint-generator. This is similar to a SOIC package and thus can be generated with the IPC generator. You can start from `Infineon_PG-DSO-8-27` and remove the exposed pad.  Then the review is really simple because just the yaml file needs to be reviewed since the generator itself is trusted.

See here: https://github.com/pointhi/kicad-footprint-generator/blob/master/scripts/Packages/Package_Gullwing__QFP_SOIC_SO/size_definitions/infineon.yaml

myfreescalewebpage:
@cronJ any news following the previous comment ?

PR #2123, opened 2020-03-02
Create Converter_DCDC_TRACO_TDN_5-xxxxWISM_SMD.kicad_mod

cronJ:
Add Tracopower TDN 5-xxxxWISM SMD footprint

https://assets.tracopower.com/20200225170127/TDN5WISM/documents/tdn5wism-datasheet.pdf#page=3

image

---

---


chmorgan:
Hi maintainers. I reviewed this footprint against the data sheet.

The data sheet does not provide a recommended footprint. This footprint meets the requirements of the pins on the part and is dimensionally correct when compared to the data sheet.

chschlue:
So the footprint pads are the same size as the package pads?

chmorgan:
@chschlue they are slightly longer than package pads (which is what I would expect), and not by a significant amount, like 1.2mm in length for each pad or so, so maybe 0.6mm on each side, but I'd have to check again to confirm this.

PR #2124, opened 2020-03-02
Create Converter_DCDC_TRACO_THD_15-xxxxWIN_THT.kicad_mod

cronJ:
Add Tracopower THD 15WIN THT footprint

https://www.tracopower.com/products/thd15win.pdf#page=3

image

---

---


chmorgan:
Hi @cronJ, thank you for your contribution. I had some feedback on the footprint that could help it be ready for merge:

- Per F4.2 pin one should be in the top left corner
- The outline dimensions don't appear correct. For example, I would have expected 2.5mm from the pin center on pin 23 to the courtyard and silkscreen lines but its about 1.5mm. Same for the outline and courtyard on the bottom. Same on the left and right sides.
- The 0.8mm hole size looks ok but maybe someone else can comment on whether it should be a little tighter to the pin size of 0.5mm.

chmorgan:
@cronJ can you move and rotate the two REF** entires to match how other parts do it? For example the Converter_DCDC_TRACO_TMR-1-xxxx_Dual_THT? The FAB one can go in the middle but the Silk one should be placed to be visible when the part is placed. And can you move the part name to match as well? Once that's in place it should be good to merge.

PR #2127, opened 2020-03-05
Add switch C&K JS202011JCQN

mibe:
Add SMD DPDT switch C&K JS202011JCQN.
[Datasheet (pg. 8 / 56)](https://dznh3ojzb2azq.cloudfront.net/products/Slide/JS/documents/datasheet.pdf)
Unbenannt

~~Pin 1 is bottom left as with the datasheet.~~

---

---


poeschlr:
Just because the datasheet shows the drawing in a particular direction does not negate the requirement of pin 1 being at the top left corner

mibe:
Well, F4.1 states that:

> Where conflict exists between datasheet recommendations and KLC requirements, then (in the general case) the datasheet should take priority.
> [...]
> If the KLC is at odds with the datasheet, follow the datasheet, and note this difference to the KiCad library team when submitting your contribution.

Why is this not the case here?
The footprint could be rotated to show pin 1 at the top, though.

chschlue:
The datasheet doesn't _recommend_ a specific orientation, it just happens to be drawn that way.
That's very different from a datasheet recommending a specific land pattern, for example.

mibe:
Oh, I see. Thanks for the explanation!

mibe:
Alright, I swapped the pad numbers. pin 1 is top left now.

mibe:
>     * Any reason the pin numbering doesn't match the datasheet? You have 4, 5, 6, datasheet has 5, 6, 7.
In this case the pin numbers wouldn't match the symbol ```SW_DPDT_x2``` of the ```Switch``` library.

>     * Am I correct to assume that its ok to install the switch in any orientation as the center are the common pins?
That is correct. The center pins are the common pins.

cp-aquila:
At least in the screenshot there is no pin1 marker on silk or Fab.

mibe:
That's true, there are no pin 1 markers.
Fab: F5.2: "Footprint polarisation / location of pin-1 is drawn": My interpretation was that this rule only applies to polarised parts. This part is not polarised, so I didn't draw the pin 1 marker.
Silk: The [example pic for non polarised parts](https://kicad-pcb.org/img/klc/silk_non_polarized.png) does not show the pin 1 designators, so I also omitted it on the silk screen.

cpresser:
We are discussing this rule (without result yet): https://github.com/KiCad/kicad-website/issues/422

I would argue that this part is 'polarized'. In the sense that the switch has a default position. The datasheet has detailed drawings of the tape the and how the switches are oriented in the tape.

The Fab layer is intended to help the SMT operator loading the assembly machine. The silkscreen can be used to check proper orientation. Unless there is a mark they might put the switch on the board in any direction.
This will be really annoying for AOI or any kind of test that relies on a default NC/NO position of the switch.

PR #1993, opened 2019-12-06
Create Fuse_1206_3216Metric_Pad1.1x1.75mm.kicad_mod

eclecticc:
A 1206 fuse for Littlefuse 12V Polyswitches: https://www.littelfuse.com/~/media/electronics/datasheets/resettable_ptcs/littelfuse_ptc_lorho_smd_datasheet.pdf.pdf

image

---

All contributions to the kicad library must follow the [KiCad library convention](http://kicad-pcb.org/libraries/klc/)

Thanks for creating a pull request to contribute to the KiCad libraries! To speed up integration of your PR, please check the following items:

- [x] Provide a URL to a datasheet for the footprint(s) you are contributing
- [x] An example screenshot image is very helpful 
- [x] If there are matching symbol or 3D model pull requests, provide link(s) as appropriate
- [x] Check the output of the Travis automated check scripts - fix any errors as required
- [x] Give a reason behind any intentional library convention rule violation.

---

Be patient, we maintainers are volunteers with limited time and need to check your contribution against the datasheet. You can speed up the process by providing all the necessary information (see above). And you can speed up the process even more by providing a dimensioned drawing of your contribution. A tutorial on how to do that is found here: https://forum.kicad.info/t/how-to-check-footprint-correctness/9279 (This is optional!)

PR #2128, opened 2020-03-06
Add scripted footprints for castellated two terminal fuses

poeschlr:
Fuses are available with castellated terminals as demonstrated by https://github.com/KiCad/kicad-footprints/pull/1993

The same size definitions as all our normal footprints are used.

script pull request https://github.com/pointhi/kicad-footprint-generator/pull/505


evanshultz:
Thanks!

PR #1055, opened 2018-10-30
Add new footprint for the Multicomp BC-2001 CR2032 battery holder

Cylindric:
I am adding a Multicomp BC-2001 CR2032 battery clip footprint and model.

Note that there is no symbol to go with this part, because the correct symbol is the generic battery cell symbol.

[Datasheet](http://www.farnell.com/datasheets/1505853.pdf)

The model was created using Fusion360, exported to STEP and converted using FreeCAD StepUp tools.

battery-footprint

[Footprint PR](https://github.com/KiCad/kicad-footprints/pull/)
[Package PR](https://github.com/KiCad/kicad-packages3D/pull/439)
[Package-Source PR](https://github.com/KiCad/kicad-packages3D-source/pull/184)

- [x] Provide a URL to a datasheet for the footprint(s) you are contributing
- [x] An example screenshot image is very helpful 
- [x] If there are matching symbol or 3D model pull requests, provide link(s) as appropriate
- [ ] Check the output of the Travis automated check scripts - fix any errors as required

PR #2129, opened 2020-03-07
Add new footprint for the Multicomp BC-2001 CR2032 battery holder

chmorgan:
See https://github.com/KiCad/kicad-footprints/pull/1055

Adjusted the courtyard based upon cr2032 having a 20mm diameter and placing a temporary circle into the part. If I had a picture of the battery in this part it could be adjusted but there is no guidance from the mfg on where the battery sits. This is a few mm conservative.

image

Here it is @poeschlr 

---

---


cnieves1:
Hi @chmorgan,
I've used a part like this before....
Coin battery is inserted until it meets one of the surfaces marked with red arrow:
Captura de pantalla de 2020-03-17 22-53-31

Another issue is that negative contact is made only with solder paste (pin 2). Depending on which solder paste you use, it may get rusted, affecting reliability. 
In order to avoid this, I prefer to use a negative contact, like:
  https://www.keyelco.com/category.cfm/Negative-Contacts/Compact-Negative-Contact-SMT/p/405/id/925/c_id/1035
But pin 2 should be removed if using this contact, so it woulr be nice if you could make two footprints: one as you did, and another without pin 2...

F5.3 requires the courtyard to have extra clearance for mating connectors. In this case, I think it should have extra clearance for inserting the coin battery...

chmorgan:
Hi @cnieves1. Someone else had opened up https://github.com/KiCad/kicad-footprints/pull/1055 and I picked it up to correct the courtyard to account for where the battery goes. If you have some interest in picking it up and improving it I'm all for it. I don't have the drive to do so at the moment though...

Franck78:
Adding the arc. Room to slide in and out a battery cell determined empirically because obviously a little room is necessary for the cell (here represented by the 20mm diam with circle not part of the footprint)

If or when done,
can close 1055 and merge the ready 3dmodel (349) and 3ssource (184)

bh
[BatteryHolder_Multicomp_BC-2001_1x2032.kicad_mod.zip](https://github.com/KiCad/kicad-footprints/files/5277113/BatteryHolder_Multicomp_BC-2001_1x2032.kicad_mod.zip)

PR #2132, opened 2020-03-09
LED_SMD: add LiteOn LTST-C295K series

karlp:
dual independent surface mount leds.

Signed-off-by: Karl Palsson <karlp@tweak.net.au>

Datasheet: http://optoelectronics.liteon.com/upload/download/DS22-2009-0072/LTST-C295KGKRKT.PDF
Screenshot from 2020-03-09 22-58-36

Silk is 0.2mm clearance from pads, courtyard is 0.2mm.  Package is "0603" not less, not more, so .15 seemed too tight, but 0.25 unnecessarily wide.


karlp:
Will check the travis notes tonight and update

karlp:
Violating F5.1 seems in line with guidelines 2a, I used .1mm silk as I feel 0603 sized led counts as "compact" but I'm happy to change it if it's a problem.

karlp:
If this needs to be repushed as a new PR, let me know, something's happened to make the travis checks not run?

chmorgan:
Hi @karlp, what version of kicad did you use to create this part? I'm having trouble loading the footprint here under kicad 5.1.5 on OS X.

karlp:
Version: 5.99.0-unknown-r17833-741783c6, debug build

I'm on the current nightlies.

chmorgan:
@karlp we are discussing internally what to do. It looks like the nightlies are saving files in a format incompatible with older kicad versions such as the one I'm using. Not sure what the fix will be at this point. One option could be to drop back to an older version or wait until its resolved in the nightlies.

karlp:
I noticed also that it's leaving timestamp droppings all over the place, even without changes, which is horrible.  Personally, I'd consider breaking footprint compatibility with 5.x to be a regression of the worst form.  I can't go running multiple versions just to contribute on this I'm afraid, it's hard enough to make things KLC compliant and self contained without having to run special blessed versions as well.

chmorgan:
@karlp so the conclusion is that you'd have to use the stable version of kicad when submitting parts. Otherwise we'll end up with parts that have half baked formats etc. Is it possible for you to use a stable version? When kicad 6.x is launched it will save to the new format but be able to load from the old format (but not save back to the old format). It's just that by using the nightly you are using an intermediate codebase that isn't complete yet.

karlp:
Well, thanks for the feedback.  I've got ~zero intention of using two versions I'm afraid.
So what's the plan when 6 comes out then? no-one using 5 will be able to use the new footprints? really?  This does not sound like a good plan.

chmorgan:
Hi @karlp. I think libraries would be fixed in time and the old versions only supportable via the old version which might not be the worst no?

PR #2136, opened 2020-03-14
Converter_DCDC_TRACO_TDN_5-xxxxWISM_SMD adjusted pad sizes to IPC-735…

chmorgan:
…1 standard using library expert pro, changed pads to have no rotation

---

---


chmorgan:
@chschlue and here is the pad size adjustment per IPC-7351

chmorgan:
@chschlue I used the defaults which selected a density of nominal

On the silk overlapping with fab I've seen that on other parts. Is there something in the KLC that would justify moving the silk outside of the fab beyond the implication that the fab is the part outline and F5.1 entry 4 says it has to be visible? I'm wondering if we should be considering a KLC update.

I see the silk to copper distance in F5.1. Would it make sense to update the F5_1 rule to check for that clearance or are there enough exceptions to warrant not doing so?

image

chmorgan:
Realized I had pushed this branch to the main repo last night, I'll correct and remove when this is merged or otherwise concluded. New PR from my fork is at https://github.com/KiCad/kicad-footprints/pull/2138

chschlue:
Ah yes, I noticed that when pulling this branch to do the review but then forgot to mention it.
It's also a good idea to actually remove the upstream push url from your git remotes to prevent accidents.

Silk should be visible for placement and debugging, that's one aspect.
Another one is: Silk is an actual layer of paint on the board and so destroys planarity, which is why SMT parts generally shouldn't have any underneath (yes, I'm aware that this particular package has its leads protruding downward).

If you find KLC rules which are unclear, don't convey their rationale or are plain stupid, you're welcome to open issues and/or PRs on kicad-website.
This is something only the new guys can do because you simply turn blind to this sort of thing over time ;)

chschlue:
@chmorgan can this be closed?

chmorgan:
@chschlue yep, closing

PR #2137, opened 2020-03-14
Added footprint for ASMB-MTB0-0A3A2

johnfwhitmore:
Added footprint for ASMB-MTB0-0A3A2

---

Datasheet - https://www.farnell.com/datasheets/2608172.pdf

Screenshot from 2020-03-14 12-49-18


johnfwhitmore:
Sorry was using a later version of Kicad. Trying to install latest stable on distro and failing as it's not happy with other package versions.
Forget it

chmorgan:
@johnfwhitmore what do you mean by a later version of kicad? An older version or a newer one?

chmorgan:
@johnfwhitmore mind holding on to see if there is a fix coming to kicad? It sounds like very new nightlies are resulting in this issue but that shouldn't be the case. We are discussing internally to see what the outcome is. Would hate to lose your contribution due to this issue with compatibility that may be resolved in the near future.

johnfwhitmore:
Newer one. I installed from source but had a later version then 5.1.5. Just managed to install the 5.1.5 version so redoing the change.

PR #2138, opened 2020-03-14
Converter_DCDC_TRACO_TDN_5-xxxxWISM_SMD adjusted pad sizes to IPC-735…

chmorgan:
…1 standard using library expert pro, changed pads to have no rotation

---

---


chmorgan:
image

chmorgan:
@chschlue damnit, sorry man. Pushed fixes for that.

chmorgan:
@chschlue is this ok? I moved the pin 1 indicator for fab inside of the outline to avoid having to expand the courtyard.

image

chmorgan:
@chschlue I think the courtyard makes sense, I missed the top line that was 0.05mm too close.

Update image with bevel and pushing updates. I kept the outer marking on the silk to ensure it would be visible when the part was placed, per KLC (and what I'd like to see when assembling a board).

image

chmorgan:
@chschlue is the idea that the courtyard should be exactly 0.25mm away? I'm using an 0.05mm grid here, should I be using a smaller grid or assigning by value to get a precise value?

chmorgan:
@chschlue also, I'm seeing 0.25mm from the pad to courtyard on the top and bottom, not 0.275mm.

Switched to an 0.1mm grid and using the measurement tool (rather than the zero origin approach). Only had to move the right and left lines, the top and bottom were exactly at 0.25mm from the pads and the left and right are 0.25mm from the fab lines.

chschlue:
CrtYd should be as accurate as possible while still fitting a 0.01mm grid.
This means in x direction ±4.8mm is right on target while in y direction ±8.12 or ±8.13mm is as good as it gets.

chmorgan:
@chschlue are you measuring from the center of lines or the edges of lines? I'm using the edges but if I use the center I get measurements closer to yours.

chschlue:
See https://github.com/KiCad/kicad-footprints/pull/2138#pullrequestreview-374737944

chmorgan:
@chschlue k, let me update and repush, that explains the differences I was seeing.

chmorgan:
@chschlue alright, that *should* do it

chschlue:
Done.

One more thing:
3slyk9
:)

chmorgan:
Hah. Seriously though it’s ok to force push so we don’t have dozens of fix commits to squash?

chschlue:
The problem with force-pushing during review iterations is that it makes it unnecessarily hard for the reviewer to track the actual changes.
Example: If I tell you that courtyard is off by 0.1mm and you add a commit fixing that, I can immediately see that you moved those lines and those lines only.

Squashing all those commits afterwards is just a single click in GH's web UI.

chmorgan:
@chschlue k, I'll avoid in the future

PR #2141, opened 2020-03-15
added ESP32-S2 SoC Module

morganrallen:
Footprint from the new ESP32-S2-WROVER module. Based on the original ESP32 WROVER with changes for new pins and changes in pin spacing.
https://www.espressif.com/sites/default/files/documentation/esp32-s2-wroom_esp32-s2-wroom-i_datasheet_en.pdf

image


chmorgan:
@morganrallen still need to update the 3d model path on the footprint. You can look at another one and copy the syntax approach in the part settings.

cpresser:
Judging from the picture: Silk is overlapping the Pads. On the top it seems like silk is not properly aligned.

morganrallen:
@cpresser Whoops, I fixed those and didn't upload new image.
image

chmorgan:
@morganrallen can you fix the 3D path issue that travis spotted? That would let us go to the next step in review here and might be one of the final gates.

chmorgan:
Courtyard is closer than 0.25mm from pads and fab (bottom is too close, sides are too far), see https://www.kicad-pcb.org/libraries/klc/F5.3/

Silk is still overlapping fab in some areas, its ok to move it out  a little more.

Can you put the a url to the datasheet for the part in the part description?

Pads on the bottom have a rotation of 90 degrees. For whatever reason we like all pads to have 0 rotation, if you could swap the height and width there and zero the rotation out that would be great.

EP looks a little bit too large no? Recommended to be 4.1mm x 4.1mm in the datasheet but you have 5x5. Any reason for the increase?

The length of the fab looks slightly off. The part drawing says 31mm but the length of the fab line is 31.045mm. The bottom line looks like it should be at a y offset of 11.825 (not 11.725) if I look at the center of pad 27 and add 1mm as shown in the drawing. I could be wrong here.

Confirmed that:
- Pinout is correct
- EP is in the correct position
- Pads are the correct size
- Pad spacing is correct
- Fab width looks correct

chmorgan:
@morganrallen thanks for the submission and the updates. I made some final adjustments to the silk but otherwise everything else checked out dimensionally.

cpresser:
Tbh, I think this still needs some work. It is not consistent with the existing part:
- [ ] Keepout area completely missing
- [ ] Pin1 Styles in F.Fab and F.SilkS differ without reason
- [ ] The Slikscreen after every 5th pin is nothing we have done in this library before. I rather not have it in the footprint, the PCB designer can add debugging help in the board design if required.
- [ ] I suggest to move `Ref**`on F.SilkS down. I its current position it is not on the PCB in most designs. This also adds consistency.

image
image

chmorgan:
Hi @cpresser. I can make the adjustments here as I introduced most of what you've raised here. The ticks are helpful for large pin parts like this. I actually like them on most any part where you can't easily visually count pins. I can remove them and make the other adjustments. I'll open another PR.

In terms of review did we just miss each other in terms of reviewing? If you want to assign yourself as the reviewer that would let me know that I should avoid overlapping review :-)

chmorgan:
Hi @cpresser. The keep outs on the other espressif parts are way off. Espressif recommends 15mm, those keep outs are only 5mm. https://www.espressif.com/sites/default/files/documentation/esp32_hardware_design_guidelines_en.pdf figure 13

cpresser:
> 
> In terms of review did we just miss each other in terms of reviewing? If you want to assign yourself as the reviewer that would let me know that I should avoid overlapping review :-)

No, this is 'your' PR. 
I just want to help with a few things that I spotted. Consider my input as regular comments.

Regarding the keepout: Looks like the WROOM-32 footprint should be fixed.

chmorgan:
@cpresser yep, looks like that footprint has far too small of a keep out

morganrallen:
> No, this is 'your' PR.

And here I thought it was mine....

@chmorgan are you running away with this or should I address the issue above?

chmorgan:
@morganrallen I have a PR open to correct and @cpresser has been reviewing. Thanks for asking though, it helps to avoid overlapping work!

PR #2144, opened 2020-03-16
Added HTSSOP-14-1EP_4.4x5.0mm_P0.65mm_EP3.4x5.0mm_Mask3.0x3.1mm

cnieves1:
Generated using: https://github.com/pointhi/kicad-footprint-generator/pull/510
Datasheet: http://www.ti.com/lit/ds/symlink/lm5161.pdf#page=34
Package drawing: http://www.ti.com/lit/ml/mpds370/mpds370.pdf
Custom TI name: R-PDSO-G14 (PWP)
imagen
Captura de pantalla de 2020-03-16 11-27-54

---

---


chmorgan:
Hi @cnieves1, can you explain why the drawing is showing a smaller EP than the part does?

cnieves1:
I don't fully understand the issue...

Central pad dimension is specified by solder mask size, as other footprints do, like, for example:
  HTSSOP-20-1EP_4.4x6.5mm_P0.65mm_EP3.4x6.5mm_Mask2.4x3.7mm
Copper under central pad is bigger than central pad in order to improve thermal transfer.
If you look at the footprint with thermal vias (HTSSOP-14-1EP_4.4x5mm_P0.65mm_EP3.4x5mm_Mask3x3.1mm_ThermalVias):
Captura de pantalla de 2020-03-16 22-46-45

Is that what you are asking?

chmorgan:
@cnieves1 ahh thats exactly it. For whatever reason the footprint isn't rendering mask correctly. It's rendering mask and then copper on top, thanks for pointing that out.

chmorgan:
@cnieves1 the recommended land pattern in the datasheet doesn't match with the lands in the part. For example the pads are 1.575mm long but the datasheet indicates 1.78mm. Am I looking in the wrong place?

cnieves1:
No, you are right. 
In this case, I think the footprint generator  calcules the land pattern based on physical dimensions (both body and legs), using IPC rules.
Therefore the generated footprint usually doesn't match the manufacturer's recommended footprint.
If the IPC rules built into the generator are right, the footprint would be ok.
In other PRs, the general recommendation is using this generator whenever possible, so I hope it is doing the right thing...

evanshultz:
@chmorgan 
Correct. See readme at https://github.com/pointhi/kicad-footprint-generator/tree/master/scripts/Packages/Package_Gullwing__QFP_SOIC_SO.

For scripted entries you will review if the package dimensions entered into the script match in the datasheet, disregarding the recommended footprint. If they do, and the footprint doesn't have an weird error that stands out, merge the footprint and add a comment on the script repo that the footprint has been merged.

chmorgan:
@evanshultz appreciate the tip and it makes sense that the the values might not match exactly.

PR #2154, opened 2020-03-17
Add Murata 78250JC

alfredosola:
Murata 78250JC. No 3D model unfortunately, as I lack the required skills.
[Datasheet](https://www.murata-ps.com/datasheet?/data/magnetics/kmp_78250j.pdf)


chmorgan:
Hi @alfredosola would you mind signing the cla and adding the anticipated 3D part path to the part? Even though there isn't a 3D part today we want the path to be in there and correct when/if the 3D part is added.

alfredosola:
Regarding the CLA, happy to comply but something is not working for me. Could it be that I have my e-mail set to private? Sorry.

Captura de pantalla 2020-03-17 a las 19 49 01

chmorgan:
Hi @alfredosola, no idea. We would require that before being able to accept the contributions. Any chance you can try it without private email? It's one of those gates that we have to have to avoid worrying about copyright around changes.

alfredosola:
> Any chance you can try it without private email?

I have tried to set my email as public, tried to give cla-assistant.io all the additional permissions it wanted, and clicked on "Sing in to Github to agree" more times than I can count, every time getting "You have signed the CLA for KiCad/kicad-footprints", but here it is still saying it is not signed yet. I am lost. Any ideas or alternative ways to sign the CLA?

chschlue:
Do your commits use an address that's linked with your Github account?

alfredosola:
> Do your commits use an address that's linked with your Github account?

This must be it. I hadn't realized the commits bear a different e-mail. I added it to github and that seems to make CLA Assistant happy. Sorry about that!

alfredosola:
Hi, thanks for your review. I have tried to address all the points raised. Looking forward to your review.

> * Why are the pads rounded vs. square? (I'm not sure that this matters but I'm curious)

Well, given that the manufacturing process is going to be unimpacted, I think rounded corners help in a few tiny, almost unmeasurable ways.

Firstly, during manufacturing they give solder a nice smooth line to stick to, so that solder easily flows to all the pad.

Secondly, when in use, and depending on the current flowing (this is a transformer, after all) and the PCB layout, avoiding corners may help in preventing tiny losses due to heating and (depending on frequency) radiation.

I believe it would be very hard to measure those benefits as they would amount to infinitesimal gains. But since there is no cost and they are best design practices, I also can't find a reason not to go for them.

chmorgan:
@alfredosola it looks like the center of the side courtyard to the edge of the pad is a little bit more than the 0.25mm we are looking for.

The center of the top and bottom courtyard looks to be 0.25 from the center of the silk line but it should be from the center of the fab line instead.

For example the courtyard should be 0.25mm away (for an SMT part) from the nearest feature (pad, part edge etc).

Can you move the part name below the part as we do with other parts?

The pin 1 indicator silk looks odd due to the line widths not being consistent there.

chmorgan:
@alfredosola can you review? I modified the silk slightly so it wasn't overlapping the courtyard (so its easier to see the courtyard), and out from under the fab to avoid having the silk cause the part to be uneven on the pcb.

PR #2158, opened 2020-03-19
ESP32-S2-WROVER: Added keepout on DWGS layer for espressif recommende…

chmorgan:
…d clearance of 15mm. Removed tick lines. Move courtyard around clearance area.

Figure 13 in https://www.espressif.com/sites/default/files/documentation/esp32_hardware_design_guidelines_en.pdf shows clearance drawing

image

---

---


chmorgan:
@cpresser I left the pin 1 indication as moving the fab indication would shift the courtyard out and moving the pin 1 silk indicator in would result in the silk being partially under the footprint which would cause unevenness as this is an smt component. Otherwise I think everything else was addressed here.

cpresser:
How about that?
- Keep-out also under the antenna
- Pin1 and Silkscreen as `ESP32-WROOM-32` 

image

chmorgan:
@cpresser can you re-review?

cpresser:
I checked again,
but I don't see consistency between `ESP32-S2-WROVER` and `ESP32-WROOM-32`.
The later has silkscreen not only on the edges, but all around the antenna part. And the chevron on F.Fab looks different.

chmorgan:
Hi @cpresser. I can add that silk around the edges but imo the triangle is a bit easier to spot visually so I went that way vs. consistent but less clear with the other parts.

chmorgan:
@cpresser I was initially going to push back a bit as I think we should get fixes into the library as long as they are moving in the right direction. In any case I decided to copy the pin 1 indicator from one of the other parts and correct the upper silk. I did leave the fab line for the antenna as that could be a helpful visual indication.

PR #2055, opened 2020-01-14
Added USB_A_Molex_67643_Horizontal

blackvladimir:
Molex USB A connector. Similar to USB_A_Stewart_SS-52100-001_Horizontal but little bit longer.

molex

datasheet https://www.molex.com/pdm_docs/sd/676433910_sd.pdf

two dimensions are not clear from datasheet so I took them from step file ( https://www.molex.com/molex/part/partModels.jsp?&prodLevel=part&partNo=676430910&channel=products )

width
sheet

And dimensions of footprint:

dim

---

All contributions to the kicad library must follow the [KiCad library convention](http://kicad-pcb.org/libraries/klc/)

Thanks for creating a pull request to contribute to the KiCad libraries! To speed up integration of your PR, please check the following items:

- [x] Provide a URL to a datasheet for the footprint(s) you are contributing
- [x] An example screenshot image is very helpful 
- [ ] If there are matching symbol or 3D model pull requests, provide link(s) as appropriate
- [x] Check the output of the Travis automated check scripts - fix any errors as required
- [ ] Give a reason behind any intentional library convention rule violation.

---

Be patient, we maintainers are volunteers with limited time and need to check your contribution against the datasheet. You can speed up the process by providing all the necessary information (see above). And you can speed up the process even more by providing a dimensioned drawing of your contribution. A tutorial on how to do that is found here: https://forum.kicad.info/t/how-to-check-footprint-correctness/9279 (This is optional!)

PR #2161, opened 2020-03-20
Small fixes on some USB connectors

cpresser:
While working on https://github.com/KiCad/kicad-footprints/pull/2055 i looked at some other USB footprints.

* USB3_A_Plug_Wuerth_692112030100_Horizontal
  * Improve description
* USB_A_CNCTech_1001-011-01101_Horizontal
  * Add Pin1 Marker
  * Improve description

image


chmorgan:
@cpresser should we add more silk around those parts to make it easier to see the borders? Those are oddly lacking in silk showing where the part is.

cpresser:
I suppose we could. 
But I don't see a usage scenario where the silk actually gets printed. Both are board-edge connectors. That is why I decided not to add more silkscreen in the first place.

chmorgan:
@cpresser gotcha, makes sense!

PR #2162, opened 2020-03-20
Add skyworks SE5004L footprint

matthuszagh:
https://www.skyworksinc.com/-/media/SkyWorks/Documents/Products/601-700/SE5004L_202393B.pdf

se5004l

The datasheet is very particular about this PA footprint since it generates a lot of heat. For that reason I've made the ThermalVia version the default and omitted that modifier from the name. I can't imagine any scenario in which you wouldn't use the thermal vias; the component would overheat and not work. Please let me know if any concerns/questions.

1. I've violated the thermal via annular ring requirement since it is connected to the EP and satisfies the requirement anyway.
2. I believe the mask errors are related to some of the pads overlapping the exposed pad, which is specified in the datasheet.

---

---


matthuszagh:
I've changed all ground footprint numbers to be the same so they can be directly connected, in accordance with the KLC. I've also added custom mask and paste since some pads are connected directly to the EP.

matthuszagh:
I'm considering making another version of this footprint with a less dense array of thermal vias: either 3x3 or 4x4. The current version is the one specified in the datasheet and should therefore (I believe) be kept. However, the vias are too closely spaced for many fab houses. Any thoughts on this idea?

PR #2174, opened 2020-03-24
Vishay BFC2808 variable cap footprint

evanshultz:
https://www.vishay.com/docs/28527/bfc2808-75mm.pdf

image

A few things to consider:
1. `Capacitor_THT` seems the best place for this footprint as I don't think we need a new library just for trimmer caps. Let me know if there is a better place for this footprint.
1. I'm open to suggestion on the footprint name. All other cap footprints start with `C`, so I tried to maintain that and chose `CVariable` as the symbol is named `C_Variable` but the footprint libraries often use camel case. We could use `CTrim` instead or something else?
1. Orientation is odd. Since there are essentially just two pins, I put pin 1 on the left like our other caps.
1. I named the footprint in this way because there doesn't seem to be a clear MPN system which separates top and side versions with just a single character. I also didn't want to include the NPTH for bottom-side adjustment.
1. The diameter across the 4.5mm tall piece on the left side measures to ~8.5mm, so it sticks ~1mm out from the main body.
1. I've tied both of the rotor pins together, even though they don't have to be, so that this footprint just works with the appropriate symbol having pins 1 and 2. If that's an issue let me know.

---

---


evanshultz:
Dimensioned drawing:
image

You will have to do a little math to match up the numbers since the orientation doesn't quite match the datasheet and I don't know how to draw angled dimension lines in this workbench in FreeCAD.

The 0.3mm courtyard clearance on the right side is because of the pick points in this workbench: it is actually the correct 0.25mm.

PR #2175, opened 2020-03-25
Add longer SIP9 package (based on existing SIP-9_21.54x3mm_P2.54mm)

dvdesolve:
I've added quite longer SIP9 package from existing one. The only change is wider footprint and link to another 3D model

- [x] Provide a URL to a [datasheet](https://www.alldatasheet.com/datasheet-pdf/pdf/39974/SANYO/LA1185.html) for the footprint(s) you are contributing
- [x] An example screenshot image is very helpful:

footprint view

- [x] If there are matching symbol or 3D model pull requests, provide [link(s)](https://github.com/KiCad/kicad-symbols/pull/2482) as appropriate


chschlue (approved):
Thanks.

PR #2178, opened 2020-03-27
Fix hole sizes in 2.54 mm headers.

garytwong:
The holes were previously 0.8 mm diameter, which is too small to
accommodate a standard header.  Normal header posts are 0.025" square
(0.635 mm), and so the diagonal (approx. 0.898 mm) will not fit into
a 0.8 mm hole.  1.0 mm is an appropriate size.

---

---


evanshultz:
Hi Gary! Welcome and thanks for making your first contribution.

The auto-added notes above ask for a screenshot and documentation. Can you provide that? 1.0mm holes should be reasonable, though IPC 2221 would say 1.1mm, so I don't have a problem with the concept. But if you could check for any documentation to go along with it that would be great.

These footprints were moved from the old repo quite some time ago and I didn't chase their history all the way back. Just FYI to let you know I looked back into the history of these files.

I see the Arduino Nano footprints also have 0.8mm holes. Would you mind checking if those should be updated too?

garytwong:
Thanks!  I'm not sure what documentation is most appropriate, but the [Arduino design files](https://store.arduino.cc/usa/arduino-uno-rev3) are available, and a [data sheet for pin headers](https://cdn.amphenol-icc.com/media/wysiwyg/files/documentation/datasheet/boardwiretoboard/bwb_bergstik.pdf) specifies 1.0 mm holes (see for instance page 5, "Please use a hole of 1.00 +/- 0.05 mm"; and the Recommended PCB Layouts on page 8 show 1.02 mm holes).

Screenshots of the footprint before and after the patch is applied:

before
after

Yes, I do believe it would be appropriate to make equivalent changes to the Nano footprints as well.  I can add a corresponding commit if that helps.

evanshultz:
I'm not sure if these footprints are meant to be for the Arduino board itself or a hat. If it's meant to be the board, then here are a smattering of Amphenol female headers which have hole sizes <1mm: https://cdn.amphenol-icc.com/media/wysiwyg/files/drawing/76342.pdf, https://cdn.amphenol-icc.com/media/wysiwyg/files/drawing/87606.pdf, and https://cdn.amphenol-icc.com/media/wysiwyg/files/drawing/66953.pdf.

So maybe we do need more info. https://github.com/KiCad/Modules.pretty/pull/2 and https://github.com/KiCad/Modules.pretty/pull/3 are the PRs for the footprints, and it seems clear to me that they are meant to be a hat/shield. https://github.com/KiCad/kicad-library/issues/1026 also talks about this. I'm really not sure why they'd be for the 'main' board, but the footprint name doesn't include anything that clearly states they're for a hat/shield. So with that hopefully cleared up, they should be using bottom-mounted male headers and 1mm holes seems reasonable.

If anybody else knows more or wants to comment, please do so. I'll give this a few days before merging just in case.

@garytwong 
Yes, if you push another commit to this branch to fix the Nano boards I will squash merge. Thanks!

garytwong:
Oh, good point about which side of the interface is targeted.  I had presumed these were all for the shield side, but can't find any explicit indication either way.

garytwong:
Would this be OK to merge, please?

evanshultz:
Oh yes. Thank you for reminding me.

There should be for a shield if they're a footprint; the mainboard should be in a template. I didn't know if that's how it was actually done.

Thanks again!

Issue #2149, opened 2020-03-16
Battery contacts - which library?


mibe:
Hey guys,

I have two new footprints ready, they are the 9 volt battery contacts Keystone [269](https://www.keyelco.com/product.cfm/product_id/1077) & [261](https://www.keyelco.com/product.cfm/product_id/1073). My question now is, which is the most fitting library for them?

Of course the ```Battery``` library with the ```BatteryHolder``` footprints first comes to mind. But on a second thought, these parts are _contacts_, not _holders_. They have (beside the actual contact) no mechanical support for holding a battery. You'd also need both to connect a battery to and you'd as a designer also need to get the distance of both parts on your PCB right.
My next idea was to maybe put them in the generic ```Connector``` library? They are mating contacts after all, so a designer could use them for whatever they like, as a fancy board-to-board interconnect for example.

What's you opinion on this?

evanshultz:
They're specifically for batteries, so put them in the `Battery` repo. Instead of `BatteryHolder_` as a prefix, which I still think might be reasonable, you could use `Contact_` or some other term you think might fit better.

mibe:
Alright, the ```Battery``` library it is. I'd prefer the prefix named ```BatteryContact_``` actually.

cp-aquila:
I assume the contacts always get used in pairs? I see some benefit in having them as seperate symbols+footprints as it makes the BOM creation easier.

**Question:** How does the designer know how much distance needs to be between the contacts?
**Related:** How does the designer know how big the battery will actually be? There is no keepout or courtyard associated with those contacts.

The first solution that comes to mind is to add some kind of marker in F.Fab for the 2nd contact into each footprint (maker example here https://github.com/KiCad/kicad-footprints/pull/1394). Perhaps also add the outline of the battery into each footprint. On the board they should perfectly overlap.
That leaves the courtyard. No clue on how to fix that :/

mibe:
> **Question:** How does the designer know how much distance needs to be between the contacts?

The designer would have to look up the distance in the specification / battery datasheet. It is half an inch. Maybe it'd good to have the distance in the footprint description?

> **Related:** How does the designer know how big the battery will actually be? There is no keepout or courtyard associated with those contacts.
> 
> The first solution that comes to mind is to add some kind of marker in F.Fab for the 2nd contact into each footprint (maker example here #1394). Perhaps also add the outline of the battery into each footprint. On the board they should perfectly overlap.
[...]

PR #2181, opened 2020-03-28
Add battery contact Keystone 261

mibe:
Keystone 261 is a battery contact for 9 volt PP3 batteries.
[Datasheet](https://www.keyelco.com/product-pdf.cfm?p=1073)

Screenshot:

---

I prefixed the footprint with ```BatteryContact_``` as discussed in #2149.

The datasheet does not have a recommended PCB layout. I used the max. part diameter (7.92 mm), added 0.5 mm to the radius and rounded the resulting diameter up to 9 mm, and used this value for the pad size. The part fits quite nicely on this:

---

KLC violations:
- F5.1: The footprint has no simplified outline on the silk layer, since it would be exposed on the pad.
- F5.1: There is no pin 1 designator because this footprint has only one pad. A designator would be futile.
- F5.2: There's again no pin 1 location drawn due to the same reason as above.

---

---


strfry:
Keystone 269 should also be included for this PR to be of value

mibe:
@strfry there's #2182 for the 269 part.
Or did you mean to include the 269 part in this PR?

PR #2184, opened 2020-03-29
Diodes_SMD: Reduce amount of exposed copper on back pads

mrnuke:
These footprints exposed rectangular pads on both front and copper layers.
This is am issue when soldering SMD diodes, as a large amount of solder can be wicked to the back pads. A mountain of solder accumulates on the back pads.

To resolve this, only expose a circular pad on the back. The front still has rectangular pads, so the functionality of this footprint is not reduced.

**Example of problem:**
kicad-wick

**Proposed solution:**
new_foot

---


mrnuke:
#### Violating F6.3
     - Pad '1' missing layer 'Paste'
     - Pad '2' missing layer 'Paste'

This is intentional because this is a handsoldering footprint.
     

#### Violating F9.3
    - 3D model name is different from footprint name

This is intentional.

mrnuke:
Ping

chschlue:
Please be patient, we have (at the time of writing) 391 open footprint PRs.

That being said, what's the point of this PR? If your issue is solder wicking through the holes, why don't you use a proper MELF/SMC footprint in the first place?

PR #2196, opened 2020-04-04
Mini-Circuits footprint HQ1157

aewallin:
https://www.minicircuits.com/case_style/HQ1157.pdf
Used in https://github.com/KiCad/kicad-symbols/pull/2604

Mini-Circuits_HQ1157

---

---


aewallin:
- rotated the footprint
- I don't understand which dimension is wrong
- The datasheet gives even numbers in inches. Insisting on metric just leads to a lot of rounding issues. For example do you consider A=9.271 mm to be correct - or A=9.27mm as in the datasheet?

hq1157

chschlue:
> * rotated the footprint

Rotated by -100<sup>g</sup>, check :)

> * I don't understand which dimension is wrong

Never mind, don't know what I did there.

> * The datasheet gives even numbers in inches. Insisting on metric just leads to a lot of rounding issues. For example do you consider A=9.271 mm to be correct - or A=9.27mm as in the datasheet?

Neither is "correct" and both are within the tolerance given. We usually prefer metric because that's what's used under the hood anyway.
But I agree, rounding errors keep adding up in this case, so forget it.

PR #2205, opened 2020-04-08
Fix ublox_NEO footprint missing paste on VCC pad 23

### LINKED ISSUE #2204: ublox_NEO footprint in RF_GPS has missing paste on VCC pad 23
As per title: ublox_NEO footprint in RF_GPS has missing paste on VCC pad 23 as shown below:

image

I can see no logical reason for this so I will resolve this issue in an imminent PR.

andrewteasdale:
Resolves #2204

The ublox_NEO footprint in RF_GPS is missing solder paste on pad 23, as shown below.

image

This is the VCC pin, and I can see no logical reason for this being the case. This PR corrects this issue.

---

---


andrewteasdale:
In fixing this issue, I noted that he existing silkscreen outlines for ublox NEO, LEA and MAX parts had variable clearances of, in some cases, just 0.01 mm to copper pads, as shown below.

Capturea

I have since modified to give 0.20 mm clearance as per KLC guidelines, as shown here:

Captureb

Because [particularly on the MAX part] this makes pin 1 less clear, I have reduced the silkscreen around pin 1 to make it more readily identifiable. If there is a preferred approach e.g. additional silkscreen around pin 1 as per e.g. SOIC parts, or even less silkscreen around pin 1 as used in some other parts, I would be happy to modify accordingly.

andrewteasdale:
I also took the opportunity to update the pads to IPC-7351C style rounded rectangle pads, but obviously if this is undesirable for some reason I can revert it:

Capturec

andrewteasdale:
Finally, I've also modified the assembly layer pin-1 ident to match KLC guidelines of a 1mm bevel:
 image

evanshultz:
I think extending the silkscreen line above pin 1, as done with the scripted SOIC footprints, would look nice.

andrewteasdale:
I completely agree, that's certainly my preference but I sensed a general trend in the opposite direction. I'll make that modification first thing tomorrow morning.

evanshultz:
Extending silk and cutting it back both work. But when the line length is too short we defer to cutting it back in the scripts. This mainly affects small parts where the line lengths are short, but here it should be no problem to use a longer line. For bigger parts the scripts do extend the line over pin 1.

andrewteasdale:
Thanks for the explanation. I've modified the silkscreen outline as discussed above - screen grab below. It's not as clear as it is on e.g. the scripted SOIC footprints because the body of the component extends over a fair bit of the pad, so let me know if you think it needs modifying further. It's only critical for the MAX part as the other two are not rotationally symmetrical.

snip1

For context, the NEO part with model:

snip2

chschlue:
I'd prefer cutting silk back because while these have pad sizes in the SOIC ballpark, they don't protrude as much, making the lines very short.
Comments? @andrewteasdale @evanshultz

andrewteasdale:
I think you are right @chschlue. @evanshultz what are your thoughts?

Would this [as per the previous commit] be sufficient or would you pull it back even further?

image

evanshultz:
You can see at https://github.com/pointhi/kicad-footprint-generator/blob/master/scripts/tools/global_config_files/config_KLCv3.0.yaml that if lines are less than 0.2mm the footprint is switched to the 'reduced silk' version. So if the silk lines are longer than 0.2mm it should be OK. That being said, cutting silk back is clear too, and most important is that the footprints are usable. So if removing silk is more clear then do it! :)

chschlue:
Alright. Merging.

PR #2213, opened 2020-04-12
Create Diode_Bridge_Vishay_KBPM.kicad_mod

evanshultz:
image

This package apparently is apparently called KBP, but KBP is also another package and I'm familiar KBPM with from Fairchild and Vishay such as http://www.farnell.com/datasheets/2238158.pdf. But it's apparently EOL by them. http://www.cdil.com/rectifiers/kbp2xx-series?rq=kbp201 is in production but with the 'G' dimension only having a max and no min. https://www.comchiptech.com/admin/files/product/KBP2005G-G%20Thru.%20KBP210G-G%20RevD.pdf also is the same except that same dimension is about half, which I suspect might be a typo. I used `Vishay` in the footprint name as many of the diode bridges do, but in reality the diode bridge packages are often shared by many vendors.

Dimensioned drawing (taken before adjusting the bevel to be 1mm instead of 0.9mm);
image

---

---


aris-kimi:
Hi @evanshultz , i think there are some dimmensions wrong here.

εικόνα
εικόνα

Pads shown to be (almost) centered, i think this needs a correction. (Sorry for not coloured wrong values, new FreeCAD user too)
About 1mm bevel,  i think this also needs to be fixed.
Last comment i have is about drill holes, i think they need to be 1mm dia. ( 0,81 + 0,2 )

@chschlue , any input welcomed.

Thank you for your time.

chschlue:
> Pads shown to be (almost) centered, i think this needs a correction.

They're not at the mean given by the drawing, but then, tolerance bands are so large you can't really tell were they're actually going to end up.

> About 1mm bevel, i think this also needs to be fixed.

Fab pin 1 mark is fixed at 1mm and is not related to an actual package bevel, see [F5.2](https://kicad-pcb.org/libraries/klc/F5.2/).

> Last comment i have is about drill holes, i think they need to be 1mm dia. ( 0,81 + 0,2 )

They have to be at least (max pin dia + 0.2), so at least 1.06, but there's also half a millimeter position tolerance, so large drill holes appear reasonable to me.

aris-kimi:
Thank you both. Merging.

PR #2215, opened 2020-04-14
Add Aavid 590302B03600G heatsink

evanshultz:
https://www.boydcorp.com/aavid-datasheets/Board-Level-Cooling-Channel-5903.pdf

image

Drill holes and pad sizes should work but I'm open to suggestions. The pin diagonal is 2.67mm so a round hole may be difficult to (wave, but not hand) solder as the barrel will be rather far from the pin at several points around the pin.

Used unique pin numbers for each pad since the user may want to float just one, both, or neither, and we already have heatsink symbols with 1, 2, and 3 pins.

Dimensioned drawing:
image

---

---


aris-kimi:
Hi @evanshultz , i want to propose to remove the internal graphics since someone might want to use it with reversed mounting, and i think courtyard should follow body shape, same as silk and fab.
I hope i made myself clear, what do you think?

PR #2228, opened 2020-04-19
Add footprint for Quectel L80 GPS module

thatoddmailbox:
Datasheet: https://www.quectel.com/UploadImage/Downlad/Quectel_L80-R_Hardware_Design_V1.2.pdf

I wasn't completely sure how to indicate pin 1. I added an extra line on the silkscreen, but I wasn't sure if there should be something more noticeable (like an arrow or circle or something...)

Screenshot: 
image
Dimensioned drawing (this is my first time creating one of these--if I'm missing anything important, please let me know!): 
l80_dimensions

---

---


thatoddmailbox:
The associated symbol for this is KiCad/kicad-symbols#2657.

myfreescalewebpage:
Hi @thatoddmailbox , thanks for contributing,

- [ ] The footprint is not vertically centered. Everything should be moved by 0.5mm
- [ ] F.CrtYd layer is too much closed to the pad, please respect 0.25mm clearance
- [ ] Datasheet to be added in the description as done for all footprints
- [ ] The keepout area is not placed correctly according to the dimensions on Figure 20 Page 31 of the datasheet

Cheers,
Joel

myfreescalewebpage:
@thatoddmailbox ping

thatoddmailbox:
@myfreescalewebpage Sorry about that, I was really busy last week! I've made fixes for the first three issues you mentioned. However, for the keepout area, I remeasured and I still got the same measurements as in that figure. (7.4 mm and 7.37 mm). I'm measuring from the center of the circle to the lines on F.Fab, since those are 16 mm x 16 mm, like it says in that figure. 

Do you mean that the keepout circle should be 7.4 mm and 7.37 mm from the bottom right corner? That's also what I originally thought, but on the actual module, the circle on the top is closer to the top left (pin 1). See: 

kicad_l80

myfreescalewebpage:
@thatoddmailbox thanks for the details and for the fixes. Just checked again the keepout is correct I think now. Maybe I have made a wrong comment because of the first issue (vertical centering of fp).

No more comment, merging.

PR #2230, opened 2020-04-19
Add Kemet Ex2 Series Mechanical Relays

yankee14:
This PR adds a family of DPDT relays from Kemet. They differ by package type, size, functionality, and coil voltage.

Datasheet: https://content.kemet.com/datasheets/KEM_R7002_EC2_EE2.pdf

Symbol PR: https://github.com/KiCad/kicad-symbols/pull/2631
Footprint PR: https://github.com/KiCad/kicad-footprints/pull/2230
Packages3D-source PR: https://github.com/KiCad/kicad-packages3D-source/pull/280

* 2 general package types
  * DIP --> PN beginning with `EC2`
  * SMD --> PN beginning with `EE2`
* 3 or 5 different coil voltages, these make up the symbol aliases
  * 3V, available on all variants
  * 4.5V, available on all variants
    * I don't like decimal point in part name, should replace with underscore? Or maybe `4V5`?
  * 5V, available on all variants except `NKX`
  * 12V, available on all variants
  * 24V, available on all variants except `NKX`
* 3 different functional variants, these must make up separate symbols since the symbol artwork will differ
  * Non-latch type --> No letter after coil voltage
  * Single coil latch type --> `S` after coil voltage
  * Double coil latch type --> `T` after coil voltage
* 4 different general footprints, length and width differ depending on relay lead type
  * THT for non-latch and single coil type
  * THT for double coil latch type
  * SMD for non-latch and single coil type
  * SMD for double coil latch type
* 4 different lead types, each of these types have unique footprint
  * Standard --> `NU` after latching type
  * High density footprint --> `NUH` after latching type
  * High solder retention --> `NUX` after latching type
  * High solder retention and high breakdown voltage --> `NKX` after latching type

Screencap of non-latching and single coil latching, standard lead, THT:
Relay_DPDT_Kemet_EC2

Screencap of double coil latching, standard lead, THT:
Relay_DPDT_Kemet_EC2_Double-Coil

Screencap of non-latching and single coil latching, standard lead, SMD:
Relay_DPDT_Kemet_EE2_NU

Screencap of double coil latching, standard lead, SMD:
Relay_DPDT_Kemet_EE2_NU_Double-Coil

Screencap of non-latching and single coil latching, high density, SMD:
Relay_DPDT_Kemet_EE2_NUH

Screencap of double coil latching, high density, SMD:
Relay_DPDT_Kemet_EE2_NUH_Double-Coil

Screencap of non-latching and single coil latching, high solder joint reliability and/or high breakdown voltage, SMD:
Relay_DPDT_Kemet_EE2_NUX_NKX

Screencap of double coil latching, high solder joint reliability, SMD:
Relay_DPDT_Kemet_EE2_NUX_Double-Coil

---


yankee14:
I'm not sure how to handle the 0.005mm anchor error without creating a lot of additional work :(

yankee14:
I'll try, but like I suggested, I wanted advice

chschlue:
You could move the anchor using the `Place footprint reference anchor` tool (while on a 10µm grid).
Or use FreeCAD + KSU to import each FP, translate the layers as required and re-export (probably less tedious).

0.05mm is nonsense, BTW. Anchors are off by a lot. CI calculates center from the pads, not the actual part body.
Also, please rotate FPs by 90° CW.

yankee14:
Okay, this will give me a good excuse to dive into KSU. Thanks.

chschlue:
You're welcome.
Good excuses are a specialty of the house.

yankee14:
Boy, I am really lost. The documentation for KSU doesn't really seem to describe what to do, and it's hard to follow.

KSU loads the footprint with the following hierarchy:
image

There's no real FreeCAD friendly hierarchy with Parts and Bodies and such. Every time I try to translate something, it moves it out of the hierarchy.

image

If I export it and load it into KiCad, it's totally corrupted, e.g. afe8548.

chschlue:
You have to convert these imported bodies to sketches first.
Thinking about it again, _modifying_ footprints this way isn't actually less tedious.

Perhaps importing just one as a template and exporting all variants as new footprints from KSU is the easier way (modifying the sketches along the way, of course).

Another method involves a text editor and possibly a pocket calculator.


I think I'll just stop giving bad advice. Sorry for the noise.

yankee14:
That's okay! I'll get it figured out eventually.

chschlue:
Wait.

chschlue:
Before you do it all twice:
Relay_DPDT_Kemet_EE2_NU_DoubleCoil

This is basically where the anchor is supposed to be for all of these. Center of the body, not center of the pads. I already tried to say that a few comments further up.

Apologies in advance if this turns out to be bad advice again:
Get one right, copy over to all variants, remove / shrink pads as needed. Body outline is the same for all of them.

yankee14:
The height dimension error that CI is finding is in the y direction, which should be correct regardless if you reference the pads or the relay body. That's the direction I'm playing around with moving things around.

> Get one right, copy over to all variants, remove / shrink pads as needed. Body outline is the same for all of them.
I've never tried that workflow before so I don't know how efficient it would be for me. Off the top of my head, it feels harder than keeping the pads where they are, moving the anchor, then fixing the remaining polygons that get moved off-grid as a result of 0.005mm anchor movement. Because of these footprints not only have different pad sizes, but different xy locations for those pads.

I planned on moving the anchor to the center of the x axis after I squash the polygon off-axis errors.

Thank you for looking into this though, let me know if I sound insane.

chschlue:
Theoretically yes, but finding the centroid of irregular fab polygons requires multivariate integration or meshing into triangles, AFAIK.
But then, irregular parts tend to have custom pick and place locations, so the payoff would be pretty low.

yankee14:
[KLC 5.2](https://www.kicad-pcb.org/libraries/klc/F5.2/) says the shape should be as simple as possible, but there are probably exceptions where it can't be simple and there are edge cases.

> multivariate integration

Fancy pants xy averaging doesn't sound that hard to me, but hey what do I know?

Anyways, I'll try to finish fixing these tonight!

chschlue:
What I'm saying is: in probably more than 90% of cases, averaging a fab rectangle and averaging the pad positions around it yields the same.
The rest is all fancy pants polygons and fancy pants arcs. Now you're screwed. Multivariately :)

Anyway, good luck with your relays.

yankee14:
@chschlue Okay, I think I am ready for you to tell me how badly I screwed it up! I will try to get the kicad-packages3d PR going later tonight or tomorrow https://github.com/KiCad/kicad-packages3D-source/pull/280

> The rest is all fancy pants polygons and fancy pants arcs. Now you're screwed. Multivariately :)

It sounds like a fun project though

yankee14:
Thanks, I forgot to rotate them from earlier.

Is the 0.25 mm courtyard clearance rule referring to the edge of the courtyard polyline, or to the center of the courtyard polyline?

chschlue:
> Is the 0.25 mm courtyard clearance rule referring to the edge of the courtyard polyline, or to the center of the courtyard polyline?

Center, because courtyard lines (and fab for that matter) don't have any real, physical width, contrary to silk, for example.

yankee14:
> Upshot: You didn't screw up too badly :)

Hooray!

> Center, because courtyard lines (and fab for that matter) don't have any real, physical width, contrary to silk, for example.

That makes sense.

chschlue:
Thanks.

PR #2232, opened 2020-04-20
ublox_NEO: Fix pad 23 type so SMD pad shows up in paste-layer

dpiegdon:
ublox_NEO: Fix pad 23 type so SMD pad shows up in paste-layer

My PCB manufacturer called and informed me that pad 23 was missing in the paste layer / stencil.
I investigated and found that pad 23 type was set to "connect" instead of "smd".


dpiegdon:
Here a screenshot from the manufacturer
pad23

dpiegdon:
And here a datasheet of the NEO-7, just in case:
https://www.u-blox.com/sites/default/files/products/documents/NEO-7_DataSheet_%28UBX-13003830%29.pdf

jneiva08:
was already open PR for this issue https://github.com/KiCad/kicad-footprints/pull/2205

dpiegdon:
true, and he has done quite a few more improvements.
so better pick that PR.

PR #2234, opened 2020-04-21
Added footprint for CUI CMR-12062S-67 speaker

cnieves1:
Datasheet: https://www.cuidevices.com/product/resource/cmr-12062s-67.pdf
3D: https://github.com/KiCad/kicad-packages3D/pull/681

imagen
Captura de pantalla de 2020-04-21 12-31-12

---

- [] Check the output of the Travis automated check scripts - fix any errors as required

---


cnieves1:
Regarding F6.2: footprint is centered, but it has only two pads on the bottom, so TravisCI calculations are wrong.
Regarding F6.3: this device is not soldered. Connection is made with compression springs.

cpresser:
Question: That 'ear'-shape on F.Fab, is that a removable tape on the device?

cnieves1:
Yes, it is. I think it should be better to draw it, just in case there is any conflict with other component nearby...

cpresser:
What If I want to put a component next to the speaker, below that flap of tape. Then the courtyard should not be there. Nevertheless, the flap should be drawn somehow.

I was about to say there might be problems during production, the speaker need to be assembled last so the tape is not interfering with another pick'n'place operation. But that does not apply since the speaker will not get assembled in a SMT line prior to reflow.
Or am I missing something here?

chschlue:
This type of speaker is usually glued to the cover first and will contact the PCB on final assembly.
I don't get the point of having that tab in the FP at all.

chschlue:
Example:
spkr

Sorry, couldn't resist using an image of a prehistoric phone :)

cnieves1:
I think you are right. Removed cover from footprint. Thanks!

cnieves1:
Captura de pantalla de 2020-04-24 23-56-29

cpresser:
Good thing this is such a simple footprint and easy to check :)
All dimensions look good. The choice of the Pin1 indicator-style is also fine.

cnieves1:
I agree this is up to the designer and hard to guess beforehand. If you think 0.5mm is better, it's fine for me...

cnieves1:
Increased courtyard clearance to 0.5mm:
Captura de pantalla de 2020-05-25 11-41-35

PR #2239, opened 2020-04-22
Fix F.Paste of SIMCom SIM800C

jneiva08:
Fixed F.Paste in accord of datasheet
Moved F.CrtYd to have 0.25mm from F.Paste pad

Datasheet: https://simcom.ee/documents/SIM800C/SIM800C_Hardware_Design_V1.05.pdf#page=16

Screenshot:
sim800c

---

---


evanshultz:
Is it weird to anyone else that the paste pads are taller than the copper pads? That seems like a relatively odd thing and I question it. During reflow the paste may wick down onto the pads and it would probably work out OK, but it's abnormal in my (questionable and limited) experience.

chschlue:
Exactly what I thought. So I decided to leave it to someone more experienced with castellated pads :)

Apart from that strange paste-on-mask thing, I'd generally leave paste clearance to the user (board setup or local FP options).

jneiva08:
I'm curious too, but i think they make this way, is because when goes to the oven, the solder melt an try to climb the pad and leaves more solder near of the module.

I will making this change for the other modules (Quectel) on repository. Because the module i need will have more complex patterns, and the F.Paste is different depending of the version.
In one project I'm envolved, the company responsible to order the pcb's with ublox SAM-M8, forgot to send the F.Paste. When be PCB's arrived they found some pads of gps was shorted.

poeschlr:
It is definitely a strange thing but the resulting footprint agrees with the datasheet so i am ok with merging it.

jneiva08:
Thanks

PR #2244, opened 2020-04-25
Fix Pin 1 Markers to fit with KLC (IPC suggestion)

poeschlr:
Fixes https://github.com/pointhi/kicad-footprint-generator/issues/278 for DFN footprints

This is a manual fix to unify all Pin 1 markers to fit with KLC (IPC). The pull request only touches silk and fab outlines (fab only where the chamfer was way too large).

---

Edit: I do  not intend to fix any of the other KLC issues in these.


evanshultz:
Took a quick look at the diff and CI. Nothing jumped out at me. If there are issues from this it will still likely be better for users and we can deal with it later. Thank you, Rene!

PR #2252, opened 2020-04-29
Add Bourns SRR6040A inductor

matthuszagh:
[Datasheet](https://www.bourns.com/docs/Product-Datasheets/SRR6040A.pdf)

srr6040a

---

---


aewallin:
for consistency of style with existing smd inductors, would you agree on short silk-lines between the pads for aligning the component along the pad-direction?
orientation of the whole footprint should probably be left (pin1) to right (pin2)?

poeschlr:
The silk lines are not just for compatibility. They are required for KLC compliance which is a necessary unnegoziable condition for inclusion.

matthuszagh:
Here's the new fp. Is the silkscreen done correctly here?

srr6040a

PR #2255, opened 2020-04-29
TRANSFORMER: Fix Wurth spelling

myfreescalewebpage:
Everything is in the title.
No modifications of the footprint itself, just fix the name of the manufacturer.

---

---

### DIFF
--- a/Transformer_THT.pretty/Transformer_Wurth_750343373.kicad_mod
+++ b/Transformer_THT.pretty/Transformer_Wurth_750343373.kicad_mod
@@ -1,10 +1,10 @@
-(module Transformer_Wuerth_750343373 (layer F.Cu) (tedit 5CA646E6)
+(module Transformer_Wurth_750343373 (layer F.Cu) (tedit 5EA9A8D9)
   (descr "Transformer, horizontal core with bobbin, 10 pin, 3.81mm pitch, 15.24mm row spacing, 22x23x17.53mm (https://katalog.we-online.com/ctm/datasheet/750343373.pdf)")
   (tags "transformer flyback")
   (fp_text reference REF** (at 0 -5.08) (layer F.SilkS)
     (effects (font (size 1 1) (thickness 0.15)))
   )
-  (fp_text value Transformer_Wuerth_750343373 (at 7.62 8.89) (layer F.Fab)
+  (fp_text value Transformer_Wurth_750343373 (at 7.62 8.89) (layer F.Fab)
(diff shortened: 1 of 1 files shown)


chschlue:
This was actually correct before (see https://en.wikipedia.org/wiki/Diaeresis_(diacritic)#Printing_conventions_in_German), but I agree that it's far from obvious.
We should probably also "fix" Mounting_Wuerth and all the other Wuerth stuff in the long run.

evanshultz:
Search around and you'll see discussions about this. I recall we mostly relied on what the vendor called themselves if diacritics weren't used and also the input of German-speaking librarians. Closing.

poeschlr:
Even the webaddress of würth uses wuerth: See https://www.we-online.com/web/en/wuerth_elektronik/start.php

We can however add the alternative spelling to the keyword list such that it appears with a search

myfreescalewebpage:
I wasn't aware of this kind of details. I was the author of this symbol and the associated footprint and I realized I have used this spelling but the logo of Wurth is the following:

WE-Logo_2007_A4_RGB

Are you sure we want to keep the existing symbol and footprint ?

myfreescalewebpage:
Also just notice this new PR: https://github.com/KiCad/kicad-footprints/pull/2256/files. The name of the file is "Wurth"...

poeschlr:
Yes i am positive. And their logo is not wurth its würth. The ü is a special sound. And with the knowledge of ue being ü have a guess what the oe in my name stands for ;).

chschlue:
My guess: Ρενε Ποισχλ, transcribed from Greek to Pöschl?

pic_gall_a_0004434_int

Perhaps we should consistently switch to ligatures. Phœnix. Wᵫrth.

Sorry for derailing this thread :)

myfreescalewebpage:
Interesting details :) thanks @poeschlr and @chschlue 

@chschlue this include special characters which is not so good in computing science...

evanshultz:
@myfreescalewebpage 
I think that is a joke above. Using `ue` instead of `ü` was done to avoid 'special' characters.

chschlue:
Jokes aside, adding `wurth` keywords / tags sounds like a good idea to me to improve searchability.

Compare https://trends.google.com/trends/explore?q=jagermeister and https://trends.google.com/trends/explore?q=lowenbrau for Jagërmeister and Lowënbraü, respectively, for example.

Note: I'm not suggesting this for Phoenix. As @poeschlr already pointed out, they never use Phönix anywhere anymore.

PR #2209, opened 2020-04-11
added SO8L Vishey

rs-development:
Added the Vishey SO8**L** Package.

https://www.vishay.com/docs/64721/an913.pdf

image

https://github.com/KiCad/kicad-symbols/pull/2576

It did not touch the fab and courtyards layer since I am not sure about dimensions. Please check if the SO8 Dimensions are matching. I used PowerPack_SO8_Single as a template.

https://www.vishay.com/docs/71655/powerpak.pdf
---

All contributions to the kicad library must follow the [KiCad library convention](http://kicad-pcb.org/libraries/klc/)

Thanks for creating a pull request to contribute to the KiCad libraries! To speed up integration of your PR, please check the following items:

- [ ] Provide a URL to a datasheet for the footprint(s) you are contributing
- [ ] An example screenshot image is very helpful 
- [ ] If there are matching symbol or 3D model pull requests, provide link(s) as appropriate
- [ ] Check the output of the Travis automated check scripts - fix any errors as required
- [ ] Give a reason behind any intentional library convention rule violation.

---

Be patient, we maintainers are volunteers with limited time and need to check your contribution against the datasheet. You can speed up the process by providing all the necessary information (see above). And you can speed up the process even more by providing a dimensioned drawing of your contribution. A tutorial on how to do that is found here: https://forum.kicad.info/t/how-to-check-footprint-correctness/9279 (This is optional!)

PR #1778, opened 2019-08-12
Add AND9137_SO-3-8_EP_5x6mm_P1.27mm footprint

flaviut:
Ref #1000, https://github.com/KiCad/kicad-footprints/issues/1000

**Datasheet**: https://www.onsemi.com/pub/Collateral/AND9137-D.PDF

2019-08-12-015010_816x684_scrot
2019-08-12-014959_874x963_scrot
2019-08-12-014950_895x987_scrot
2019-08-12-014945_865x977_scrot

3d is already-existing `${KISYS3DMOD}/Package_SO.3dshapes/Vishay_PowerPAK_SO-8.wrl`, which this footprint is specifically designed for.

This footprint almost exactly implements the footprint described in [AND9137](https://www.onsemi.com/pub/Collateral/AND9137-D.PDF); the only difference is that this footprint has a .175mm radius on the top 4 pins [because some guy on stackexchange suggested it's a good idea](https://electronics.stackexchange.com/a/284549/).

Another point of note is the pin numbering. Instead of numbering this as 1-8, I've numbered it 1-3. All packages that this claims compatibility for have source=1-3, gate=4, and drain=5-8. If you'd like to confirm compatibility yourself, here are the datasheets:

- http://www.farnell.com/datasheets/1704644.pdf
- http://www.vishay.com/docs/62839/si7149adp.pdf
- https://www.infineon.com/dgdl/irfhm9331pbf.pdf?fileId=5546d462533600a4015356237f7e1f59
- https://www.ti.com/lit/ds/symlink/csd18537nq5a.pdf
- http://aosmd.com/res/data_sheets/AON6236.pdf
- https://www.st.com/content/ccc/resource/technical/document/datasheet/d7/9b/35/84/f3/31/44/19/DM00098815.pdf/files/DM00098815.pdf/jcr:content/translations/en.DM00098815.pdf
- https://www.infineon.com/dgdl/DS_BSC014N04LS_2_1.PDF?folderId=db3a304313b8b5a60113cee8763b02d7&fileId=db3a3043353fdc16013552e99a8147f1
- https://assets.nexperia.com/documents/data-sheet/PSMN030-60YS.pdf
- https://www.renesas.com/us/en/doc/products/transistor/002/r07ds0765ej0200_rjk03m1dpa.pdf
- https://www.onsemi.com/pub/Collateral/FDMS86202-D.pdf
- https://toshiba.semicon-storage.com/info/docget.jsp?did=15504&prodName=TPW4R50ANH
[...]

PR #1699, opened 2019-07-15
PowerPAK-8 footprints to custom pads

evanshultz:
Breaking apart https://github.com/KiCad/kicad-footprints/pull/981.

Changes several pads into a single custom pad and sets the zone connection to "solid". Otherwise, the pads don't connect to zone at all in KiCad 5.1.2.

See top two footprints here:
image

---

All contributions to the kicad library must follow the [KiCad library convention](http://kicad-pcb.org/libraries/klc/)

Thanks for creating a pull request to contribute to the KiCad libraries! To speed up integration of your PR, please check the following items:

- [ ] Provide a URL to a datasheet for the footprint(s) you are contributing
- [ ] An example screenshot image is very helpful 
- [ ] If there are matching symbol or 3D model pull requests, provide link(s) as appropriate
- [ ] Check the output of the Travis automated check scripts - fix any errors as required
- [ ] Give a reason behind any intentional library convention rule violation.

PR #2266, opened 2020-05-02
Added bevel to corner of package on FAB layer. Bevel is 1mm. (1.414mm…

slagdang:
… in diagonal length)

Without this the orientation of the package is not apparent using only the FAB layer.

This the only change. And I changed the single and dual versions of PowerPAKSO8 even though the below picture shows dual without the bevel that is just because I placed the old version of the foot before I saved the change so that they could be compared.

The bevel is 1mm (across and down, 1.414mm diagonally) as specified.

There are no other changes.

Screen Shot 2020-05-02 at 2 14 15 PM

I have another foot I use myself which is supposedly a more "universal" PowerSO8. I made it from this document:

https://www.onsemi.com/pub/Collateral/AND9137-D.PDF

I can submit that as a replacement upon request. But do note that I did not do that in this case, I only changed the FAB layer of the previously published KiCAD symbol. I would only do that on request.

---

---


chschlue:
Thanks for the fix.

You're welcome to submit further PowerPAK fixes / additions, but please note existing PRs #1699, #1778 and #2209.

PR #2268, opened 2020-05-04
Add Harwin S1941-46R spring contact

kacper-ka:
This merge adds **S1941-46R** spring contact footprint manufactured by **Harwin**. The recommended side deflection is marked using `Dwgs.User` layer.  
CAD model is available at manufacturer site.

#### Datasheet
https://www.harwin.com/products/S1941-46R/  
https://cdn.harwin.com/pdfs/S1941R.pdf

#### Screenshot
Screenshot from 2020-05-04 17-54-35

#### F6.2 - Footprint anchor should be placed in the middle of the component body
Footprint anchor is located at the center of a pick and place area, as designed by the manufacturer and specified in the datasheet.

---

---


diegoherranz:
Thanks for this submission.

I've started the review of this footprint. A few comments:

- [ ] This vertical line in the Fab layer should be at X= -6.89mm. I know it is very close, but just to match the drawing exactly.
fab_
- [ ] Nice touch adding the deflection indications to Dwgs.User. This will need to be adjusted slightly after the previous point.
- [ ] Given this is a connector (or connector-like), the clearance on the courtyard should be 0.5mm all around (see [KLC F5.3](https://kicad-pcb.org/libraries/klc/F5.3/)).

Thanks!

diegoherranz:
When I first started reviewing, I saw this 9.45 .. 9.85 mm which would give you a 9.65mm nominal if you average as you say:
details

But then I saw the footrprint recommendation and if you add 3.49 + 4.60 +1.50, you get 9.59mm which is where I was coming from:
footprint

Not sure why those 2 don't match. Go for the one you prefer, sincerely. But that vertical line is a at x=-6.9mm so either way it needs fixing.

Thanks!

kacper-ka:
Oh, you're right, I missed the second drawing when I was replying to your comment. I'll fix it right away.

diegoherranz:
Thanks for the fixes.

Just a minor thing I had overlook. Could you make the description something like the following instead of just the URL?
"7.25mm SMT Multi-directional Spring Contact (T+R), https://cdn.harwin.com/pdfs/S1941R.pdf"

Thanks!

diegoherranz:
Thanks again for fixing the description.
For the record, a few checks done:
- PAD 4.60x5.70 - OK
- Fab dimensions OK
- Courtyard 0.5mm clearance all around - OK
- Deflection indications on Dwgs.Usr: 0.4 to 1mm - OK
- The pick and place point (origin in this footprint) has been placed at 2.7mm from the right hand side and is pretty much in the centre of the flat pick and place area. I think that following this drawing it would be (4.45-1)/2 + 1 = 2.725mm from the right hand side.
panp
But if you consider the 3.5mm diameter pick and place area shown on this drawing, you can get to the 2.7mm figure (3.5/2 + 4.45 - 3.5). I think the drawings contradict each other and we're talking 25um here, so I think 2.7mm is OK. 
panp2

Thanks for your contribution. Merging.

PR #2274, opened 2020-05-09
Round Drills and Remove Pitch Singles for Solderwires

poeschlr:
Fixes for what i discovered shortly after https://github.com/KiCad/kicad-footprints/pull/2062 got merged

- The drill sizes are now rounded to 0.05mm
- The footprint name for single wire footprints does not need to include the pin pitch


chschlue:
Letting it settle for a couple minutes.

poeschlr:
I have now also pushed the changes to the script PR so you can check them there

chschlue:
Looks good to me once more...

PR #2286, opened 2020-05-15
Added D_SMP

clnhlzmn:
Added a diode footprint (SMP, DO-220AA) for Vishay SS3P3.

Datasheet: https://www.vishay.com/docs/88944/ss3p3.pdf

Screenshot:
image

image

No errors from Travis CI.


poeschlr:
Silkscreen must have at least the same clearance as its line width to exposed copper (pads).

Courtyard clearance must be 0.25mm all around (rounded to 0.01mm grid)
Screenshot from 2020-06-11 21-25-29

clnhlzmn:
> Silkscreen must have at least the same clearance as its line width to exposed copper (pads).
> 
> Courtyard clearance must be 0.25mm all around (rounded to 0.01mm grid)

Fixed! I think.

poeschlr:
There is a misunderstanding. For courtyard the center of the line is what matters (as it only has thickness for the user to see it, it is treated as a mathematical line of 0 thickness by DRC)

For Silk the full line is taken into account so there it is correct to ensure the inside edge is far enough away.

clnhlzmn:
> For courtyard the center of the line is what matters (as it only has thickness for the user to see it, it is treated as a mathematical line of 0 thickness by DRC)

Ok. I think I have it now.

Are these things not checked by Travis CI?

poeschlr:
Checking courtyard against pads would be possible. Even checking a simple rectangular outline against the body outline would be feasible. It gets however way to complex if one allows closely contoured courtyards. Still not impossible but way too much effort to implement at this point in time.

PR #2289, opened 2020-05-17
Add Ai-Thinker ESP32-CAM

stefanct:
- There is no suitable datasheet that shows precise layout information, but the general board dimensions are specified here: http://wiki.ai-thinker.com/_media/esp32/docs/esp32-cam_%E4%BA%A7%E5%93%81%E8%A7%84%E6%A0%BC%E4%B9%A6.pdf
 - The symbol has been submitted here: https://github.com/KiCad/kicad-symbols/pull/2737
 
PS: There is an excellent 3D model available here: https://cdn.thingiverse.com/assets/20/fe/9e/9e/e7/ESP3-CAM_v4.step (under CC-BY) that could be facilitated but it needs some minor fixes and correct translation. Instead I have added the pin sockets and headers with the right offsets to visualize the proportions.


myfreescalewebpage:
Hi @stefanct , thanks for contributing,

- [ ] I'm not sure it is allowed in the library to add multiple 3D models, do you have another example on which it has been done ?
- [ ] Pin numbering to be fixed according to the last page of the datasheet.
- [ ] 0.25mm clearance between F.CrtYd and F.Fab is enought
- [ ] What do you ant to represent with the two circles on F.Fab layer ?

Cheers,
Joel

myfreescalewebpage:
Symbol Abandoned

PR #2294, opened 2020-05-22
Add USB Micro-B connector Amphenol 10118194

vsergeev:
Add USB Micro-B connector Amphenol 10118194

Datasheet: https://cdn.amphenol-icc.com/media/wysiwyg/files/drawing/10118194.pdf

Screenshot:
usb

Recommended layout:
footprint

---

- [x] (N/A) If there are matching symbol or 3D model pull requests, provide link(s) as appropriate
- [x] (N/A) Give a reason behind any intentional library convention rule violation.

---


cpresser:
Hi & sorry for the huge delay of the review.

- [ ] The Pin1 mark on F.SilkS could be smaller, perhaps copy the style from `USB_Micro-B_Molex-105017-0001`. I am aware that the arrow is present in other similar parts. But I think smaller is better.

vsergeev:
Hi @cpresser, thanks for the review. I've amended the commit with the following fixes:

* [x] The Pin1 mark on F.SilkS could be smaller, perhaps copy the style from USB_Micro-B_Molex-105017-0001. I am aware that the arrow is present in other similar parts. But I think smaller is better.
* [x] There is no paste on two of the shield pins (Pin6)
* [x] The body-width of 7.5mm is not in the datasheet. How did you come up with that value? My current assumption is that the metal thickness is 0.2mm. With the 6.9mm inside measurement that would make the body 7.3mm wide.
* [x] Courtyard clearance is 0.25mm, but it should be 0.5mm since this is a connector

footprint

I wasn't sure which shield pins you were referring to, but I enabled F.Paste on all four for now.

I had to eyeball and use Euclidean distance for the clearance between shield pin copper and silkscreen. Not sure if there is a better way.

Let me know if there is anything else I should change.

vsergeev:
I got an error from CI on the paste layers for the shield pads:

```
  Violating F7.4

    Pad requirements for THT footprints

    Some THT pads have incorrect layer settings
     - Pad '6' has extra layer 'F.Paste'
     - Pad '6' has extra layer 'F.Paste'
     - Pad '6' has extra layer 'F.Paste'
     - Pad '6' has extra layer 'F.Paste'
```

so I disabled them for now. I'm not sure how to proceed on that issue.

cpresser:
Through-Hole-Reflow is somewhat difficult. Afaik KLC does not specify anything special.
Unfortunately KiCad does not have a clever way to handle this. Same goes for the travis-CI scripts.
A similar part (`USB_Micro-B_Amphenol_10103594-0001LF_Horizontal`) has SMD pads that overlap the THT holes.

One possible solution here is to stack another SMD pad on top of the holes, select 'no copper layers' and only activate the paste layer.
The datasheet of this part does not say how much paste should be there, So I suggest the same size as the copper.

vsergeev:
OK, I created two pads, with the same size as the oval through-hole, but only with the top paste layer and no copper layers. The paste layer now looks like this:

pastelayer

cpresser:
That looks good. Please also add them for the other two shield-pins and we are good.
Travis-CI will most likely complain about them. Its because its just a stupid script that does not handle special cases. I will review that manually.

vsergeev:
OK, done. New paste layer:

paste

cpresser:
>
> I had to eyeball and use Euclidean distance for the clearance between shield pin copper and silkscreen. Not sure if there is a better way.

There is a trick for that. Increase the size of the Pad by the clearance. Like so:
image

cpresser:
For future contributions, please do not force-push your changes but make incremental commits instead. That makes review way easier. In theory I would have to re-check everything with each force-push since I don't know what you changed.

I did use 
`git diff f4c4592415d97c11652aa0bc30b89844bbf64fad Connector_USB.pretty/USB_Micro-B_Amphenol_10118194_Horizontal.kicad_mod` 
to get a rough Idea what changed. Which worked here.

Once we merge, commits will get squashed to make the history of the master branch nicer.

PR #2306, opened 2020-05-29
Footprint for ON Semiconductors SC-74A Package

ParticleBert:
- https://www.onsemi.com/pub/Collateral/318BQ.PDF was used as reference

Created the footprint for the ON Semi package SC-74A.
grafik

---

---


aewallin:
- pin1 could be indicated by extending the top silk line left (see TSOT-23-5 footprint)
- check that no silkscreen is hidden by the component body (KLC "no silkscreen allowed under component")
- does anyone know the relation (if any) between the standards TSOT-23-5 and this one here SC-74A ?

chschlue:
> * does anyone know the relation (if any) between the standards TSOT-23-5 and this one here SC-74A ?

What's commonly called SOT-23-5 is a JEDEC MO-178AA package, it's thinner cousin TSOT-23-5 is MO-193AB.
I don't have access to JEITA documents, but I believe SC-74A is identical or at least compatible footprint-wise.

Either way, I don't think a manufacturer-specific footprint is justified here.
See also https://www.nxp.com/docs/en/package-information/SOT753.pdf (which also references SC-74A) and https://www.nxp.com/docs/en/application-note/AN10161.pdf#page=2

evanshultz:
Yes, as @chschlue said this should not be manufacturer-specific.

JEITA gives a footprint but I believe the gullwing script can make a footprint from this package drawing as well and this might even be the best way to go:
image

PR #2308, opened 2020-06-02
Add Senseair S8 CO2 sensor

yankee14:
Add Senseair S8 CO2 sensor.

Symbol PR: https://github.com/KiCad/kicad-symbols/pull/2777
Footprint PR: https://github.com/KiCad/kicad-footprints/pull/2308

3 footprints:
1. Upward facing mount
2. Downward facing mount
3. Downward facing mount with board cutout

* Fabrication layer is nominal part outline
* Courtyard is part outline tolerance + 0.25 mm
* Silkscreen is nominal part outline + 0.15 mm
* Board cutout is rough due to incomplete drawing

Datasheet: https://rmtplusstoragesenseair.blob.core.windows.net/docs/publicerat/PSP0113.pdf
Mechanical drawing: http://www.co2meters.com/Documentation/Datasheets/DS-S8-3.2.pdf

The mechanical drawing is pretty poor, with some of it incompletely specified and defined. There is a better one available IIRC, but I don't think it is public. I am asking Senseair to provide me a link to the better drawing.

Users may wish to mount the sensor facing upwards or downwards. In the downwards case, I have added an additional footprint with polygons on the Edge.Cuts layer for a cutout to allow better airflow to the sensor, through the PCB. I understand if you do not accept these into the library.

I need help specifying:
* Thickness of Edge.Cuts lines. They are currently 0.05 mm.
* 3D part path.

Screencaps:

Senseair_S8_up

Senseair_S8_down

Senseair_S8_down_cutout

---

---


cpresser:
I feel like this could be done in one footprint for all use-cases.
The upward/downward variant is the same, it is just mirrored. On the other hand, using the same footprint for both use-cases has the downside that the 3D rendering (assuming there is a matching 3D-Model) won't be correct.
Tbh, I am not sure if reducing them to one footprint is the correct choice.

Regarding the cutout I would suggest to not draw it on Edge.Cuts but Dwgs.User and add some text like "Diffusion area". Then the board-designer can add that cutout themselves. And we save one variant.

chschlue:
@cpresser do you want to take this one?
I only just noticed I apparently assigned it to myself and then left it to rot.

cpresser:
> @cpresser do you want to take this one?
> I only just noticed I apparently assigned it to myself and then left it to rot.

Yep, will do.

@yankee14 Do I assume correctly that you generated the `down` version by mirroring the `up` version (or vice versa)? Then I will just have to thoroughly inspect one of them. 


**Senseair_S8_up**
- [ ] Please rotate it so that Pin1 is at the correct position
- [x] F.Fab Dimensions are okay
- [x] Pin-Positions, shape and diameters are okay
- [ ] Location of Pin1 relative to F.Fab seems to be okay. I could not find that info in the data-sheet. But it is well centered in the part.
- [ ] The courtyard seems off. On the left and right part it is bigger. Also, I did not find it to be 0.25mm as expected on the other sides.
image
- [ ] I am not sure about the pin-numbering. I could not find anything in the datasheet for that. Why did you choose to number them the way they are?
- [ ] 3D-Model path needs to be set. Even if there is no matching model yet. The name based on the path and filename of the footprint. Set it accordingly. Then travis will be green.
- [ ] Names. Please use `CamelCase` for `Up` and `Down`

yankee14:
@cpresser 

> Did you have success in obtaining a better datasheet?

Unfortunately, no. I never got a response.

> Do I assume correctly that you generated the down version by mirroring the up version (or vice versa)? Then I will just have to thoroughly inspect one of them.

Sorry, I don't remember. I think I drew it manually.

> Why did you choose to number them the way they are?

I wasn't sure how to number them because, as you noticed, it is not mentioned in the datasheet. I went with what I thought was a sane default. Please let me know what I should name them.

cpresser:
I spend some time to figure out if there is a "agreed upon pin numbering". But my google-foo did not return anything useful. Just this one Project with a schematic: https://github.com/merlokk/SmartHome/wiki/ESP8266-air-quality
image

That makes your numbering the official standard :+1: 
Regarding PascalCase you are right. I did read the wikipedia-article, I was not aware of the difference. I will try to remember that for future use.

cpresser (approved):

PR #2317, opened 2020-06-12
added Wuerth SMD LED WL-SFSW

Knochi:
*added Würth RGB SMD side LED WL-SFSW*

---

2020-06-05 19_19_25-Footprint Editor — Knochi_Opto_LED_WE_WL-SFSW_3 2x1 5mm_Horizontal
---


koenraad-voltza:
The pinout numbering of this footprint doesn't match the datasheet. The numbering in the datasheet matches the numbering of existing multicolor led schematic symbols (LED:ASMB-MTB0-0A3A2) so there is no reason to change the numbering.

Knochi:
I'm not shure how the outline and the center is defined.. the origin should be in the middle of the component body, which includes the round part.

Knochi:
grafik
Screenshot to show real center and calculated center

koenraad-voltza:
> I'm not shure how the outline and the center is defined.. the origin should be in the middle of the component body, which includes the round part.

It seems like the scripts are calculating center based on copper area instead of bounding box, I'm not sure what the reason for that is..

poeschlr:
@koenraad-voltza we are working on adding support for center detection via body outline for asymmetric footprints. The main reason why only copper is used right now is because handling outlines is a much harder problem to solve.

PR #2157, opened 2020-03-19
Add RJ45-Molex vertical connector footprint

floesche:
The Molex 85513 series is a surface-mount only (no mechanical via/holes)
RJ45 unshielded vertical plug. Product webpage at
https://www.molex.com/molex/products/part-detail/modular_jacks_plug/0855135013.

The datasheet is at https://www.molex.com/pdm_docs/sd/855135013_sd.pdf

image

---

All contributions to the kicad library must follow the [KiCad library convention](http://kicad-pcb.org/libraries/klc/)

Thanks for creating a pull request to contribute to the KiCad libraries! To speed up integration of your PR, please check the following items:

- [x] Provide a URL to a datasheet for the footprint(s) you are contributing
- [x] An example screenshot image is very helpful 
- [ ] If there are matching symbol or 3D model pull requests, provide link(s) as appropriate
- [x] Check the output of the Travis automated check scripts - fix any errors as required
- [ ] Give a reason behind any intentional library convention rule violation.

---

Be patient, we maintainers are volunteers with limited time and need to check your contribution against the datasheet. You can speed up the process by providing all the necessary information (see above). And you can speed up the process even more by providing a dimensioned drawing of your contribution. A tutorial on how to do that is found here: https://forum.kicad.info/t/how-to-check-footprint-correctness/9279 (This is optional!)

PR #2324, opened 2020-06-21
Fix reversed pin numbering

floesche:
The initial submission had the pin numbers reversed. This new submission corrects that.

The problem was pointed out by @Ripae in #2157.

---

---


floesche:
Screenshot for 1ff6137

Screenshot_2020-06-21T12-52-24

cpresser:
Do I assume correctly that you used the rotate-function in the library-editor for this?

To review this I did rotate the original part myself and then compared it to yours. But that diff is also really hard to read because the line-ordering of kicad is not consistent.

floesche:
Sorry, I cant give a direct answer. But compared to the original file, I had to mirror it (because of the wrong pin ordering) and then rotated it. It's been a while since I did that so I don't remember all details, but I did some manual corrections for both steps eg to keep it centered. I also remember going through some of the steps in the documentation again and double checking that the new part is OK with that.

cpresser:
> 
> Is there any particular question about the part or something that sticks out?

No it looks good. Visually. Also the diff looks okay.
But I will have to look at all the changed lines in this case. Perhaps you made a small mistake  - thats what reviews are for.

If you would have said: `I just used the "Move-Exactly" function` I would have trusted you and just merged it.

cpresser:
Looking at diffs is a PITA. Those here are the same lines, but in different orientation and different line-ordering:
image
I really hope the KiCad devs find a way to normalize the files in the future, so that those comparisons are simpler.

Anyway, I checked the graphics and confirmed that all lines are where I expect them to be. Will merge. Thanks!

PR #2325, opened 2020-06-21
Added ARJM11D7-114-AB-EW2 footprint

Misaka0x2730:
Added ARJM11D7-114-AB-EW2 - Single port 100Base-T connector with PoE.
https://abracon.com/Magnetics/ARJM11.pdf
image

---

---


cpresser:
Hi and thanks for you contribution.

I added a few screenshots here for reference.
image
image
image


I took a look and found a few issues:
* [ ] The name should not contain `AB`. The footprint will also work with different LED arrangements
* [ ] The name should include `_Horizontal`
* [ ] **TBD**: Please use pin-numbers for the LEDs instead of `L1`.. that will make it consistent with other parts in the library.
* [ ] The arrow on silkscreen and the Chevron on F.Fab are on the top left. But Pin1 is on the right side. That does not match.
* [ ] The pin numbers seem to be okay, but please rotate the part so that pin1 is in the top left quadrant (https://kicad-pcb.org/libraries/klc/F4.2/) `RJ45_Wuerth_7499010121A_Horizontal` is a good example for that.
* [ ]  The courtyard clearance should be 0.5mm. Measuring from F.Fab and Pin11 (clearance to the copper)
* [ ] The Body-Dimensions should be 21.3x15.9mm
* [ ] I did not yet check the exact location of pins. Please fix the above issues first and re-check your drawing. Then I will give the geometry a full review.

Open questions, can you please help @chschlue 
* Should Pin11 be named `SH`. It is a shield per KLC S3.7, but also electrically connected via some caps. I would argue yes, it is called SHIELD in the datasheet.
* The LED pins are called `L1`... and so on in the datasheet. But it is inconsistent with all other parts currently in the lib. For which kind of consistency should we go? I am good with either one.

Misaka0x2730:
@cpresser Hello, thank you for comments!
But your picture show A & B mechanical options, this footprint for C & D as on picture below:
image

cpresser:
> @cpresser Hello, thank you for comments!
> But your picture show A & B mechanical options, this footprint for C & D as on picture below:

You are right. Sorry about that. I exchanged the first screenshot.

chschlue:
> Should Pin11 be named SH. It is a shield per KLC S3.7, but also electrically connected via some caps. I would argue yes, it is called SHIELD in the datasheet.

LGTM

> The LED pins are called L1... and so on in the datasheet. But it is inconsistent with all other parts currently in the lib. For which kind of consistency should we go? I am good with either one.

I'm for prefixes. That way it doesn't get too messy if something like `RJ45_Amphenol_RJHSE5380-08` with LEDs is added.
Perhaps we should even use A1/K1 and so on, although there are connectors with multi-color LEDs as well.

myfreescalewebpage:
@Misaka0x2730 what is the status here ? Thanks

myfreescalewebpage:
@Misaka0x2730 ping ?

PR #2333, opened 2020-06-24
Molex 503480 1200

openelectron:
Created new footprint.

Screenshot from 2020-06-24 21-53-04

Datasheet here;
https://www.molex.com/pdm_docs/sd/5034801200_sd.pdf

---

---


poeschlr:
From the screenshot it looks like the silk outline is right on top of the fab outline. However fab must represent the nominal body size and silk must be outside of the nominal body (but still close enough to be of use for alignment)

From the screenshot it also seems as if the body outline is not quite correct as the body should not be that far out from the pads (i did not measure it yet).

openelectron:
I agree, Not sure how I missed that. I have corrected the F.Fab layer to follow the body outline closer.
Screenshot from 2020-07-06 11-55-42

cp-aquila:
How do you feel about scripting this footprint? So the whole series can be added? This one is rather simple and should not be to much work.
Perhaps start from here: https://github.com/pointhi/kicad-footprint-generator/blob/master/scripts/Connector/Connector_Molex/conn_ffc_molex_200528.py

It looks like F.Fab is missing a Pin1-indicator. How about the "cevron-style" that other connectors in the lib have?

At first it seems odd to me that name has `-1MP` in the name because there are two Mechanical-Pins. But that is consistent with other parts in the library. Perhaps because both MP have the same pin-number. I just could not find this in KLC.

openelectron:
@cp-aquila  I'd love to try the script. I didn't know whose tools existed. I will try and find some time to look at it.

PR #2349, opened 2020-07-04
Fix issue #2348, added missing F.Paste for pad number 3

### LINKED ISSUE #2348: Infineon_PG-HSOF-8-1 missing f.Paste for pad 3
Footprint for the Infineon_PG-HSOF-8-1 in Package_TO_SOT_SMD is missing the Front-Paste for the pad number 3. 
The footprint Infineon_PG_HSOF-8-1_ThermalVias has the correct F.Paste layer.

If nobody else can correct this I can submit a pull request with the correction.

aracosta1:
Fixes #2348, Missing F.Paste on pad 3.

I copied the F.Paste layer from the  Infineon_PG_HSOF-8-1_ThermalVias footprint who has the correct mask.


poeschlr:
A screenshot showing the resulting paste layout would help a lot for a review. Also does this layout come from a datasheet, if not what is the area coverage factor (paste area over copper are given in percent).

aracosta1:
The paste layout is taken from [https://www.infineon.com/cms/en/product/packages/PG-HSOF/PG-HSOF-8-1/#!downloads](https://www.infineon.com/cms/en/product/packages/PG-HSOF/PG-HSOF-8-1/#!downloads). Unfortunatey there is no pdf.

This is the resulting layout 

image

cpresser:
This is a regression from https://github.com/KiCad/kicad-footprints/pull/1704

The coordinates are not exactly the same:
```
-  (pad "" smd circle (at -0.08 -4.2 270) (size 1 1) (layers F.Paste))
+  (pad "" smd circle (at -0.07 -4.2) (size 1 1) (layers F.Paste))
```
The newer version does not have a rotation.

cpresser:
Ping @evanshultz  or @chschlue as both of you were involved in #1704

evanshultz:
Yep, I missed this up when doing #1704. My mistake, and thanks for catching it. I may have gotten bleary-eyed, but it looks like they are at the same coordinates as the HSOF-8-1 with thermal vias from #1704 so I assume that's good. Thank you!

PR #2350, opened 2020-07-04
Added LQFN-48 package with 16 centered ground pins and thermal vias

matyas1995:
for NXP MKW41Z part to Package_DFN_QFN.

This part requires a 7x7mm 48 pin VQFN footprint with 16 centered ground pads, which are created as one 3,5x3,5mm ground pad with 16 solder mask cutouts

Package datasheet: https://www.nxp.com/docs/en/package-information/98ASA00694D.pdf

I also added thermal and RF grounding vias as recommended in the design considerations: https://www.nxp.com/docs/en/application-note/AN5377.pdf

This footprint was not scripted with the kicad-footprint-generator. I have asked there and it was not deemed possible nor necessary to script it: https://github.com/pointhi/kicad-footprint-generator/issues/551

I still have a question regarding the correct naming of the footprint: Since it has 16 ground pads, would I define its name as 1EP and if yes, which size do I define then?

Thanks for your help,
Matyas

image

---

---


poeschlr:
Footprints for standardized packages like QFN must be made with our IPC generator scripts: https://github.com/pointhi/kicad-footprint-generator/tree/master/scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON

matyas1995:
Yes, you are right, but as I stated in the pull request, there is no way to script this package currently with the IPC generator scripts.

I have opened an issue in that repository and the reply that while the package belongs to the QFN library, there is no current way to script the center pad cutouts:

https://github.com/pointhi/kicad-footprint-generator/issues/551

So, thats the issue here...

> Footprints for standardized packages like QFN must be made with our IPC generator scripts: https://github.com/pointhi/kicad-footprint-generator/tree/master/scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON

poeschlr:
The screenshot does not look special to me. This is therefore absolutely possible with the script.

Also notice that we don't aim to rebuild suggested footprints for standardized packages exactly. This is completely unnecessary as there are industry standards available.

matyas1995:
So, in your opinion it would be enough to script a package with a single 3,5x3,5 mm center pad and vias, even though the chip package has 16 individual pads. Any excess paste would then drain through the vias, preventing the package from floating?

Thanks for the help,
Matyas

poeschlr:
If the package has 16 center pads then it is not a QFN ;). Also the -1EP might then no longer apply.

matyas1995:
Now I'm a bit confused, sorry for having to deal with me ':D

image

This is how the package looks like, I'm really not sure what type it is and in which library it would fit best. 

NXP also calls it QFN, but the center pad is not one single large pad like it is usual... I have found options in the script to define the cutouts for the paste mask, but not for the solder mask.

poeschlr:
The thing here is that sadly manufacturers don't really know what to call their packages. Which is why they often call things a QFN or LGA that are not really following the standards of such packages (If one would be strict than nothing that is currently called LGA is truly what LGA was meant to represent). I am not really sure yet how we should handle this specific instance to be honest. Maybe the easiest option would be to give the package a manufacturer prefix for now.

Is this package used for general components or is it special to a certain group (example for RF applications, sensors, ...)? If it is specialized to an application then maybe we can place it in a specialized library to avoid adding this strange package to the generic QFN lib.

matyas1995:
Hi,

According to the package brief (https://www.nxp.com/packages/SOT1649-1), this package is mainly used for some of the NXP Kinetis brand of microcontrollers, all of which integrate various 2.4GHz radio like Bluetooth, Thread etc.

Greetings,
Matyas

matyas1995:
So, I have thought a bit about your suggestions and here is what I would do:

I would rename the footprint package to "NXP_VQFN-48-1EP_7x7mm_P0.5mm_EP3.5x3.5mm"

I would still leave it in the DFN_QFN library. It could be moved to the RF library, but the NXP Kinetis chips using this package are no Bluetooth / Zigbee chips in the sense that this is not their main functionality, but just a peripheral option.

What do you think?

poeschlr:
I would suggest a name that makes it clear that the mask cutout is split and has a small size. Something like `NXP_VQFN-48-1EP_7x7mm_P0.5mm_EP3.5x3.5mm_16xMask[size x]x[size y]` with size x and size y indicating the size of an individual mask cutout.

matyas1995:
Okay, thank you for the help.

I changed the name now to the suggested one. I also separated the package into two versions, one with thermal vias and one without. This has the reason that the thermal vias are not part of the package footprint as in the NXP specification, but their layout is recommended in the hardware design recommendation document of the IC: https://www.nxp.com/docs/en/application-note/AN5377.pdf

chschlue (changes requested):
Please fix 3D model path (`NXP_LQFN-48-1EP_7x7mm_P0.5mm_EP3.5x3.5mm_16xMask0.45x0.45`)

matyas1995:
Thanks for the suggestion, since I have found no matching 3D model, I have removed it from the footprints altogether.

Greetings,
Matyas

chschlue:
Please keep it so a 3D model can be added later without touching the footprint again.

matyas1995:
So, you mean to add the 3D model path ${KISYS3DMOD}/Package_DFN_QFN.3dshapes/NXP_LQFN-48-1EP_7x7mm_P0.5mm_EP3.5x3.5mm_16xMask0.45x0.45.wrl even though no such 3D shape exists?

> Please keep it so a 3D model can be added later without touching the footprint again.

chschlue (approved):
Exactly.

PR #2351, opened 2020-07-06
Added silkscreen version of numerous symbols

Gabriel-06:
Added a number of silkscreen versions of symbols that I commonly use. I basically just copied the copper version and charged it to top silkscreen. I also fixed a couple of the current silkscreen symbols with problems:

Before:
Screen Shot 2020-07-06 at 10 01 11 pm

After:
Screen Shot 2020-07-06 at 10 33 42 pm

I also corrected symbol Symbol_Attention_CopperTop_Big as there was a gap in the exclamation point.

Before: 
Screen Shot 2020-07-06 at 10 21 13 pm

After:
Screen Shot 2020-07-06 at 10 21 23 pm

---

---


cp-aquila:
Holy shit... those CC-Logos are so shitty. This heap of line-segments makes my eyes hurt.
**Not your fault by any means. They were in the library like that.**
But I would like to have proper vector versions of them instead of this. But I am not sure this can easily be done.

Unfortunately KiCad does not support beziers, only arcs and lines. The original vector-logos here use beziers: https://creativecommons.org/about/downloads/
For example: https://mirrors.creativecommons.org/presskit/icons/cc.svg

With some tinkering around in various CAD programs one can create something like this:
grafik

Its not filled yet, just outlines. I will look into that next.

PR #2353, opened 2020-07-07
HDFN-12

cmcquinn:
Add HDFN-12 footprint for Microchip MIC33050-4YHL. Footprint from land pattern on page 16 of [datasheet](http://ww1.microchip.com/downloads/en/DeviceDoc/MIC33050-4MHz-Internal-Inductor-PWM-Buck-Power-Module-with-HyperLight-Load-DS20006120A.pdf). Matching symbol is included in PR https://github.com/KiCad/kicad-symbols/pull/2847.

image


cpresser:
Hi and thanks for you contribution.
I did a quick check (no full review) and have a few questions:

This is what I see in the data-sheet: 
image
image

I do not quite understand how this is supposed to work. Why are there pads that are not electrically or thermally connected to anything? What was microchip thinking when making that `recommended land pattern` drawing?
I would rather assume that pads 4,5,6 and the small strip should be one copper plane, but with 4 explicit paste pads. The same for the other island.

---

For the lulz, this is what **ultra librarian** thinks this should look like.
image

cpresser:
There seems to be an eval board from micrel where I found this potato-quality top-copper-layout:
http://ww1.microchip.com/downloads/en/DeviceDoc/MIC33050-4YHL%20Evaluation%20Board%20User%20Guide.pdf
image

One can see that there are big copper islands that connect the pins together.

myfreescalewebpage:
@cmcquinn ping are you still working on this ?

myfreescalewebpage:
No news of the author, indicate this is Abandoned

PR #2355, opened 2020-07-14
Added D_DO-247-2_Vertical

aris-kimi:
This footprint is a copy of TO-247-2_Vertical from Package_TO_SOT_THT library renamed.

Found in this diode datasheet: https://www.st.com/resource/en/datasheet/stbr3012-y.pdf

3D PR: https://github.com/KiCad/kicad-packages3D/pull/710

EDIT:
d_do-247

---

---


chschlue:
Please add a cathode silk marker for this diode-specific version.
Also, would you like to add horizontal variants for completeness?

aris-kimi:
Hi @chschlue , i am not sure how to mark cathode. Is just a vertical line between pads, near cathode pin, acceptable?
I am also not sure i understood you correctly. By horizontal variants you mean horizontal lines on the top of the symbol to indicate heatsink position?

diode

chschlue:
> Hi @chschlue , i am not sure how to mark cathode. Is just a vertical line between pads, near cathode pin, acceptable?

Yes. It's more or less up to you, but that's how it's been done for most THT diodes, I think.

> I am also not sure i understood you correctly. By horizontal variants you mean horizontal lines on the top of the symbol to indicate heatsink position? (noted area)

No, I was talking about diode versions of TO-247-2_Horizontal* (leads bent 90 degress and package flat on the board).

Another thing: I'm not sure the `-2` pin count suffix is necessary.

aris-kimi:
Oh, cool, yes i am on it. I will deal with it after work. I'll add cathode mark, remove `-2 `and add the other variants in the same branch.

aris-kimi:
I think that was the best approach i could thought.. If ok i will proceed with symbol's footprint name and footprint's 3D symbols.
I was wondering, do we need extra a _TadDown with an exposed pad?

chschlue:
LGTM so far.
What do you mean, extra _TabDown?

aris-kimi:
Something with an exposed pad for solder connected... If you understood what i mean, it will be ok i guess to add in a future PR...

chschlue:
I think I get it now...
AFAIK, TO-247 is isolated all around and has no conductive tab like, say, TO-220. Am I mistaken?

aris-kimi:
Honestly i don't know.. I am not sure if conductive tabs are used same way from all manufacturers per package type, but we can consider heat dissipation as a need, and the lack of heat sink in a design, if can be avoided, is a pro in my opinion. If you think we can add something for this purpose, inform me and i'll deal with it in a new PR.
Thank you for your comments and i am sorry for not making myself clear in the first place.

chschlue:
I don't think that's very useful as a one-size-fits-all solution. Thermal management is very project-specific.

Our DPAK FP, for example, has copper under the tab, but first and foremost as an electrical connection. It can (and is supposed to) of course be used as a heatsink by adding a copper zone around it, vias, and so on, manually, case-by-case, according to your specific requirements.
I don't see how adding a random patch of copper (and perhaps mask cutout) under a TO-247 package takes any work off users' backs.

PR #2359, opened 2020-07-16
update 3D L1008_2520Metric by scaling and using L_1206 3D file

florent1080:
the File L_1008_2520Metric.wrl is not present in 3Dpackage file.
https://github.com/KiCad/kicad-packages3D/tree/master/Inductor_SMD.3dshapes

So I change the 3D file to L_1206_3216Metric.wrl and scale it to shape the 1008 package.
Scaling it from 1, 1, 1 to 0.8333, 1.333 0.8333

image

Edit: Travis has error cause Component name is not the same as 3D model name.
So maybe we should create a 3D model rescaled, but those are not my skills.

---

---


pixeldexter:
This should really be directed towards kicad-3d-models-in-freecad@easyw. I already made a pull request https://github.com/easyw/kicad-3d-models-in-freecad/pull/378

chschlue:
@florent1080 Thanks for your first contribution, but @pixeldexter is right, we require an additional, fitting model.

@pixeldexter Are you going to submit the 1006 3d files?

pixeldexter:
@chschlue If someone can point me to a proper datasheet then why not?

chschlue:
@pixeldexter https://search.murata.co.jp/Ceramy/image/img/P02/JELF243B-0076.pdf?

pixeldexter:
@chschlue I believe this is the same 2520 metric size as I submitted, only the thickness is different. Do we have different footprints/models for different component heights?

BTW, I did a little research on the subject. IMO, there is a very little selection of 10**06** inductors on the market as opposed to 10**08**. Some distributors even do not have them at all. Do we still care?

chschlue:
Sry, I was confused. I meant 1008, this PR is about 1008, after all.
Now I get why you asked for another datasheet.

chschlue:
And no, we don't usually have different height chip components.
What we do have is electrolytic caps and such with different heights.

pixeldexter:
Makes sense, thx.

chschlue:
Still, are you going to submit the output of https://github.com/easyw/kicad-3d-models-in-freecad/pull/378?

pixeldexter:
Hmmm.. I don't think I really understand how the model makes its way from generator to repository. My guess is that someone from core team periodically checks out fresh generator sources, regenerates all the models and footprints and checks them into kicad/kicad-footprints and kicad/kicad-packages3d respectively. I.e. for the general user this should happen 'automagically' with the next release. Is this more or less correct?

As for submitting generated models I have a few doubts. First, getting generator to work required some fiddling to make it compatible with python3 and FC 0.18 (this is the configuration I have on Debian) and this is not finished yet. I got it _somewhat_ working, but this needs more effort. Second, I checked a few models and they are different from what is shipped with KiCAD. Not sure if this because my configuration is different, or this is just things evolving normally. In any case I'd prefer someone with "proper" setup did this.

chschlue:
> for the general user this should happen 'automagically' with the next release. Is this more or less correct?

That would be great, right? But unfortunately, that's not how it works. Never mind.

> Second, I checked a few models and they are different from what is shipped with KiCAD.

Interesting. Which ones?
There shouldn't have been many replacements since 5.1.6's release.

chschlue:
https://github.com/KiCad/kicad-packages3D/pull/717

florent1080:
Thank you guys for submitting it

pixeldexter:
@chschlue Thanks for submitting this.

WRT to model differences, I'd say all of them are different. E.g. if I compare L_0805 from master with the one I generated I get this:
```diff
@@ -93,8 +93,8 @@
 appearance Appearance{material USE PIN-01 }
 }
 Shape { geometry IndexedFaceSet 
-{ creaseAngle 0.50 coordIndex [0,1,2,-1,3,4,5,-1,6,0,2,-1,7,8,9,-1,7,9,
---
+{ creaseAngle 0.50 coordIndex [0,1,2,-1,3,4,5,-1,6,0,2,-1,7,8,9,-1,7,10,8,
```
Note the last index is different and there are more further down the line, I skipped them for clarity. I understand something is changed under the hood, but tracing this involves more time than I'd like to invest at the current moment. I'm not sure how critical this is, the model _appears_ to look normal, but still....

Any ideas?

chschlue:
@pixeldexter I'm not that much of a 3d model expert, so take this with a grain of salt:
Your diff is from the VRML model, which is meshed and therefore not an exact science. I think the particular mesh generated may depend on FreeCAD version, StepUp version, FreeCAD libraries versions, lunar phase, qi flow, quantum spin and whatnot.
How do the STEP models compare?

pixeldexter:
@chschlue You seem to be right. Meshing operation could be unstable wrt to sorting order and such, this is perfectly understandable. I wasn't able to determine is this due to mesh generation only, or underlying representation is somehow involved. However, comparing STEP models shows they are identical (up to dates and version numbers) so I think there isn't any major breakage and I have workable setup. Thanks for sorting this out.

PR #2370, opened 2020-07-23
Update SOIC-8_5.23x5.23mm_P1.27mm.kicad_mod

timkoers:
Fixed default 3D package (version 5.1.6 uses non existant 3d model)

---

---


aris-kimi:
Hi @timkoers , thank you for notice that.
Footprint's 3D file *.wrl/step must keep same naming with the footprint. This is also pointed by travis as `Violating F9.3`.
Reason behind this is that now it is only left to upload 3D files with correct naming(and dimensions) with a new PR to 3D models repo.
So this PR i think has to close... You should wait for a librarian for a better suggestion.

myfreescalewebpage:
@timkoers yes @aris-kimi is right, the 3D name should be the same used for the footprint. There are many footprint with no 3D package matching, but the field is filled so if somebody add the 3D package, we have not to modify the footprint again.
Rejecting the issue.
Joel

PR #2375, opened 2020-07-26
Footprint D_SOD-80 added

LHSmicius:
Datasheet: https://www.vishay.com/docs/84013/minimelf.pdf
D_SOD-80
Travis gives no errors

---

---


chschlue:
Can this be reconciled with the existing `D_MiniMELF`?

LHSmicius:
It's the same package (SOD-80 = miniMELF) according to datasheets I found so far:
https://www.st.com/resource/en/datasheet/tmmbat41.pdf#page=4
https://www.vishay.com/docs/85557/ll4148.pdf#page=3
https://www.vishay.com/docs/85630/ll103a.pdf#page=3

But in KiCAD it is little bit smaller:
4,8 mm instead of 5 mm,
1,7 mm instead of 2 mm,
1,3 mm instead of 1,25 mm,
2,2 mm instead of 2,5 mm.

NXP gives different dimensions for miniMELF (SOD-80) package
https://www.nxp.com/docs/en/package-information/SOD80C.pdf
But still it's not what's in KiCAD library.

Just interesting for me... But D_MiniMELF model is suitable. Differences between D_MiniMELF and what I draw are negligible.

evanshultz:
IPC 7351 gives fillet goals for MELF packages, so any change to the footprint I'd like to see is scripting that footprint generator. That will get rid of manufacturer variations as well.

However, https://github.com/KiCad/kicad-footprints/blob/master/Diode_SMD.pretty/D_MiniMELF.kicad_mod sadly doesn't have `SOD-80` in the description and/or keywords. @LHSmicius if you want to submit a PR to help this footprint it would be nice to get both common package names in the existing footprint for easier searching. If you want to do that, it think it would be best to close this PR and open another one for it.

LHSmicius:
I will edit D_MiniMELF footprint.

PR #2376, opened 2020-07-28
Added Lite-On SOP6

lorenz:
This adds the Lite-On SOP6 footprint, common to a lot of Lite-On optical isolation devices (like all parts and series [here](https://optoelectronics.liteon.com/en-global/Led/LED-Component/Detail/404?param4=11&param5=56)) and some more.

Sadly Lite-On doesn't actually maintain separate package documentation and doesn't name their footprints so you have to look at the dimensions to figure out what belongs to what. Considering that they call a variant of this the SOP4 I'm calling this one SOP6 since this is what it most closely represents.

I've intentionally deviated from their recommended footprint since it's extremely simplistic and certain things are just plain bad (pad length 1.3mm with a lead length of 1.0mm +- 0.25) in favour of a more IPC-compliant footprint.

Lite-On_SOP6_6 5x7 3mm_P2

[Datasheet](https://optoelectronics.liteon.com/upload/download/DS70-2001-026/MOC306X%20series%20Mar19.PDF)
---

---


chschlue:
Please use the FP generator found at https://github.com/pointhi/kicad-footprint-generator/tree/master/scripts/Packages/Package_Gullwing__QFP_SOIC_SO.

PR #2383, opened 2020-08-07
Added  Diode_Bridge_Comchip_SCVB-L

aris-kimi:
Datasheet: https://www.comchiptech.com/admin/files/product/SC35VB80S-G%20Thru506369.%20SC35VB160S-G%20RevB.pdf

Pads: 5x10mm
Holes(Oval): 1.1x6.7mm
~MountHoles: 10mm Dia~
45 Deg 3.70mm line is actually 3.68 in my design. couldn't make it precise. If not accepted i will try something different.
Countryard kept 0.2mm from fab (exept pin1 indicator lines) with square shape.

SCVB-L
Edit:
εικόνα

As far as i've checked FreeCAD's dimensions i am confident about this one.

Edit:
  ```
Violating F7.3

    Pad 1 should be denoted by rectangular pad

    Only pad 1 should be rectangular
```
I think this violation cannot be avoided.

---

---


myfreescalewebpage:
@aris-kimi starting the review here:

- [ ] Datasheet must be add in the description of the footprint, like all other footprints
- [ ] The two holes in the top right and bottom left corner are too big !!! It's currently 10mm but based on the datasheet I think 5mm is enough
- [ ] F.Fab : the triangle just beside pin 1 is not required. You have already the corner indication to indicate pin 1
- [ ] F.SilkS: the clearance with F.Fab lines should be 0.11mm by default
- [ ] F.CrtYd: the clearance with F.Fab lines should be 0.25mm by default

Else looks good.

Joel

aris-kimi:
Hi @myfreescalewebpage , thank you for your review.

Holes kept big enough to easy fit a screw with its washer.
For example, [M3.5](http://www.farnell.com/datasheets/1800969.pdf), that screw has a 6.2mm head Dia, so the minimum should be kept at 7mm for screws only.

But a washer(with lock-washer also) is also a good thing to have and [this](https://www.amazon.com/Plain-Finish-Stainless-Steel-Washers/dp/B076GW8RDW/ref=sr_1_2?dchild=1&qid=1601035068&refinements=p_n_feature_twenty_browse-bin%3A17420945011&s=industrial&sr=1-2) washer has an outside Dia of 11mm (i think this is too much), but for an 8mm outside Dia washer, i think 9mm drill hole is perfect.

Tell me your thoughts and i will update this later today.

myfreescalewebpage:
> Holes kept big enough to easy fit a screw with its washer.
For example, M3.5, that screw has a 6.2mm head Dia, so the minimum should be kept at 7mm for screws only.

What is the purpose of the screw if it goes over the pcb ? Do you have a photo or something to explain the mounting ?

> As presented, silk clearance is 0.14mm, if you think this is ok inform me to leave it as it is, else i will change that to 0.11mm later today.

Silkscreen clearance is to be measured from center to center of the lines, not from silkscreen border. Anyway this means you have clearance of about 0.2mm which is acceptable.

aris-kimi:
For the ability to mount and unmount after soldering.
I do not have a product image, but imagine those parts shown below and a big heatsink/metal enclosure.
εικόνα

Most IGBT modules i saw in motor drives, are mounted like this. I think it is better to take that responsibility from the designer.

myfreescalewebpage:
Thanks for the details, understood.
Because the size of the hole maybe different depending of the screw type, or maybe some other mounting possibilities can be found, I suggest to not make the hole, but just to suggest it on the Dwgs.User layer, has we have done with BUS_PCI_Express_Mini_* footprints at https://github.com/KiCad/kicad-footprints/pull/1124 with evanshultz (I was the author, he was the reviewer):

half

aris-kimi:
Very good idea, thanks. I will update this later today, and same logic will be applied to all of my footprints with this option.

aris-kimi:
First comment's image updated @myfreescalewebpage .

myfreescalewebpage:
Great job, looks very nice.
No more comment, merging.

PR #2389, opened 2020-08-11
Add support for Hybrid Oval  Cherry MX/Alps/Matias  footprints.

egladman:
Add support for Hybrid Oval Cherry MX/Alps/Matias footprints.

I'm the maintainer of [keebs.pretty](https://github.com/egladman/keebs.pretty). I'll like to merge my footprints into the official KiCAD library.

Please see egladman/keebs.pretty#7 for additional context and justification.

Screenshot from 2020-08-11 07-24-50

Datasheets/Specs:
- https://cdn.sparkfun.com/datasheets/Components/Switches/MX%20Series.pdf
- http://matias.ca/switches/quiet/

The keebs.pretty repo is licensed under CERN Open Hardware Licence. I'm more than happy to dual license it to comply with this Repo's license.

PR #2394, opened 2020-08-17
Keyboard footprints rework

perigoso:
This PR aims at reworking the current footprints related to keyboard switches, as well as it's structure, as in, directories.

It adds footprints for Cherry MX keyswitches, Alps / Matias keyswitches, Kailh Choc keyswitches and Kailh hotswap sockets for the Cherry MX format.

It Overrides the existing footprints, but they are mostly the same.

These footprints come from a [repository](https://github.com/ai03-2725/MX_Alps_Hybrid) that has been widely used by the community, and is generally used as opposed to the official libraries, given it's limited amount of footprints. They have suffered quite a bit of modification to better comply with KLC and fit with the rest of the official libraries.

Due to the high amount of footprints added i felt it was necessary to split the original Button_Switch_Button_Keyboard.pretty into:
- Switch_Keyboard_Alps_Matias.pretty
- Switch_Keyboard_Cherry_MX.pretty
- Switch_Keyboard_Hybrid.pretty
- Switch_Keyboard_Kailh.pretty 

There is some KLC breakage, but it's mostly on purpose:
- Violating F7.2 (Pad '1' not located at origin)
Given the nature of these components it is extremely convenient to have their origin at the center, and i feel it's justified.
- Violating F7.4 (Some THT pads have incorrect layer settings)
It is typical for keyboard switch footprint to not have mask on the switch side, ie, the solder mask is supposed to cover the pads on the side the switch is placed on.

F7.3 and F9.3 are just warnings and should not be an issue.

[Reference for Alps/Matias keyswitches](http://matias.ca/switches/quiet/)
[Reference for Cherry MX keyswitches](https://www.cherrymx.de/en/dev.html)
[Reference for Kailh Choc keyswitches](http://www.kailh.com/en/Products/Ks/CS/321.html)
[Reference for Kailh Hotswap Socket](https://github.com/KiCad/kicad-footprints/files/5082879/PG151101S11.pdf)

I am aware that this conflicts with a current PR #2389 Though I am submitting some footprints that directly conflict with those, they can still be merged given they are a slightly different implementation that may be of interest.

I'm sorry for the big PR/submission, I tried my best to have a clean set of footprints so that this process may go smoothly.


perigoso:
I split the libs because we have 200+ footprints with the new additions, and since on the KLC a limit of 250 is stated i assumed it was a hard requirement. But when i was looking at the issues i found this:

> The 250 requirement was purely arbitrary - mostly chosen because currently in the footprint selector window there is no filtering > option so scrolling through > 250 items to find the one you want is tedious.
> 
> Increasing this number should be fine, though.
> 
_Originally posted by @SchrodingersGat in https://github.com/KiCad/kicad-footprints/issues/544#issuecomment-386151827_

Since it's apparently not that big of a deal, maybe it's not necessary to split them.

chschlue:
250 is not a hard limit, but splitting by manufacturer may still be a good idea.

That being said, upstreaming these footprints is a great idea, just one question: Is this PR completely independent from https://github.com/KiCad/kicad-footprints/pull/2389? If so, it would be even better to coordinate both efforts.

@diegoherranz

perigoso:
@chschlue It is independent from that PR, though they add compatibility for the same kind of components, with slight differences:

image
image

cp-aquila:
Can we split this PR please? How about one PR per manufacturer, perhaps even split by series.
Reviewing this monster will not be fun.

perigoso:
I'll split it tonight

perigoso:
See #2416

perigoso:
Split into several PRs and a central discussion issue.

perigoso:
#2416

PR #2396, opened 2020-08-18
TO-92: move F.Fab ref des under part body

evanshultz:
Almost all vertical TO-92 footprints didn't have the F.Fab ref des under the part body, going against KLC F5.2.4.a (https://kicad-pcb.org/libraries/klc/F5.2/). This made the assembly layer of the board look rather ugly and confusing because the ref des was floating somewhere outside of the part. So pretty simple fix.

Since this is in violation of KLC I'm marking it as a bug.

Here's an example before and after:
image
image

I also unrotated pads of these footprints when I noticed them.

---

---


evanshultz:
Man, some of these footprints had loads of latent issues... I fixed most of the KLC warning and errors. I'm not sure there's anything more to be done with `TO-92Mini-2` as we know arcs don't close well and the thin annular rings are needed due to pin spacing. 'TO-92_HandSolder` appears to be hung up on pin 1 being rectangular and the other pins being round rect; I think this is an unlikely enough situation that we should allow it for now with the only possible fix being to make the check script realize that a round rect is OK for non-pin 1 pins if there is an offset as an oval pad would be a bad choice.

poeschlr:
We could define the rule as:
Pin 1 must either be rectangle or roundrect.
All other pins must be different from pin 1. If pin 1 is rectangular than all others can be roundrect as long as they are visually different enough (maybe require a larger minimum rounding factor than we normally do)
I would not allow pin 1 roundrect and the rest rectangular.

---

For v6 we might also look into chamfered (or basically octagon) pads as an option.


---

Thanks for these improvements. It is ok for me. (Further improvements will always be possible and welcome)

Issue #2381, opened 2020-08-06
SOT-143 footprint


matianfu:
I have check several documents for this package, including:

+ onsemi: https://www.onsemi.com/pub/Collateral/318A.PDF
+ diodes: https://www.diodes.com/assets/Package-Files/SOT143.pdf
+ nxp/nexperia: https://assets.nexperia.com/documents/package-information/SOT143B.pdf

Apparently, the sot-143 footprint in current KiCAD library is way too large. It is for wave soldering, not for reflow. Furthermore, Pad for pin 1 is misplaced. The edge is not well aligned. There is no reason to do so.

chschlue:
Are you willing to submit a fix?

matianfu:
I don't know if there is a standard? Including:

1. the line width of silkscreen, fab, and courtyard.
2. the fab and courtyard enlargement compared with the physical size and occupied area provided by the manufacturer.

Is there any agreement on those value? If so, I am happy to try it. :)

Also, the 3D model seems no appropriate for this package. All three documents listed in the first post agrees that the maximum distance between pin edges are no more than 2.5mm. But the SOT-143.wrl model seems to be at least 2.8mm width. I think this should be fixed either.

chschlue:
Short answer: you can use the line widths of the existing FP, they're fine; fab should be the nominal body outline while courtyard should be 0.25mm from fab and copper.
Long answer: https://kicad-pcb.org/libraries/klc/

If anything is unclear, feel free to ask.

perigoso:
I was working on this because i needed this package, i added a suffix to the originals `_WaveSolder` but it seems it's not done anywhere else, should we keep the old footprints? under what name? maybe specify the pad sizes

chschlue:
If there is demand (is SMT wave soldering still a thing?), I guess it would be fine to add `_WaveSolder` footprints.

perigoso:
I don't think it's relevant anymore, we don't have wave solder footprints for anything else, I'll submit the PR without them.

PR #2415, opened 2020-08-25
Correct SOT-143

perigoso:
Corrected footprint dimensions for standard SMD instead of wave soldering.

Related to  #2381 .

Datasheets fo this package have been posted in the issue.

The 3D model seems to not be correct according to the datasheets.

Before:
image
image

After:
image
image

The same applies to the `_Reversed` and `_Reversed_Handsolder` footprints


cpresser:
I assumed you used the diodes datasheet? I checked against that one.

- [x] Pad numbering and Position is good.
- [x] F.Fab is good
- [x] Courtyard is good
- [ ] Please add a datasheet link into the description
- [ ] Silkscreen should move outwards a little. It should not overlap the F.Fab line. Like so 
image
- [ ] The reverse part should be rotated so that pin1 is in the upper-left quadrant
- [ ] I am unsure how to check the handsoldering parts. Is there any rule you applied to elongate the pads?

perigoso:
> I assumed you used the diodes datasheet? I checked against that one.

I don't remember which datasheet i ended up using as a guideline, but i believe i checked them all against each other and they were the same or very close to the same dimensions (if I'm remembering correctly) so it shouldn't matter too much.

> Please add a datasheet link into the description

done. I added NXP's references because they had a both the normal SOT-143 and the reverse pinning version and so it is more consistent.

>  Silkscreen should move outwards a little. It should not overlap the F.Fab line. Like so

done.
image

> The reverse part should be rotated so that pin1 is in the upper-left quadrant

done.
image

> I am unsure how to check the handsoldering parts. Is there any rule you applied to elongate the pads?

I looked for guidelines, but since i didn't find any i just did what felt right to me, the pins are slightly shorter than the original hand-solder versions

Issue #2416, opened 2020-08-25
Keyboard footprint merge discussion


perigoso:
I'm creating this issue to have a centralized place for discussion regarding the massive keyboard footprints merge that i now split into multiple PRs.

This way we have somewhere to discuss things that affect all the fp's and we can better coordinate things.

The original PR #2394

perigoso:
The original `Button_Switch_Keyboard.pretty` Includes a README.md, is this required? What's it's purpose? It seems most footprint libs don't include this file.

diegoherranz:
> The original `Button_Switch_Keyboard.pretty` Includes a `README.md`, is this required? What's it's purpose? It seems most footprint libs don't include this file.

No. In fact, I would remove it. It may come from the old repo structure in which every .pretty was a separate repo.

perigoso:
In that case maybe we'd remove it from all libs that include one? can I submit a PR with that or Just for the `Button_Switch_Keyboard.pretty` lib, since it would be the last file in that folder after the keyboard switch PRs move things around.

perigoso:
So I had `_NoSilk` variants of footprints, because some people, including me, don't include silkscreen on the switch footprints for aesthetic reasons, but i was unaware that it was possible to hide it in bulk in `pcbnew`, so after some digging around and trying things out i came to this:

![image](https://user-images.githubusercontent.com/39195157/91366888-e3edaa80-e7fc-11ea-8244-fff18b60cee2.png)

This is a good versatile solution, and i feel the `_NoSilk` variants don't make sense given this option, maybe it might not be immediately clear to people but still, any input?

evanshultz:
We require assets to comply with http://kicad-pcb.org/libraries/klc/, unless there's a specific reason not to or that doc needs an update. In this case, it's not. So yes, please only keep versions with silk as those do comply with the library guidelines.

perigoso:
I am familiar with [KLC](http://kicad-pcb.org/libraries/klc/), I see no reason the `_NoSilk` versions of the footprints do not comply, hence why i was ok with including them in the first place.

chschlue:
[Thou shalt not leave out silk](https://kicad-pcb.org/libraries/klc/F5.1/)

perigoso:
The reference was still included, they just didn't have the square outline

chschlue:
Yes, if you want to be really, really finnicky about the wording, you could argue that outline silk is not explicitly required.

[...]

PR #2418, opened 2020-08-25
Added Cherry MX kbd switch fp, replaced existing

perigoso:
See #2416 and #2394

This PR Adds Cherry MX keyboard switch footprints, replacing some existing ones.


perigoso:
> So I had `_NoSilk` variants of footprints, because some people, including me, don't include silkscreen on the switch footprints for aesthetic reasons, but i was unaware that it was possible to hide it in bulk in `pcbnew`, so after some digging around and trying things out i came to this:
> 
> image
> 
> This is a good versatile solution, and i feel the `_NoSilk` variants don't make sense given this option, maybe it might not be immediately clear to people but still, any input?

[original comment](https://github.com/KiCad/kicad-footprints/issues/2416#issuecomment-681177486)

perigoso:
Generated footprints with kicad-footprint-generator.

image
image

perigoso:
Re-generated footprints with kicad-footprint-generator after fixups.

PR #2420, opened 2020-08-25
Add Kailh Choc low profile kbd switch

perigoso:
See #2416 and #2394

This PR Adds Kailh Choc low profile keyboard switch footprints.


cp-aquila:
I think I will start with those, since it is just 5 footprints.

So far all I could find for documentation were those pictures:
grafik
* Is there more, better documentation?
* Where are the pins located (x/y) coordinates?
* According to KLC the footprint-origin needs to be on Pin1 for THT components. I am thinking about deviating from that rule. For a keyboard-designer the center of the keycap would be a better choice. @chschlue What do you think?
* That 5.5x2.95 area, what is it?

chschlue:
> * According to KLC the footprint-origin needs to be on Pin1 for THT components. I am thinking about deviating from that rule. For a keyboard-designer the center of the keycap would be a better choice. @chschlue What do you think?

I'm not too familiar with the contemporary keyboard industry but I assume your typical off-the-shelf keyboard is not hand-assembled by wage slaves. Meaning if some sort of PnP origin is documented somewhere, that would solve the problem.
Other than that, I'm with you here.

perigoso:
Firstly I want to state that I did not originally create most of these footprints, and though i went through most of them and made some extensive modifications, I assumed the footprint was correctly made according to documentation in the first place.

I think they have already been tested in real life, so i did not actually check it though i probably should've anyway.

That being said I made an effort to try to find the documentation for review, which i posted in the first PR #2394 [Reference for Kailh Choc keyswitches](http://www.kailh.com/en/Products/Ks/CS/321.html) it comes from the official manufacturers site so assume it's the best we got, unless maybe we tried contacting them.

You found the documentation for the old version of these switched which i believe are not manufactured anymore, this should explain some of the discrepancies.
image

> * According to KLC the footprint-origin needs to be on Pin1 for THT components.

Like I mentioned in the original PR this was intentional as I felt it made sense given the nature of these components.

> * That 5.5x2.95 area, what is it?

That is a cutout the component has that is meant for an LED to be mounted behind for retroilumination. this could be added to a documentation layer, but I don't find it necessary.

> * What is the 1.00u thing in the name? Imho the name should be something like CPG135301. The library-name already has the manufacturers name and series.

That is the dimension of the Keycap 1.00u in the keyboard world means 19.05mm this is standard nomenclature, the documentation layers have the keycap represented this is for layout and alignment purposes. such a specific label does not make sense in my eyes given that this footprint fits multiple references, and Kailh Choc is a unique series and is representative enough I feel.

perigoso:
ups.

perigoso:
Generated footprints with kicad-footprint-generator.

image

perigoso:
[datasheet](https://github.com/keyboardio/keyswitch_documentation/blob/master/datasheets/Kailh/CPG123201D01--Choc_Blue.pdf)

cpresser:
I added myself as reviewer. But I doubt I will be able to wrap this up before our move to gitlab. If somebody else wants to take this over, please go ahead.

Related: https://github.com/pointhi/kicad-footprint-generator/pull/605

perigoso:
Re-generated footprints with kicad-footprint-generator after fixups.

PR #2444, opened 2020-08-31
EA DOGM128 and DOGL128 displays added

kajusK:
EA DOGM128 and DOGL128 displays added
datasheets:
 * https://www.lcd-module.de/eng/pdf/grafik/dogm128e.pdf
 * https://cdn.sos.sk/productdata/84/8a/86e59451/ea-dogl128b-6-1.pdf

2020-08-31-220553_1920x1080_scrot


myfreescalewebpage:
Hi @kajusK , thanks for contributing,

EA_DOGL128-6
- [ ] F.SilkS: clearance of 0.11mm with F.Fab is enougth
- [ ] F.CrtYd: top and bottom lines are not horizontal

EA_DOGM128-6
- [ ] I disagree with the distance between pin 1 and the F.Fab layer. (48.26-47.96)/2=0.15mm instead of 0.075. Can you check ?

Will continue after the changes

Cheers,
Joel

kajusK:
Hi @myfreescalewebpage,
I've updated the distance from pin 1 to F.Fab (thank you, I guess I've made some mistake during calculations) and made the crtyd lines horizontal. However I'm not sure about the clearance 0,11mm you mentioned. Should I move all silkscreen lines to be 0,11mm from the F.Fab?

Thank you,
Jakub

myfreescalewebpage:
> However I'm not sure about the clearance 0,11mm you mentioned. Should I move all silkscreen lines to be 0,11mm from the F.Fab?

Yes. This is the default value also used when we are generating the footprints in python

Joel

kajusK:
@myfreescalewebpage Thanks, updated as requested. Might be worth adding to KLC?

myfreescalewebpage:
KLC at https://kicad-pcb.org/libraries/klc/F5.2/. the 0.11mm value is not detailed however, I know it mainly because I practice scripting :)

myfreescalewebpage:
Thanks for the update, it's ok for me. One detail I have just notice: there is a small rectangle to be added at the bottom of the footprint, shown on the drawing in the datasheets:

Capture

I think it will be better to add it please. The dimensions are not fully indicated, let's say based on the drawing it will be 0.8x5mm. If you have one display you measure on it.

Joel

kajusK:
Good point. Unfortunately I don't have the physical parts, so I can't measure it, so made it the dimension you proposed.

myfreescalewebpage:
Looks good ! Great job done here. Merging.

PR #2447, opened 2020-09-03
Added OnSemi custom WDFN6 footprint

Duckle29:
Added OnSemis custom WDFN6 footprint used in the LC05111CMT chip

---

## :warning: Deprecation warning

On Oct 1, 2020 this repository, including issues and pull requests, will be archived and transferred to [gitlab.com](https://gitlab.com/kicad/libraries/kicad-footprints/). In order for your pull requests and issues to be imported into GitLab you must set your email address on GitHub to public and use the same address for your GitLab account. Or login to GitLab at least once using the GitHub icon. Otherwise the importer can't correlate the account information and the issues/comments on GitLab will be owned by `kicad-bot` ([importer documentation](https://docs.gitlab.com/ee/user/project/import/github.html#how-it-works)).

---

[Datasheet](https://www.onsemi.com/pub/Collateral/511BZ.PDF)

Symbol PR: https://github.com/KiCad/kicad-symbols/pull/2968

## Screenshots
image
image
image
image

---


Duckle29:
A note on the pad naming. The datasheet doesn't explicitly state which is which, however they say that "pin 8" is a pad, and is connected to vss, and that "pin 7" is the fet drains, which shouldn't be connected to anything in use.

This is mirrored in the example layout:
image
It also makes sense to link the hottest part "power fet drain" to a thermal pad, as they've done in their example layout. The thermal connection to pin 7 is purely a thermal, as it isn't electrically connected to anything else.

cpresser:
The Pin1 style on Silkscreen should be like the one used for `WDFN-8_2x2mm_P0.5mm`.
image

I will do a full review soon.

Duckle29:
image
image
Updated :)

cpresser:
- [x] F.Fab is okay
- [x] Pads 1-6 position, layers and size okay. They have an orientation of -90° which can just be set to 0 instead. But there is no explicit rule demanding that. So its okay.
- [x] Exposed Pads Position and Size are okay.
- [x] Silkscreen is okay, it could be slightly tighter around F.Fab. But that is not mandatory.
- [ ] The courtyard on the left- and right side is to close to the copper.
- [ ] Naming: I would like to drop the OnSemi prefix. We already have a few footprints in the library without a prefix that would fall into the same category (`WDFN-8-1EP_2x2.2mm_P0.5mm_EP0.80x0.54`).
- [ ] Naming again, it should be `4x2.6mm`
- [x] EP-Pad numbering is consistent with other resources I found:
image
image

Duckle29:
Done :)

cpresser:
You removed a file, but did not add a new one^^

Duckle29:
Ahh fun. untracked files. Should be added and pushed now :)

cpresser:
I am not quite happy with the courtyard. 
Your lines are at y=2.33, they should be at 2.35

A quick way to visually inspect that is to increase the pad-size by 0.5mm:
image

While you are at it, can you please also move the silkscreen closer to F.Fab. That would make me happy :heart:

Duckle29:
Done, and thanks for the tip on expanding the pad. Great way to do that! :)

cpresser:
Good work, thank you very much!

PR #2450, opened 2020-09-04
Add TI PWQFN N16 with 1.7x1.7mm EP

zonque:
Used by this OpAmp: https://www.ti.com/lit/ds/symlink/tlv9064.pdf


zonque:
Generated with https://github.com/pointhi/kicad-footprint-generator/pull/590

PR #2461, opened 2020-09-09
Add TSSOP16 variant with EP

einball:
Scripted footprint for Allegro A4954:

Symbol PR: https://github.com/KiCad/kicad-symbols/pull/2399
Generator PR: https://github.com/pointhi/kicad-footprint-generator/pull/615
Datasheet: https://www.allegromicro.com/-/media/Files/Datasheets/A4954-Datasheet.ashx

Bildschirmfoto vom 2020-09-09 16-58-24


myfreescalewebpage:
Hello @einball thanks for the footprint,

I try to make one of my first footprint review so basically this is more questions :)

Source : checked
Dimensions : checked
Pins : checked
EP : checked

I'm wondering about the thermal vias:
- How the drilling diameter is determined ? (0.3mm)
- What is "paste_avoid_via" and why it is "false" ?
- What is "EP_num_paste_pads" and why it is "[2, 2]" ?

@chschlue if you can help me to learn on this, thanks! Travis reports some warnings and this also surprising me because I was expecting to have none for this kind of simple footprint generated with the generator.

Joel

chschlue:
- drill diameter: There is no rule. I usually request 0.3mm (at least for large EPs like this one) because that's kind of a common lower limit cheap fabs have (meaning it allows more people to use the FP out of the box)
- paste_avoid_via: makes the algorithm place paste pads outside vias (to prevent solder wicking through the holes). Last I checked, this leads to strange results by the generator. Also, via tenting/filling avoids that problem as well. Generally not used.
- EP_num_paste_pads: array defining solder paste pads on the EP in x/y direction (to achieve a paste coverage of 65% by default.) As with drill diameter above, there's no rule. Ask for a reasonable value (a lot of paste pads on a tiny EP leading to 10µm stencil cutouts don't make sense, for example)
- Travis error: The way EP lands are created, the FP contains a copper/mask only pad along with several smaller paste pads whereas the CI script expects an SMD pad to have copper/mask/paste layers enabled. False alarm in these cases, error can be ignored.
- output comparison: As long there's nothing obviously wrong with the FP PR, I assume good faith and don't generate the FP myself for comparison

HTH

einball:
I set the drill to 300um for the reason @chschlue said. The paste probably does not have to be arranged in a 2x2 pattern. However, it may prevent solderbridges with a higher stencil. Same goes for the avoid_via - When the holes have a bit of paste already in them they are more likely to suck excess solder down.

There are also still errors in the kicad-library-utils due to line widths I can't do anything about. I'm not sure if the library utils need an update for this or the footprint-generator.

myfreescalewebpage:
@chschlue and @einball thanks so much for the details and the time you have spent to explain me all this stuff.

My last question for this PR: the thermal vias version has the four Paste pads (2x2 with 65% of coverage) but not the first footprint. It has only a single and very large Paste pad instead. I have checked across the TSSOP footprints in the lib, some are in the same case, some other have 2x2 Paste pads for both variants.

According to the previous details and my experience with reflow soldering (I have a oven at home), 4 small Paste pads is better. Any rules to follow for the new submissions ?

Thanks again,
Joel

einball:
I doubt, there can or should be a rule for pad coverage. Some parts need more, some less - especially the ones with a tight clearance between pins and pad. 

The ThermalVias footprint is also autogenerated from the footprint generator, maybe there is an option to change it?

chschlue:
`EP_num_paste_pads` is only specified for the thermal_vias variant in the generator PR.
There should be an `EP_num_paste_pads: [2, 2]` for the whole size definition (i.e., one level above.)

myfreescalewebpage:
Thanks @chschlue 
@einball I think you should move the EP num paste pad definition one level up to have the 4 Paste pad on the non thermal via variant.

einball:
I tried to but it will not let me:

```yaml
  EP_size_x:
    nominal: 3
  EP_size_y:
    nominal: 3
  EP_num_paste_pads: [2, 2]
  thermal_vias:
    count: [2, 2]
    drill: 0.3
    paste_avoid_via: false
```
gives me the following output (since it seems to require the argument)
```
generating part for parameter set TSSOP-16-1EP_4.4x5mm_P0.65mm_EP3x3mm
Traceback (most recent call last):
  File "./ipc_gullwing_generator.py", line 622, in <module>
    gw.generateFootprint(cmd_file[pkg], header)
  File "./ipc_gullwing_generator.py", line 144, in generateFootprint
    self.__createFootprintVariant(device_params, header, dimensions, True)
  File "./ipc_gullwing_generator.py", line 275, in __createFootprintVariant
    EP = ExposedPad(
  File "/home/einball/git/kicad-footprint-generator/scripts/Packages/Package_Gullwing__QFP_SOIC_SO/../../../KicadModTree/nodes/specialized/ExposedPad.py", line 153, in __init__
    self._initPaste(**kwargs)
  File "/home/einball/git/kicad-footprint-generator/scripts/Packages/Package_Gullwing__QFP_SOIC_SO/../../../KicadModTree/nodes/specialized/ExposedPad.py", line 260, in _initPaste
    self.paste_layout = toIntArray(kwargs.get('paste_layout', [1, 1]))
  File "/home/einball/git/kicad-footprint-generator/scripts/Packages/Package_Gullwing__QFP_SOIC_SO/../../../KicadModTree/util/paramUtil.py", line 86, in toIntArray
    return toNumberArray(value, length, min_value, member_type=int)
  File "/home/einball/git/kicad-footprint-generator/scripts/Packages/Package_Gullwing__QFP_SOIC_SO/../../../KicadModTree/util/paramUtil.py", line 58, in toNumberArray
    raise TypeError('Unsupported type: {}'.format(type(value)))
TypeError: Unsupported type: <class 'NoneType'>
```

Having it in both levels produces an output but no paste on the ThermalPad variant. I have not yet had the time to look at the fp-generator code yet though.

einball:
I added a second definition. If I only move the existing one a level above, the Python script fails with the following error:

```
generating part for parameter set TSSOP-16-1EP_4.4x5mm_P0.65mm_EP3x3mm
Traceback (most recent call last):
  File "./ipc_gullwing_generator.py", line 622, in <module>
    gw.generateFootprint(cmd_file[pkg], header)
  File "./ipc_gullwing_generator.py", line 144, in generateFootprint
    self.__createFootprintVariant(device_params, header, dimensions, True)
  File "./ipc_gullwing_generator.py", line 275, in __createFootprintVariant
    EP = ExposedPad(
  File "/home/einball/git/kicad-footprint-generator/scripts/Packages/Package_Gullwing__QFP_SOIC_SO/../../../KicadModTree/nodes/specialized/ExposedPad.py", line 153, in __init__
    self._initPaste(**kwargs)
  File "/home/einball/git/kicad-footprint-generator/scripts/Packages/Package_Gullwing__QFP_SOIC_SO/../../../KicadModTree/nodes/specialized/ExposedPad.py", line 260, in _initPaste
    self.paste_layout = toIntArray(kwargs.get('paste_layout', [1, 1]))
  File "/home/einball/git/kicad-footprint-generator/scripts/Packages/Package_Gullwing__QFP_SOIC_SO/../../../KicadModTree/util/paramUtil.py", line 86, in toIntArray
    return toNumberArray(value, length, min_value, member_type=int)
  File "/home/einball/git/kicad-footprint-generator/scripts/Packages/Package_Gullwing__QFP_SOIC_SO/../../../KicadModTree/util/paramUtil.py", line 58, in toNumberArray
    raise TypeError('Unsupported type: {}'.format(type(value)))
TypeError: Unsupported type: <class 'NoneType'>

```

myfreescalewebpage:
@einball not understood your last comment.
The https://github.com/pointhi/kicad-footprint-generator/pull/615 contains a single definition only.
Else, footprints seems correct. Just waiting your answer to the previous question to merge this.
Joel

myfreescalewebpage:
@einball ping

einball:
I updated both PRs at the same time. Please see: https://github.com/pointhi/kicad-footprint-generator/pull/615/commits/5422025961267f0c25ba220348d09a1f8c48f372#diff-c961c19afa1f1d5d0dfccc38d5b35fa8R499

It contains 2 definitions for EP_num_paste_pads, one on the component level and one indented for the ThermalVias variant. If I only have one definition on the component level but none on the 'ThermalVias' level the above mentioned error appears.

myfreescalewebpage:
Understood, thanks for the details!
Merging.

PR #2466, opened 2020-09-13
Fix pin numbering in 1.5mm pitch Pico-Lock connectors (504050 series)

### LINKED ISSUE #2465: Incorrect pin numbering on Molex Pico-Lock 504050 series connectors
The pin numbering of the 1.5mm pitch versions of the Molex Pico-Lock 504050 connectors is backwards.

The files are in this repository at `Connector_Molex.pretty/Molex_Pico-Lock_504050-**91_1x0*-1MP_P1.50mm_Horizontal.kicad_mod`.

The connector family page is [here](https://www.molex.com/molex/search/deepSearch?pQuery=category%253A%2522PCB%2BHeaders%2522%2540orientation%253A%2522Right%2BAngle%2522%2540pitchmatinginterface%253A%25221.50mm%2522%2540productname%253A%2522Pico-Lock%2522), the datasheet [here](https://www.molex.com/pdm_docs/sd/5040500291_sd.pdf).

As well as the numbering being specified in the datasheet, circuit 1 is also marked with a physical indicator on the connectors.

These footprints are generated using [kicad-footprint-generator](https://github.com/pointhi/kicad-footprint-generator), using [smd_single_row_plus_mounting_pad.py](https://github.com/pointhi/kicad-footprint-generator/blob/master/scripts/Connector/Connector_SMD_single_row_plus_mounting_pad/smd_single_row_plus_mounting_pad.py) with values from [conn_molex.yaml](https://github.com/pointhi/kicad-footprint-generator/blob/master/scripts/Connector/Connector_SMD_single_row_plus_mounting_pad/conn_molex.yaml).

The fix is really simple, the value [here](https://github.com/pointhi/kicad-footprint-generator/blob/491dfee3abab24f1449327aae393530a9c6d4fa2/scripts/Connector/Connector_SMD_single_row_plus_mounting_pad/conn_molex.yaml#L321) for `pad1_position` must be changed from `bottom_left` to `top_left`.

I'd be happy open a PR with the fixed footprints but I'm not sure what the process is given that changing the pin numbering would break any existing designs.

This issue came up because I am trying to add the 2mm pitch version of this family ([PR here](https://github.com/pointhi/kicad-footprint-generator/pull/625)).

image

image

justyn:
The existing footprints have reversed pin ordering, so that pad 1 does
not match that on the datasheet or indicated by the pin 1 indicator on
the connector itself.

The footprints are created with kicad-footprint-generator with:

```
smd_single_row_plus_mounting_pad.py conn_molex.yaml
```

Footprints generated by this script must always have pin 1 on the
left-most side, so in order to fix the numbering the connectors are
rotated by 180 degrees.

These files are generated from updated value tracked in https://github.com/pointhi/kicad-footprint-generator/pull/628

Fixes #2465

This update also fixes incorrect tags that specified "top entry" instead
of "horizontal".

image

image


chschlue (approved):
Thanks.

PR #1304, opened 2019-01-16
Add various Samtec mezzanine footprints from the Q Strip/Q Pairs, Basic Blade & Beam, Razor Beam, and mPOWER families

calebreister:
This pull request contains footprints for [Altera HSMC cards](https://www.intel.com/content/dam/www/programmable/us/en/pdfs/literature/ds/hsmc_spec.pdf). There are two variations, which use different connectors:

1. [Samtec ASP-122952-01](http://suddendocs.samtec.com/prints/asp-122952-01-mkt.pdf): a modified Samtec QTH-series connector with every third pin removed in Bank 1 to improve signal integrity for transceivers.
2. [Samtec QTH-090-01-x-D-A](https://www.samtec.com/products/qth-090-01-l-d-a-tr): identical to the ASP-122952-01, except that all of the pins were left in Bank 1. It is not as good for transceivers, but is much more common.

The following footprints were added...

Samtec_HSMC_ASP-122952-01_P0.5mm_Vertical
screenshot_20190115_182233

Samtec_QTH-090-01-x-D-A_P0.5mm_Vertical
screenshot_20190115_182042

Samtec_HSMC_Card_QTH-090-01-x-D-A_3.076x0.932in_P0.5mm_Vertical
screenshot_20190115_181745

**KLC Notes:**
- The `HSMC_Card` footprint anchors are centered on the connector, not the courtyard. Changing this would cause issues with the 3D models.
- The `HSMC_Card` courtyards do not align to a 0.01mm grid, since the specifications were provided by the HSMC specification.
- Some 3D models are reused, so the footprint names do not always match.

**See also:**
- [Pull request](https://github.com/KiCad/kicad-symbols/pull/1428) for `HSMC_Card` symbols
- [Pull request](https://github.com/KiCad/kicad-packages3D/pull/512) for the 3D Models

------------

Thanks for creating a pull request to contribute to the KiCad libraries! To speed up integration of your PR, please check the following items:

- [x] Provide a URL to a datasheet for the footprint(s) you are contributing
- [x] An example screenshot image is very helpful 
- [x] If there are matching symbol or 3D model pull requests, provide link(s) as appropriate
- [x] Check the output of the Travis automated check scripts - fix any errors as required

PR #2471, opened 2020-09-15
Create Samtec_QFS-52-3MP_0.26x2mm_P0.625mm_MH2.6mm.kicad_mod

UsairemAlamgeer:
image

http://suddendocs.samtec.com/productspecs/qms-qfs.pdf

---


calebreister:
Is that the only Q2 connector you are adding? I've been working on #1304, which adds a wide variety of Samtec connectors. It shouldn't be too difficult for me to write a script that generates *all* of the vertical Q2 connectors, including that one. Also, it appears that none of those would meet the KLC, since the silkscreen lines are too small and would be covered by the connector bodies.

UsairemAlamgeer:
So are you suggesting that I add the F.SilkS outside the connector premises so that it is visible?

calebreister:
Yes. Since the CI test passed, I assume you did. You will also need to consult the KLC for the correct part naming convention. For example, the Samtec part at the top should actually be named `Samtec_QFS-026-xx.xx-x-D_P0.635mm_Vertical`, except that the part corresponding to the exact footprint you've created doesn't exist (according to the [product page](https://www.samtec.com/products/qfs) there is no mounting hole option for the QFS series). What I as suggesting is that you remove the QFS part from this PR, and I will create a script that generates every standard connector in the QFS/QMS family and add all of them to #1304.

PR #2475, opened 2020-09-19
Change sot563 pad y dimension to 0.34mm to make sure the pad to pad clearance

jxltom:
As mentioned in https://github.com/KiCad/kicad-footprints/pull/581, all recommended layout has pady smaller than 0.34mm, usually 0.3mm. Though setting pady as 0.4mm matches all datasheet, this makes pad to pad clearance to only 0.1mm. Usually the pcb fabrication in production requires pad to pad clearance as 5mil which is 0.127mm. So this PR changes the pady as 0.34mm, make sure the pad to pad clearance as 0.16mm, then fix the issue

https://www.centralsemi.com/PDFS/CASE/SOT-563PD.PDF
https://www.diodes.com/assets/Package-Files/SOT563.pdf
https://www.torexsemi.com/file/SOT-563/SOT-563-pkg.pdf
https://www.monolithicpower.com/pub/media/document/MP5075L_r1.0.pdf
http://www.bourns.com/docs/Product-Datasheets/CDSOT563-T05C.pdf


jxltom:
No idea why the CI is failing, I didn't change the courtline layer

evanshultz:
There was an issue with our check script that is now fixed. That was causing the bogus courtyard failure.

While some datasheets did show 0.4mm padY, you're right that 0.1mm spacing rather tight. The lead widths I see in the link above are 0.27mm or 0.3mm in width. So surely a 0.34mm high pad is fine. We can extend our generators to support SOTFL package types and then use IPC rules to make footprints, but this is better in the interim. Thank you!

PR #2477, opened 2020-09-20
Added footprints for two solar cells by IXYS

jlehmke:
Hello,
this is my first contribution to KiCAD. I've added two SMD solar cells by IXYS to a new Library (Solar.pretty) as I didn't found an existing library suitable for solar cells.

SM111K04L

Added parts:
- SM111K04L - https://ixapps.ixys.com/DataSheet/SM111K04L.pdf
- KXOB25 - https://ixapps.ixys.com/DataSheet/KXOB25_14X1F.pdf

KXOB25 solar cells are available in many different configurations, sharing the same form factor. The attached data sheet is an example of one of those variants. 

To ease manual soldering the pad length has been increased by 1mm to the side. There is no recommended pad layout mentionend in the data sheets. As there is barely no self centering during reflow due to high weight/pad ratio, 0.5mm distance has been used for CrtYrd.

The footprints are compatible to the Device/Solar_Cell symbol. The check script ran without errors.


myfreescalewebpage:
Hello @jlehmke and welcome for this first contribution. I will review it and here are the first comment to integrate your contribution:

General:
- [ ] "fp-lib-table" file should be completed to indicate the new library
- [ ] Datasheet should be added to the description of the footprints (have a look to other footprint to have example)

IXYS_KXOB25
- [ ] Name should be "IXYS_KXOB25-14X1F"
- [ ] "+" and "-" on F.Fab are not useful
- [ ] "+" on F.SilkS should be fully visible after mounting, please move the symbol outside of the F.Fab area
- [ ] Pin 1 should have marker on F.Fab. I suggest a corner at 45deg as you can see on transformer "Pulse_P0926NL" for example
- [ ] Distance between F.Fab and F.SilkS lines is 0.11mm by default (instead of 0.25mm actually)
- [ ] Distance between F.Fab and F.CrtYd should be 0.25mm (instead of 0.5mm actually)
- [ ] Pads are too large, they should be 2x7.4mm. If you want to have a version to solder yourself, you can do a "_HandSoldering" version after this one.

IXYS_SM111K04L 
- [ ] The same comments apply to this footprint too

Will continue review after those changes

Joel

PR #2481, opened 2020-09-22
Added TO-46-4 + ThermalShield for LM399

PatrickBaus:
This is the footprint for the [Analog Devices LM399](https://www.analog.com/media/en/technical-documentation/data-sheets/199399fc.pdf). It comes in a TO-46-4 package but is inside an enclosure made of Polysulfone.

This PR comes with the footprint PR over here : https://github.com/KiCad/kicad-symbols/pull/1548. I will also add the 3d model as soon as I have updated it according to the design rules. Should take a day or so.

The first pad is oval, because this makes soldering easier than the rectangular version and the current TO-46-4 package has an oval pad for pin 1 as well.

Screenshot:
Screenshot from 2020-09-22 10-13-01

---

---


evanshultz:
Is this a copy of the existing TO-46-4 footprint, but with a couple circles added, the courtyard changed, and the description/keywords changed? It looks that way but I want to confirm.

1. It looks to me like the nominal heatshield radius is 5.207mm, so this one is too big. Can you confirm?
1. Please add a circle on F.Fab at the same location. Right now there is only a silk circle.
1. Revert the keywords to the existing TO-46-4 footprint. I don't expect this heatshield is _only_ for LM399.
1. For the description add a link to the datasheet. And if the heatshield is specific to ADI (LT) then we may want to make this a manufacturer-specific footprint. Do you know?

PatrickBaus:
1. Regarding the silkscreen, yes indeed it is bigger, than the expected radius. The reason is, that the heat shield of the LM399 is loosely packaged around the diode. The diode is not mounted inside and therefore the legs of the diode hold it in place. The drill holes in the heat shield are large than the diode legs to accomodate for tolerance. Therefore I have increase the the circle radius by about half the leg diameter tolerance.

2. done. Thx for mentioning
3. If *only* ever seen this package on the LM399
4. Indeed, I guess you are right and is manufacturer specific. I have changed the name accordingly.

evanshultz:
1 Alright. I'll take your word for it since I don't have one in my hand.
4 Please add a link to the datasheet in the description.

Can you confirm this is the existing TO-46-4 footprint with a few changes?

PatrickBaus:
No need to take my word. How about the two thousand of those pictures. I have a couple of them mounted in a project of mine. I just pushed the case lightly with a screwdriver:

LM399_left
LM399_right

I also added the datasheet link

PatrickBaus:
I think the footprint is complete now. I have also added the 3d model. Although this seemed to be too late for 5.17, but maybe in next version.

Thank you so much for your help @evanshultz

evanshultz:
Alright. We'll go with this. Thank you!

The 3D model is indeed a bit too late, but the symbol and footprint will make it into 5.1.7. :)

PR #2487, opened 2020-09-23
Fix L_1806 by reverting to previous version

cpresser:
Fix L_1806 by reverting to previous version


chschlue:
@evanshultz reverting for now. Sort this out later if necessary.

PR #2500, opened 2020-09-28
Replace WSON-8_4x4mm_P0.8mm with generated one

cpresser:
Naming of the legacy footprint was wrong (missing the exposed pad).
Also, it was not generated.


chschlue (approved):
Thanks.

PR #2507, opened 2020-09-30
More feathers

matthijskooijman:

This provides additional variations on the Feather footprint from #1657. It also applies some fixes to the original footprint.

I used the original board files from Adafruit to doublecheck the board dimensions and pin placement (by overlaying the new footprint).

### Documentation
 - https://learn.adafruit.com/adafruit-feather-32u4-fona
 - https://learn.adafruit.com/adafruit-feather-m0-radio-with-rfm69-packet-radio
 - https://learn.adafruit.com/adafruit-feather-32u4-radio-with-rfm69hcw-module
 - https://learn.adafruit.com/adafruit-feather-m0-wifi-atwinc1500/
 - https://learn.adafruit.com/introducing-the-adafruit-wiced-feather-wifi

### Matching symbols
Some of these footprints are needed by https://github.com/KiCad/kicad-symbols/pull/1933, but I added a few others (32u4_RFM and 32u4_FONA) for completeness.

### Policy violations
The FONA footprint says "Courtyard lines are not on 0.01mm grid". This is because that board is 2.375" = 60.325mm long and I just kept the standard 0.25mm or so clearance. What's the recommended way to solve this? Round the board dimensions? Or just the courtyard lines?

### 3D Models
I just left all of these at the default feather model, since if models are ever added, I'm doubtful that all variants are also added, so then just the default might be better? Or should I just use the footprint name for the 3D models everywhere?

### Mounting holes
I noticed that the mounting holes in the original footprint had a slightly different (drill and copper) size than the actual feather boards, so I updated them to match. For the topmost holes, this seems to match *all* feather boards ([Adafruit documentation says](https://learn.adafruit.com/adafruit-feather/feathers) "All Feathers have the same two mounting holes near the USB port"). For the bottommost holes, most of them seem to be 2.54mm without plating (checked against the board files of some feathers), so that's what I used for the common footprint. Looking at [this incomplete image](https://cdn-learn.adafruit.com/assets/assets/000/041/768/original/arduino_Feather_Family_Instagram_ORIG.jpg?1494734735), there seems to be one that use plated holes and one using 2.2mm holes (ESP8266), but I'm not so sure if it's useful to create all these variations too (also, I'd have to check the board files for each one to see what their hole size is, which I'm not too eager to do). The FONA also uses 2.2mm bottom holes and has its own footprint, so I use 2.2mm in that footprint.

### Screenshots of the mounting hole versions

image
image
image
image
image


myfreescalewebpage:
@matthijskooijman review of each fp:

Adafruit_Feather_32u4_FONA
- [ ] When converting from inches to mm, you should round the value (this should respond the above question)
- [ ] 3D package: Adafruit_Feather_32u4_FONA.wrl

Adafruit_Feather_32u4_FONA_WithMountingHoles
- [ ] When converting from inches to mm, you should round the value (this should respond the above question)
- [ ] 3D package: Adafruit_Feather_32u4_FONA.wrl

For the above footprint:
- [ ] We don't have the symbol, isn't it ?

Adafruit_Feather_32u4_RFM
- [ ] 3D package: Adafruit_Feather_32u4_RFM.wrl

Adafruit_Feather_32u4_RFM_WithMountingHoles
- [ ] 3D package: Adafruit_Feather_32u4_RFM.wrl

Adafruit_Feather_M0_RFM
- [ ] 3D package: Adafruit_Feather_M0_RFM.wrl

Adafruit_Feather_M0_RFM_WithMountingHoles
- [ ] 3D package: Adafruit_Feather_M0_RFM.wrl

Adafruit_Feather_M0_Wifi
- [ ] 3D package: Adafruit_Feather_M0_Wifi.wrl

Adafruit_Feather_M0_Wifi_WithMountingHoles
- [ ] 3D package: Adafruit_Feather_M0_Wifi.wrl

Adafruit_Feather_WICED
- [ ] 3D package: Adafruit_Feather_WICED.wrl

Adafruit_Feather_WICED_WithMountingHoles
- [ ] 3D package: Adafruit_Feather_WICED.wrl

Else looks good
Joel

matthijskooijman:
Ok, fixed everything, except:

>Adafruit_Feather_32u4_FONA
> Adafruit_Feather_32u4_RFM
>  We don't have the symbol, isn't it ?

Yeah, there are still some symbols missing, but it seemed easier to just do all footprints now while I already needed most of them. I was assuming this is ok, since it makes it easier to add the other symbols later.

myfreescalewebpage:
Module:Adafruit_Feather_M0_Wifi_WithMountingHoles
- [ ] The bottom holes are not correctly placed, they should move down

matthijskooijman:
Bugger, I should have seen that, sorry. Now I realize that was the only footprint I didn't check against the original schematics, because the drawings seemed clear enough, but I messed up the vertical position and, checking against the schematic, now noticed that the horizontal position is also different for this module (they are slightly further to the outside to fit the wifi chip, I suppose).

Anyway, fixed now.

image

myfreescalewebpage:
Double checked, now it's correct. Yes probably because of the WiFi antenna. Now merging your contribution, very good and efficient job done here, thanks!

